PR #15, opened 2020-06-24
fix: prismic-vue to @prismicio/vue package

lihbr:
Bootstrapping Prismic Vue SDK with new package name: from `prismic-vue` to `@prismicio/vue`, should avoid errors like that: https://github.com/nuxt-community/prismic-module/issues/70#issuecomment-648665532

Wasn't able to test if that's enough as I have no idea how to test this part of the CLI locally haha~

@hypervillain

### DIFF
--- a/packages/sm-api/bootstrap/nuxt/index.js
+++ b/packages/sm-api/bootstrap/nuxt/index.js
@@ -29,7 +29,7 @@ module.exports = {
     libraries: [],
     bootstraper: ["npx", ["create-nuxt-app"]],
     transpile: ["vue-slicezone", "nuxt-sm"],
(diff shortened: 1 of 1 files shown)


hypervillain:
To test a CLI + API feature, you need to:

- clone, install and npm run watch `prismic-cli`
- use `bin/prismic.js` as entry file
- pass param `--sm-endpoint [endpoint]` to the CLI

For your case, it would be:

`node ~/prismic-cli/bin/prismic.js sm --setup --sm-endpoint https://sm-api-g88ayrhm7.vercel.app/api`

hypervillain:
(no real need to test here but for next times 😉)

hypervillain (approved):

PR #16, opened 2020-06-25
Fix: next slicezone error when no component registered

chriz001:


hypervillain (approved):

hypervillain:
Good catch @chriz001, thanks!

PR #27, opened 2020-07-02
[sm-api] Return JSON file on setup

hypervillain:
* allow external libraries to be fetched
* use `unpkg` to fetch SM manifest
* return JSON instead of zip file
* remove /job endpoint and clean code


arnaudlewis:
Can you delete all the unused code please @hypervillain

PR #36, opened 2020-08-04
fix(sm-commons): use platform-aware path parser

lihbr:
## Types of changes

## Description

Now getting directory name with platform aware `path.basename` method.

Parsing paths with `/` does not work with win32 paths.


hypervillain:
Did you look for other similar cases?

lihbr:
Glob patterns are expeting POSIX paths, double checked rest of the code (looking for `/`) should be good, we'll likely discover other bug with win32 in the future but apart from this command most of my past work with SM was done on Windows without OS issues so should be good for now 👌

hypervillain:
Perfect, thank you!

PR #50, opened 2020-10-13
Adding revalidate to getStaticProps

FredvanRijswijk:
Next.js will attempt to re-generate the page:
- When a request comes in
- At most once every second

## Types of changes

revalidate: 1, // In seconds

- [x ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

Support revalidate getStaticProps for Nextjs 9.5 for incremental builds

### DIFF
--- a/packages/next-slicezone/hooks/useGetStaticProps.js
+++ b/packages/next-slicezone/hooks/useGetStaticProps.js
@@ -36,7 +36,8 @@ export const useGetStaticProps = ({
           error: null,
           slices: doc ? doc.data[body] : [],
           registry
-        }
+        },
+        revalidate: 1,
       }
 
     } catch(e) {


brybeecher:
Is there any reason `useGetStaticProps`'s can't accept a value to be used in the result of nextjs's `getStaticProps`?
I'm mostly concerned with the `revalidate` prop as well but I'm sure a case could be made for the others.

lihbr:
Hey @FredvanRijswijk, @brybeecher, thank you so much for contributing!

I get the point of this change and this makes me wonder if the current implementation of Next Slicezone has to be reworked as it should not be responsible for handling that kind of opinionated configuration (both ours and yours). What do you think @a-trost, @hypervillain?

hypervillain:
We have a `getStaticPropsParams` prop that is spread in the return object of getStaticProps. It's been merged with 0.1.2

PR #70, opened 2020-12-23
Pass DocData to slices

Duaner:
I would like to use DocData in my slices in order to be able to use in my Components elements like FirstPublicationDate or tags.

## Types of changes
Pass Document data through a prop (DocData) to the slice in order to use it in the slice level. 
Might be interesting to pass your custom data in the future

- [ ] New feature : Pass DocData at the Slicelevel

## Description

In my use case i would like to use FirstPublicationDate or tags in a slice. For that i need data coming fro the document not scoped in a slice only.

### DIFF
--- a/packages/next-slicezone/index.js
+++ b/packages/next-slicezone/index.js
@@ -3,7 +3,7 @@ import React from 'react'
 import { pascalize } from 'sm-commons/utils/str'
 
 const Empty = () => <p>Your SliceZone is empty!</p>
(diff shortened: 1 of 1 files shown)


lihbr:
Perhaps just calling it `data` at the slice level too? `slices` > `slice`, `data` > `data`

Also we need not to forget to update the props definition for the attribute not to bleed in the front-end 🙂 (for Vue at least)

Miniplop:
checked with @hypervillain, this PR is now obsolete. I close it

PR #96, opened 2021-04-25
fix: allow for slashes in libPath

Bram-Zijp:
I didn't like having linted JS/TS outside of my `src` folder. Therefor I'm using the following config inside of `sm.json`:

```
{
  "apiEndpoint": "https://***.cdn.prismic.io/api/v2",
  "libraries": [
    "@/src/slices"
  ],
  "_latest": "0.0.45",
}
```
This made the import path incorrect inside of `sm-resolver.js` e.g.:

```
import { Fragment } from 'react'
import * as Src/Slices from './src/slices'

const __allSlices = {  ...Src/Slices, }
...
```

This MR changes the above to:
```
import { Fragment } from 'react'
import * as SrcSlices from './src/slices'

const __allSlices = {  ...SrcSlices, }
...
```

## Types of changes

## Description

Not sure if this is breaking anywhere else, otherwise, enough should be mentioned. Perhaps it's an idea to fix this in the `pascalize` function.

Edit: 2 commits, pick your poison :smile:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ nbproject
 .project
 .settings
 .vscode
+.history
 composer.lock
 *.sublime-workspace
 *.swp
--- a/packages/next-slicezone/features/resolver/index.js
+++ b/packages/next-slicezone/features/resolver/index.js
@@ -26,7 +26,7 @@ async function handleLibraryPath(libPath) {
   return {
     isLocal,
     from,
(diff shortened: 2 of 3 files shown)


lihbr:
Hey again @Bram-Zijp, thank you so much for taking the time to open a PR! We'll have a look at it soon~

cc @arnaudlewis, @hypervillain

PR #133, opened 2021-06-21
nuxt-sm. Allow custom sm.json file location. Read sm.json file from nuxt rootDir by default

### LINKED ISSUE #22: Nuxt srcDir conflict between nuxt-sm & prismic-cli
### Version

- Prismic CLI version (know it by running `prismic --version`): 3.8.0-beta.0 & previous

### Reproduction

1. `npx create-nuxt-app` (default preset is fine)
2. `prismic sm --setup`
3. add `slice-zone` component inside your page (if you stop here everything works fine)
4. add key `srcDir` to `nuxt.config.js` and set it to `src` (whatever), then move all folder inside it (except `node_modules` and `.nuxt` of course)

### What is expected?

`nuxt-sm` should handle `srcDir` option correctly.

**Two cases:**

- If you don't move `sm.json` to new `src` folder you get this error when running Nuxt: `[Vue warn]: Error in render: "TypeError: Cannot read property 'resolver' of undefined"` because [`nuxt-sm` expect `sm.json` to be at `srcDir` level](https://github.com/prismicio/slice-machine/blob/master/packages/nuxt-sm/src/index.js#L49)

- If you move `sm.json` to `src` folder then Nuxt works well but `prismic sm --create-slice` fail:
  - At root level: SM file not found
  - Inside `src` folder: `Something (really) wrong happened. If the problem persists, feel free to contact us!
[Full error] Error: [Error] Please specify a framework by using "--framework <framework_name>"` because `sm.json` is not at the same level `package.json`, using the `--framework` flag doesn't help

### Any additionnal information
1. What does the `@` mean in `sm.json` libraries array? (`@/slices`) it's the path relative to this directory from `sm.json`?
2. `nuxt-sm` should expect `sm.json` to be at `Nuxt.options.rootDir` instead of `Nuxt.options.srcDir` by default ([`rootDir` has the same value as `srcDir` by default](https://nuxtjs.org/api/configuration-srcdir)), should be an easy fix if that's just that? Can you confirm scope @hypervillain?

shainegordon:
This PR adds 2 pieces of functionality

1. Allow a user to define the location `sm.json` using `options.pathToSmFile`, similar to `options.pathToResolver`
2. Read the `sm.json ` from `options.rootDir` instead of `options.srcDir` - See issue https://github.com/prismicio/slice-machine/issues/22

## Types of changes

## Description

The reason for this is that many of us configure nuxt.js to use a `./src` folder for our nuxt code, while keeping configuration and non-nuxt code (tests, storybook, etc) in the root.

Currently this is not possible, due to how `sm.json`

Resolves: #22

### DIFF
--- a/packages/nuxt-sm/package-lock.json
+++ b/packages/nuxt-sm/package-lock.json
@@ -1,8 +1,5711 @@
 {
   "name": "nuxt-sm",
   "version": "0.0.6",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "version": "0.0.6",
+      "license": "MIT",
+      "dependencies": {
+        "consola": "^2.11.3",
+        "jest": "^25.5.4",
+        "slash": "^3.0.0",
+        "sm-commons": "^0.0.21"
+      },
+      "devDependencies": {
+        "vue-essential-slices": "0.0.7"
+      }
+    },
+    "node_modules/@babel/code-frame": {
+      "version": "7.10.4",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.4.tgz",
+      "integrity": "sha512-vG6SvB6oYEhvgisZNFRmRCUkLz11c7rp+tbNTynGqc6mS1d5ATd/sGyV6W0KZZnXRKMTzZDRgQT3Ou9jhpAfUg==",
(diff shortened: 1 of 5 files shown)


lihbr:
That's awesome @shainegordon! I'll try to have a look at it this week 🔥

lihbr (approved):
LGTM, thanks!

PR #146, opened 2021-07-22
feat: add Gatsby support

angeloashmore:
> Note: This PR depends on PR #144

## Types of changes

## Description

This PR adds a minimal Gatsby supported framework. It creates a very basic React component for a newly created Slice.


packages/slice-machine/lib/env/index.ts
```diff
@@ -79,6 +80,17 @@ function validate(config: { apiEndpoint: string, storybook: string }): ConfigErr
       run: 'Add "storybook" property with a localhost url'
     }
   }
+
+  if (config.framework && !isValidFramework(config.framework)) {
```
Miniplop: What's happening if the framework is not set ?

packages/slice-machine/templates/slice/gatsby/index.js
```diff
@@ -0,0 +1,30 @@
...
+import * as React from 'react'
+import { graphql } from 'gatsby'
+
+// TODO: Implement the {{componentName}}Slice component.
+//
+// → Learn how to render this component with <SliceZone>
+//   https://prismic.io/docs/technologies/template-slices-gatsby
+export default function {{componentName}}Slice({ slice }) {
+  return (
+    <section>
```
Miniplop, line 12: I see that you generate a model with a **title** and a **description** field but those fields are not use in the template component. Is it something intended ?

Miniplop (commented):
Nice one, I have only 2 questions on the PR :)

packages/slice-machine/templates/slice/gatsby/index.js
angeloashmore: Yes, this is intentional. Gatsby is unique in that it requires a GraphQL query to fetch data. If the query contains invalid fields, the project fails to start or build.

Since `title` and `description` are default model fields, they are very likely to be deleted immediately by the developer. Since they may be deleted, it means the Gatsby project would quickly be in a broken state. The developer would need to go into the component and edit the query and component before moving forward.

By keeping `title` and `description` out of the component and GraphQL query, it will automatically support any immediate changes the developer makes to their model within Slice Machine.

Hope that makes sense! Let me know if I should describe in more detail. :)

packages/slice-machine/lib/env/index.ts
angeloashmore: If a framework is not set in `config` (read from the `sm.json` file), it will be populated in the `getEnv` function.

https://github.com/prismicio/slice-machine/blob/d376bdef73c606f8884732e3005e7bddf1d96b5d/packages/slice-machine/lib/env/index.ts#L205

Basically, if an explicit framework is not provided, it will attempt to detect the framework by reading the project's `package.json` file. If a framework cannot be detected, it falls back to a default "vanillajs" framework.

(Maybe it should fall back to the "none" framework that was just merged: https://github.com/prismicio/slice-machine/pull/145)

I think I understand this correctly, but @hypervillain let me know if this is inaccurate 😉

angeloashmore:
Closing this PR as it will be superseded by a plugin.

PR #163, opened 2021-09-13
Mm/yarn workspaces

MarcMcIntosh:
Added yarn workspaces so deps and individual package scripts can be run be run with one command from the top repository

## Description
Deps can be installed with one command from the root directory by running 'yarn', shared dependancies will be installed at the root level node_modules and sym-linked to where they are needed, same for the packages in this repo.

run tests in all projects with `yarn test` or `yarn workspaces run test`

## TODO
We should consider moving the test-project in `packages/slice-machine/tests/project` to it's own place, in the project either at the root level, or with in the packages directory.

Fix failing tests

remove as much of the top level `dependancies` as posable, as we want to make sure they are properly linked to where they are needed.

## Checklist:

- [x] My change requires a change to the documentation.

Is there a developer guide or section in the readme?
also some of the tests are failing, (but being run so that's a success).


Miniplop:
I tested the PR and I get this error during the next build

image

### How to reproduce
- I removed the SM node-modules and reinstall them with yarn workspaces (run at the root of the project). 
- I run the watch-server command, it is running well
- I run the yarn run build command it crash

Did you also get this error ? 

start command should be test after the building is passing.

PR #172, opened 2021-09-20
Kje/improve mock tests

Miniplop:
1) I refacto the mock test to get precise error when a mock builder is failing

## Types of changes

## Checklist:


packages/slice-machine/lib/models/common/widgets/Color/Mock/index.js
```diff
@@ -2,8 +2,17 @@ import { createDefaultHandleMockContentFunction } from "../../../../../utils";
 
 export const initialValues = null;
 
+function getRandomColor() {
+  const letters = "0123456789abcdef";
```
Heyrows: I would just define these outside of the function.
Between full random and generated value based on an array it seems to be the same.

packages/slice-machine/package.json
```diff
@@ -10,6 +10,7 @@
   "scripts": {
     "dev": "next dev",
     "test": "jest",
+    "watch-test": "jest --watch",
```
Heyrows, line 13: This might be a bit heavy, i wouldn't use it personally i think.

packages/slice-machine/tests/mocks.test.js
```diff
@@ -1,16 +1,315 @@
...
 
-test("Can create mocks", () => {
-  const primary = Object.entries(model.variations[0].primary);
-  const mockEntries = Object.entries(mockFields(primary));
-  expect(mockEntries.length).toBe(primary.length);
+describe("handleFields", () => {
+  const mockFields = handleFields(Widgets);
+  test("can create a text mock", () => {
+    const textField = [
+      [
```
Heyrows: I would use the constructor of the widget directly

packages/slice-machine/tests/mocks.test.js
```diff
@@ -1,16 +1,315 @@
...
+      ],
+    ];
 
-  const expector = createExpector(expect);
-  mockEntries.forEach(([key, value]) =>
-    expector[key] ? expector[key](value) : null
-  );
+    const textFieldMocked = Object.entries(mockFields(textField))[0];
+    const [key, value] = textFieldMocked;
+    expect(Array.isArray(value)).toBe(true);
```
Heyrows: I think to not have all of these, I would put the mock of the field in a mock file and then test the equality of it.
See if the function mock field for a given field gives the same result.

Heyrows: These two comments apply for all the fields i think.

Heyrows (changes requested):

packages/slice-machine/lib/models/common/widgets/Color/Mock/index.js
Miniplop: why not :)

packages/slice-machine/tests/mocks.test.js
Miniplop: Good idea :)

packages/slice-machine/tests/mocks.test.js
Miniplop: Yeah I use now match object with expect inside it, it makes clearer the object that should be returned :)

packages/slice-machine/lib/models/common/widgets/index.ts
```diff
@@ -1,33 +1,33 @@
...
+import { StructuredTextWidget } from "./StructuredText";
+import { TextWidget } from "./Text";
+import { TimestampWidget } from "./Timestamp";
+import { UIDWidget } from "./UID";
 
 export {
   BooleanWidget as Boolean,
-  Color,
-  ContentRelationship,
+  ColorWidget as Color,
```
Heyrows, line 19: Thanks for keeping the same names as before 👍 , `Wigets.ColorWidget` would have been bad.

Heyrows (approved):

PR #173, opened 2021-09-20
AAuser, I want to get the run script of Slicemachine in my package.json

Heyrows:
## Types of changes

- [X] Bug fix (a non-breaking change which fixes an issue)

## Description

We just want to make sure that the user has the script to run SM, even if he doesn't have an sm.json.

## Checklist:

wolfgang-hasselmann-_EG_Th77EmQ-unsplash

### DIFF
--- a/packages/slice-machine/postinstall.js
+++ b/packages/slice-machine/postinstall.js
@@ -57,11 +57,11 @@ function installSMScript(pkg) {
   const smModuleCWD = require.main.paths[0].split("node_modules")[0];
   const { pkg, smConfig } = retrieveConfigFiles(projectCWD);
 
-  if (pkg && smConfig) {
-    installSMScript(pkg);
-    writeSMVersion(smModuleCWD, smConfig);
-    return;
-  }
-  if (!pkg) console.error("Missing file package.json");
-  if (!smConfig) console.error("Missing file sm.json");
+  if (pkg.value) installSMScript(pkg);
+  else return console.error("[postinstall] Missing file package.json");
+
+  if (smConfig.value) writeSMVersion(smModuleCWD, smConfig);
+  else return console.error("[postinstall] Missing file sm.json");
+
+  return;
 })();
--- a/packages/slice-machine/postinstall.js
+++ b/packages/slice-machine/postinstall.js
@@ -61,7 +61,7 @@ function installSMScript(pkg) {
   else return console.error("[postinstall] Missing file package.json");
 
   if (smConfig.value) writeSMVersion(smModuleCWD, smConfig);
-  else return console.error("[postinstall] Missing file sm.json");
+  else console.error("[postinstall] Missing file sm.json");
 
   return;
 })();
--- a/packages/slice-machine/yarn.lock
+++ b/packages/slice-machine/yarn.lock
@@ -1600,7 +1600,7 @@
     "@nodelib/fs.scandir" "2.1.5"
     fastq "^1.6.0"
 
-"@npmcli/arborist@*", "@npmcli/arborist@>=2.8.2", "@npmcli/arborist@^2.3.0", "@npmcli/arborist@^2.5.0":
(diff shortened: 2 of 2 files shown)


packages/slice-machine/yarn.lock
```diff
@@ -1600,7 +1600,7 @@
     "@nodelib/fs.scandir" "2.1.5"
```
Miniplop, line 1600: Why is there so many changes in the yarn lock ?

packages/slice-machine/postinstall.js
```diff
@@ -57,11 +57,11 @@ function installSMScript(pkg) {
...
   const { pkg, smConfig } = retrieveConfigFiles(projectCWD);
 
-  if (pkg && smConfig) {
-    installSMScript(pkg);
-    writeSMVersion(smModuleCWD, smConfig);
-    return;
-  }
-  if (!pkg) console.error("Missing file package.json");
-  if (!smConfig) console.error("Missing file sm.json");
+  if (pkg.value) installSMScript(pkg);
```
Miniplop, line 60: what happen if pkg is not defined ? With this code it will crash trying to call value on undefined ?

packages/slice-machine/postinstall.js
```diff
@@ -57,11 +57,11 @@ function installSMScript(pkg) {
...
+  else return console.error("[postinstall] Missing file package.json");
+
+  if (smConfig.value) writeSMVersion(smModuleCWD, smConfig);
+  else return console.error("[postinstall] Missing file sm.json");
```
Miniplop: return useless here as there is a return at the end no ?

packages/slice-machine/yarn.lock
Heyrows: I used lerna to reinstall all the dependancies on all the packages :)

packages/slice-machine/postinstall.js
Heyrows: The property pkg already exist, it's the value inside that might be null if the file doesn't exist https://github.com/prismicio/slice-machine/blob/c690dfb5c65842be50540294b3a664eb06502a8a/packages/slice-machine/postinstall.js#L16

packages/slice-machine/postinstall.js
Heyrows: Yes :) this one is useless.

Miniplop (approved):

packages/slice-machine/postinstall.js
MarcMcIntosh: In the older version of the code it's possible for `console.error` to be called twice.
is that the behaviour we want to remove?

MarcMcIntosh:
When trying to test this i noticed a few things.

1) postinstall script is node postinstall  should it be node ./postinstall.js accorind to the docs on npm just ./postinstall.js will work.

2) after running npm pack and installing the tar-ball into a project the postinstall script doesn't seem to be run, i can force it to run with `npm explore slice-machine-ui -- npm run postinstall` but this changes process.cwd I'll try and find an other solution if not maybe we test this with an alpha / private release?

MarcMcIntosh:
when running with `node node_modules/slice-machine-ui/postinstall.js` (in a fresh project)

I get this result.

```
Added script "slicemachine" to package.json
[postinstall] Missing file sm.json
```

MarcMcIntosh:
according to `https://docs.npmjs.com/cli/v7/commands/npm-run-script` we can use `process.env.INIT_CWD` to get the directory where npm is being run from

but if it already works (in prod) then no need :)

Heyrows:
@MarcMcIntosh I tested the `process.env.INIT_CWD`
It seems to be undefined when launching it from another directory.
So in my case, I want to test a new version of SM with an external project, the postinstall has the right directory for the project and has the right directory for the SM module (different from node modules).
but the `process.env.INIT_CWD` is undefined, so I wouldn't use it.

image

MarcMcIntosh:
I was using it like `proccess.env.INIT_CWD || process.cwd` as INIT_CWD is supposed to be set by npm when it runs scripts

PR #174, opened 2021-09-20
fix key accessors in Field zones + fix Svelte hints

hypervillain:
I tested slices and stories created with Svelte integration.
I fixed fields that did not render correctly. I also spotted that fields containing a '-' would always break in JS, I added a small utility which formats these fields using brackets and leaves other fields untouched

### DIFF
--- a/packages/slice-machine/lib/utils/str.ts
+++ b/packages/slice-machine/lib/utils/str.ts
@@ -1,6 +1,14 @@
 import camelCase from "lodash/camelCase";
 
 const camelizeRE = /-(\w)/g;
+
+export function transformKeyAccessor(str: string): string {
+  if (str.includes('-')) {
+    return `["${str}"]`
+  }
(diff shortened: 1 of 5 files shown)


MarcMcIntosh:
``["${str}"]`` would over all cases right?

hypervillain:
I think it would not look like what people except in most cases

packages/slice-machine/lib/utils/str.ts
```diff
@@ -1,6 +1,14 @@
 import camelCase from "lodash/camelCase";
 
 const camelizeRE = /-(\w)/g;
+
+export function transformKeyAccessor(str: string): string {
```
Miniplop, line 5: Maybe we should this function in other frameworks no ? I think the problem can occur in React and Vue also no ?

Miniplop (approved):

PR #182, opened 2021-10-04
Mm/easier setup

MarcMcIntosh:
Adds `lerna bootstrap` to the top level post install script, this makes it easier to install everything in one command.
Also removed the call to `lerna bootstrap` from the ci as it's no longer needed

## Types of changes

## Checklist:


Miniplop (approved):
+1 Don't forget to mention the new command behavior to the other dev :)

packages/next-slicezone/package.json
```diff
@@ -10,7 +10,7 @@
     "build": "rollup -c",
     "prepublishOnly": "yarn build && yarn test",
     "lint": "eslint --ext .js,.jsx .",
-    "security-audit:prod": "yarn audit --production --only=prod --audit-level=low",
+    "security-audit:prod": "yarn audit --groups dependencies --audit-level=high",
```
Miniplop, line 13: I don't think this is the way to solve this problem I think. This line mean that we will ignore moderate security issues.

Miniplop (changes requested):
I don't think hiding the problem is a good solution. Which moderate issues are hard to fix or introduce regressions ? What are the options for these libs ?

MarcMcIntosh:
closed in favour of https://github.com/prismicio/slice-machine/pull/185

PR #188, opened 2021-10-07
Ls repos

MarcMcIntosh:
Added a method to get the repo names :)

## Types of changes

## Checklist:


packages/sm-init/src/index.ts
```diff
@@ -27,6 +32,9 @@ async function init() {
 
   await Auth.login(base);
   validatePkg(cwd);
+  const config = FileSystem.getOrCreateAuthConfig();
+  const repoName = await maybeExistingRepo(config.cookies, config.base);
```
Miniplop: why we don't use the base constant previously set ? And I think the FileSystem.getOrCreateAuthConfig(); can be put inside the fonction, seems to me that we always want to get the cookie from the authConfig right ?

packages/sm-init/src/steps/index.ts
```diff
@@ -1,3 +1,4 @@
 export { installSm } from './install-sm'
 export { validatePkg } from './validate-pkg'
-export { addScriptToPkg } from './add-script-to-pkg'
\ No newline at end of file
+export { addScriptToPkg } from './add-script-to-pkg'
+export { maybeExistingRepo } from './maybe-existing-repo'
```
Miniplop: new line here, prettier don't catch this ?

packages/sm-init/src/steps/maybe-existing-repo.ts
```diff
@@ -0,0 +1,58 @@
...
+        {name: "Create a new Repository", value: CREATE_REPO},
+        new inquirer.Separator("---- Use an existing Repository ----"),
+        ...repos.map((d) => ({name: d, value: d})),
+      ]
+    }, 
+  ])
+
+  if(res.repoName === CREATE_REPO) return promptForCreateRepo(base)
+  return Promise.resolve(res.repoName)
+}
```
Miniplop: no new line here

packages/slicemachine-core/index.ts
```diff
@@ -1,3 +1,3 @@
 export * as FileSystem from './src/filesystem'
 export * as Utils from './src/utils'
-export * from './src/core'
\ No newline at end of file
+export * from './src/core'
```
Miniplop, line 3: nice 😊

packages/slicemachine-core/src/core/auth.ts
```diff
@@ -125,29 +125,34 @@ export async function startServerAndOpenBrowser(
...
-    console.log("Opening browser to " + underline(url));
-    s.start();
-    void open(url);
+  return new Promise((resolve) => {
+    const s = spinner("Waiting for the browser response");
+
+    function onSuccess(data: HandlerData) {
+      s.succeed(`Logged in as ${bold(data.email)}`).stop();
+      setAuthConfigCookies(base, data.cookies);
+      resolve(); // todo add cookies here
```
Miniplop: I don't see the goal of creating a promise to return cookies if the promise don't return the cookies.

.github/workflows/test.yml
```diff
@@ -14,11 +14,8 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node-version }}
-          cache: "yarn"
```
Miniplop: I think we can keep the cache in CI. The problem in the CI is already solved :)

packages/sm-init/README.md
```diff
@@ -20,4 +20,9 @@ This package makes heavy use of the `slicemachine-core` package, you'll have to
 ### Solution 2
 - Watch the Core & the Init
 - run the `index.js` of the build in your project
-- enjoy coding
\ No newline at end of file
+- enjoy coding
+
+### Solution 3
+- from te top level run `yarn`
+- then `pushd packages/slicemachine-core && yarn build && popd && pushd packages/sm-init && yarn build && popd`
```
Heyrows, line 27: what's pushd ?

packages/sm-init/package.json
```diff
@@ -28,9 +27,16 @@
...
     "jest": "^27.2.4",
-    "ts-jest": "^27.0.5"
+    "mock-stdin": "^1.0.0",
+    "nock": "^13.1.3",
+    "stdout-stderr": "^0.1.13",
+    "ts-jest": "^27.0.5",
+    "typescript": "^4.4.3"
   },
   "dependencies": {
+    "chalk": "^4.1.2",
```
Heyrows: We already have Chalk and Ora on the Core side because the login requires it, do you think we need it here or we could export some helpers ?
Not saying one is better than the other but I don't think it's a good idea to have on both sides.

packages/slicemachine-core/src/core/auth.ts
MarcMcIntosh: the cookies are needed in a few places, but i guess they can be read form disk

packages/sm-init/README.md
MarcMcIntosh: like cd but will work on windows too :)
https://en.wikipedia.org/wiki/Pushd_and_popd

packages/sm-init/package.json
MarcMcIntosh: I think libraries like chalk and ora and make for cli's like the init script, rather than have them in the core it's self?

packages/slicemachine-core/src/core/auth.ts
Miniplop: My point was for now the resolve don't return nothing `resolve();` so what the purpose of using promise here instead of regular fonction ?

MarcMcIntosh: so we can await later on

packages/sm-init/src/index.ts
MarcMcIntosh: can do i guess

MarcMcIntosh: Although, this may cause issues if some-one logs in using the `prismic-cli` the runs sm init :/

packages/sm-init/src/steps/index.ts
MarcMcIntosh: re-ran prettier, gitup doesn't use git-hooks (yet)

packages/sm-init/src/index.ts
MarcMcIntosh: yes a few functions, the create repo function will also need that cookie

packages/sm-init/package.json
MarcMcIntosh: ora isn't used in the sm-api anymore

packages/sm-init/src/steps/index.ts
Miniplop: That's a problem, you put the pre-commit hook to catch those errors and avoid code style inconsistencies. So how can you ensure that your code respect our standard ?

Miniplop (approved):
LGTM 💪

packages/sm-init/package.json
Heyrows: The only use in the Core is the login function that make it works in both the CLI and the init. I don't know if it's a good thing to remove it since the style is the same between CLi and Init.
Don't really know what to do here tbh 🤷‍♂️

packages/slicemachine-core/.eslintrc.json
```diff
@@ -17,29 +17,5 @@
...
-    "@typescript-eslint/no-empty-function": "warn",
-    "prefer-const": "warn",
-    "no-var": "warn",
-    "@typescript-eslint/no-unnecessary-type-assertion": "warn",
-    "jest/no-identical-title": "warn",
-    "@typescript-eslint/triple-slash-reference": "warn",
-    "@typescript-eslint/await-thenable": "warn",
-    "@typescript-eslint/no-misused-promises": "warn"
-  }
+  "ignorePatterns": ["build", "templates", "**/__tests__/**"]
```
Heyrows: I know it doesn't come from you but I don't think tests should be excluded.
It's important to right clear tests, it's part of the code.
To debate ^^

packages/slicemachine-core/__tests__/src/core/communication.test.ts
```diff
@@ -9,14 +9,13 @@ describe("communication", () => {
...
   const fakeCookie = "prismic-auth=biscuits;";
-
   test("validateSession, default base", async () => {
     const responseData = {
       email: "fake@prismic.io",
       type: "USER",
       repositories: {
-        "foo-repo": { dbid: "abcd", role: communication.Roles.OWNER },
-        qwerty: { dbid: "efgh", role: communication.Roles.WRITER },
+        "foo-repo": { dbid: "abcd", role: "OWNER" },
```
Heyrows: incoherence with the lines below.
`OWNER` is not a valid value here, `Owner` is.
This is something that would have been caught by the linter and i'm surprised typescript didn't warn about it.
It's always preferable to use an Enum, that way if the value change you have to change it at one place only.

packages/slicemachine-core/__tests__/src/core/communication.test.ts
```diff
@@ -47,4 +46,136 @@ describe("communication", () => {
         expect(data).toEqual(responseData);
       });
   });
+
+  test("refreshSession", async () => {
+    const token = "biscuits";
+    const base = "https://prismic.io";
+    const wanted = "some-new-token";
+    nock("https://auth.prismic.io")
```
Heyrows, line 55: Just a suggestion, but could it be better to use another value than prismic.io, just so we can detect any hard coded prismic.io we might have ?

It's just a suggestion, i didn't applied it myself yet.

packages/slicemachine-core/__tests__/src/core/communication.test.ts
```diff
@@ -47,4 +46,136 @@ describe("communication", () => {
...
+    expect(result).toEqual(wanted);
+  });
+
+  test("listRepositories", async () => {
+    const base = "https://prismic.io";
+    const responseData = {
+      email: "fake@prismic.io",
+      type: "USER",
+      repositories: {
+        "foo-repo": { dbid: "abcd", role: "OWNER" },
```
Heyrows: I would use the enum here. And maybe define that at the top level of all tests since they seem to all use the same.

packages/slicemachine-core/__tests__/src/core/communication.test.ts
```diff
@@ -47,4 +46,136 @@ describe("communication", () => {
...
+
+    const result = await communication.listRepositories(fakeCookie, base);
+    expect(result).toEqual(responseData.repositories);
+  });
+
+  describe("validateRepositoryName", () => {
+    const fakeBase = "https://prismic.io";
+
+    test("should fail if subdomain is not defined", async () => {
+      const fn = () => communication.validateRepositoryName();
```
Heyrows, line 87: I think the function (with typescript) should validate a value, the value should always be defined.

packages/slicemachine-core/__tests__/src/core/communication.test.ts
```diff
@@ -47,4 +46,136 @@ describe("communication", () => {
...
+      expect(fn).rejects.toThrow("start with a letter");
+    });
+
+    test("should fail if the name ends with a hyphen", async () => {
+      const fn = () => communication.validateRepositoryName("abc-");
+      expect(fn).rejects.toThrow("Must end in a letter or a number");
+    });
+
+    test("Max length 30 characters", async () => {
+      const repoName = Array.from({ length: 31 }, () => "a").join("");
```
Heyrows, line 116: funny, creative 😂

packages/slicemachine-core/__tests__/src/core/communication.test.ts
```diff
@@ -47,4 +46,136 @@ describe("communication", () => {
...
+
+    test("should fail if subdomain is not defined", async () => {
+      const fn = () => communication.validateRepositoryName();
+      expect(fn).rejects.toThrow("repository name is required");
+    });
+
+    test("should fail if name length is less than 4", async () => {
+      const fn = () => communication.validateRepositoryName("abc");
+      expect(fn).rejects.toThrow(
+        "Must have four or more alphanumeric characters and/or hyphens."
```
Heyrows, line 94: All these errors are really very precise.
If we decide to change one of them the test would fail automatically.
I think an enum would be good for this. More robust.

packages/slicemachine-core/__tests__/src/core/communication.test.ts
```diff
@@ -47,4 +46,136 @@ describe("communication", () => {
...
+      const fn = () => communication.validateRepositoryName(repoName);
+      expect(fn).rejects.toThrow(
+        "(1: Must start with a letter. (2: Must contain only lowercase letters, numbers and hyphens."
+      );
+    });
+
+    test("should fail if repo name is not available", async () => {
+      const repoName = "test";
+      nock(fakeBase)
+        .get(`/app/dashboard/repositories/${repoName}/exists`)
```
Heyrows, line 132: We have an endpoint modules in the Core and I think these should be inside of it.

packages/slicemachine-core/__tests__/src/core/communication.test.ts
```diff
@@ -47,4 +46,136 @@ describe("communication", () => {
...
+
+      const fn = () => communication.validateRepositoryName(repoName);
+
+      expect(fn).rejects.toThrow("already in use");
+    });
+
+    test("existing repo", () => {
+      const repoName = "test";
+      nock(fakeBase)
+        .get(`/app/dashboard/repositories/${repoName}/exists`)
```
Heyrows, line 143: same

packages/slicemachine-core/__tests__/src/core/communication.test.ts
```diff
@@ -47,4 +46,136 @@ describe("communication", () => {
...
+        .reply(200, () => false);
+      expect(
+        communication.validateRepositoryName(repoName, fakeBase, true)
+      ).resolves.toEqual(repoName);
+    });
+
+    test("existing repo, does not exist", () => {
+      const repoName = "test";
+      nock(fakeBase)
+        .get(`/app/dashboard/repositories/${repoName}/exists`)
```
Heyrows, line 153: same

packages/slicemachine-core/__tests__/src/core/communication.test.ts
```diff
@@ -47,4 +46,136 @@ describe("communication", () => {
...
+    });
+
+    test("different base", () => {
+      const repoName = "test";
+      nock("https://example.com")
+        .get(`/app/dashboard/repositories/${repoName}/exists`)
+        .reply(200, () => true);
+
+      expect(
+        communication.validateRepositoryName(repoName, "https://example.com")
```
Heyrows, line 178: should this function be in the core also ?

Heyrows: which doesn't mean it require more test because it would just call the one you made.

packages/slicemachine-core/src/core/communication.ts
```diff
@@ -50,10 +56,7 @@ export async function validateSession(
   return axios
     .get<{ email: string; type: string; repositories?: string }>(url)
     .then((res) => {
-      const repositories: RepoData =
-        res.data.repositories && typeof res.data.repositories === "string"
-          ? JSON.parse(res.data.repositories)
-          : res.data.repositories;
+      const repositories = maybeParseRepoData(res.data.repositories);
```
Heyrows, line 78: good things to have extracted it in a function but I think that the case where `data.repositories` doesn't exist should be handled here.

packages/slicemachine-core/src/core/communication.ts
```diff
@@ -41,6 +41,12 @@ export enum Roles {
 export type RepoData = Record<string, { role: Roles; dbid: string }>;
 export type UserInfo = { email: string; type: string; repositories: RepoData };
 
+function maybeParseRepoData(repos?: string | RepoData): RepoData {
+  if (!repos) return {};
+  if (typeof repos === "string") return JSON.parse(repos) as RepoData;
```
Heyrows: You should use IO-TS because here if the contract change you wouldn't find it out until it breaks somewhere.

packages/slicemachine-core/src/core/index.ts
```diff
@@ -29,8 +31,13 @@ export interface Core {
   },*/
 
   Repository: {
-    list: (token: string) => Promise<string[]>;
-    create?: (apiEndpoint: string, token: string) => Promise<void>;
+    list: (token: string, base?: string) => Promise<Communication.RepoData>;
```
Heyrows: If it's inside the core, you don't need the base as parameter, the Core will provide :)

packages/slicemachine-core/src/core/index.ts
```diff
@@ -29,8 +31,13 @@ export interface Core {
...
+    // create: (apiEndpoint: string, token: string) => Promise<void>
+    validateName: (
```
Heyrows, line 36: same

packages/slicemachine-core/src/core/index.ts
Heyrows: I wonder if we need the token also or can it come from the Filesystem directly

packages/slicemachine-core/src/core/index.ts
```diff
@@ -47,8 +54,13 @@ export default function createCore({ cwd, base, manifest }: CoreParams): Core {
...
 
     Repository: {
-      list: async (token: string): Promise<string[]> =>
+      list: async (token: string): Promise<Communication.RepoData> =>
         Communication.listRepositories(token, base),
+      validateName: (
+        name: string,
+        base = CONSTS.DEFAULT_BASE,
+        existingRepo = false
+      ) => Communication.validateRepositoryName(name, base, existingRepo),
```
Heyrows: return type is missing.

packages/slicemachine-core/src/utils/cookie.ts
```diff
@@ -1,11 +1,18 @@
...
+  opts?: cookie.CookieParseOptions
+): { [key: string]: string } {
+  return cookie.parse(cookieString, { decode: noEscape, ...opts });
 }
 
-export function serialize(name: string, value: string, opts?: cookie.CookieSerializeOptions): string {
-  return cookie.serialize(name, value, {encode: noEscape, ...opts})
-}
\ No newline at end of file
+export function serialize(
```
Heyrows, line 12: lots of prettier stuff, it's weird that it was triggered in your branch but not before, did you make a changes to the prettier conf ?

packages/sm-init/.eslintrc.json
```diff
@@ -17,29 +17,5 @@
...
-    "@typescript-eslint/no-empty-function": "warn",
-    "prefer-const": "warn",
-    "no-var": "warn",
-    "@typescript-eslint/no-unnecessary-type-assertion": "warn",
-    "jest/no-identical-title": "warn",
-    "@typescript-eslint/triple-slash-reference": "warn",
-    "@typescript-eslint/await-thenable": "warn",
-    "@typescript-eslint/no-misused-promises": "warn"
-  }
+  "ignorePatterns": ["build", "templates", "**/__tests__/**"]
```
Heyrows: same as the other eslint file

packages/slicemachine-core/src/utils/poll.ts
```diff
@@ -4,20 +4,26 @@ export default async function poll<Result>(
...
-      return reject(new Error("Exceeded max attempts"));
-    } else {
-      setTimeout(executePoll, interval, resolve, reject);
-    }
-  };
-
-  return new Promise(executePoll);
+  let attempts = 0
+  const checkCondition = (
+    resolve: (value: void | PromiseLike<void>) => void,
```
Heyrows, line 9: how can value be a void ?

Heyrows: I think the code before was clearer and it returned the result as well, i'm not sure why you changed it like this.
The arguments were a lot clearer and the Result type was used as a universal return type.

packages/sm-init/__tests__/maybe-existing-repo.test.ts
```diff
@@ -0,0 +1,336 @@
...
+      .get("/validate?token=biscuits")
+      .reply(200, {
+        email: "fake@prismic.io",
+        type: "USER",
+        repositories: JSON.stringify({ dbid: "foo", role: "Owner" }),
+      });
+
+    jest
+      .spyOn(inquirer, "prompt")
+      .mockResolvedValueOnce({ repoName: CREATE_REPO })
```
Heyrows, line 81: does this test that the user is asked again when the repo name is invalid ?
It so the name of the test is not clear enough, should be something like "it ask the users fora valid domain until it receives one"

packages/sm-init/__tests__/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "noEmit": true
+  },
+  "references": [{ "path": ".." }]
```
Heyrows, line 6: You should exclude the test from typescript original config and include them in this one so they wo'nt be transpiled and run twicer

packages/sm-init/src/index.ts
```diff
@@ -16,6 +21,8 @@ async function init() {
 
   await loginOrBypass(base);
   validatePkg(cwd);
+  const config = FileSystem.getOrCreateAuthConfig();
+  await maybeExistingRepo(config.cookies, cwd, base);
```
Heyrows, line 25: The step should only takes what's coming from the core, the config should be retrieved in side the step itself to keep a clear execution in this file.

packages/sm-init/src/steps/maybe-existing-repo.ts
```diff
@@ -0,0 +1,174 @@
+import * as inquirer from "inquirer";
+import Separator from "inquirer/lib/objects/separator";
+import { Communication, Utils, FileSystem } from "slicemachine-core";
+import { RepoData } from "slicemachine-core/src/core/communication";
+
+export const CREATE_REPO = "$_CREATE_REPO"; // not a valid domain name
```
Heyrows, line 6: I understand that this value would be used in tests but i don't see what it has to do here.
I think it shouldn't be here at all, and tests shouldn't affect the code if that the only use of this variable.
But maybe i'm missing something.

packages/sm-init/src/steps/maybe-existing-repo.ts
```diff
@@ -0,0 +1,174 @@
...
+const DEFAULT_BASE = Utils.CONSTS.DEFAULT_BASE;
+
+export function prettyRepoName(address: URL, value?: string): string {
+  const repoName = value ? Utils.cyan(value) : Utils.dim.cyan("repo-name");
+  const msg = [
+    Utils.cyan.dim(`${address.protocol}//`),
+    repoName,
+    Utils.cyan.dim(`.${address.hostname}`),
+  ];
+  return msg.join("");
```
Heyrows: please return a dynamic string for clarity.

packages/sm-init/src/steps/maybe-existing-repo.ts
```diff
@@ -0,0 +1,174 @@
...
+            false
+          );
+          return result === name || result;
+        },
+      },
+    ])
+    .then((res) => res.repoName);
+}
+
+export function maybeRepoNameFromSMFile(cwd: string, base: string) {
```
Heyrows: I think this should be an helper from the Core package.
Maybe even one to get a repo from an endpoint and a helper on the Manifest file to retrieve it.
I'm guessing this will be used a bunch.
And i'm guessing if you need it then maybe you already have the Core at this point and it could be called from the directly.

packages/sm-init/src/steps/maybe-existing-repo.ts
```diff
@@ -0,0 +1,174 @@
...
+  if (!maybeSMFile.content?.apiEndpoint) return null;
+
+  const repoUrl = new URL(maybeSMFile.content.apiEndpoint);
+  const correctBase = repoUrl.hostname.includes(baseUrl.hostname);
+  if (correctBase === false) return null;
+
+  return repoUrl.hostname.split(".")[0];
+}
+
+export function canUpdateCustomTypes(role: Communication.Roles) {
```
Heyrows: same, I would define that as a helper in the communication module.

packages/sm-init/src/steps/maybe-existing-repo.ts
```diff
@@ -0,0 +1,174 @@
...
+    return {
+      name: `${Utils.purple("Use")} ${Utils.bold(repoName)} ${Utils.purple(
+        address.toString()
+      )}`,
+      value: repoName,
+    };
+  };
+}
+
+export function orderPrompts(maybeName?: string | null) {
```
Heyrows, line 80: I do understand the need for disabled but not for the Invalid repo name. I don't see how we could end up having to log that.

packages/sm-init/src/steps/maybe-existing-repo.ts
```diff
@@ -0,0 +1,174 @@
...
+  cwd: string
+): RepoPrompts {
+  const createNew = {
+    name: `${Utils.purple("Create a")} ${Utils.bold("New")} ${Utils.purple(
+      "Repository"
+    )}`,
+    value: CREATE_REPO,
+  };
+
+  // const sep: Separator = new inquirer.Separator(
```
Heyrows: remove unused code :)

packages/sm-init/src/steps/maybe-existing-repo.ts
```diff
@@ -0,0 +1,174 @@
...
+  //   "---- Use an existing Repository ----"
+  // );
+
+  const start: RepoPrompts = [
+    createNew,
+    //  sep
```
Heyrows: same

packages/sm-init/src/steps/maybe-existing-repo.ts
```diff
@@ -0,0 +1,174 @@
...
+  const choices = sortReposForPrompt(repos, base, cwd);
+
+  const numberOfDisabledRepos = choices.filter((repo) => {
+    if (repo instanceof Separator) return false;
+    return repo.disabled;
+  }).length;
+
+  const maybeConfiguredRepoName = maybeRepoNameFromSMFile(cwd, base);
+
+  const res = await inquirer.prompt<Record<string, string>>([
```
Heyrows, line 139: If there is no configured repo name then why wait for inquirer at all, why not just promptForCreateRepo directly ?

Heyrows (changes requested):

Heyrows (approved):

packages/sm-init/src/steps/maybe-existing-repo.ts
```diff
@@ -15,6 +15,12 @@ export function prettyRepoName(address: URL, value?: string): string {
 
 export async function promptForRepoName(base: string): Promise<string> {
   const address = new URL(base);
+
+  const hint = `${Utils.yellow("ℹ")}  ${Utils.dim(
```
Heyrows, line 19: I would have put it in the Core because I don't think this is the last hint we ever write.

Heyrows (approved):

PR #190, opened 2021-10-13
[V0.1.1] - AAuser, I want SM init to recognise if i'm logged in, so that I can bypass login

Heyrows:
## Types of changes

## Checklist:

### DIFF
--- a/packages/slicemachine-core/__tests__/src/core/auth.test.ts
+++ b/packages/slicemachine-core/__tests__/src/core/auth.test.ts
@@ -1,12 +1,13 @@
 import { describe, expect, test, afterAll, afterEach } from "@jest/globals";
 import * as authHelpers from "../../../src/core/auth";
 import { Auth } from "../../../src/core";
-import { buildEndpoints } from "../../../src/utils";
+import * as Utils from "../../../src/utils";
 import * as communication from "../../../src/core/communication";
 import * as filesystem from "../../../src/filesystem";
 
 jest.mock("../../../src/filesystem");
 jest.mock("../../../src/core/communication");
+jest.mock("../../../src/utils/poll");
 
 describe("communication", () => {
   afterAll(() => {
@@ -19,7 +20,7 @@ describe("communication", () => {
   });
 
   const fakeBase = "https://fake.io";
-  const endpoints = buildEndpoints(fakeBase);
+  const endpoints = Utils.buildEndpoints(fakeBase);
 
   test("login should always have the same parameters", async () => {
     const spy = jest.spyOn(authHelpers, "startServerAndOpenBrowser");
@@ -27,7 +28,7 @@ describe("communication", () => {
       expect(url).toEqual(endpoints.Dashboard.cliLogin);
       expect(action).toEqual("login");
       expect(base).toEqual(fakeBase);
-      return Promise.resolve();
+      return Promise.resolve({ onLoginFail: () => null });
     });
 
     await Auth.login(fakeBase);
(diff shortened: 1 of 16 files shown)


packages/slicemachine-core/src/core/index.ts
```diff
@@ -66,4 +71,18 @@ export const Auth = {
     );
   },
   logout: (): void => removeAuthConfig(),
+  validateSession: async (
+    requiredBase: string
+  ): Promise<Communication.UserInfo | null> => {
+    console.log("validate session");
```
Miniplop: Do we want to keep this console.log ? :p

packages/slicemachine-core/src/core/index.ts
```diff
@@ -66,4 +71,18 @@ export const Auth = {
...
+    const config: AuthConfig = getOrCreateAuthConfig();
+    console.log(config);
```
Miniplop: same question here

packages/slicemachine-core/src/core/index.ts
```diff
@@ -66,4 +71,18 @@ export const Auth = {
...
+    if (!config.cookies.length) return Promise.resolve(null); // default config, logged out.
+    if (requiredBase != config.base) return Promise.resolve(null); // not the same base so it doesn't count.
+
+    console.log("validating now");
```
Miniplop: Re same question hihi

packages/slicemachine-core/__tests__/src/core/auth.test.ts
```diff
@@ -69,4 +78,84 @@ describe("communication", () => {
...
+          qwerty: { dbid: "efgh", role: communication.Roles.WRITER },
+        },
+      })
+    );
+
+    const result = await Auth.validateSession(fakeBase);
+    expect(result).toBe(null);
+  });
+
+  test("validate session shouldn't work for different base", async () => {
```
Miniplop: maybe we can change the name to "validate session should return null if there is different bases"

packages/slicemachine-core/__tests__/src/core/auth.test.ts
```diff
@@ -69,4 +78,84 @@ describe("communication", () => {
     expect(authHelpers.isHandlerData(nonHandlerData4)).toBe(false);
     expect(authHelpers.isHandlerData(nonHandlerData5)).toBe(false);
   });
+
+  test("validate session shouldn't work for empty cookies", async () => {
```
Miniplop: maybe we can change the name to "validate session should return null for empty cookies"

packages/slicemachine-core/__tests__/src/core/auth.test.ts
```diff
@@ -69,4 +78,84 @@ describe("communication", () => {
...
+          qwerty: { dbid: "efgh", role: communication.Roles.WRITER },
+        },
+      })
+    );
+
+    const result = await Auth.validateSession(fakeBase);
+    expect(result).toBe(null);
+  });
+
+  test("validate session shouldn't work for throw in the validation", async () => {
```
Miniplop: maybe we can change the name to "validate session should return null when validate session reject the promise"

packages/slicemachine-core/__tests__/src/core/auth.test.ts
```diff
@@ -69,4 +78,84 @@ describe("communication", () => {
...
+    });
+
+    const mockedValidate = communication.validateSession as jest.Mock;
+    mockedValidate.mockReturnValue(Promise.reject("unauthorized"));
+
+    const result = await Auth.validateSession(fakeBase);
+    expect(result).toBe(null);
+  });
+
+  test("validate session should work", async () => {
```
Miniplop, line 140: Is there others case where the validate session should works ?

Heyrows: the only other case i can think of is if the .prismic file doesn't exist but it has been covered by the getConfig which creates a new one if it doesn't exist.

Miniplop (approved):
Nice :)

PR #193, opened 2021-10-15
AAUser, when my connexion token expires, I can login from SM UI

Miniplop:
## Types of changes

## Checklist:


Miniplop:
image

.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-yarn lint-staged
+npx lint-staged
```
Heyrows: I think we should use `npm lint-staged`.
That way we controle which version of the package is used and if one day we decide to change this will throw an error right away because the package isn't installed.

Makefile
```diff
@@ -0,0 +1,22 @@
+.PHONY: install
```
Heyrows, line 1: What's the use of this file ?
Seems like a CI thing but i'm not sure.

package.json
```diff
@@ -22,14 +20,16 @@
   },
   "dependencies": {
     "@theme-ui/color": "^0.3.1",
+    "fp-ts": "^2.11.4",
```
Heyrows: I don't think we need this io-ts and fp-ts at this level, there is no code here :).

package.json
```diff
@@ -2,15 +2,13 @@
   "name": "root",
   "private": true,
   "scripts": {
-    "::postinstall": "manypkg check",
     "bootstrap": "lerna bootstrap",
     "postinstall": "lerna bootstrap",
     "test": "jest",
-    "postinstall": "lerna bootstrap",
```
Heyrows, line 9: i think this command was there to be executed automatically by npm.
Are we sure we want to delete it.

packages/next-slicezone/package.json
```diff
@@ -10,9 +10,8 @@
     "build": "rollup -c",
     "prepublishOnly": "yarn build && yarn test",
     "lint": "eslint --ext .js,.jsx .",
-    "security-audit:prod": "yarn improved-yarn-audit --ignore-dev-deps --min-severity high",
-    "security-audit:dev": "yarn improved-yarn-audit --min-severity moderate",
-    "audit-fix": "yarn-audit-fix",
+    "audit": "better-npm-audit audit -l high -p",
```
Heyrows, line 13: I know that for yarn you need to do a `yarn module` to call a node module, isn't it the same for Npm ? (tbh I don't know, that's a question 😄 )

packages/slice-machine/components/LoginModal/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+import { useToasts } from "react-toast-notifications";
+
+Modal.setAppElement("#__next");
+
+type LoginModalProps = {
+  onClose: () => void;
+  isOpen: boolean;
+};
+
+const poll = async (
```
Heyrows, line 25: I'm pretty sure you made a Poll function in the core already.

Heyrows: Or at least I saw one on a PR of Marc and it had been refactored for linter issues.

packages/slice-machine/components/LoginModal/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+import React, { useContext, useState } from "react";
+import { Button, Card, Close, Flex, Heading, Spinner, Text } from "theme-ui";
+import Prismic from "components/AppLayout/Navigation/Icons/Prismic";
+import SliceMachineModal from "@components/SliceMachineModal";
+import { ConfigContext } from "@src/config-context";
+import { useToasts } from "react-toast-notifications";
+
+Modal.setAppElement("#__next");
+
+type LoginModalProps = {
```
Heyrows: type vs interface that is the question
https://stackoverflow.com/questions/37233735/interfaces-vs-types-in-typescript

I don't mind if it remains a type, I usually use interface rather than type.

packages/slice-machine/components/LoginModal/index.tsx
Heyrows: I would at least put it in another file so that when Marc's PR is merged, we just need to change the import

packages/slice-machine/components/LoginModal/index.tsx
```diff
@@ -0,0 +1,154 @@
+import Modal from "react-modal";
+import React, { useContext, useState } from "react";
+import { Button, Card, Close, Flex, Heading, Spinner, Text } from "theme-ui";
+import Prismic from "components/AppLayout/Navigation/Icons/Prismic";
```
Heyrows: I think the path should be relative.

packages/slice-machine/components/LoginModal/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const { addToast } = useToasts();
+
+  const onClick = async () => {
+    if (!env) {
+      return;
+    }
+
+    try {
+      setIsLoading(true);
+      await fetch("/api/auth/start", { method: "POST" });
```
Heyrows: It could be good to have an endpoint module with all the routes shared by the serveur and the UI maybe.
That way if one route change it doesn't break everything.

packages/slice-machine/components/LoginModal/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+      const checkStatus = async () =>
+        await fetch("/api/auth/status", { method: "POST" }).then((response) =>
+          response.json()
+        );
+      const isAuthStatusOk = ({ status }: { status: string }) =>
+        status === "ok";
+      const apiUrl = new URL(env.baseUrl);
+
+      window.open(
```
Heyrows: just a suggestion but maybe we could close the window when the poll return a true.
Originally the dashboard page has a success message when you're logged in but maybe we can close it here and the user would just see the popup disappear

Heyrows:
```suggestion
      const tab = window.open('blabla')
      tab.close()
```

packages/slice-machine/components/LoginModal/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+        `https://prismic.io/dashboard/cli/login?port=${apiUrl.port}&path=/api/auth`,
+        "_blank"
+      );
+      await poll(checkStatus, isAuthStatusOk, 3000, 60);
```
Heyrows: i'm wondering if we can still nagivate while it polls. To test

packages/slice-machine/components/LoginModal/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+      setIsLoading(false);
+      addToast("Logging successfully", { appearance: "success" });
```
Heyrows: Are sure about the success/fail messages.
There are so many ways to write the action of logging in that i'm never sure which one is good.

packages/slice-machine/components/LoginModal/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+          }}
+        >
+          <Text
+            variant={"xs"}
+            sx={{
+              mb: 3,
+              maxWidth: 280,
+            }}
+          >
+            In to order to perform this action, you’ve to be connected into your
```
Heyrows: `you've got to be connected` or `you need to be connected`.
Maybe `with` instead of `Into`

packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header.tsx
```diff
@@ -45,9 +47,13 @@ const SliceHeader = ({
           if (!isLoading) {
             setIsLoading(true);
             store.push(Model, (data: ToastPayload): void => {
-              if (data.done) {
-                setIsLoading(false);
-                handleRemoteResponse(addToast)(data);
+              if (!data.done) {
```
Heyrows, line 50: if the done symbolise that everything is Pushed, if you return before an error or the status is checked, the popup will never open no ?

packages/slice-machine/lib/builders/SliceBuilder/index.jsx
```diff
@@ -40,6 +42,13 @@ const Builder = ({ openPanel }) => {
     error: null,
   });
 
+  const onPush = (data) => {
+    setData(data);
```
Heyrows, line 46: I would set the data only if there is no error.

packages/slice-machine/lib/builders/SliceBuilder/index.jsx
```diff
@@ -40,6 +42,13 @@ const Builder = ({ openPanel }) => {
...
+    if (data.error && (data.status === 401 || data.status === 403)) {
```
Heyrows: To me 401 means you're not connected while 403 means you don't have the right to do that.
I wouldn't open the login for a 403

packages/slice-machine/package.json
```diff
@@ -14,20 +14,20 @@
...
-    "export-build": "next export && yarn clean-build",
-    "bundle": "yarn build && yarn test && yarn export-build",
+    "export-build": "next export && npm run clean-build",
+    "bundle": "npm run build && npm run test && npm run export-build",
     "slicemachine": "start-slicemachine --port 9999",
-    "security-audit:prod": "yarn improved-yarn-audit --ignore-dev-deps --min-severity high --exclude GHSA-4jqc-8m5r-9rpr",
-    "security-audit:dev": "yarn improved-yarn-audit --min-severity moderate",
     "lint": "eslint --max-warnings 3279 --ext .ts,.tsx .",
-    "audit-fix": "yarn-audit-fix"
+    "lint:precommit": "eslint",
```
Heyrows, line 28: wouldn't the commit fail if we don't have the max-warning ?
Why not put it there ?

packages/slice-machine/pages/[lib]/[sliceName]/[variation].tsx
```diff
@@ -1,6 +1,8 @@
 import Builder from "@builders/SliceBuilder";
 
-const SliceEditor = ({ openPanel }) => {
+const SliceEditor: React.FunctionComponent<{
```
Comment (Heyrows) [inline packages/slice-machine/pages/[lib]/[sliceName]/[variation].tsx]: That's a personal preference but I like to have the interface/type of a component above it.

packages/slice-machine/pages/_app.tsx
```diff
@@ -171,44 +172,49 @@ function MyApp({
...
-                              open={drawerState.open}
-                              onClose={() =>
-                                setDrawerState({ ...drawerState, open: false })
-                              }
-                            >
-                              <Warnings
-                                priority={drawerState.priority}
-                                list={data.warnings}
-                                configErrors={data.configErrors}
+                <ToastProvider>
```
Heyrows, line 175: Just for my knowledge because I didn't worked much with contexte before, the provider is a component that adds the context to the property of the children ?

packages/slice-machine/pages/cts/[ct].tsx
```diff
@@ -5,47 +5,32 @@ import { CustomTypesContext } from "src/models/customTypes/context";
...
-  const [Model, store] = useModelReducer({
-    customType,
-    remoteCustomType,
-    initialMockConfig,
-  });
-  return <Ct Model={Model} store={store} onLeave={onLeave} />;
 };
 
-const WithRouter = () => {
+const CustomTypeBuilderWithProvider: React.FunctionComponent<CustomTypeBuilderWithProviderProps> =
```
Comment (Heyrows) [inline packages/slice-machine/pages/cts/[ct].tsx:18]: I like the changes on this file, a quick refacto but a nice one !!

packages/slice-machine/server/src/api/auth/post.ts
```diff
@@ -0,0 +1,34 @@
+import * as t from "io-ts";
+import { fold } from "fp-ts/Either";
+import { FileSystem } from "slicemachine-core";
+
+const AuthRequest = t.type({
```
Heyrows, line 5: I love that

packages/slice-machine/server/src/api/auth/post.ts
```diff
@@ -0,0 +1,34 @@
...
+  email: t.string,
+  cookies: t.array(t.string),
+});
+
+type AuthRequest = { email: string; cookies: string[] };
```
Heyrows: You can do this, it's cooler and we never redefine twice the type :)
```suggestion
 type AuthRequest = t.TypeOf<typeof AuthRequest>
```

packages/slice-machine/server/src/api/auth/post.ts
```diff
@@ -0,0 +1,34 @@
...
+
+type PostAuthResponse = {
+  err?: Error;
+};
+
+export default async function handler(authRequest: {
```
Heyrows: Io-Ts usually takes an object, a record or an unknown to decode.

Here I'm guessing you're casting the shape of the authRequest but it seems that here you know already it's an AuthRequest.

I would suggest:
- if you know the type of auth request for sure, like you call the function yourself, then don't use io-ts because you already have you data
- If you don't know and it's casting, then put authRequest as Record<string, unknown>, that would be more accurate to the situation.

packages/slice-machine/server/src/api/auth/post.ts
```diff
@@ -0,0 +1,34 @@
...
+  email: string;
+  cookies: string;
+}): Promise<PostAuthResponse> {
+  try {
+    const authPayload = fold(
```
Heyrows: Here I would just return the fold without a try.
In case of error, you can send back the object { err }
In case of success you can setTheAuthConfigCookie and then return the error

I would also set the base here.
If someone had previously logged in stage and now use the prod, this wouldn't work.

packages/slice-machine/server/src/api/auth/start.ts
```diff
@@ -0,0 +1,17 @@
+import { FileSystem } from "slicemachine-core";
+
+type StartResponse = {
+  err?: Error;
+};
+
+export default async function handler(): Promise<StartResponse> {
+  try {
+    // Reset the prismic auth cookie
```
Heyrows, line 9: I would not set just the cookie, the base might not be right.

Let's say you were connected on stage and now you're playing with the prod, the UI will detect you're not connected but this will only reset the cookies, not the base.

packages/slice-machine/server/src/api/custom-types/push.ts
```diff
@@ -28,13 +28,25 @@ export default async function handler(query: { id: string }) {
...
 
-  if (state.clientError || state.isFake) {
+  // When the user is connected but the token expired
+  if (state.clientError) {
     const message =
       "[custom-types/push] Could not fetch remote custom types. Are you logged in to Prismic?";
     return {
       err: new Error(message),
       reason: message,
-      status: 500,
```
Heyrows, line 37: I would put back the code there was with
```
 status: state.clientError ? state.clientError.status : 403
```

Should it be 403 tho ?
No connected usually means 401

packages/slice-machine/server/src/api/index.ts
```diff
@@ -151,7 +154,7 @@ router.get(
   async function (req: express.Request, res: express.Response) {
     const payload = await pushCustomType(req.query);
     if (payload.err) {
-      return res.status(400).json(payload);
+      return res.status(payload.status).json(payload);
```
Heyrows, line 157: I did not check all the files yet, but if this send a 401 and not a 400 which is. good, I hope the front will still react the same way

packages/slice-machine/server/src/api/index.ts
```diff
@@ -168,6 +171,36 @@ router.get(
   }
 );
 
+router.post(
+  "/auth/start",
+  async function (_req: express.Request, res: express.Response) {
+    const payload = await startAuth();
+    if (payload.err) {
+      return res.status(400).json({});
```
Heyrows: i guess it was used like that before but i don't like the idea of sending a 400 when we have a potential error.

packages/slice-machine/server/src/api/index.ts
```diff
@@ -168,6 +171,36 @@ router.get(
...
+    }
+    return res.status(200).json({});
+  }
+);
+
+router.post(
+  "/auth/status",
+  async function (_req: express.Request, res: express.Response) {
+    const payload = await statusAuth();
+    return res.status(200).json(payload);
```
Heyrows, line 192: I would return an error if the status was an error.

packages/slice-machine/server/src/api/slices/push.ts
```diff
@@ -116,9 +116,23 @@ export default async function apiHandler(query: {
   const { sliceName, from } = query;
   const { env } = await getEnv();
   const { slices, err } = await getSlices(env.client);
+
+  if (env.client.isFake()) {
+    console.error(
+      "[slice/push] An error occurred while fetching slices.\nCheck that you're properly logged in and that you have access to the repo."
+    );
+    const message =
+      "[slice/push] Could not fetch remote custom types. Are you logged in to Prismic?";
```
Heyrows: remote slices*

packages/slice-machine/server/src/api/slices/push.ts
```diff
@@ -116,9 +116,23 @@ export default async function apiHandler(query: {
...
+    return {
```
Heyrows, line 124: I would use the OnError function that was made and used in this same file.

packages/slice-machine/server/src/api/slices/push.ts
```diff
@@ -116,9 +116,23 @@ export default async function apiHandler(query: {
   const { sliceName, from } = query;
   const { env } = await getEnv();
   const { slices, err } = await getSlices(env.client);
+
+  if (env.client.isFake()) {
```
Heyrows, line 120: I'm not sure to see why we need a specific case for the fake env.

What's a fake env ?
I can't think of anything but the dev environment

packages/slice-machine/server/src/index.ts
```diff
@@ -33,6 +35,7 @@ const api = require("./api");
 const app = express();
 app.use(bodyParser.json({ limit: "64mb" }));
 app.use(bodyParser.urlencoded({ extended: true }));
+app.use(cors());
```
Heyrows: I would use Cors as the first middleware of the app, so we don't use the bordy parser for nothing :)

Heyrows: It's a good point to have added that.

packages/slicemachine-core/src/filesystem/authConfig.ts
```diff
@@ -73,25 +73,53 @@ export function removeAuthConfig(directory?: string): void {
...
 ): void {
   const { cookies: currentCookies } = getOrCreateAuthConfig(directory);
-  const oldCookies = cookie.parse(currentCookies || "");
+  const mergedCookie = mergeCookies(cookies, currentCookies);
+
+  return updateAuthConfig({ base, cookies: mergedCookie });
+}
 
-  const newCookies = cookies
+export function setAuthConfigCookies(
```
Heyrows: I think this function shouldn't exist, it was wrong of me to use the name cookies

packages/slicemachine-core/src/filesystem/authConfig.ts
```diff
@@ -73,25 +73,53 @@ export function removeAuthConfig(directory?: string): void {
  * @param {string} [cookies] - the list of new cookies
  * @returns {void} nothing
  */
-export function setAuthConfigCookies(
+export function setAuthConfig(
   base: string,
```
Heyrows: maybe we could send an authConfig instead of base + cookie directly.

Heyrows: or maybe event a Partial AuthConfig

packages/sm-init/package.json
```diff
@@ -8,12 +8,13 @@
   "scripts": {
     "build": "tsc --project tsconfig.json",
     "watch": "tsc --project tsconfig.json --watch",
-    "prepublishOnly": "yarn run build",
+    "prepublishOnly": "npm run build",
     "lint": "eslint --ext .ts .",
+    "lint:precommit": "eslint",
+    "audit": "better-npm-audit audit -l high -p",
```
Heyrows, line 14: I noticed that all subpackages don't have an audit fix, it might be worth adding one to use lerna to fix everything quickly

packages/slice-machine/server/src/api/auth/status.ts
```diff
@@ -0,0 +1,20 @@
...
+import { parseAuth } from "@lib/auth";
+
+type StatusResponse = {
+  status: "error" | "ok" | "pending";
+};
+
+export default async function handler(): Promise<StatusResponse> {
+  try {
+    const authConfig = FileSystem.getOrCreateAuthConfig();
+    const authResult = parseAuth(authConfig.cookies);
```
Heyrows: We have a cookie Module inside the Core, i think this function should be on it.

Heyrows (changes requested):

package.json
Miniplop: oh, that lerna troll

Miniplop: I will fix it

package.json
Miniplop: Look at 2 lignes above :p

packages/next-slicezone/package.json
Miniplop: in npm you can call a module directly

packages/slice-machine/server/src/index.ts
Miniplop: I don't know if the use order change the priority order for the middleware, but I don't think so. Anyway I can change it, if it make it clearer

packages/sm-init/package.json
Miniplop: good point 👍

packages/slice-machine/server/src/api/auth/status.ts
Miniplop: good catch

Makefile
Miniplop: Nop it's a file for creating make command to ease the installation process and other painfull processes

package.json
Miniplop: Done

packages/slice-machine/server/src/index.ts
Miniplop: Done

packages/slice-machine/components/LoginModal/index.tsx
Miniplop: I can change it yes

packages/slice-machine/components/LoginModal/index.tsx
Miniplop: The icon was deleted

packages/slice-machine/components/LoginModal/index.tsx
Miniplop: The wording was changed

packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header.tsx
Miniplop: done in this case simbolise that the request ended, but the response can be an error

packages/slice-machine/lib/builders/SliceBuilder/index.jsx
Miniplop: I need to check that but I think there is a confusion in the API around the status code

packages/slice-machine/package.json
Miniplop: As we run the linter only on file staged the max warnings flag is not useful. It will be ok except if you change all the files in one commit 🤯.

I think the issue is more  that the lint:precommit command won't crash because every linter errors is now cast to warnings in this package :)

packages/slice-machine/pages/_app.tsx
Miniplop: Yes 💪

packages/slice-machine/pages/cts/[ct].tsx
Comment (Miniplop) [reply packages/slice-machine/pages/cts/[ct].tsx:18]: ♥️

packages/slice-machine/server/src/api/auth/post.ts
Miniplop: Oh yeah 🤟

packages/slice-machine/server/src/api/auth/post.ts
Miniplop: Ohhh love that

.husky/pre-commit
Miniplop: Good point I follow the lint-staged but we can improve this :)

Miniplop: Done

packages/slice-machine/components/LoginModal/index.tsx
Miniplop: Yes I think that should be changed when we will merge the branch 🧙‍♂️

packages/slice-machine/components/LoginModal/index.tsx
Miniplop: Yes good point I just start this by creating a apiClient module

packages/slice-machine/components/LoginModal/index.tsx
Miniplop: I think this is a little bit brutal to close the dashboard success login page without asking :p

packages/slice-machine/components/LoginModal/index.tsx
Miniplop: I tested and tested the polling continue if you navigate :) You think this is a problem ?

packages/slice-machine/components/LoginModal/index.tsx
Miniplop: I change it to "logged in" => I take this wording from the dashboard :)

packages/slice-machine/lib/builders/SliceBuilder/index.jsx
Miniplop: For push slice : 
- You have a 403 if don't have any credentials in your ~/.prismic file or if your token has expired
- You have a 401 if don't have the right to access to the repository

Same for the custom types. I notice this is the opposite of what 401 and 403 is :)

So for me we should display the login only on 403 and not on 401 where we should display a error message to ask for changing to repo. Are you ok with that ?

packages/slice-machine/server/src/api/auth/post.ts
Miniplop: Yeah I think Record<string, unknown> is more accurate

packages/slice-machine/server/src/api/auth/post.ts
Miniplop: Fixed :)

packages/slice-machine/server/src/api/slices/push.ts
Miniplop: fixed :)

packages/slice-machine/server/src/api/slices/push.ts
Miniplop: Fake env is when the user don't have any token in the prismic file

Miniplop: I don't like this concept, I found it very confusing and not scalable as the fake client should always follow the same interface that the Default client. I think the case where the user don't has a correct prismic file should be handle by an another solution.

packages/slice-machine/server/src/api/index.ts
Miniplop: Yep I send now a 500

packages/slice-machine/server/src/api/index.ts
Miniplop: Yeah good one :)

packages/slice-machine/server/src/api/index.ts
Miniplop: There isn't any condition on the fronts on error status (except login now)

packages/slice-machine/lib/builders/SliceBuilder/index.jsx
Miniplop: I don't like this callback but it seems the interface need to have information even if there is an error

packages/slice-machine/pages/[lib]/[sliceName]/[variation].tsx
Comment (Miniplop) [reply packages/slice-machine/pages/[lib]/[sliceName]/[variation].tsx]: Done :)

packages/slice-machine/server/src/api/custom-types/push.ts
Miniplop: I change this to match the slice route behavior : 
- 401 => Mean you are not logged in
- 403 => Mean you don't have the good right

packages/slice-machine/server/src/api/slices/push.ts
Miniplop: We cannot use this function, because the status of the error is coming from the error itself. 

`export const onError = (
  r: Response | FakeResponse | null,
  message = "Unspecified error occurred."
) => ({
  err: r || new Error(message),
  status: r && r.status ? r.status : 500,
  reason: message,
})`

At the moment, we don't have any tool for creating http error with a status as the `http-errors` lib. And that seens a little bit overkill to do so. We can also add a new param `status` to this function. What do you think ?

packages/slice-machine/components/LoginModal/index.tsx
Heyrows: Ok

packages/slicemachine-core/src/filesystem/authConfig.ts
Miniplop: Fixed :)

packages/slicemachine-core/src/filesystem/authConfig.ts
Miniplop: I used optional param to do so

packages/slice-machine/server/src/api/auth/start.ts
Miniplop: As we saw the SM server use for now this information for calling the right endpoint. We will change this behavior when using SM.json repo as the reference for the base

packages/slice-machine/server/src/api/slices/push.ts
Miniplop: Fixed

packages/slice-machine/lib/auth.ts
```diff
@@ -69,10 +37,18 @@ export function getPrismicData(): Result<PrismicData, ErrorWithStatus> {
...
     .map<PrismicData>((prismicFile) => {
-      const authResult = parseAuth(prismicFile.cookies);
-      if (authResult.isOk())
-        return { base: prismicFile.base, auth: authResult.value };
-      else return { base: prismicFile.base, authError: authResult.error };
+      const authResult = Utils.cookie.parsePrismicAuthToken(
+        prismicFile.cookies
+      );
+      if (!!authResult)
+        return { base: prismicFile.base, auth: { auth: authResult } };
```
Heyrows: are you sure about this ?
Does that means we are calling `object.auth.auth` ?

packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header.tsx
```diff
@@ -45,9 +47,13 @@ const SliceHeader = ({
...
+                return;
+              }
+              setIsLoading(false);
+              handleRemoteResponse(addToast)(data);
```
Heyrows, line 54: i'm still not understanding totally why we have to call handlers if the response can be an error.
Shouldn't all error case be handled here ? (might not be due to your development but i'm asking the question again).

packages/slice-machine/pages/[lib]/[sliceName]/[variation].tsx
```diff
@@ -0,0 +1,13 @@
+import Builder from "@builders/SliceBuilder";
+
+interface SliceEditorProps {
+  openPanel: (priority?: any) => void;
```
Comment (Heyrows) [inline packages/slice-machine/pages/[lib]/[sliceName]/[variation].tsx:4]: ouch, that's an any.

packages/slice-machine/server/src/api/custom-types/push.ts
```diff
@@ -29,12 +29,23 @@ export default async function handler(query: { id: string }) {
...
   if (state.clientError || state.isFake) {
-    const message =
-      "[custom-types/push] Could not fetch remote custom types. Are you logged in to Prismic?";
+    const isAnAuthenticationError =
+      state.isFake || (state.clientError && state.clientError.status === 403);
+    const errorExplanation = isAnAuthenticationError
+      ? "Please log in to Prismic!"
+      : `You don\'t have access to the repo \"${state.env.repo}\"`;
+
+    const errorCode = state.isFake
```
Heyrows, line 38: I think here we could do `errorCode = state.clientError?.status || 403`

packages/slicemachine-core/__tests__/src/utils/cookie.test.ts
```diff
@@ -0,0 +1,48 @@
...
+import { parsePrismicAuthToken, serialize } from "../../../src/utils/cookie";
+
+describe("cookie", () => {
+  afterAll(() => {
+    jest.clearAllMocks();
+    return;
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
```
Heyrows, line 10: If you have an afterEach, an afterAll is not needed :)

packages/slicemachine-core/src/utils/cookie.ts
```diff
@@ -1,11 +1,18 @@
-import * as cookie from 'cookie'
+import * as cookie from "cookie";
 
-const noEscape = (str: string) => str
+const noEscape = (str: string) => str;
 
-export function parse(cookieString: string, opts?: cookie.CookieParseOptions): { [key: string]: string } {
-  return cookie.parse(cookieString, {decode: noEscape, ...opts})
+const AUTH_KEY = "prismic-auth";
```
Heyrows, line 5: this could / should be in the const file or the utils

packages/slicemachine-core/src/filesystem/authConfig.ts
```diff
@@ -23,75 +18,50 @@ export function createDefaultAuthConfig(directory?: string): AuthConfig {
...
   directory?: string
 ): void {
+  const currentConfig = getOrCreateAuthConfig(directory);
+  const formattedCookies = formatCookies(cookies);
   const configPath = PrismicConfigPath(directory);
-  const oldConfig = getOrCreateAuthConfig(directory);
+
+  const newConfig = {
+    cookies: formattedCookies,
+    base: !!base ? base : currentConfig.base,
```
Heyrows: Base is a `string | undefined`, you can just have
```suggestion
    base: base ? base : currentConfig.base,
```

or 

```suggestion
    base: base || currentConfig.base,
```

packages/slicemachine-core/src/filesystem/authConfig.ts
```diff
@@ -23,75 +18,50 @@ export function createDefaultAuthConfig(directory?: string): AuthConfig {
...
-export function setAuthConfigCookies(
-  base: string,
-  cookies: ReadonlyArray<string> = [],
-  directory?: string
-): void {
-  const { cookies: currentCookies } = getOrCreateAuthConfig(directory);
-  const oldCookies = cookie.parse(currentCookies || "");
-
-  const newCookies = cookies
+function formatCookies(cookies: ReadonlyArray<string>) {
```
Heyrows: no type of return.

I'm not sure but shouldn't this be in the Cookie file ?

Heyrows (changes requested):

packages/slice-machine/lib/auth.ts
Miniplop: unfornately yes :/

packages/slice-machine/pages/[lib]/[sliceName]/[variation].tsx
Comment (Miniplop) [reply packages/slice-machine/pages/[lib]/[sliceName]/[variation].tsx:4]: oui :/

Comment (Miniplop) [reply packages/slice-machine/pages/[lib]/[sliceName]/[variation].tsx:4]: openPanel is coming from the _app.tsx with this typing : 

```
const openPanel = useCallback(
    (priority?: any) =>
      setDrawerState({
        ...drawerState,
        open: true,
        ...(priority ? { priority } : null),
      }),
    []
  );
```

Miniplop:
100 ! 🤟

Heyrows (approved):
disruption

Miniplop:
Loading state with the manual link 

image

packages/slice-machine/components/LoginModal/index.tsx
```diff
@@ -118,16 +128,28 @@ const LoginModal: React.FunctionComponent<LoginModalProps> = ({
...
             }}
           >
-            Your session has expired.
-            <br />
-            Please log in again.
+            {isLoading ? (
+              <>
+                Not seeing the browser tab? <br />
+                <Link target={"_blank"} href={loginRedirectUrl}>
+                  Go back and try again
```
Heyrows, line 135: i'm not an expert in design so what i'm saying might be complete shit. 😄 
But usually I see something more like `Didn't the tab open automatically ? <link>Open it manually</link>`

Heyrows (approved):

PR #194, opened 2021-10-20
AAUser, I want to launch SM on local host 9999 when I run SM start

Miniplop:
## Types of changes

## Checklist:


Miniplop:
The CI is failing because of yarn link not working with the audit, we should merge the PR on login before this one

MarcMcIntosh:
Why would it work on other branches but fail on this one?

Miniplop:
Because in your branches you didn't have the slicemachine core installed in the sm ui which make the audit fail. And in the login branch I migrate to npm to solve this problem :)

Miniplop:
I installed the sm in a project with node v12.22.7, works fine. I also noticed with node 12 with have issue to run the local env due to next platform requirement.

packages/slice-machine/lib/models/common/http/DefaultClient.ts
```diff
@@ -23,7 +23,7 @@ const AuthApi = {
 
 const TrackingApi = {
   STAGE: "https://2p29q0kam4.execute-api.us-east-1.amazonaws.com/stage/",
-  PROD: "https://2p29q0kam4.execute-api.us-east-1.amazonaws.com/stage/",
+  PROD: "https://tracking.prismic.io",
```
Heyrows: stupid question probably but has the tracking repository domain be reserved ? (PR on wroom).
Because if the repo already exists, we cannot use that address :D

packages/slicemachine-core/src/filesystem/manifest.ts
```diff
@@ -1,10 +1,10 @@
 import { Framework, Files } from "../utils";
-import { FileContent, SMConfigPath } from './paths';
+import { FileContent, SMConfigPath } from "./paths";
 
 export interface Manifest {
   apiEndpoint: string;
   storybook?: string;
-  framework?: Framework;
+  framework?: Framework.Framework;
```
Heyrows: feels like it's prettier that wrote that, but it's giving me chills 😂

packages/slicemachine-core/src/utils/index.ts
```diff
@@ -1,5 +1,5 @@
 export { default as Files } from "./files";
-export { Framework } from "./framework";
+export * as Framework from "./framework";
```
Heyrows, line 2: I usually love to have a module, but here to avoid `Framework.Framework` or `Framework.detectFramework` I would `export * from './framework'`

packages/slice-machine/start/index.js
```diff
@@ -79,11 +78,11 @@ function start({ cwd, port }, callback) {
...
   });
 
-  start.stderr.on("data", function (data) {
+  smServer.stderr.on("data", function (data) {
     console.log("[slice-machine] " + data.toString());
   });
 
-  start.on("exit", function (code) {
+  smServer.on("exit", function (code) {
     console.log("[slice-machine] Thanks for using SliceMachine");
```
Heyrows: never actually saw that one because I usually stop with `Ctrl+C`
I think it would be a nice touch to have it on `Ctrl+C` too :)
https://stackoverflow.com/questions/14031763/doing-a-cleanup-action-just-before-node-js-exits

packages/slice-machine/start/index.js
```diff
@@ -147,9 +141,9 @@ async function run() {
...
 
   const nodeVersion = process.version.slice(1).split(".")[0];
-  if (parseInt(nodeVersion) < 15) {
+  if (parseInt(nodeVersion) < 12) {
     console.error(
-      `\n🔴 Slicemachine requires node version >= 15 to work properly.\nCurrent version: ${process.version}\n`
+      `\n🔴 Slicemachine requires node version >= 12 to work properly.\nCurrent version: ${process.version}\n`
     );
     process.exit(-1);
   }
```
Heyrows: This comment is for 2 lines below the `const userConfig = handleManifest(cwd);`

I know it's not you but Man the handleManifest function is imported from the build directly T_T
I think the function that gives a state about the manifest should be in the core.

packages/slice-machine/start/index.js
```diff
@@ -109,14 +108,9 @@ See below for more info 👇`,
...
     case ManifestStates.NotFound: {
-      console.log(`Slicemachine requires an "sm.json" config file, at the root of your project.
-      
-Required properties:
-* apiEndpoint, eg. "https://repo.prismic.io/api/v2"
-* libraries, eg. ["~/slices"]\n\n`);
-
-      const exit = await createManifest(cwd);
-      return { exit };
+      console.log(`Run "npx sm init" command to configure your project`);
```
Heyrows: The current message we have everywhere is `console.log(Run ${bold("npx slicemachine init")} again!);`
I guess we either need to have the command as a CONST so that we don't all write different things, and we need to have it bold here for you :).

Heyrows (changes requested):

packages/slice-machine/lib/models/common/http/DefaultClient.ts
Miniplop: The tracking service is already deployed on this address :)

packages/slice-machine/start/index.js
Miniplop: Can be a good idea yes

packages/slice-machine/start/index.js
Miniplop: Yeah this is normal as this file is not yet transpiled by webpack

packages/slicemachine-core/src/utils/index.ts
Miniplop: I think we should choose one way for accessing sub module and don't a mix  : 

- option 1 : we defined namespace for each
- option 2 : we use wildcard to have one big utils object

What is your opinion ? 

For the type I'm not comfortable of having the types spread into the submodules, because that also mean importing all the functions with it. I think we should have a type module for sharing types.

packages/slicemachine-core/src/filesystem/manifest.ts
Miniplop: As I said in the next comment, I think we should have something like Types.Framework or exposing types on the core itself

packages/slice-machine/lib/models/common/http/DefaultClient.ts
Heyrows: We have a system in Wroom that keep some domain from being taken, that's what I meant :)

packages/slicemachine-core/src/utils/index.ts
Heyrows: I like defined namespace for each, I prefer it that way actually but i know hugo was against.
I like the idea of having a type submodule, we could import it and use it like `Type.Framework`, or even import like `{ Framework } from type`.
But that would require making it :) and eventually taking all the type we currently have into it.

packages/slice-machine/start/index.js
Miniplop: Ctrl+C already works to shut down the server

packages/slice-machine/start/index.js
Heyrows: hmmmm Ok i see.

packages/slice-machine/start/index.js
Miniplop: As we removed the message this is not relevant now :)

packages/slice-machine/lib/models/common/http/DefaultClient.ts
Miniplop: So I imagine this domain is not taken as this endpoint is working :p 😂

packages/slice-machine/start/index.js
Miniplop: Done

packages/slicemachine-core/src/utils/index.ts
Miniplop: Done

packages/slice-machine/package.json
```diff
@@ -70,6 +70,7 @@
     "npmlog": "^5.0.0",
     "oembed-parser": "^1.4.5",
     "open": "^8.2.1",
+    "ora": "^6.0.1",
```
Heyrows, line 73: are you sure you need that ?
Because Utils.spinner is basically calling ora.
I think it's better to use the utils so if one day we change we need to change one place.

packages/slicemachine-core/src/utils/const.ts
```diff
@@ -1,6 +1,7 @@
-export const SCRIPT_NAME = 'slicemachine'
-export const SCRIPT_VALUE = 'start-slicemachine'
-export const SM_PACKAGE_NAME = 'slice-machine-ui'
-export const MANIFEST_FILE_NAME = 'sm.json'
-export const DEFAULT_SERVER_PORT = 5555
-export const DEFAULT_BASE = 'https://prismic.io'
\ No newline at end of file
+export const SCRIPT_NAME = "slicemachine";
+export const SCRIPT_VALUE = "start-slicemachine";
+export const INIT_COMMAND = "npx sm init";
```
Heyrows, line 3: I'm pretty sure this is being used in most of the error message of the init so we might have to verify before releasing that this is used everywhere.

packages/slicemachine-core/package.json
```diff
@@ -22,6 +22,9 @@
     "audit-fix": "npm audit fix",
     "test": "jest"
   },
+  "peerDependencies": {
+    "ora": "^5.4.1"
```
Heyrows, line 26: it's already a dependency, why does it need to be a peer dependency ?

packages/slicemachine-core/src/utils/const.ts
Miniplop: mmmm I don't think it in the sm-init, maybe this is added by the ls-repos branch or non merged branch no ?

packages/slicemachine-core/package.json
Miniplop: Because I want that when you install the core you install ora too

Miniplop: But this is not compatible with lerna, so we can remove this

packages/slice-machine/package.json
Miniplop: Because we need ora to import utils in the start

Heyrows (approved):

Heyrows (approved):

PR #197, opened 2021-10-25
Keep the library helper synchronous

arnaudlewis:
This code has nothing asynchronous so we just remove any piece of code that converts it to async for no reason


Miniplop:
This function was removed from the @slicemachine/core in the release V0.1.1. I close this PR. 

cc @arnaudlewis

PR #199, opened 2021-10-26
AAuser, I want SM to be already configured after downloading SM in terminal

Heyrows:
## Types of changes

- [X] New feature (a non-breaking change which adds functionality)

## Checklist:

- [X] All new and existing tests are passing.

Screenshot of what it gives before the merge with an updated version.
image


packages/sm-init/src/index.ts
```diff
@@ -7,13 +7,16 @@ import {
...
+  configureProject,
+  displayFinalMessage,
   detectFramework,
 } from "./steps";
 import { findArgument } from "./utils";
 
 async function init() {
   const cwd = findArgument(process.argv, "cwd") || process.cwd();
-  const base = findArgument(process.argv, "base") || Utils.CONSTS.DEFAULT_BASE;
+  const base = (findArgument(process.argv, "base") ||
```
MarcMcIntosh, line 18: No io-ts?

packages/sm-init/sm.json
```diff
@@ -0,0 +1,3 @@
+{
```
MarcMcIntosh: How did this happen?

packages/slicemachine-core/src/utils/endpoints.ts
```diff
@@ -17,3 +17,19 @@ export function buildEndpoints(base: string): Endpoints {
     },
   };
 }
+
+export type Base = `${"http" | "https"}://${string}${"/" | ""}`;
```
MarcMcIntosh, line 21: i didn't know patterns could be typed

MarcMcIntosh (changes requested):
Few changes requested

packages/sm-init/sm.json
Heyrows: hmmm that's a good question, i'm wondering if the test was not mocked at some point, that's a good catch

Heyrows: I'll remove it and re-run the test to check if it's created again

packages/slicemachine-core/src/utils/endpoints.ts
Heyrows: Me neither, discovered it here :) thought it would be easier to understand what's a base and what's an endpoint.

packages/sm-init/src/index.ts
Heyrows: That's interesting, I always thought of Io-Ts as a lib to validate an object to have the right properties.
Here we have an argument that is potentially not there, how do you think it could be validated ?

Because I guess there is no data on which we can actually make a validation on a specific type, we have an array that can be empty.

So what would be the data structure for you ?

MarcMcIntosh: argv is io isn't it?

:trollface:

packages/slicemachine-core/src/utils/endpoints.ts
MarcMcIntosh: neat

MarcMcIntosh (approved):

PR #207, opened 2021-11-02
AADev, when A user send request to the CT API, I send "slice-machine" as "User-Agent" header in the request

MarcMcIntosh:
Conditionally set the user agent in fetch, and fix some linter warnings

## Types of changes

## Checklist:


packages/slice-machine/lib/models/common/http/DefaultClient.ts
```diff
@@ -42,26 +47,28 @@ function createApiUrl(base: string, { STAGE, PROD }: ApiSettings): string {
...
-    body?: object | string,
+    body?: Body,
     action = "",
     method = "get"
   ): Promise<Response> {
     const headers = {
       repository: repo,
       Authorization: `Bearer ${auth}`,
-      UserAgent: "slice-machine",
+      ...(typeof window !== undefined ? { "User-Agent": "slice-machine" } : {}),
```
Heyrows: the key is much better now 👍.
If I'm not mistaken, this code will be executed only on the server, not in the browser.
So this condition will always be false isn't it ?

Why do we need to differentiate cases, in which case do we want to send the information and in which case do we not ?

packages/slice-machine/lib/models/common/http/DefaultClient.ts
```diff
@@ -42,26 +47,28 @@ function createApiUrl(base: string, { STAGE, PROD }: ApiSettings): string {
   return PROD;
 }
 
+type Body = Slice<AsObject> | Record<string, unknown> | string;
```
Heyrows, line 47: I'm guessing it was for the linter that you changed those stuff 👍 😄 
Isn't `Slice<AsObject>` already included in `Record<string, unknown>`, why do you need to specify for this type ?

packages/slice-machine/lib/models/common/http/DefaultClient.ts
```diff
@@ -90,30 +97,30 @@ const initFetcher = (
...
 
-  static validate(base: string, auth: string) {
+  static validate(base: string, auth: string): Promise<Response> {
     return fetch(
       `${createApiUrl(base, AuthApi)}${ValidatePrefix}?token=${auth}`,
       {
         method: "GET",
       }
-    ).catch((e) => e);
+    ); // .catch((e) => e);
```
Heyrows: I think your forgot to remove your comments 😄

packages/slice-machine/lib/models/common/http/DefaultClient.ts
```diff
@@ -122,12 +129,15 @@ export default class DefaultClient {
...
     readonly auth: string
   ) {
-    const devConfig = (() => {
+    const devConfig: Record<string, DevConfig> = (() => {
       if (!cwd) {
         return {};
       }
       try {
-        return Files.readJson(path.join(cwd, "sm.dev.json"));
+        return Files.readJson(path.join(cwd, "sm.dev.json")) as Record<
```
Heyrows, line 134: :o never heard of an sm.dev.json, do you know what it does ?

packages/slice-machine/lib/models/common/http/DefaultClient.ts
```diff
@@ -158,52 +168,52 @@ export default class DefaultClient {
...
       return this.aclFetcher("", undefined, "create", "get");
     },
-    deleteFolder: async (body: object | string) => {
+    deleteFolder: async (body: Body): Promise<Response> => {
       return this.aclFetcher("", body, "delete-folder", "post");
     },
     post: async (params: {
       url: string;
-      fields: { [key: string]: string };
+      fields: Record<string, string>;
```
Heyrows, line 209: very nice 👍

Heyrows (commented):
I took the liberty of changing the PR name to the user story as discussed this morning :).

packages/slice-machine/lib/models/common/http/DefaultClient.ts
MarcMcIntosh: Apparently not, it seems to be an array of some sort

packages/slice-machine/lib/models/common/http/DefaultClient.ts
MarcMcIntosh: not a clue

packages/slice-machine/lib/models/common/http/DefaultClient.ts
MarcMcIntosh: I'm not 100% if it is or isn't used on the client side, since the code base looks like it's isomorphic as global fetch is defined in `server/src/index.ts` when/if the code is used client side it will be true

hypervillain: this is called solely server-side

packages/slice-machine/lib/models/common/http/DefaultClient.ts
Heyrows: hmmm ok, weird because everywhere it is used, the type use to be `object | string` which seems coherent.

packages/slice-machine/lib/models/common/http/DefaultClient.ts
hypervillain: this might be how @arnaudlewis or @Miniplop have solved the issue of using Slicemachine with a local instance of wroom

packages/slice-machine/lib/models/common/http/DefaultClient.ts
Heyrows: thanks hugo, then I guess we don't need to check if the window exists.

packages/slice-machine/lib/models/common/http/DefaultClient.ts
MarcMcIntosh: Yes object causes linter warnings

packages/slice-machine/lib/models/common/http/DefaultClient.ts
MarcMcIntosh: done

Heyrows (approved):

PR #208, opened 2021-11-02
Kje/add default librairies

Miniplop:
## Types of changes

## Checklist:

### DIFF
--- a/packages/slicemachine-core/src/filesystem/manifest.ts
+++ b/packages/slicemachine-core/src/filesystem/manifest.ts
@@ -4,6 +4,7 @@ import { FileContent, SMConfigPath } from "./paths";
 export interface Manifest {
   apiEndpoint: Endpoints.ApiEndpoint;
   storybook?: string;
+  libraries?: string[];
   framework?: Framework.FrameworkEnum;
   chromaticAppId?: string;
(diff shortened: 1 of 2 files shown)


packages/slicemachine-core/src/filesystem/manifest.ts
```diff
@@ -4,6 +4,7 @@ import { FileContent, SMConfigPath } from "./paths";
 export interface Manifest {
   apiEndpoint: Endpoints.ApiEndpoint;
   storybook?: string;
+  libraries?: string[];
```
Heyrows, line 7: to go a little further, do you think the manifest file should be more like the AuthConfig ?
Having function such as "createOrDefault"

Heyrows (approved):

packages/slicemachine-core/src/filesystem/manifest.ts
Heyrows: Also, i'm pretty sure I took this model from the UI, so there might be a wrong model on the "env" part of the ui

Miniplop: As we don't update this file after creating it for now  at least this is less relevant I think

PR #212, opened 2021-11-03
Display onboarding intro screen

hypervillain:
Display introduction screen in onboarding:

87789d859c78cb0a5229a7b3318ae2e6

I've updated the steps in the onboard to prevent videos from breaking the design on load.

Linear Ticket: https://linear.app/prismic/issue/SM-296/aauser-in-the-onboarding-i-see-a-intro-screen


packages/slice-machine/pages/onboarding.tsx
```diff
@@ -93,7 +92,10 @@ const PushPagesSlide = () => (
 const OnboardingGrid = ({
   children,
 }: {
-  children: ReactElement | ReadonlyArray<ReactElement>;
+  children:
```
Miniplop: If you want to type functional component, you can use React.FunctionComponent<Props> this type includes children as props :)

packages/slice-machine/pages/onboarding.tsx
```diff
@@ -198,17 +200,22 @@ export default function Onboarding(): JSX.Element {
...
+        <Flex
+          key={`step-${i + 1}`}
+          sx={{
+            gridArea: "content",
+            alignItems: "center",
+            justifyContent: "center",
+            alignContent: "center",
+            flexDirection: "column",
+            opacity: i === state.step ? "1" : "0",
+            zIndex: i === state.step ? "9999" : "-1",
```
Miniplop: Why z -index is needed if the element is not visible ?

Miniplop (commented):
I don't see the intro screen in this PR, maybe you forget to push some file no ? 

Also can you remove the cypress debug video from the PR ?

packages/slice-machine/pages/onboarding.tsx
hypervillain: As is, the element is invisible but still present in the DOM, z-index makes sure you click on the current element no matter the position of the others. There are probably other strategies (display: none) but I know this one would work if we wanted to implement CSS animations int he slideshow

Miniplop: Ok not very sexy but good enough

hypervillain: Yeah! I can look for something else

hypervillain: Quoting an article I just read: "display: none has been the favorite solution to hide elements for many years, but it’s been superseded by more flexible, animatable options. It’s still valid, but perhaps only when you want to permanently hide content from all users. transform or opacity are better choices when considering performance."

I can maybe change the zIndex to pointer-events: none if that seems more appropriate to you

Miniplop (approved):
good 🍾

packages/slice-machine/tests/project/cypress/integration/0-is-running/00-onboarding.spec.js
```diff
@@ -37,12 +37,11 @@ describe('onboarding', () => {
     })
     cy.get('[data-cy=get-started]').click()
 
-    const continueSelector = '[data-cy=continue]'
-
-    const clickContinue = () => cy.get(continueSelector).click()
-    clickContinue()
-    clickContinue()
-    clickContinue()
+    const clickContinue = () => cy.get('[data-cy=continue]').click()
```
MarcMcIntosh, line 40: maybe rename this tes or create another,
as when `is-onboarded` is in local-storage the user should see the first onboarding slide?

packages/slice-machine/tests/project/cypress/integration/0-is-running/00-onboarding.spec.js
```diff
@@ -37,12 +37,11 @@ describe('onboarding', () => {
...
+    
+    for (let i = 0; i <= 2; i++) {
```
MarcMcIntosh, line 42: I'm surprised this works :)

MarcMcIntosh (changes requested):
There's a few videos files and pictures added by cypress. 
Theses can be removed :)

maybe we should add a `.gitignore` in the videos directory ?

packages/slice-machine/tests/project/cypress/integration/0-is-running/00-onboarding.spec.js
hypervillain: No actually we don't want to have this logic, it's fine to see the first screen in any case

MarcMcIntosh (approved):

PR #215, opened 2021-11-04
AAPM I can track users who have seen/skip the Intro Onboarding

MarcMcIntosh:
## Types of changes

## Checklist:


packages/core/package.json
```diff
@@ -22,7 +22,7 @@
     "audit": "better-npm-audit audit -l high -p",
     "prepublishOnly": "npm run test && npm run build",
     "audit-fix": "npm audit fix",
-    "test": "jest"
+    "test": "jest --verbose"
```
Heyrows: I don't think verbose should be here all the time, it's something we want to use only for debug, not when all the tests passes.

packages/slice-machine/jest.config.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+  silent: false,
```
Heyrows: Why would you add that, it would prevent a developer from logging stuff ?

packages/slice-machine/lib/env/index.ts
```diff
@@ -111,7 +111,7 @@ function parseStorybookConfiguration(cwd: string) {
 export async function getEnv(
   maybeCustomCwd?: string
 ): Promise<{ errors?: { [errorKey: string]: ServerError }; env: Environment }> {
-  const cwd = maybeCustomCwd || process.env.CWD;
+  const cwd = maybeCustomCwd || process.env.CWD || process.cwd();
```
Heyrows, line 114: In which case do you need that last option ?

packages/slice-machine/lib/models/common/http/DefaultClient.ts
```diff
@@ -197,6 +199,10 @@ export default class DefaultClient {
     return this.trackingFetcher("", review, "", "post");
   }
 
+  async sendOnboarding(data: OnboardingTrackingEvent) {
+    return this.trackingFetcher("", data, "", "post");
```
Heyrows, line 203: OPT : Wondering if this could be improved because the first and third parameters are always empty string and it's always a post.

packages/slice-machine/lib/models/common/TrackingEvent.ts
```diff
@@ -12,3 +13,12 @@ export type ReviewTrackingEvent = TrackingEvent & {
...
   comment: string;
 };
+
+export type OnboardingTrackingEvent = TrackingEvent & {
+  id: TrackingEventId.ONBOARDING;
+  lastStep: number;
+  maxSteps: number;
+  startTime: number;
+  endTime: number;
+  totalTime: number;
```
Heyrows: A question, how is the repository ID found by the CDP project ?
I saw that it's part of the Codec but couldn't figure out how it was sent.

packages/slice-machine/pages/onboarding.tsx
```diff
@@ -140,6 +140,40 @@ const StepIndicator = ({
...
+      const totalTime = endTime - startTime;
+      const data = {
+        lastStep: step,
+        maxSteps,
+        startTime,
+        endTime,
+        totalTime,
+      };
+
+      fetch("/tracking/onboarding", {
```
Heyrows: Isn't there a module with these API calls ? (I really don't know that's why i'm asking).

packages/slice-machine/pages/onboarding.tsx
```diff
@@ -140,6 +140,40 @@ const StepIndicator = ({
...
+  startTime: number;
+  maxSteps: number;
+}) {
+  const state = useRef(props);
+
+  useEffect(() => {
+    state.current = props;
+  }, [props]);
+
+  useEffect(
```
Heyrows: The Goal of this PR is to track the end of the onboarding.
Why did you make a useEffect that handles himself instead of adding something to the existing functions that skip / end the onboarding ?

packages/slice-machine/server/src/api/index.ts
```diff
@@ -62,6 +63,17 @@ router.post(
   }
 );
 
+router.post(
+  "/tracking/onboarding",
+  async (req: express.Request, res: express.Response) => {
+    return onboarding(req.body)
```
Heyrows: looking at how the other routes are made in this file, I think we should have the same.

packages/slice-machine/server/src/api/tracking/onboarding.ts
```diff
@@ -0,0 +1,34 @@
...
+export class HTTPResponseError extends Error {
+  response: Response | FakeResponse;
+  constructor(response: Response | FakeResponse) {
+    super(`HTTP Error Response: ${response.status} ${response.statusText}`);
+    this.response = response;
+  }
+}
+
+export default async function (
+  query: Pick<
```
Heyrows: Didn't know the Type Pick before, it's interesting.
Although if every Key can be optional, I would use `Partial<T>`

I'm wondering, why would the query object be `"Picked"` or `Partial` ?
The information are sent by the Client, you know the object structure so i'm wondering what's that about.

packages/slice-machine/pages/onboarding.tsx
Heyrows: I think useEffect are uncontrolled and not very stable compare to a skip button trigering once the call.

packages/core/package.json
MarcMcIntosh: I was trying to get more info in the out put from the ci to make sure the contract test had run with the right env vars

packages/slice-machine/lib/env/index.ts
MarcMcIntosh: in one of the tests

packages/slice-machine/lib/models/common/TrackingEvent.ts
MarcMcIntosh: those have been removed in favour o more events

MarcMcIntosh: reop-id is found by the authentication server

packages/slice-machine/pages/onboarding.tsx
MarcMcIntosh: tbh, I don't know if there is

packages/slice-machine/pages/onboarding.tsx
MarcMcIntosh: onunmount is called in more cases than pressing the skip button

they could close the screen or find another way to exit, like changing the url.

packages/slice-machine/server/src/api/tracking/onboarding.ts
MarcMcIntosh: it's been removed

packages/slice-machine/server/src/api/index.ts
MarcMcIntosh: you mean with `await`?

MarcMcIntosh: fine, but we'll lose some of the error handling in `packages/slice-machine/server/src/api/tracking/onboarding.ts` and also type checking of errors will be more tricky

packages/slice-machine/lib/models/common/http/DefaultClient.ts
MarcMcIntosh: yeah, I think the idea here was have some generic functions idk

packages/slice-machine/pages/onboarding.tsx
```diff
@@ -61,32 +68,32 @@ const WelcomeSlide = ({ onClick }: { onClick: () => void }) => (
     </Button>
   </>
 );
-const BuildSlicesSlide = () => (
+const BuildSlicesSlide = ({ onEnded }: { onEnded: () => void }) => (
```
Miniplop: Maybe you can create a SlicesSlideProps type for { onEnded: () => void } and using it in the 3 slides ?

packages/slice-machine/pages/onboarding.tsx
```diff
@@ -140,29 +147,126 @@ const StepIndicator = ({
...
+      return TrackingEventId.ONBOARDING_FIRST;
+    case 2:
+      return TrackingEventId.ONBOARDING_SECOND;
+    default:
+      return TrackingEventId.ONBOARDING_THIRD;
+  }
+}
+
+function postTracking(
+  data:
```
Miniplop: we should avoid using generic name as data, stuff, things, etc when we can, I was thinking about "onboardingEvent" to be more specific, what do you think about it ?

packages/slice-machine/tests/onboarding.spec.js
```diff
@@ -0,0 +1,51 @@
+import nock from "nock";
+import fs from "fs";
+import onboarding from "../server/src/api/tracking/onboarding";
+import { name } from "../package.json";
+import { TrackingEventId } from "../lib/models/common/TrackingEvent";
+import fetch from "node-fetch";
```
Miniplop, line 6: I found it weird that node-fetch is not in our package.json and that you can use it like this ? Do you know how this is possible ?

packages/slice-machine/tests/project/cypress/integration/0-is-running/00-onboarding.spec.js
```diff
@@ -15,18 +15,37 @@ describe('onboarding', () => {
   });
 
   it('should redirect to /onboarding when is-onboared is not in local storage', () => {
+    cy.intercept('POST', '/tracking/onboarding', ({body}) => {
```
Miniplop: good :)

packages/slice-machine/tests/project/cypress/integration/0-is-running/00-onboarding.spec.js
```diff
@@ -15,18 +15,37 @@ describe('onboarding', () => {
...
+
+    const ids = [
+      "slicemachine_onboarding_start",
+      "slicemachine_onboarding_continue_screen_intro",
+      "slicemachine_onboarding_continue_1",
+      "slicemachine_onboarding_continue_2",
+      "slicemachine_onboarding_continue_3",
+    ]
+    cy.intercept('POST', '/tracking/onboarding', ({body}) => {
+      expect(body.id).equal(ids.shift())
```
Miniplop: 👍

packages/slice-machine/tests/project/cypress/integration/0-is-running/01-onboarding-skip.spec.js
```diff
@@ -0,0 +1,25 @@
...
+    cy.get('[data-cy=get-started]').click()
+  
+    cy.get('[data-cy=skip-onboarding]').click()
+  
+    cy.location('pathname', {timeout: 1000}).should('eq', '/')
+  
+    cy.getLocalStorage("is-onboarded").should('eq', 'true')
+
+  })
+})
```
Miniplop: new line here 😈

packages/slice-machine/types/cloudinary.d.ts
```diff
@@ -7,7 +7,11 @@ interface ImageProps extends JSX.IntrinsicElements.img {
   crop: string;
 }
 
-interface VideoProps extends JSX.IntrinsicElements.video {
+// interface VideoProps extends JSX.IntrinsicElements.video {
```
Miniplop: Can you remove this ? We avoid commented line in the code : https://refactoring.guru/fr/smells/dead-code

packages/slice-machine/tests/onboarding.spec.js
MarcMcIntosh: node-fetch should be in the package.json, it's used as a global in the server.

packages/slice-machine/types/cloudinary.d.ts
MarcMcIntosh: can do

packages/slice-machine/pages/onboarding.tsx
MarcMcIntosh: maybe, could call it something like `WithOnEnded`

PR #218, opened 2021-11-05
AAUser, during the init script, I install all framework dependencies the required dependencies for the slice machine ui

Miniplop:
## Types of changes

## Checklist:

### DIFF
--- a/packages/init/src/steps/install-required-dependencies.ts
+++ b/packages/init/src/steps/install-required-dependencies.ts
@@ -24,6 +24,9 @@ export async function installRequiredDependencies(
     case Utils.Framework.FrameworkEnum.next:
       `${installDependencyCommand} ${Utils.CONSTS.PRISMIC_REACT_PACKAGE_NAME}`;
       break;
+    case Utils.Framework.FrameworkEnum.svelte:
+      `${installDependencyCommand} ${Utils.CONSTS.PRISMIC_DOM_PACKAGE_NAME}`;
+      break;
   }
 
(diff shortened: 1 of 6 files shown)


packages/init/src/steps/install-required-dependencies.ts
```diff
@@ -2,17 +2,33 @@ import path from "path";
...
-    `${command} ${Utils.CONSTS.SM_PACKAGE_NAME}`
+    `${installDevDependencyCommand} ${Utils.CONSTS.SM_PACKAGE_NAME}`
   );
 
+  switch (framework) {
+    case Utils.Framework.FrameworkEnum.react:
+    case Utils.Framework.FrameworkEnum.next:
+      `${installDependencyCommand} ${Utils.CONSTS.PRISMIC_REACT_PACKAGE_NAME}`;
+      break;
+    case Utils.Framework.FrameworkEnum.svelte:
```
MarcMcIntosh, line 27: I guess later on we could include `prismic/client`, but configuring it could be tricky.

MarcMcIntosh, line 27: I'm trying to remember  if svelte need any extra deps to work, I'll add a not on the ticket to check this to make sure it works

MarcMcIntosh (approved):

packages/init/src/steps/install-required-dependencies.ts
MarcMcIntosh: Also what about the slice-zone

PR #219, opened 2021-11-05
[Preview Branch] Slice canvas init

arnaudlewis:
Generate a state in a json file each time a slice is saved to provide data for sliceCanvas


packages/slice-machine/server/src/api/common/SliceCanvas.ts
```diff
@@ -0,0 +1,112 @@
...
+
+import Files from "@lib/utils/files";
+import Environment from "@lib/models/common/Environment";
+import * as Core from '@slicemachine/core'
+
+function stateExists(cwd: string) {
+  return Files.exists(Core.FileSystem.SliceCanvasStatePath(cwd));
+}
+
+export async function generateState(env: Environment) {
```
hypervillain: async

arnaudlewis:
Once the linter will pass, I still need to update previewURL with relative paths is posix

.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-npm run lint-staged
+# npm run lint-staged
```
Miniplop: 😭

lihbr: Savage

lihbr: My POV, but git CLI hooks are a pain for outside contributors on public projects.

Someone fixing a bug/adding a feature should just have to worry about fixing the bug/adding the feature. Then the lint CI would eventually fail on the PR and maintainers can help to straighten that before merging because they're the ones who are familiar with their linting rules and know how to fix that quickly~

hypervillain:
Code related to Storybook you be taken care of

lihbr:
Hey there, just tested this PR for Slice Canvas-related work, everything's working correctly on my end!

Slice Canvas has been updated accordingly (@bapmrl let me know how the update goes on your end! I've updated the online demo):
- [@prismicio/slice-canvas-renderer@0.0.2](https://github.com/prismicio/slice-canvas/blob/master/packages/renderer/CHANGELOG.md)
- [@prismicio/slice-canvas-com@0.0.4](https://github.com/prismicio/slice-canvas/blob/master/packages/com/CHANGELOG.md)
- [@prismicio/slice-canvas-renderer-react@0.0.2](https://github.com/prismicio/slice-canvas/blob/master/packages/renderer-react/CHANGELOG.md) (cc @angeloashmore to update the Next library boilerplate)
- [@prismicio/slice-canvas-renderer-vue@0.0.3](https://github.com/prismicio/slice-canvas/blob/master/packages/renderer-vue/CHANGELOG.md)

Although I had some issues while using @slicemachine/core which aren't related to this PR (cc @hypervillain & @Miniplop I guess):
- Package does not export an ESM build, it only exports a CJS one, exporting an ESM build would ease the usage of this package by packages integrating it (having only CJS forces us to use default export, not named export for example);
- Package subpath exports aren't working great as-is, I added some minimal exports to help in my above commit: 380fd46
- Might sound like I'm complaining, but the two above made it difficult to use this package in a browser environment because you cannot simply import it this way there: `import Core from "@slicemachine/core"` as this will load parts of the package that are relying on Node built-in libraries that are not available once in the browser (`fs`, `path`, `child_process`, etc.)

I don't know if you have settled on a build system yet for libraries of the Slice Machine project, for our other open-source kits and Slice Canvas we've been using [siroc](https://github.com/unjs/siroc) for a while now with great success. In any case you can have a look at our template here: https://github.com/prismicio/prismic-typescript-template

Cheers!

packages/core/src/filesystem/authConfig.ts
```diff
@@ -25,7 +32,9 @@ export function getOrCreateAuthConfig(directory?: string): AuthConfig {
   const configPath = PrismicConfigPath(directory);
   if (!Files.exists(configPath)) return createDefaultAuthConfig(directory);
 
-  const conf = Files.readJson(configPath);
+  const conf = Files.safeReadEntity(configPath, (payload) => {
+    return getOrElseW(() => null)(AuthConfig.decode(payload));
```
Heyrows, line 36: I've watched [the doc ](https://gcanti.github.io/fp-ts/modules/Option.ts.html#getorelsew)
and I believe this will trigger an error if the configuration file is not of decoded the right way.

I would use the default Config as default value and use `getOrElse` and not `getOrElseW`.
And with that you don't event need the line 38.

packages/core/src/libraries/component.ts
```diff
@@ -0,0 +1,197 @@
...
+function getMeta(model: SliceAsObject): ComponentMetadata {
+  return {
+    id: model.id,
+    name: model.name,
+    description: model.description,
+  };
+}
+
+/** take a path to slice and return its name  */
+/* TODO: REFACTOR! so error prone to pop everywhere */
```
Heyrows: I'm not sure to understand, this is a new function but you're already planning some refactor 😆

packages/core/src/libraries/component.ts
```diff
@@ -0,0 +1,197 @@
...
+  if (pop.indexOf("index.") === 0) {
+    return split.pop();
+  }
+  if (pop.indexOf(split[split.length - 1]) === 0) {
+    return slicePath.split(path.sep).pop();
+  }
+  return pop.split(".")[0];
+}
+
+/** naive method to find component file in a folder */
```
Heyrows, line 43: We do have a way to know the framework, why would we need that naive method ?

packages/core/src/libraries/component.ts
```diff
@@ -0,0 +1,197 @@
...
+
+function matchPossiblePaths(
+  files: ReadonlyArray<string>
+): boolean {
+  const modelFilename = "model.json";
+
+  return files.includes(modelFilename);
+}
+
+function splitExtension(str: string): {
```
Heyrows: I think this should be in a utils file because i'm sure it won't just be used here.
Extracting the return type is good too.

packages/core/src/libraries/component.ts
```diff
@@ -0,0 +1,197 @@
...
+  const hasFile = Files.exists(fullPath);
+
+  if (hasFile) {
+    return Files.readEntity(fullPath, validate)
+  }
+  return null;
+}
+
+/** returns component file/directory info */
+function getFileInfoFromPath(
```
Heyrows, line 103: this seems like a useful method about the file system, shouldn't it be in the Filesystem part of the core ?

packages/core/src/libraries/component.ts
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  const [fileName, extension] = fullName.split(".");
+  return {
+    fileName,
+    extension,
+  };
+}
+
+function fromJsonFile<T extends unknown>(
```
Heyrows, line 88: seems like a filysystem function

packages/core/src/libraries/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    pathToSlices.split(path.sep).join(path.posix.sep)
+  )
+    .map((curr) => path.join(pathToSlices, curr))
+    .filter((e) => {
+      const f = e.split(path.sep).pop();
+      return fs.lstatSync(e).isDirectory() && !f?.startsWith(".");
+    });
+
+  // relative path to slice folder, to be appended with sliceName
+  const pathToSlice = `${isLocal ? "./" : ""}${from}${pathToSlices
```
Heyrows: I think this could be a helper, i'm sure it could be reusued.

packages/core/src/libraries/index.ts
```diff
@@ -0,0 +1,74 @@
...
+  };
+}
+
+export function libraries(
+  cwd: string,
+  libraries: string[]
+): ReadonlyArray<Library> {
+  return (libraries || [])
+    .map((lib) => handleLibraryPath(cwd, lib))
+    .filter(Boolean) as ReadonlyArray<Library>;
```
Heyrows, line 70: I don't think you need to cast it here, the `HandleLibraryPath` return type is declared.

Heyrows: I'm guessing this is used by the front to display libraries.
I'm wondering if it wouldn't be better here to send back an object with
```
{ [LibName]:  (LibInfo: Library) }
```

That way we could keep the information that some libraries declared in the manifest are not installed locally.

packages/core/src/libraries/screenshot.ts
```diff
@@ -0,0 +1,48 @@
...
+  cwd,
+  from,
+  sliceName,
+  variationId,
+}: {
+  cwd: string;
+  from: string;
+  sliceName: string;
+  variationId: string;
+}): { exists: boolean; path: string; isCustom: boolean } | undefined {
```
Heyrows: I guess if you return more than the type then the name of the function is wrong.
Also, Please define the return type of the function outside.

Defining the return type inside the function doesn't allow it to be used outside :❌

Heyrows: I think there is something wrong with this function.
Instead of targeting a specific file because we know for example what's the status of the ressource (custom or regular), or what's the framework, we try every possible file that exist and return the first one.

I think we need something more precise that takes specific parameters.

packages/core/src/mocks/index.ts
```diff
@@ -0,0 +1,18 @@
...
+import { SliceMock } from "../models/Slice";
+
+/** This only works with local libraries atm */
+export function sliceMocks(cwd: string, lib: string, sliceName: string): SliceMock | undefined {
+
+  const pathToLocalMock = CustomPaths(cwd).library(lib).slice(sliceName).mocks()
+  const pathToDataMocks = GeneratedPaths(cwd).library(lib).slice(sliceName).mocks()
+
+  return Files.readFirstOf<SliceMock | undefined, Record<string, unknown>>([pathToLocalMock, pathToDataMocks])((mocks: string) => {
+    const decoded = SliceMock.decode(JSON.parse(mocks))
```
Heyrows: Use fold

packages/core/src/models/CustomType/fields.ts
```diff
@@ -0,0 +1,60 @@
+import * as t from 'io-ts';
```
Heyrows, line 1: I'm guessing taken from the custom type API, shouldn't we have a common lib ?

packages/core/src/models/Slice.ts
```diff
@@ -0,0 +1,27 @@
+import * as t from "io-ts";
+import { VariationMock } from "./Variation";
+import { AsArray, AsObject, Variation } from "./Variation";
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export const Slice = <T = AsArray | AsObject>(formatReader: t.Type<T>) => {
```
Heyrows, line 6: AsArray & AsObject should not appear here.
They are a specific need of the Slice-machine UI, they shouldn't appear in the Core.

packages/core/src/models/Variation.ts
```diff
@@ -0,0 +1,46 @@
+import * as t from "io-ts";
+import { Field } from "./CustomType/fields";
+
+export enum WidgetsArea {
+  Primary = "primary",
+  Items = "items",
+}
+
+export const AsArray = t.array(t.type({ key: t.string, value: Field }));
```
Heyrows, line 9: same

packages/core/src/utils/const.ts
```diff
@@ -7,3 +7,5 @@ export const PRISMIC_DOM_PACKAGE_NAME = "prismic-dom";
 export const MANIFEST_FILE_NAME = "sm.json";
 export const DEFAULT_SERVER_PORT = 5555;
 export const DEFAULT_BASE = "https://prismic.io";
+
+export const ACCEPTED_IMAGE_TYPES = ["png", "jpg", "jpeg"];
```
Heyrows, line 11: I would have put screenshot in the name like VALID_SCREENSHOTS_EXTENSIONS.

packages/core/src/utils/endpoints.ts
```diff
@@ -18,8 +18,8 @@ export function buildEndpoints(base: string): Endpoints {
   };
 }
 
-export type Base = `${"http" | "https"}://${string}${"/" | ""}`;
-
+export type Base = string;
+export type ApiEndpoint = string;
```
Heyrows, line 22: I think it loose all of it's sense and use to just put them as String, they should just be removed.

packages/core/src/utils/errors.ts
```diff
@@ -0,0 +1,54 @@
+import { Errors } from "io-ts";
+
+interface PathType {
+  type: "index" | "path";
+  value: string;
+}
+
+const ErrorsHandler = {
```
Heyrows, line 8: I would call this file like validationErrors, because it's really specific to IoTs

packages/core/src/utils/files.ts
```diff
@@ -32,17 +32,40 @@ const Files = {
...
-    return JSON.parse(this.readString(pathToFile)) as Record<string, unknown>;
+  readEntity<T extends unknown>(
+    pathToFile: string,
+    validate: (payload: unknown) => Error | T
+  ): Error | T {
+    const entity = this.safeReadJson(pathToFile)
+    if (entity) {
+      return validate(entity);
+    }
+    return new Error(`Could not parse file "${path.basename(pathToFile)}"`)
```
Heyrows: I would keep the extension in case the user needs to find the file.

packages/core/src/utils/files.ts
```diff
@@ -32,17 +32,40 @@ const Files = {
...
+  },
+  safeReadEntity<T>(
+    pathToFile: string,
+    validate: (payload: unknown) => null | T
+  ): null | T {
+    try {
+      const result = this.readEntity(pathToFile, validate);
```
Heyrows, line 54: Or the type of result here is unknown and not T which means this function returns unknown and not T

packages/core/src/utils/lib.ts
```diff
@@ -0,0 +1,71 @@
...
+  C = "/",
+}
+
+const Identifiers: Record<Prefix, number> = {
+  "@/": 2,
+  "~/": 2,
+  "/": 1,
+};
+
+export const findIndexFile = (libPath: string): string | null => {
```
Heyrows, line 16: This could be a generic file function where you just give a directory.

packages/core/src/utils/lib.ts
```diff
@@ -0,0 +1,71 @@
+import path from "path";
+import Files from "./files";
+
+enum Prefix {
```
Heyrows, line 4: This file seems redundant with the libraries from the Core and it definitely shouldn't be in utils.

packages/slice-machine/lib/builders/SliceBuilder/layout/Header/VariationsPopover/index.tsx
```diff
@@ -60,7 +63,7 @@ const VarationsPopover: React.FunctionComponent<{
             defaultValue={current}
             variations={variations}
             onChange={handleChange}
-            MenuItemAction={MenuItemAction}
+            MenuItemAction={onNewVariation ? MenuItemAction : undefined}
```
Heyrows, line 66: Should it be undefined or null ?

packages/slice-machine/lib/builders/SliceBuilder/layout/Header/index.js
```diff
@@ -66,11 +66,11 @@ const Header = ({ Model, store, variation, onSave, onPush, isLoading }) => {
                     onNewVariation={() => setShowVariationModal(true)}
                     onChange={(v) => {
                       router.push(
-                        ...Links.variation(
-                          Model.href,
-                          Model.infos.sliceName,
-                          v.id
-                        ).all
+                        ...Links.variation({
```
Heyrows, line 69: cleaner 👍

packages/slice-machine/lib/builders/SliceBuilder/links.ts
```diff
@@ -1,16 +1,25 @@
...
-  options: object = {}
-): {
+export function variation({
+  lib,
+  sliceName,
+  variationId,
+  options = {},
+  isPreview,
+}: {
+  lib: string;
```
Heyrows, line 8: put the type above please.
It is good practice to never define a type like that but put it above to be reusable.

packages/slice-machine/lib/utils/files.ts
```diff
@@ -84,8 +100,8 @@ const Files = {
   exists(pathToFile: string) {
     try {
       return Boolean(fs.lstatSync(pathToFile));
-    } catch (e) {
-      if (e.code === ERROR_CODES.ENOENT) return false;
+    } catch (e: unknown) {
```
Heyrows, line 103: OPT : e should be at least an error, you can find it's type on the fs lib I believe

packages/slice-machine/package.json
```diff
@@ -37,8 +37,9 @@
   },
   "dependencies": {
     "@babel/register": "^7.11.5",
+    "@prismicio/slice-canvas-com": "^0.0.4",
```
Heyrows, line 41: why `-com ?` and not just slice-canvas, is it different packages ?

packages/slice-machine/server/src/api/common/LibrariesState.ts
```diff
@@ -0,0 +1,72 @@
+import Files from "@lib/utils/files";
+import Environment from "@lib/models/common/Environment";
+import * as Core from '@slicemachine/core'
+
+export function generateState(env: Environment): void {
```
Heyrows, line 11: I think this should be in the Core, this configuration file being stable, I believe it should be part of the Core.
Especially if you need to manipulate or read it like in the Canvas.

packages/slice-machine/server/src/api/common/LibrariesState.ts
```diff
@@ -0,0 +1,72 @@
...
+  const libraries = (env.userConfig.libraries || [])
+    .map(lib => Core.Libraries.handleLibraryPath(env.cwd, lib))
+    .filter(Boolean) as ReadonlyArray<Core.Models.Library.Library>;
+
+  const state = formatLibraries(libraries);
+  Files.write(Core.FileSystem.LibrariesStatePath(env.cwd), state);
+}
+
+export function formatLibraries(
```
Heyrows, line 20: same, it's formating 2 types from the Core itself, should be in the Core.

packages/slice-machine/server/src/api/common/LibrariesState.ts
```diff
@@ -0,0 +1,72 @@
...
+  libraries: ReadonlyArray<Core.Models.Library.Library>
```
Heyrows: libraries has a type I believe.

packages/slice-machine/server/src/api/common/LibrariesState.ts
```diff
@@ -0,0 +1,72 @@
...
+  return library.components.reduce(
+    (acc, component) => ({
+      ...acc,
+      [component.model.id]: formatComponent(component),
+    }),
+    {}
+  );
+}
+
+export function formatComponent(
```
Heyrows, line 53: Same

packages/slice-machine/server/src/api/common/LibrariesState.ts
```diff
@@ -0,0 +1,72 @@
...
+  slice: Core.Models.Library.Component
+): Core.Models.LibrariesState.Component {
```
Heyrows: I'm not sure, but if that "LibraryState" is only destined to be used by the front then it shouldn't be in the Core but should be in here with all the function that transform from the Core model to it.

packages/slice-machine/server/src/api/common/generate.ts
```diff
@@ -0,0 +1,50 @@
...
+  env: Environment,
+  libraries: ReadonlyArray<Library>
+): Promise<void> {
+  try {
+    const components = libraries.reduce<ComponentWithLibStatus[]>(
+      (acc, curr) => [...acc, ...curr.components],
+      []
+    );
+
+    components.forEach(async (c) => {
```
Heyrows, line 21: What if mocks are not up to date because a slice has been modified ?

packages/slice-machine/server/src/api/common/generate.ts
```diff
@@ -0,0 +1,50 @@
...
+      if (
+        !Files.exists(
+          GeneratedPaths(env.cwd)
+            .library(c.from)
+            .slice(c.infos.sliceName)
+            .mocks()
+        )
+      ) {
+        const mocks = await mock(
```
Heyrows, line 30: I think it could be good to have once and for all a Mock module that does the mock and write in the file.

packages/slice-machine/src/modules/Preview/components/ScreenSizes/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export enum Size {
+  FULL = "FULL",
+  TABLET = "TABLET",
+  PHONE = "PHONE",
+}
+
+const screens = [
+  {
+    Icon: <span>A</span>,
```
Heyrows, line 11: Let's not forget those 😄

packages/slice-machine/src/modules/Preview/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+    return <div />;
+  }
+
+  const sliceView: SliceView = [
+    { sliceID: Model.infos.model.id, variationID: variation.id },
+  ];
+
+  console.log({ sliceView });
+
+  if (!Model) {
```
Heyrows: you have that twice.

packages/slice-machine/src/modules/Preview/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  if (!Model || !variation) {
+    return <div />;
+  }
+
+  const sliceView: SliceView = [
+    { sliceID: Model.infos.model.id, variationID: variation.id },
+  ];
+
+  console.log({ sliceView });
```
Heyrows: burnnn 💧 🎤

packages/slice-machine/src/modules/Preview/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+    <div>
+      <Header
+        title={Model.infos.sliceName}
+        Model={Model}
+        variation={variation}
+        handleScreenSizeChange={handleScreenSizeChange}
+        size={state.size}
+      />
+      <IframeRenderer size={state.size} sliceView={sliceView} />
+      {/* <Flex
```
Heyrows: to remove if it's a comment.

packages/slice-machine/src/modules/Preview/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+        }}
+      >
+        <iframe src="http://localhost:3001/_canvas" style={{ border: 'none', height: '100%', width: '100%' }} />
+      </Flex> */}
+    </div>
+  );
+}
+
+/**
+ * .iframe-container {
```
Heyrows: is it css ?
comments to remove.

Heyrows (changes requested):

packages/core/src/filesystem/authConfig.ts
arnaudlewis: The getOrElse must have a default value that has the same type as the one if right, getOrElseW allows you to return a value of a different type, null here.

packages/core/src/libraries/component.ts
arnaudlewis: I don't know, I brought changes from poc-simple-story that Hugo did, he's the one that will tell you what is planned

Heyrows:
there is a build issue on the Init also

packages/core/src/models/Slice.ts
arnaudlewis: on the contrary, it's part of the type of slice and variation, you absolutely need em in the core, it's exposed everywhere even for LibrariesState

packages/core/src/models/LibrariesState.ts
```diff
@@ -0,0 +1,46 @@
...
+    description: t.string,
+  })
+])
+export type Component = t.TypeOf<typeof Component>
+
+// dictionnary of componentId -> component
+export const Library = t.record(t.string, Component)
+export type Library = t.TypeOf<typeof Library>
+
+// dictionnary of libraryId -> components
```
Heyrows, line 46: `libraryId -> library` i guess

packages/core/src/models/LibrariesState.ts
```diff
@@ -0,0 +1,46 @@
+import * as t from 'io-ts';
```
Heyrows: if this file represent the LibraryState but we don't have a type LibraryState being exported, this is a bit weird I think.

packages/slice-machine/server/src/api/common/LibrariesState.ts
```diff
@@ -0,0 +1,72 @@
...
+    .map(lib => Core.Libraries.handleLibraryPath(env.cwd, lib))
+    .filter(Boolean) as ReadonlyArray<Core.Models.Library.Library>;
+
+  const state = formatLibraries(libraries);
+  Files.write(Core.FileSystem.LibrariesStatePath(env.cwd), state);
+}
+
+export function formatLibraries(
+  libraries: ReadonlyArray<Core.Models.Library.Library>
+): Core.Models.LibrariesState.Library {
```
Heyrows: Should the return type be a LibrariesState and not Library ?

packages/slice-machine/server/src/api/common/generate.ts
arnaudlewis: it's the generate at start time, we upgrade the state when a slice is saved

packages/slice-machine/server/src/api/common/LibrariesState.ts
```diff
@@ -0,0 +1,72 @@
+import Files from "@lib/utils/files";
+import Environment from "@lib/models/common/Environment";
+import * as Core from '@slicemachine/core'
+
+export function generateState(env: Environment): void {
+  const libraries = (env.userConfig.libraries || [])
+    .map(lib => Core.Libraries.handleLibraryPath(env.cwd, lib))
+    .filter(Boolean) as ReadonlyArray<Core.Models.Library.Library>;
+
+  const state = formatLibraries(libraries);
```
Heyrows, line 16: A bit weird, I would expect the type of State to be LibraryState and not LibraryState.library

packages/core/src/libraries/component.ts
hypervillain: I copy-coded the code as-is. I left the comment that was written by someone else out of respect but the code works and is extensively tested.

packages/core/src/models/LibrariesState.ts
arnaudlewis: I don't think so, if you have a module `utils` you don't have to export a const or a function named `utils`.
It's just what it represents but it contains way more than one type, otherwise I would have put a default export

packages/core/src/libraries/component.ts
hypervillain: The method is naive because it does not rely on user's possibly custom resolution. This means that a user could define that `.foo` is properly loaded by webpack or equivalent. We cannot know that so it's best effort

packages/core/src/libraries/component.ts
hypervillain: Let's wait to have a need for this then

packages/core/src/libraries/component.ts
hypervillain: This function is just a local helper to interact with FileSystem functions

packages/core/src/libraries/index.ts
arnaudlewis: we have to cast because of the filter, ts is lost

packages/core/src/libraries/screenshot.ts
arnaudlewis: I won't modify this because first it was there as is and I made an extensive refacto on type on the next PR that will be merged in this one later

packages/core/src/libraries/component.ts
hypervillain: It is useful _here_ indeed. I don't think moving "everything" to helpers makes things clearer in any way until we have a clear use for this.

packages/slice-machine/src/modules/Preview/components/ScreenSizes/index.tsx
hypervillain: :)

packages/core/src/models/CustomType/fields.ts
arnaudlewis: no it's not but at some point it should and that won't be on me ;)

packages/slice-machine/src/modules/Preview/index.tsx
hypervillain: right 😇

packages/slice-machine/src/modules/Preview/index.tsx
hypervillain: 😇

packages/core/src/utils/endpoints.ts
arnaudlewis: kind of agree except for code clarity and I didn't introduce it myself so I won't remove them for now if they are used somewhere.

packages/core/src/utils/files.ts
arnaudlewis: no it returns T coming from readEntity

packages/core/src/mocks/index.ts
arnaudlewis: no x) no need

packages/core/src/libraries/component.ts
Heyrows: then let's remove the comment if it's useless and is pointless now

packages/core/src/filesystem/authConfig.ts
Heyrows: We used to not have the config validated, that's why we had to spread, make sure we had the good type in the end.

Here you validate, so you could return the default if it doesn't valide and there would be no extra operation of spreading in case of a right file

packages/core/src/models/LibrariesState.ts
Heyrows: but then it becomes uneasy to understand.
Honestly I wouldn't know what's a LibrariesStates and what's on the file withouth the function in SM-UI that writes in the file.

So if this is to be used outside then I find it totally un-understandable and wrong.

packages/core/src/models/Slice.ts
Heyrows: We needed to make a difference between Object and Array only for the SM-UI.
But here in the Core, we only have as Object normally.

As Array is just a representation of the models for the SM-UI, nobody else should know about it.

packages/core/src/utils/files.ts
Heyrows: Ok , didn't knew it could take the type automatically, thought you had to declare `readEntity<T>`

Heyrows:
i'm closing this, we have a preview branch now.

PR #220, opened 2021-11-08
Improve Login Modal / Refresh token on get state

Miniplop:
## Types of changes

## Checklist:


packages/slice-machine/components/AppLayout/index.tsx
```diff
@@ -11,14 +10,15 @@ const AsIs: { [x: string]: boolean } = {
   "/onboarding": true,
 };
 
-const AppLayout = ({
+type AppLayoutProps = {
+  env: Environment;
+  serverState: ServerState;
+};
+
+const AppLayout: React.FunctionComponent<AppLayoutProps> = ({
```
Heyrows, line 18: OPT : I don't see the import of react, how does this not "fail" due to React being unknown here ?

packages/slice-machine/components/LoginModal/index.tsx
```diff
@@ -11,28 +11,48 @@ import {
...
-interface LoginModalProps {
-  onClose: () => void;
-  isOpen: boolean;
-}
+const LoginModal: React.FunctionComponent = () => {
+  const { env, isOpen, isLoginLoading } = useSelector(
+    (store: SliceMachineStoreType) => ({
+      isOpen: isModalOpen(store, ModalKeysEnum.LOGIN),
+      isLoginLoading: isLoading(store, LoadingKeysEnum.LOGIN),
+      env: getEnvironment(store),
```
Heyrows, line 39: This is only my second file that I review but i'm not sure why the "env" would be different depending on the store.
Because the env here is the tokens I believe so I don't see why it would change

packages/slice-machine/components/LoginModal/index.tsx
```diff
@@ -45,17 +65,23 @@ const LoginModal: React.FunctionComponent<LoginModalProps> = ({
     }
 
     try {
-      setIsLoading(true);
+      startLoadingLogin();
       await startAuth();
-      const isAuthStatusOk = ({ status }: { status: string }) =>
-        status === "ok";
+      const isAuthStatusOk = (
```
Heyrows: I think this can be define outside, i don't think it will change.

packages/slice-machine/components/ReviewModal/index.tsx
```diff
@@ -49,11 +69,65 @@ const SelectRatingComponent = ({ field, form }: FieldProps) => {
   );
 };
 
-const ReviewModal: React.FunctionComponent<ReviewModalProps> = ({
-  close,
-  isOpen,
-  onSubmit,
-}) => {
+const ReviewModal: React.FunctionComponent<ReviewModalProps> = () => {
```
Heyrows, line 65: OPT: We call it Rating in Linear, during every call so I think it should be rename.

packages/slice-machine/components/ReviewModal/index.tsx
```diff
@@ -49,11 +69,65 @@ const SelectRatingComponent = ({ field, form }: FieldProps) => {
...
+      isLoginModalOpen: isModalOpen(store, ModalKeysEnum.LOGIN),
+      hasSendAReview: userHasSendAReview(store),
+    })
+  );
+
+  const dispatch = useDispatch();
+  const openLogin = () =>
+    dispatch(modalOpenCreator({ modalKey: ModalKeysEnum.LOGIN }));
+  const startReviewLoading = () =>
+    dispatch(startLoadingActionCreator({ key: LoadingKeysEnum.REVIEW }));
```
Heyrows: OPT: should be call `modalKey` or `modalId` for consistancy

packages/slice-machine/components/ReviewModal/index.tsx
```diff
@@ -49,11 +69,65 @@ const SelectRatingComponent = ({ field, form }: FieldProps) => {
...
+  const stopReviewLoading = () =>
+    dispatch(stopLoadingActionCreator({ key: LoadingKeysEnum.REVIEW }));
+  const skipReview = () => dispatch(skipReviewCreator());
+  const sendAReview = () => dispatch(sendAReviewCreator());
+
+  const { addToast } = useToasts();
+
+  const sliceCount =
```
Heyrows, line 90: Can this be in the core ?
Counting the slices in the libraries.

packages/slice-machine/components/ReviewModal/index.tsx
```diff
@@ -49,11 +69,65 @@ const SelectRatingComponent = ({ field, form }: FieldProps) => {
...
+
+  const dispatch = useDispatch();
+  const openLogin = () =>
+    dispatch(modalOpenCreator({ modalKey: ModalKeysEnum.LOGIN }));
+  const startReviewLoading = () =>
+    dispatch(startLoadingActionCreator({ key: LoadingKeysEnum.REVIEW }));
+  const stopReviewLoading = () =>
+    dispatch(stopLoadingActionCreator({ key: LoadingKeysEnum.REVIEW }));
+  const skipReview = () => dispatch(skipReviewCreator());
+  const sendAReview = () => dispatch(sendAReviewCreator());
```
Heyrows: I think these events could be redefined in some custom hooks, a function that would do the useDispatch and return functions like "skipReview", that would be so cool and would avoid having to redefine them each time.

Plus we wouldn't have to import all these Creators 👍

packages/slice-machine/components/ReviewModal/index.tsx
```diff
@@ -49,11 +69,65 @@ const SelectRatingComponent = ({ field, form }: FieldProps) => {
...
+          }
+
+          return count + lib.components.length;
+        }, 0)
+      : 0;
+
+  const customTypeCount = !!customTypes ? customTypes.length : 0;
+  // Deactivate for this release
+  const userHasCreateEnoughContent =
+    sliceCount >= 1 && customTypeCount >= 1 && false;
```
Heyrows: I think we'll reactive it for 0.1.2 :)

packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header.tsx
```diff
@@ -25,7 +23,9 @@ const CustomTypeHeader = ({
   store: CustomTypeStore;
 }) => {
   const [isLoading, setIsLoading] = useState(false);
-  const { openLogin } = useContext(LoginModalContext);
+  const dispatch = useDispatch();
+  const openLogin = () =>
```
Heyrows: This comment is for when the "openLogin" is being called.

I think the fact that it opens the modal and ask to login should be part of the calls like "store.push".

That way, if the user is logged in the call go through, if it's not then the user is asked to log in and nothing is sent back from the function.

packages/slice-machine/lib/builders/common/Zone/Card/index.js
```diff
@@ -32,7 +32,9 @@ const FieldZone = ({
 }) => {
   const {
     env: { framework },
```
Heyrows, line 34: I don't think you need to type the env here.

Heyrows: especially becasue the type here is only part of the real type.

packages/slice-machine/lib/models/common/http/FakeClient.ts
```diff
@@ -66,14 +66,9 @@ export default class FakeClient {
...
-  async sendReview(): Promise<FakeResponse> {
+  async sendReview() {
     return {
-      status: 200,
-      statusText: "ok",
-      fake: true,
-      json() {
-        return [];
-      },
+      status: 403,
```
Heyrows, line 71: I think for the fake client, every call on this page should send back a 403, it should be like a general response.

packages/slice-machine/package.json
```diff
@@ -135,10 +140,8 @@
     "next-videos": "^1.4.0",
     "nock": "^13.1.4",
     "nodemon": "^2.0.12",
-    "npm": "^7.6.3",
```
Heyrows, line 138: loool wtf, good catch 😂

packages/slice-machine/pages/_app.tsx
```diff
@@ -124,107 +129,101 @@ function MyApp({
...
-                                    open: false,
-                                  })
-                                }
-                              >
-                                <Warnings
-                                  priority={drawerState.priority}
-                                  list={data.warnings}
-                                  configErrors={data.configErrors}
+    <Provider store={store}>
+      <PersistGate loading={null} persistor={persistor}>
```
Heyrows: What does this do ?

packages/slice-machine/pages/_app.tsx
```diff
@@ -124,107 +129,101 @@ function MyApp({
...
+                                    priority={drawerState.priority}
+                                    list={serverState.warnings}
+                                    configErrors={serverState.configErrors}
+                                  />
+                                </Drawer>
+                              </SliceHandler>
+                            </AppLayout>
+                          </CustomTypesProvider>
+                        </LibrariesProvider>
+                        <LoginModal />
```
Heyrows: OPT : if we add more modal it's gonna be a pain, there should only be one I believe. And we could make new ones pop on demand.

packages/slice-machine/pages/_app.tsx
```diff
@@ -124,107 +129,101 @@ function MyApp({
...
-              {!payload || !payload.libraries ? (
-                <Renderer
-                  Component={Component}
-                  pageProps={pageProps}
-                  {...payload}
-                  openPanel={openPanel}
-                />
-              ) : (
-                <ToastProvider>
-                  <LoginModalProvider>
```
Heyrows, line 180: NOTE: It was all worth to remove this Provider, but I don't see any other that has been removed I think.
It would be cool to have others store using redux so that the code you make bring more values.

packages/slice-machine/pages/cts/[ct].tsx
```diff
@@ -17,7 +19,9 @@ type CustomTypeBuilderWithProviderProps = {
 
 const CustomTypeBuilderWithProvider: React.FunctionComponent<CustomTypeBuilderWithProviderProps> =
   ({ customType, remoteCustomType, onLeave }) => {
-    const { env } = useContext(ConfigContext);
+    const { env } = useSelector((store: SliceMachineStoreType) => ({
+      env: getEnvironment(store),
```
Comment (Heyrows) [inline packages/slice-machine/pages/cts/[ct].tsx:23]: NOTE: I think env is a wrong term, because honestly I believe it's the manifest which seems logical but i'm not sure if it's the same data structure or if it's not the Prismic Config.

packages/slice-machine/src/redux/store.ts
```diff
@@ -0,0 +1,31 @@
...
+import { persistStore, persistReducer } from "redux-persist";
+import storage from "redux-persist/lib/storage"; // defaults to localStorage for web
+
+const persistConfig = {
+  key: "root",
+  storage,
+  whitelist: ["userContext"],
+};
+
+declare var window: {
```
Heyrows, line 13: I think this should be made in a d.ts file.

packages/slice-machine/src/redux/store.ts
```diff
@@ -0,0 +1,31 @@
...
+    process.env.NODE_ENV !== "production" &&
+    typeof window === "object" &&
+    window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__
+      ? window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__
+      : compose;
+
+  const rootReducer = createReducer();
+
+  const persistedReducer = persistReducer(persistConfig, rootReducer);
+  const store: any = createStore(persistedReducer, {}, composeEnhancers());
```
Heyrows: NOTE: Redux seems really complex just to create a store.

packages/slice-machine/pages/_app.tsx
```diff
@@ -124,107 +129,101 @@ function MyApp({
   );
 
   useEffect(() => {
```
Heyrows, line 86: I see you've removed the part that redirect you to the onboarding but I haven't seen yet where it is implemented back.

packages/slice-machine/pages/onboarding.tsx
```diff
@@ -235,6 +225,9 @@ export default function Onboarding(): JSX.Element {
     <PushPagesSlide onEnded={handleOnVideoEnd} />,
   ];
 
+  const dispatch = useDispatch();
+  const finishOnboarding = () => dispatch(finishOnboardingCreator());
```
Heyrows: Same as before i would put all of these in a custom hook :)

packages/slice-machine/pages/onboarding.tsx
```diff
@@ -267,7 +259,7 @@ export default function Onboarding(): JSX.Element {
       ...(state.step > 0 ? { completed: state.videoCompleted } : {}),
     };
 
-    postTracking(data);
+    sendTrackingOnboarding(data).catch(console.error);
```
Heyrows, line 255: Is the catch right ? (just asking because I don't know what you've been meaning to do)

packages/slice-machine/pages/warnings.tsx
```diff
@@ -45,13 +51,7 @@ export default function WarningsPage() {
...
           {warnings.map((warning) => {
             const Component = Renderers[warning.key];
-            return (
-              <Component
-                key={warning.key}
-                errorType={warning.title}
-                {...warning}
-              />
-            );
+            return <Component errorType={warning.title} {...warning} />;
```
Heyrows, line 54: you're missing the key attribute, it's a map on several warnings, it could throw a react error

packages/slice-machine/server/src/api/auth/validate.ts
```diff
@@ -1,10 +1,11 @@
...
 
 type ValidateResponse = {
   connected: boolean;
   reason: string;
   err?: any;
 };
+
 export default async function handler(): Promise<ValidateResponse> {
-  return validate();
+  return validateUserAuth();
```
Heyrows, line 10: If the handler function just return the execution of a function, the return type should come from that function and not be defined in this file.

packages/slice-machine/server/src/api/index.ts
```diff
@@ -54,12 +59,12 @@ router.get(
...
+  async function (
+    req: express.Request<TrackingReviewRequest>,
+    res: express.Response<TrackingReviewResponse>
+  ) {
     const payload = await trackReview(req.body);
-    if (payload.clientError) {
-      return res.status(payload.clientError.status).json(payload);
-    }
-    return res.status(200).json(payload);
+    return res.status(payload.status).json({});
```
Heyrows, line 67: you don't return the payload anymore.

packages/slice-machine/server/src/api/services/getEnv.ts
```diff
@@ -173,18 +155,12 @@ export async function getEnv(
...
-        cwd,
-        prismicData.value.base,
-        repo,
-        prismicData.value.auth?.auth
-      );
-    } else {
-      return new FakeClient();
-    }
-  })();
+  const client = initClient(
```
Heyrows, line 158: It seems that we don't use the FakeClient anymore, can it be deleted ?

packages/slice-machine/server/src/api/services/getEnv.ts
```diff
@@ -193,7 +169,7 @@ export async function getEnv(
       repo,
       userConfig,
       hasConfigFile: true,
-      prismicData: prismicData.isOk() ? prismicData.value : undefined,
+      prismicData: prismicData.value,
```
Heyrows, line 172: Now I see why it's not optional anymore 👍

packages/slice-machine/server/src/api/tracking/review.ts
```diff
@@ -1,23 +1,22 @@
...
-  comment: string;
-}): Promise<{ err: FakeResponse | Response | null }> {
-  const { env } = await getEnv();
+export default async function handler(
+  requestBody: TrackingReviewRequest
+): Promise<{ status: number }> {
   try {
+    const { env } = await getEnv();
     const res = await env.client.sendReview({
       id: TrackingEventId.REVIEW,
```
Heyrows: OPT : If there was a function to send a specific event, I wouldn't expect to have to pass an Event ID.
To me it would be more like
```
function sendEvent({ id, payload })
function sendReview({ payload }) => sendEvent({ id: TrackingEventId.REVIEW, payload })
```

packages/slice-machine/server/src/api/tracking/review.ts
```diff
@@ -1,23 +1,22 @@
...
       framework: env.framework,
-      ...query,
-    });
-    if (res.status !== 200) {
-      return { err: res };
-    }
-    return { err: null };
+      ...requestBody,
+    } as ReviewTrackingEvent);
+    return { status: res.status };
```
Heyrows, line 14: You don't use errors anymore, I understand why you need the status, but the error message could be useful to debug somehow.

packages/slice-machine/src/apiClient.ts
```diff
@@ -1,18 +1,42 @@
...
+    "Content-Type": "application/json",
+  },
 };
 
 /** Auth Routes **/
 
-export const startAuth = async () =>
-  await fetch("/api/auth/start", { headers: jsonHeaders, method: "POST" });
+export const startAuth = (): Promise<AxiosResponse<{}>> =>
+  axios.post("/api/auth/start", {}, jsonHeaders);
```
Heyrows: ➕  OPT: I would use axios with a big objet to keep every parameter clear.

packages/slice-machine/src/apiClient.ts
```diff
@@ -1,18 +1,42 @@
...
+import {
+  OnboardingContinueEvent,
+  OnboardingContinueWithVideoEvent,
+  OnboardingSkipEvent,
+  OnboardingStartEvent,
+  TrackingReviewResponse,
+} from "@models/common/TrackingEvent";
+import { CheckAuthStatusResponse } from "@models/common/Auth";
+
 const jsonHeaders = {
```
Heyrows: I would rename it to `defaultAxiosConfig` since it can return more than headers

packages/slice-machine/src/modules/environment/index.tsx
```diff
@@ -0,0 +1 @@
+export * from "./environment";
```
Heyrows: I think this is useless, you could just rename the environment file to index.tsx

packages/slice-machine/src/modules/loading/index.tsx
```diff
@@ -0,0 +1 @@
+export * from "./loading";
```
Heyrows: same as environment.

packages/slice-machine/src/modules/loading/loading.tsx
```diff
@@ -0,0 +1,52 @@
+import { Reducer } from "redux";
+import mapValues from "lodash/mapValues";
+import { SliceMachineStoreType } from "src/redux/type";
+import { LoadingStoreType, LoadingKeysEnum } from "./types";
+import { ActionType, createAction, getType } from "typesafe-actions";
+
+const initialState: LoadingStoreType = {
+  ...(mapValues(LoadingKeysEnum, () => false) as Record<
```
Heyrows: NOTE: didn't knew the mapValue but you could have used  `Object.Keys` / `Object.Values` I believe

packages/slice-machine/src/modules/loading/loading.tsx
```diff
@@ -0,0 +1,52 @@
...
+>;
+
+// Selectors
+export const isLoading = (
+  store: SliceMachineStoreType,
+  key: LoadingKeysEnum
+): boolean => store.loading[key];
+
+// Reducer
+export const loadingReducer: Reducer<LoadingStoreType, LoadingActions> = (
```
Heyrows: It took time to understand redux, but I think that the data cutting might be too thin.

packages/slice-machine/src/modules/modal/index.tsx
```diff
@@ -0,0 +1 @@
+export * from "./modal";
```
Heyrows: same as environment

packages/slice-machine/src/modules/loading/types.tsx
```diff
@@ -0,0 +1,6 @@
+export type LoadingStoreType = Record<LoadingKeysEnum, boolean>;
+
+export enum LoadingKeysEnum {
```
Heyrows: Guessing this is more like an enum of the modals that are common to all the screens.
Once again it tells me that this might be too thin.

Like having a specific enum for Open and a specific enum for Loading seems too much.
I would expect a general "ModalEnum"

packages/slice-machine/src/modules/modal/modal.ts
```diff
@@ -0,0 +1,54 @@
+import mapValues from "lodash/mapValues";
+import { Reducer } from "redux";
+import { SliceMachineStoreType } from "@src/redux/type";
+import { ActionType, createAction, getType } from "typesafe-actions";
+
+export enum ModalKeysEnum {
```
Heyrows: This could be made from the "ModalEnum" that I described in another comment.
So here we would have an object / array that selects modals that are part of this "open".

I'm wondering tho why we don't have the review modal here too.

packages/slice-machine/src/modules/userContext/index.tsx
```diff
@@ -0,0 +1 @@
+export * from "./userContext";
```
Heyrows: definitely no index.tsx just to import one file

packages/slice-machine/src/routerProvider.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+const RouterProvider: React.FunctionComponent = ({ children }) => {
+  const router = useRouter();
+
+  const { isOnboarded } = useSelector((store: SliceMachineStoreType) => ({
+    isOnboarded: userHasDoneTheOnboarding(store),
+  }));
+
+  useEffect(() => {
+    if (!isOnboarded && router.pathname !== "/onboarding") {
```
Heyrows: Now I see where it's gone

packages/slice-machine/src/routerProvider.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { useEffect } from "react";
+
+import { useSelector } from "react-redux";
+import { SliceMachineStoreType } from "@src/redux/type";
+import { userHasDoneTheOnboarding } from "@src/modules/userContext";
+import { useRouter } from "next/router";
+
+const RouterProvider: React.FunctionComponent = ({ children }) => {
```
Heyrows: I've seen how you use this Provider and I think it's wrong.
I would prefer to have it not as a provider but as a custom hook.
The only thing that differentiate it from a custom hook here is that you return the children as it is, you don't modify them.

packages/slice-machine/src/stateDispatcher.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { useDispatch } from "react-redux";
+import { getEnvironmentCreator } from "@src/modules/environment";
+
+type StateProviderProps = {
+  serverState: ServerState;
+};
+
+export const StateDispatcher = React.createContext({});
+
+const StateProvider: React.FunctionComponent<StateProviderProps> = ({
```
Heyrows: same thing, it's basically just a useEffect to spread the serverState, it doesn't need to be a Provider

packages/slice-machine/tests/project/cypress/integration/onboarding/00-redirects.spec.js
```diff
@@ -14,21 +14,18 @@ describe('onboarding redirects and local storage', () => {
...
     cy.intercept('POST', '/tracking/onboarding', ({body}) => {
       expect(body.id).equal("slicemachine_onboarding_start")
     })
     
     cy.visit('/')
 
     cy.location('pathname', {timeout: 1000}).should('eq', '/onboarding')
-
-    cy.wait(1500)
-    cy.getLocalStorage("is-onboarded").should('eq', 'true')
```
Heyrows, line 27: why removing this ?
If you're onboarded it should be reflected on the local storage no ?

Heyrows: As the external source of truth that makes sure you won't be onboarded again.

packages/slice-machine/tests/project/cypress/integration/onboarding/00-redirects.spec.js
```diff
@@ -14,21 +14,18 @@ describe('onboarding redirects and local storage', () => {
...
   })
 
   it('when is-onboarded is in local storage it should not redirect', () => {
-    cy.setLocalStorage("is-onboarded", true)
+    cy.setLocalStorage("persist:root", JSON.stringify({userContext: JSON.stringify({isOnboarded: true})}))
```
Heyrows, line 28: isn't it a little weird to have to stringify twice ? Shouldn't it be a big object ?

packages/slice-machine/pages/onboarding.tsx
```diff
@@ -245,20 +238,19 @@ export default function Onboarding(): JSX.Element {
     return setState({ ...state, videoCompleted: true });
   }
 
-  useEffect(() => {
-    localStorage.setItem(LocalStorageKeys.isOnboarded, "true");
```
Heyrows, line 245: Could you point me to where you did this again ?
I couldn't find it anymore.
I found the part about the state but couldn't find the part about the local storage.

Heyrows (changes requested):

packages/slice-machine/components/AppLayout/index.tsx
Miniplop: Typescript import it on every file

packages/slice-machine/components/LoginModal/index.tsx
Miniplop: env is an object returned by the state route, there are many stuff inside it. Take a look at this type `Environment`

packages/slice-machine/components/LoginModal/index.tsx
Miniplop: outside the component ?  What is the gain to do so ?

packages/slice-machine/components/ReviewModal/index.tsx
Miniplop: You right

packages/slice-machine/components/ReviewModal/index.tsx
Miniplop: As the key refer to the loading key, a good name can be loadingKey

packages/slice-machine/components/ReviewModal/index.tsx
Miniplop: Nop I don't think this is a good idea, because the count here his linked to how the front manage the data and this is specific to the front side only.

Only generic stuff should be written in the core

packages/slice-machine/components/ReviewModal/index.tsx
Miniplop: Yes can do it now, it will be easier to test the feature in staging

packages/slice-machine/components/ReviewModal/index.tsx
Miniplop: I don't know if this is a good idea, because the every custom hooks will call useDispatch, so this hook will be call many times on one component.

If we create custom hooks for every actions creator, we will have to import them as well so this is the same right ?

packages/slice-machine/lib/models/common/http/FakeClient.ts
Miniplop: Yes but it will break a lot of stuff to do so, and this ticket is not the right moment to do that :)

packages/slice-machine/lib/builders/common/Zone/Card/index.js
Miniplop: it is not a typing annotation but a nested destructuration

packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header.tsx
Miniplop: But the front doesn't know if the user is logged in before the call

packages/slice-machine/package.json
Miniplop: hihihi :)

packages/slice-machine/pages/_app.tsx
Miniplop: PresistGate re-hydrate the userContext store part with persisted data in the local Storage

packages/slice-machine/pages/_app.tsx
Miniplop: I removed these 3 providers : 
- trackingProvider
- LoginModalProvider
- ConfigProvider

packages/slice-machine/pages/onboarding.tsx
Miniplop: avoiding app crash on tracking auth error

packages/slice-machine/server/src/api/index.ts
Miniplop: the payload is empty

packages/slice-machine/server/src/api/services/getEnv.ts
Miniplop: not yet, it's still used in the InitClient function used in the getEnv function, so almost on every routes :p

packages/slice-machine/server/src/api/tracking/review.ts
Miniplop: good point :)

packages/slice-machine/src/apiClient.ts
Miniplop: fixed

packages/slice-machine/src/modules/environment/index.tsx
Miniplop: it's for helping the navigation with our IDE, we can look for envirnonment.ts, index.ts is not a specific name but I can change it ;)

packages/slice-machine/src/modules/loading/loading.tsx
Miniplop: yes :)

packages/slice-machine/src/redux/store.ts
Miniplop: haha, hopefully you won't have to modify this file :)

packages/slice-machine/src/routerProvider.tsx
Miniplop: Found this idea sexy

packages/slice-machine/src/routerProvider.tsx
Miniplop: hahaha

packages/slice-machine/src/stateDispatcher.tsx
Miniplop: You right :)

packages/slice-machine/components/ReviewModal/index.tsx
Miniplop: Seen together, now it's fixed :)

packages/slice-machine/components/ReviewModal/index.tsx
Miniplop: Done

packages/slice-machine/pages/_app.tsx
Miniplop: Seen together :)

packages/slice-machine/pages/warnings.tsx
Miniplop: The key attribute was already present in the warning object so it was redundant. I don't think this a good practice to spread object like that on props. But I think this is out of scope to refactor this part right now

packages/slice-machine/server/src/api/index.ts
Miniplop: There is no body return by the tracking api as it's 201 :)

packages/slice-machine/server/src/api/services/getEnv.ts
Miniplop: Not yet 😭 The Fake client is still use in the init client :)

packages/slice-machine/src/stateDispatcher.tsx
Miniplop: Fixed

packages/slice-machine/tests/project/cypress/integration/onboarding/00-redirects.spec.js
Miniplop: because the feature changes, now the onbaording flag is set to true only when the user skip on finish the onboarding

packages/slice-machine/pages/onboarding.tsx
Miniplop: Redux persist handle this right now

packages/slice-machine/src/routerProvider.tsx
Miniplop: Fixed :)

packages/slice-machine/components/ReviewModal/index.tsx
Miniplop: Fixed

packages/slice-machine/components/ReviewModal/index.tsx
Miniplop: fixed

packages/slice-machine/pages/onboarding.tsx
Miniplop: done

packages/slice-machine/server/src/api/services/getEnv.ts
Miniplop: good 🍾

packages/slice-machine/src/modules/environment/index.tsx
Miniplop: seen together we name file index.ts

packages/slice-machine/src/modules/loading/loading.tsx
Miniplop: seen together : modal modules scope is to handle only all modal states, loading only for api call states

packages/slice-machine/src/modules/loading/types.tsx
Miniplop: Seen together : The loading modules scope should only be for loading state

packages/slice-machine/pages/cts/[ct].tsx
Comment (Miniplop) [reply packages/slice-machine/pages/cts/[ct].tsx:23]: env for now is a big bloater object with ~/.prismic info + sm.json info + other stuffs, should be clean later for sure 🧼

packages/slice-machine/src/redux/store.ts
Miniplop: Why ? The variable we defined `__REDUX_DEVTOOLS_EXTENSION_COMPOSE__` is used only on the store configuration and should not be use elsewhere.

packages/slice-machine/src/redux/reducer.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Combine all reducers in this file and export the combined reducers.
+ * If we were to do this in store.js, reducers wouldn't be hot reloadable.
+ */
+
+import { combineReducers } from "redux";
+import { modalReducer } from "@src/modules/modal";
```
MarcMcIntosh, line 7: why not `../modules/modal` ?

packages/core/src/filesystem/authConfig.ts
```diff
@@ -54,3 +54,32 @@ export function setAuthConfig(
     { recursive: false }
   );
 }
+
+export function updateAuthCookie(authCookie: string, directory?: string): void {
```
Heyrows, line 58: I remember that we talked about having this function before and we decided that it was the same as the function above.
So why did you needed to add it back ?

A merge mistake maybe ?

packages/core/src/utils/cookie.ts
```diff
@@ -1,8 +1,7 @@
 import * as cookie from "cookie";
 
 const noEscape = (str: string) => str;
-
-const AUTH_KEY = "prismic-auth";
+export const AUTH_KEY = "prismic-auth";
```
Heyrows, line 4: I didn't see where it is being used yet but I believe it wasn't exported on purpose so people would only use the helpers to manipulate the cookie :)

packages/slice-machine/components/LoginModal/index.tsx
```diff
@@ -43,13 +37,8 @@ const LoginModal: React.FunctionComponent = () => {
...
   );
 
-  const dispatch = useDispatch();
-  const onClose = () =>
-    dispatch(modalCloseCreator({ modalKey: ModalKeysEnum.LOGIN }));
-  const startLoadingLogin = () =>
-    dispatch(startLoadingActionCreator({ key: LoadingKeysEnum.LOGIN }));
-  const stopLoadingLogin = () =>
-    dispatch(stopLoadingActionCreator({ key: LoadingKeysEnum.LOGIN }));
+  const { closeLoginModal, startLoadingLogin, stopLoadingLogin } =
```
Heyrows, line 43: YEAHHH

packages/slice-machine/components/App/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+    (priority?: any) =>
+      setDrawerState({
+        ...drawerState,
+        open: true,
+        ...(priority ? { priority } : null),
+      }),
+    []
+  );
+
+  useOnboardingRedirection();
```
Heyrows, line 53: noiiice

packages/slice-machine/components/RatingModal/index.tsx
```diff
@@ -80,15 +70,13 @@ const ReviewModal: React.FunctionComponent<ReviewModalProps> = () => {
...
-    dispatch(startLoadingActionCreator({ key: LoadingKeysEnum.REVIEW }));
-  const stopReviewLoading = () =>
-    dispatch(stopLoadingActionCreator({ key: LoadingKeysEnum.REVIEW }));
-  const skipReview = () => dispatch(skipReviewCreator());
-  const sendAReview = () => dispatch(sendAReviewCreator());
+  const {
+    skipReview,
+    sendAReview,
+    openLoginModal,
+    startLoadingReview,
```
Heyrows: Since you renamed Review to Rating (thanks for that), do you think we could rename the functions as well ?

packages/slice-machine/lib/models/common/http/DefaultClient.ts
```diff
@@ -122,6 +123,15 @@ export default class DefaultClient {
     );
   }
 
+  static refreshToken(base: string, auth: string): Promise<Response> {
+    return fetch(
```
Heyrows, line 128: didn't you wanted to use axios ? or was it only for client side ?

packages/slice-machine/pages/_app.tsx
```diff
@@ -153,77 +108,23 @@ function MyApp({
...
+      </Head>
+      <Provider store={store}>
+        <PersistGate loading={null} persistor={persistor}>
+          <SliceMachineApp
+            theme={theme}
+            serverState={serverState}
+            payload={payload}
+            pageProps={pageProps}
+            Component={Component}
+            Renderer={Renderer}
```
Heyrows, line 123: What's the renderer for ?

packages/slice-machine/server/src/api/state.ts
```diff
@@ -104,6 +106,26 @@ export default async function handler() {
     env
   );
 
+  // Refresh auth
+  if (!isFake && env.prismicData.auth) {
+    try {
+      const newTokenResponse: Response = await DefaultClient.refreshToken(
```
Heyrows, line 112: I think we should have a function in the core to refresh the token and change the cookies, like the login or signup does.

Because i'm guessing this logic won't only be used here.

packages/slice-machine/src/hooks/useServerState.tsx
```diff
@@ -0,0 +1,14 @@
+import { useEffect } from "react";
+
+import { ServerState } from "@lib/models/server/ServerState";
+import useSliceMachineActions from "@src/modules/useSliceMachineActions";
+
+const useServerState = (serverState: ServerState | undefined) => {
+  const { getEnvironment } = useSliceMachineActions();
+
+  useEffect(() => getEnvironment(serverState), [serverState]);
```
Heyrows, line 9: Not sure to understand why we need to update the Environment when the State changes.
I'm talking from the last project where I did the same and It ended up badly with useEffect depending on each other.
I'm not asking you to change, just let's be careful not to have more useEffect depending on each other

packages/slice-machine/src/modules/modal/types.ts
```diff
@@ -0,0 +1,10 @@
+export enum ModalKeysEnum {
+  LOGIN = "LOGIN",
+}
+
+export type ModalStoreType = Record<ModalKeysEnum, boolean>;
+
+export type UserContextStoreType = {
```
Heyrows: I'm not sure to understand why this is here, it doesn't seems to be modal related.

packages/slice-machine/src/modules/useSliceMachineActions.ts
```diff
@@ -0,0 +1,66 @@
...
+    dispatch(startLoadingActionCreator({ loadingKey: LoadingKeysEnum.LOGIN }));
+  const stopLoadingLogin = () =>
+    dispatch(stopLoadingActionCreator({ loadingKey: LoadingKeysEnum.LOGIN }));
+
+  // UserContext Store
+  const skipReview = () => dispatch(skipReviewCreator());
+  const sendAReview = () => dispatch(sendAReviewCreator());
+  const finishOnboarding = () => dispatch(finishOnboardingCreator());
+  const getEnvironment = (serverState: ServerState | undefined) => {
+    if (!serverState) return;
```
Heyrows, line 41: I think the ServerState should not be undefined here but the function should not be called if it is underfined.

Heyrows (changes requested):
Small things

packages/slice-machine/components/ReviewModal/index.tsx
Miniplop: Seen with amaury : The front type shouldn't be in the Core.

packages/slice-machine/server/src/api/auth/validate.ts
Miniplop: Seen together we type the basic function : validateUserAuth

packages/slice-machine/src/modules/useSliceMachineActions.ts
Miniplop: Mistake fixed :)

Miniplop: I fix that

packages/slice-machine/src/modules/modal/types.ts
Miniplop: fixed

Heyrows (approved):

PR #228, opened 2021-11-19
Mm/prompt to update

MarcMcIntosh:
## Types of changes

## Checklist:


packages/slice-machine/components/App/index.tsx
```diff
@@ -116,7 +117,7 @@ const SliceMachineApp: React.FunctionComponent<AppProps> = ({
                           </Drawer>
                         </SliceHandler>
                       </AppLayout>
-                      <LoginModal />
+                      {<UpdateModal /> || <LoginModal />}
```
Miniplop: Why we need to choose between the two modals ?

packages/slice-machine/components/AppLayout/Navigation/Menu/Desktop.tsx
```diff
@@ -34,6 +87,7 @@ const Desktop = () => {
         <Logo />
         <ItemsList mt={4} links={navCtx?.links as []} />
         <Box sx={{ position: "absolute", bottom: "3" }}>
+          <UpdateInfo />
```
Miniplop: As the NotLoggedIn Component, it's better to let the context choose if the <UpdateInfo /> should appear or not instead of giving this responsability to the component itself.

packages/slice-machine/components/AppLayout/Navigation/Menu/Desktop.tsx
```diff
@@ -21,6 +25,55 @@ const warnings = (len: number) => ({
   Icon: FiZap,
 });
 
+const UpdateInfo = (): JSX.Element | null => {
```
Miniplop: All components should be React.FunctionComponent

packages/slice-machine/server/tsconfig.json
```diff
@@ -11,6 +11,7 @@
     "noUnusedParameters": true,
     "baseUrl": ".",
     "moduleResolution": "node",
+    "resolveJsonModule": true,
```
Miniplop, line 14: Why we need that ?

packages/slice-machine/src/modules/modal/types.ts
```diff
@@ -1,5 +1,6 @@
 export enum ModalKeysEnum {
   LOGIN = "LOGIN",
+  NEW_VERSION = "NEW_VERSION",
```
Miniplop: I think this key name should be coherent with the Modal name to make it more explicit, so I will change the modalName from "updateModal" => "newVersionModal"

packages/slice-machine/lib/utils/version.ts
```diff
@@ -0,0 +1,49 @@
+import path from "path";
```
Miniplop: I think we should avoid creating file in lib if the functions defined inside are used at only one place in the server. Create this file inside the services folder in the server folder.

packages/slice-machine/lib/utils/version.ts
```diff
@@ -0,0 +1,49 @@
...
+import semver from "semver";
+import fs from "fs";
+import fetch, { Response } from "node-fetch";
+import {
+  name as packageName,
+  version as installedVersion,
+} from "../../package.json";
+
+export class HTTPResponseError extends Error {
```
Miniplop: We don't have already a type for Error ????

packages/slice-machine/lib/utils/version.ts
```diff
@@ -0,0 +1,49 @@
...
+    throw new HTTPResponseError(res);
+  });
+}
+
+export function hasYarn() {
+  // could use lookpath to check yarn is installed as well ?
+  return fs.existsSync(path.join(process.cwd(), "yarn.lock"));
+}
+
+export async function checkVersion(): Promise<{
```
Miniplop: this function can be defined in this file : `packages/slice-machine/server/src/api/versions.ts`

packages/slice-machine/server/src/api/index.ts
```diff
@@ -222,6 +223,17 @@ router.post(
   }
 );
 
+router.get(
```
Miniplop: The /state route already return the current version and the last version of the slicemachine, why you didn't use it ?

packages/slice-machine/server/src/api/versions.ts
```diff
@@ -0,0 +1,27 @@
+import { hasYarn, checkVersion } from "@lib/utils/version";
+import { name } from "../../../package.json";
```
Miniplop: I think we can defined a name in a const instead of doing that. Package.json shouldn't be used in runtime

packages/slice-machine/server/src/api/versions.ts
```diff
@@ -0,0 +1,27 @@
...
+  | {
+      err: unknown;
+    };
+
+export default async function versions(): Promise<VersionInfo> {
+  try {
+    const details = await checkVersion();
+    const packageManager = hasYarn() ? "yarn" : "npm";
+    const updateCommand =
+      packageManager === "yarn" ? `yarn upgrade ${name}` : `npm i -D ${name}`;
```
Miniplop: Move this logic to the component

packages/slice-machine/src/apiClient.ts
```diff
@@ -47,3 +49,9 @@ export const sendTrackingOnboarding = (
     | OnboardingContinueWithVideoEvent
 ): Promise<AxiosResponse<TrackingReviewResponse>> =>
   axios.post(`/api/tracking/onboarding`, onboardingEvent, defaultAxiosConfig);
+
+/** Version info Route **/
+
+export function getVersionInfo(): Promise<AxiosResponse<VersionInfo>> {
```
Miniplop: Good but I think we don't need this route

packages/slice-machine/src/modules/update/index.ts
```diff
@@ -0,0 +1,36 @@
+import { Reducer } from "redux";
```
Miniplop: I don't think we need a state for that : 
- "current" property is already present in the environment modules
- "recent" property is already present in the environment modules
- "update command" doesn't need to be stored in a global store, as we can get if from the packageCommand, this command should be handle by the NewVersionCommand
- "packageManager" should be added to the environment variable.

Miniplop: Also I like the fact that the backend add an information on is an update is needed or not. So maybe we should add this to the env object too

packages/slice-machine/src/modules/useSliceMachineActions.ts
```diff
@@ -13,15 +13,26 @@ import {
...
+
+  const updateNotification = (payload: VersionInfo) =>
+    dispatch(getUpdateNotificationCreator(payload));
+
   // Modal store
   const closeLoginModal = () =>
     dispatch(modalCloseCreator({ modalKey: ModalKeysEnum.LOGIN }));
   const openLoginModal = () =>
     dispatch(modalOpenCreator({ modalKey: ModalKeysEnum.LOGIN }));
+  const closeUpdateModal = () =>
```
Miniplop: Name of the action creator should be coherent with the modal key : `closeNewVersionModal`

packages/slice-machine/src/modules/useSliceMachineActions.ts
```diff
@@ -13,15 +13,26 @@ import {
...
+    dispatch(modalCloseCreator({ modalKey: ModalKeysEnum.NEW_VERSION }));
+  const openUpdateModal = () =>
```
Miniplop: Name of the action creator should be coherent with the modal key : `openNewVersionModal`

packages/slice-machine/tests/components/Desktop.spec.tsx
```diff
@@ -0,0 +1,85 @@
+/**
```
Miniplop: I don't the value of this test because you create a new App and you test it. So it' not link to the real context

packages/slice-machine/tests/lib/utils/utils.test.js
```diff
@@ -0,0 +1,65 @@
+import {
```
Miniplop: good one :)

packages/slice-machine/tests/test-utils.tsx
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import { render as rtlRender, RenderOptions } from "@testing-library/react";
+import { Store, AnyAction } from "redux";
+
+// import { configureStore } from '@reduxjs/toolkit'
```
Miniplop: remove this line

packages/slice-machine/tests/test-utils.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { Provider } from "react-redux";
+import type { SliceMachineStoreType } from "../src/redux/type";
+import configureStore from "@src/redux/store";
+// import createReducer from '@src/redux/reducer'
```
Miniplop: remove this line

packages/slice-machine/tests/components/UpdateModal.spec.tsx
```diff
@@ -0,0 +1,47 @@
+/**
```
Miniplop: I like the idea of the test. But I'm not sure to see the value, we are testing that useSelector is working correctly here right ?

This test can only fail if the user don't display the update command or disconnect the text to the selector, this looks a pretty weak protection against regression. I don't think this component need test because there is no logic inside it except reacting to the props which is not very interesting to test.

packages/slice-machine/components/UpdateModal/index.tsx
```diff
@@ -0,0 +1,122 @@
+import React from "react";
```
Miniplop: I think this component should have a more explicit name like NewVersionModal or NewVersionAvailableModal

packages/slice-machine/components/AppLayout/Navigation/Menu/Desktop.tsx
```diff
@@ -21,6 +25,55 @@ const warnings = (len: number) => ({
...
+  const { openUpdateModal } = useSliceMachineActions();
+  const { update } = useSelector(getUpdateNotification);
+  if (!update) return null;
```
Miniplop: this should be in the parent component

Miniplop (changes requested):
I think there is architecture problem in this PR : 
- We don't reuse the current Env variable that already send the information for the SliceMachine Version. So we duplicated some data and that's not what we want.
- We create a new state for handling NewVersionModal component property, that should be avoid, we have to use the store if we need to share property between components and that's not the case

I'm available for help

packages/slice-machine/components/App/index.tsx
MarcMcIntosh: We don't want two modals popping up at the same time, i think it was part of the spec?

packages/slice-machine/server/tsconfig.json
MarcMcIntosh: package.json is imported on in the server to get the current version and name of the package

packages/slice-machine/tests/components/UpdateModal.spec.tsx
MarcMcIntosh: I could add more to this one to check user actions.

At them moment it checks that if given a condition it will display the popup, it's a condition we're unlikely to see during development. 

The main purpose of these tests are to implement some component based tests.
I added a readme that links to best practices, one of which was ignored here as it would involve testing the main App component.

packages/slice-machine/lib/utils/version.ts
MarcMcIntosh: Nope, it got removed during review

packages/slice-machine/server/src/api/index.ts
MarcMcIntosh: looks like the env pin doesn't provided the correct (or any info).
So it was a trade off of fixing that part of the env (and what else it impacts) or creating a replacement that works and then dealing with the issues in the `env`

```
{
  modal: {
    LOGIN: false,
    NEW_VERSION: false
  },
  loading: {
    LOGIN: false,
    REVIEW: false
  },
  userContext: {
    hasSendAReview: true,
    isOnboarded: true
  },
  environment: {
    warnings: [
      {
        key: 'STORYBOOK_NOT_RUNNING',
        title: 'Storybook is not running',
        description: 'Run `yarn run storybook` from your terminal'
      },
      {
        key: 'CLIENT_ERROR',
        title: 'CLIENT_ERROR:UNAUTHORIZED'
      }
    ],
    configErrors: {},
    env: {
      cwd: './tests/project',
      repo: 'sm-env-example',
      userConfig: {
        libraries: [
          '~/slices',
          '@/slices2/src/slices'
        ],
        apiEndpoint: 'https://sm-env-example.prismic.io/api/v2',
        storybook: 'http://localhost:6006',
        chromaticAppId: '....',
        _latest: '0.1.0'
      },
      hasConfigFile: true,
      prismicData: {
        base: 'https://wroom.io',
        auth: '....'
      },
      chromatic: {
        storybook: 'https://mm/prompt-to-update--5f5b34f06f304800225c4e17.chromatic.com',
        library: 'https://chromatic.com/library?appId=5f5b34f06f304800225c4e17&branch=mm/prompt-to-update'
      },
      currentVersion: '0.1.1',
      updateAvailable: {
        current: '',
        next: '',
        message: 'Could not fetch remote version'
      },
      mockConfig: {},
      hasGeneratedStoriesPath: true,
      framework: 'vue',
      baseUrl: 'http://localhost:9999',
      client: {
        cwd: './tests/project',
        base: 'https://wroom.io',
        repo: 'sm-env-example',
        auth: '...',
        images: {}
      }
    }
  },
  updateVersionInfo: {
    updateCommand: 'npm i -D slice-machine-ui',
    packageManager: 'npm',
    update: false,
    current: '0.1.1',
    recent: '0.1.1'
  },
  _persist: {
    version: -1,
    rehydrated: true
  }
}
```

packages/slice-machine/src/apiClient.ts
MarcMcIntosh: if we can fix the env we won't

packages/slice-machine/tests/components/Desktop.spec.tsx
MarcMcIntosh: yeah adding a large context is on my list of things to do, but now it checks that if the version is out of date the notification shows up (which could break when refactoring)

packages/slice-machine/lib/utils/version.ts
MarcMcIntosh: done

Miniplop (approved):
Thanks marc for your patience and your work, this PR is ready to be merged :)

PR #236, opened 2021-11-25
ci: cd add alpha deployment

MarcMcIntosh:
## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/deploy-alpha.yml
+++ b/.github/workflows/deploy-alpha.yml
@@ -22,7 +22,7 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
       - name: Install Dependencies
       - run: npm ci
-      - name: Version An Publish
+      - name: Version And Publish
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
       - run: |
--- a/.github/workflows/deploy-alpha.yml
+++ b/.github/workflows/deploy-alpha.yml
@@ -21,11 +21,11 @@ jobs:
           node-version: '16.x'
           registry-url: 'https://registry.npmjs.org'
       - name: Install Dependencies
-      - run: npm ci
+      run: npm ci
       - name: Version And Publish
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
-      - run: |
+      env:
+        NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      run: |
           git config user.name "${{ github.actor }}"
           git config user.email "${{ github.actor }}@users.noreply.github.com"
           lerna version prerelease --preid alpha --exact --yes
--- a/.github/workflows/deploy-alpha.yml
+++ b/.github/workflows/deploy-alpha.yml
@@ -16,16 +16,16 @@ jobs:
     steps: 
       - uses: actions/checkout@v2
       - name: Setup Node
-      - uses: actions/setup-node@v2
+        uses: actions/setup-node@v2
         with:
           node-version: '16.x'
           registry-url: 'https://registry.npmjs.org'
       - name: Install Dependencies
-      run: npm ci
+        run: npm ci
       - name: Version And Publish
-      env:
-        NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
-      run: |
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-alpha.yml
```diff
@@ -0,0 +1,33 @@
+name: deploy-alpha
+
+concurrency: alpha
+
+on:
+  pull_request:
```
Miniplop, line 6: why we are not deploying on every push on release branches ?

.github/workflows/deploy-alpha.yml
```diff
@@ -0,0 +1,33 @@
...
+    types: [closed]
+    branches:
+      - v[0-9]+.[0-9]+.[0-9]+
+
+jobs:
+  deployment:
+    if: github.event.pull_request.merged == true
```
Miniplop, line 13: I think this will return false except when the release branch will be merged in the master no ?

.github/workflows/deploy-alpha.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Setup Node
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Install Dependencies
+      - run: npm ci
+
+      - name: Bump Versions
```
Miniplop: npm login is not needed before ?

.github/workflows/deploy-alpha.yml
```diff
@@ -0,0 +1,33 @@
...
+    environment: alpha
+    steps: 
+      - uses: actions/checkout@v2
+      - name: Setup Node
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Install Dependencies
```
Miniplop, line 23: Why we don't use lerna bootsrap here, like in the test CI

.github/workflows/deploy-alpha.yml
MarcMcIntosh: I don't think it's needed for bumping versions, the token will be needed for pushing

.github/workflows/deploy-alpha.yml
MarcMcIntosh: Ah yes this part, when lerna does the version bump it'll push to the branch, it'll cause a chain reaction

.github/workflows/deploy-alpha.yml
MarcMcIntosh: When a pull request is close (for what ever reason) and the branch name matches the glob and the pull request was merged this will run :)
The merge check is there because lerna also makes some commits.... which I'll need to look into to make sure it has access

.github/workflows/deploy-alpha.yml
MarcMcIntosh: the bootstrap command is in the postinstall script ;)

.github/workflows/deploy-alpha.yml
Miniplop: ahhhh

.github/workflows/deploy-alpha.yml
Miniplop: Good :)

.github/workflows/deploy-alpha.yml
MarcMcIntosh: Added a token, I've read around a few places and the ci should be able to push updates to github too :)

.github/workflows/deploy-alpha.yml
```diff
@@ -20,14 +20,13 @@ jobs:
...
-
       - name: Install Dependencies
       - run: npm ci
-
-      - name: Bump Versions
-      - run: lerna version prerelease --preid alpha --exact --yes
-
-      - name: Publish Alpha Versions
-      - run: lerna publish from-package --dist-tag alpha --yes
+      - name: Version An Publish
```
Miniplop: And *

Miniplop (changes requested):
can we add a condition on this workflow to be run only if the test workflow is ok ?

MarcMcIntosh:
> can we add a condition on this workflow to be run only if the test workflow is ok ?
That's an option in the settings on this repo, I don't ave access to the settings but I have asked for `master` and `v[0-9]+.[0-9]+.[0-9]` to added to the `Require status checks before merging` option

Miniplop (approved):

Miniplop:
Let's start with that

PR #237, opened 2021-11-25
remove the client, cwd and the auth key from the server state

Miniplop:
## Types of changes

## Checklist:

### DIFF
--- a/packages/slice-machine/components/App/index.tsx
+++ b/packages/slice-machine/components/App/index.tsx
@@ -14,7 +14,7 @@ import ToastProvider from "../../src/ToastProvider";
 
 import LoginModal from "@components/LoginModal";
 import ReviewModal from "@components/ReviewModal";
-import { AppPayload, ServerState } from "@models/server/ServerState";
+import { ServerState } from "@models/server/ServerState";
 import useOnboardingRedirection from "@src/hooks/useOnboardingRedirection";
 import useServerState from "@src/hooks/useServerState";
 import UpdateVersionModal from "../UpdateVersionModal";
@@ -33,7 +33,6 @@ const RemoveDarkMode = ({ children }: { children: React.ReactElement }) => {
 type AppProps = {
   theme: () => Theme;
   serverState: ServerState | undefined;
-  payload: AppPayload | null;
   pageProps: any; // This is coming from next
   Component: (props: any) => JSX.Element;
   Renderer: (props: any) => JSX.Element;
@@ -42,7 +41,6 @@ type AppProps = {
 const SliceMachineApp: React.FunctionComponent<AppProps> = ({
   theme,
   serverState,
-  payload,
   pageProps,
   Component,
   Renderer,
@@ -70,34 +68,34 @@ const SliceMachineApp: React.FunctionComponent<AppProps> = ({
       <BaseStyles>
         <RemoveDarkMode>
           {!serverState ? (
-            <Renderer {...payload} />
+            <Renderer {...serverState} />
           ) : (
             <>
-              {!payload || !payload.libraries ? (
+              {!serverState || !serverState.libraries ? (
                 <Renderer
                   Component={Component}
                   pageProps={pageProps}
-                  {...payload}
+                  {...serverState}
                   openPanel={openPanel}
                 />
               ) : (
                 <ToastProvider>
                   <LibrariesProvider
-                    remoteSlices={payload.remoteSlices}
-                    libraries={payload.libraries}
-                    env={payload.env}
+                    remoteSlices={serverState.remoteSlices}
+                    libraries={serverState.libraries}
+                    env={serverState.env}
                   >
                     <CustomTypesProvider
-                      customTypes={payload.customTypes}
-                      remoteCustomTypes={payload.remoteCustomTypes}
+                      customTypes={serverState.customTypes}
+                      remoteCustomTypes={serverState.remoteCustomTypes}
                     >
                       <AppLayout>
-                        <SliceHandler {...payload}>
+                        <SliceHandler {...serverState}>
                           <Renderer
                             Component={Component}
                             pageProps={pageProps}
-                            {...payload}
                             openPanel={openPanel}
+                            {...serverState}
                           />
                           <Drawer
                             placement="right"
--- a/packages/slice-machine/components/LoginModal/index.tsx
+++ b/packages/slice-machine/components/LoginModal/index.tsx
@@ -44,10 +44,10 @@ const LoginModal: React.FunctionComponent = () => {
     useSliceMachineActions();
 
   const { addToast } = useToasts();
-  const prismicBase = !!env ? env.prismicData.base : "https://prismic.io";
+  const prismicBase = !!env ? env.prismicAPIUrl : "https://prismic.io";
   const loginRedirectUrl = !!env
     ? `${buildEndpoints(prismicBase).Dashboard.cliLogin}&port=${
-        new URL(env.baseUrl).port
+        new URL(env.sliceMachineAPIUrl).port
       }&path=/api/auth`
     : "";
 
--- a/packages/slice-machine/lib/models/common/Environment.ts
+++ b/packages/slice-machine/lib/models/common/Environment.ts
@@ -13,23 +13,22 @@ export interface UpdateVersionInfo {
   updateAvailable: boolean;
 }
 
-export default interface Environment {
+export default interface BackendEnvironment extends FrontEndEnvironment {
   cwd: string;
-  userConfig: UserConfig;
-  hasConfigFile: boolean;
-  repo?: string;
   prismicData: PrismicData;
+  client: DefaultClient | FakeClient;
+}
+
+export interface FrontEndEnvironment {
+  userConfig: UserConfig;
+  repo?: string;
   chromatic?: Chromatic;
   updateVersionInfo: UpdateVersionInfo;
   mockConfig: any;
   framework: Framework;
   baseUrl: string;
   hasGeneratedStoriesPath: boolean;
-  client: DefaultClient | FakeClient;
-}
-
-export type FrontEndEnvironment = {
   prismicData: {
     base: string;
   };
-} & Omit<Environment, "client" | "cwd" | "prismicData">;
+}
--- a/packages/slice-machine/lib/models/common/Environment.ts
+++ b/packages/slice-machine/lib/models/common/Environment.ts
@@ -1,6 +1,5 @@
 import PrismicData from "./PrismicData";
 import UserConfig from "./UserConfig";
-import Chromatic from "./Chromatic";
 import { Framework } from "./Framework";
 import DefaultClient from "./http/DefaultClient";
 import FakeClient from "./http/FakeClient";
@@ -13,7 +12,13 @@ export interface UpdateVersionInfo {
   updateAvailable: boolean;
 }
 
-export default interface BackendEnvironment extends FrontEndEnvironment {
+export default interface BackendEnvironment {
+  userConfig: UserConfig;
+  repo?: string;
+  updateVersionInfo: UpdateVersionInfo;
+  mockConfig: any;
+  framework: Framework;
+  baseUrl: string;
   cwd: string;
   prismicData: PrismicData;
   client: DefaultClient | FakeClient;
@@ -22,13 +27,9 @@ export default interface BackendEnvironment extends FrontEndEnvironment {
 export interface FrontEndEnvironment {
   userConfig: UserConfig;
   repo?: string;
-  chromatic?: Chromatic;
   updateVersionInfo: UpdateVersionInfo;
   mockConfig: any;
   framework: Framework;
-  baseUrl: string;
-  hasGeneratedStoriesPath: boolean;
-  prismicData: {
-    base: string;
-  };
+  sliceMachineAPIUrl: string;
+  prismicAPIUrl: string;
 }
--- a/packages/slice-machine/lib/models/common/Environment.ts
+++ b/packages/slice-machine/lib/models/common/Environment.ts
@@ -27,3 +27,9 @@ export default interface Environment {
   hasGeneratedStoriesPath: boolean;
   client: DefaultClient | FakeClient;
 }
+
+export type FrontEndEnvironment = {
+  prismicData: {
+    base: string;
+  };
+} & Omit<Environment, "client" | "cwd" | "prismicData">;
--- a/packages/slice-machine/lib/models/server/ServerState.ts
+++ b/packages/slice-machine/lib/models/server/ServerState.ts
@@ -24,11 +24,3 @@ export interface ServerState {
   warnings: ReadonlyArray<Warning>;
   isFake: boolean;
 }
-
-export interface AppPayload {
-  env: FrontEndEnvironment;
-  libraries?: ReadonlyArray<Library>;
-  customTypes?: ReadonlyArray<CustomType<ObjectTabs>>;
-  remoteCustomTypes?: ReadonlyArray<CustomType<ObjectTabs>>;
-  remoteSlices?: ReadonlyArray<Slice<AsObject>>;
-}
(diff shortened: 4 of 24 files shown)


packages/slice-machine/server/src/api/state.ts
```diff
@@ -139,4 +142,20 @@ export default async function handler() {
...
     warnings,
   };
+};
+
+export default async function handler(): Promise<ServerState> {
+  const state = await getBackendState();
+  const { client, cwd, prismicData, ...frontEnv } = state.env;
+  const frontEndEnv: FrontEndEnvironment = {
+    ...frontEnv,
+    prismicData: {
```
Heyrows: do we have to send PrismicData if it's just for sending the base ?
IF it's not many changes, could we change it ?

packages/slice-machine/lib/models/common/Environment.ts
```diff
@@ -13,17 +13,22 @@ export interface UpdateVersionInfo {
...
   userConfig: UserConfig;
-  hasConfigFile: boolean;
   repo?: string;
-  prismicData: PrismicData;
   chromatic?: Chromatic;
   updateVersionInfo: UpdateVersionInfo;
   mockConfig: any;
   framework: Framework;
   baseUrl: string;
   hasGeneratedStoriesPath: boolean;
```
Heyrows: I think this is unused except on the tests x)
image

packages/slice-machine/lib/models/common/Environment.ts
```diff
@@ -13,17 +13,22 @@ export interface UpdateVersionInfo {
...
+export interface FrontEndEnvironment {
   userConfig: UserConfig;
-  hasConfigFile: boolean;
   repo?: string;
-  prismicData: PrismicData;
   chromatic?: Chromatic;
   updateVersionInfo: UpdateVersionInfo;
   mockConfig: any;
   framework: Framework;
   baseUrl: string;
```
Heyrows: Not sure of what this is.
But i think either it's very similar to `PrismicData.base` and one of them can be removed.
From the only use I could find, i'd say we could just export something like "backendport".
image

packages/slice-machine/lib/models/common/Environment.ts
```diff
@@ -13,17 +13,22 @@ export interface UpdateVersionInfo {
...
+
+export interface FrontEndEnvironment {
   userConfig: UserConfig;
-  hasConfigFile: boolean;
   repo?: string;
-  prismicData: PrismicData;
   chromatic?: Chromatic;
   updateVersionInfo: UpdateVersionInfo;
   mockConfig: any;
   framework: Framework;
```
Heyrows, line 32: only used in "LackStoryBook" component, to consider removing after the preview stuff :D

packages/slice-machine/lib/models/common/Environment.ts
```diff
@@ -13,17 +13,22 @@ export interface UpdateVersionInfo {
...
+  prismicData: PrismicData;
+  client: DefaultClient | FakeClient;
+}
+
+export interface FrontEndEnvironment {
   userConfig: UserConfig;
-  hasConfigFile: boolean;
   repo?: string;
-  prismicData: PrismicData;
   chromatic?: Chromatic;
```
Heyrows: couldn't find any use for this.

packages/slice-machine/server/src/api/custom-types/push.ts
```diff
@@ -26,7 +26,7 @@ const createOrUpdate = (
 export default async function handler(query: { id: string }) {
   const { id } = query;
 
-  const state = await fetchState();
+  const state = await getBackendState();
```
Heyrows, line 29: like the renaming, makes it much clearer

packages/slice-machine/lib/models/common/Environment.ts
```diff
@@ -13,17 +13,22 @@ export interface UpdateVersionInfo {
   updateAvailable: boolean;
 }
 
-export default interface Environment {
+export default interface BackendEnvironment extends FrontEndEnvironment {
```
Heyrows: OPT: I think it should be a named export so that you import explicitly either the Backend or Frontend Env

packages/slice-machine/lib/models/server/ServerState.ts
```diff
@@ -16,15 +16,17 @@ export interface ConfigErrors {
 export interface ServerState {
   libraries: ReadonlyArray<Library>;
   customTypes: ReadonlyArray<CustomType<ObjectTabs>>;
+  remoteCustomTypes: ReadonlyArray<CustomType<ObjectTabs>>;
   remoteSlices: ReadonlyArray<Slice<AsObject>>;
-  clientError?: ErrorWithStatus;
-  configErrors: ConfigErrors;
-  env: Environment;
+  clientError: ErrorWithStatus | undefined;
+  configErrors: ConfigErrors | undefined;
```
Heyrows: the type changes, are you sure this is right ?
I never saw configErrors as optional, I thin I saw it's an object wth keys, so i think it can be an empty object.

packages/slice-machine/src/modules/useSliceMachineActions.ts
```diff
@@ -48,7 +48,9 @@ const useSliceMachineActions = () => {
       getEnvironmentCreator({
         env: serverState.env,
         warnings: serverState.warnings,
-        configErrors: serverState.configErrors,
+        configErrors: !!serverState.configErrors
```
Heyrows: I think you adapted this part to always have an object for this but you also changed the type of serverState yourself.
And I didn't saw any changes when affecting the state itself.

packages/slice-machine/server/src/api/custom-types/push.ts
Miniplop: Good

packages/slice-machine/lib/models/common/Environment.ts
Miniplop: Good catch ✅

packages/slice-machine/server/src/api/state.ts
Miniplop: Done :)

packages/slice-machine/lib/models/common/Environment.ts
Miniplop: This 2 are not the same, prismicData.base contains the prismic API url and baseUrl contain the SM API endpoint, so we cannot merge them into one, but at least we can baseUrl to slicemachineAPIUrl and the prismicData.baseUrl to prismicAPIUrl

packages/slice-machine/lib/models/common/Environment.ts
Miniplop: 👍

packages/slice-machine/lib/models/common/Environment.ts
Miniplop: Like this

image

packages/slice-machine/lib/models/common/Environment.ts
Miniplop: Me neither, I delete all the code associated to this

packages/slice-machine/lib/models/common/Environment.ts
Miniplop: I removed the inherit between the two types, I think the idea is to have two separates envs right ?

packages/slice-machine/lib/models/server/ServerState.ts
Miniplop: Client error are coming from this function : 

```
export default async function handler(env: BackendEnvironment): Promise<{
  remoteSlices: ReadonlyArray<Slice<AsObject>>;
  clientError: ErrorWithStatus | undefined;
  libraries: ReadonlyArray<Library>;
}>
```

So before the type doesn't match well the reality. 


For the config errors, the getEnv return type told that the function can return `configErrors: ConfigErrors | undefined;` but this is not the case the reality. So I changed the returnType and force the serverState to `configErrors: ConfigErrors` ✅


Good for you ?

packages/slice-machine/src/modules/useSliceMachineActions.ts
Miniplop: Fixed :)

packages/slice-machine/components/App/index.tsx
```diff
@@ -33,7 +33,6 @@ const RemoveDarkMode = ({ children }: { children: React.ReactElement }) => {
 type AppProps = {
   theme: () => Theme;
   serverState: ServerState | undefined;
-  payload: AppPayload | null;
   pageProps: any; // This is coming from next
   Component: (props: any) => JSX.Element;
   Renderer: (props: any) => JSX.Element;
```
Heyrows, line 38: I think any can be replace because we spread the ServerState actually.

Heyrows: with Components, PageProps and OpenPanel as optional it seems.

Heyrows: Anyway it seems maybe a bit too big for this PR that isn't meant for that. So OPTIONAL

packages/slice-machine/components/LoginModal/index.tsx
```diff
@@ -44,10 +44,10 @@ const LoginModal: React.FunctionComponent = () => {
     useSliceMachineActions();
 
   const { addToast } = useToasts();
-  const prismicBase = !!env ? env.prismicData.base : "https://prismic.io";
+  const prismicBase = !!env ? env.prismicAPIUrl : "https://prismic.io";
```
Heyrows, line 47: i like that 😄

packages/slice-machine/lib/models/common/Environment.ts
```diff
@@ -13,17 +12,24 @@ export interface UpdateVersionInfo {
...
   mockConfig: any;
   framework: Framework;
   baseUrl: string;
-  hasGeneratedStoriesPath: boolean;
+  cwd: string;
+  prismicData: PrismicData;
   client: DefaultClient | FakeClient;
 }
+
+export interface FrontEndEnvironment {
```
Heyrows, line 27: Should extend BackendEnv with the specific properties you want :)

Heyrows: Or have a common Env type maybe.
But tbh I don't know if it makes sense, but that would definitely makes it clearer.

packages/slice-machine/lib/models/server/ServerState.ts
```diff
@@ -16,17 +16,11 @@ export interface ConfigErrors {
...
   remoteSlices: ReadonlyArray<Slice<AsObject>>;
-  clientError?: ErrorWithStatus;
+  clientError: ErrorWithStatus | undefined;
   configErrors: ConfigErrors;
-  env: Environment;
+  env: FrontEndEnvironment;
   warnings: ReadonlyArray<Warning>;
-}
-
-export interface AppPayload {
```
Heyrows, line 26: nice one

packages/slice-machine/lib/models/server/ServerState.ts
```diff
@@ -16,17 +16,11 @@ export interface ConfigErrors {
 export interface ServerState {
```
Heyrows, line 16: I hope the Preview PR doesn't use stuff that we removed here :x

packages/slice-machine/pages/_app.tsx
```diff
@@ -20,30 +20,14 @@ import "react-datepicker/dist/react-datepicker.css";
 import "src/css/modal.css";
 import "src/css/tabs.css";
 
-import { AppPayload, ServerState } from "@lib/models/server/ServerState";
+import { ServerState } from "@lib/models/server/ServerState";
 import ServerError from "@lib/models/server/ServerError";
-import { Library } from "@lib/models/common/Library";
 import Head from "next/head";
 
 async function fetcher(url: string): Promise<any> {
```
Heyrows: only used once, not sure we need it.

packages/slice-machine/pages/_app.tsx
```diff
@@ -20,30 +20,14 @@ import "react-datepicker/dist/react-datepicker.css";
 import "src/css/modal.css";
 import "src/css/tabs.css";
```
Heyrows, line 21: Hidden but there is a `ts-ignore` on this file. I guess lots of types are not defined yet but at some point we should consider it :)

packages/slice-machine/server/src/api/services/getEnv.ts
```diff
@@ -65,49 +62,9 @@ function extractRepo(parsedRepo: ParseResult): string | undefined {
   }
 }
 
-function handleBranch(): Promise<{ branch?: string; err?: Error }> {
```
Heyrows, line 68: i'm not sure what this does.

packages/slice-machine/pages/_app.tsx
```diff
@@ -76,27 +60,16 @@ function MyApp({
     fetcher
   );
 
-  const [sliceMap, setSliceMap] = useState<any | null>(null);
```
Heyrows, line 79: I asked Hugo about deleting that.
He explained me where it is used and why.

Basically each slice has a personal useState, so React which does not allow conditional hooks needs this.

I think this needs more thinking than just removing this code.

Heyrows (changes requested):

packages/slice-machine/components/LoginModal/index.tsx
Miniplop: Good

packages/slice-machine/lib/models/server/ServerState.ts
Miniplop: 👍

packages/slice-machine/pages/_app.tsx
Miniplop: I can check with Hugo but, this state isn't shared anywhere, so I think this is not used anymore.

The store for every slice is part of the librairiesContext construction and that not use this sliceMap.

packages/slice-machine/server/src/api/services/getEnv.ts
Miniplop: Checking some ref in github for chromatic. But as chromatic is not used anymore, we should remove this function

packages/slice-machine/pages/_app.tsx
Miniplop: Yes, there are many ts-ignore hidden in the code

packages/slice-machine/pages/_app.tsx
Miniplop: You mean we don't need to declare a function for it ? 

We had the same discussion for the loginModal and we choose to create an external function. This case is the same, so we should use the same tech strat, to stay coherent. If we want to remove the function in this case, we should also change the LoginModal.

What do you want to do ?

packages/slice-machine/lib/models/server/ServerState.ts
Miniplop: There will be conflicts because the AppPayload was renamed and changed in the PreviewBranch. But I think I can handle the conflicts :)

packages/slice-machine/components/App/index.tsx
Miniplop: I also removed the spread of the serverState in the renderer in a another PR, so we should change this on the PR where I removed the server State.

arnaudlewis: The server state now extends the AppState. AppState should be the only state that we share in the UI, the server state should remain on server side or on client side when we fetch it and then converted to AppState. (cf: preview-experimental branch)

packages/slice-machine/lib/models/common/Environment.ts
arnaudlewis: It depends on what's needed in these two, unless they need all the same data, you shouldn't have extra data on the front that is useful only for the server or the other way around so 2 types makes sense if that's the case

packages/slice-machine/server/src/api/services/getEnv.ts
arnaudlewis: if so, remove also all references of chromatic, in the manifest, env etc

packages/slice-machine/pages/_app.tsx
Miniplop: @hypervillain Your help here :)

Heyrows: Debt*

Heyrows (approved):

PR #243, opened 2021-12-01
add extra deps for nextjs and svelte

MarcMcIntosh:
## Types of changes

## Checklist:

### DIFF
--- a/packages/init/src/steps/install-required-dependencies.ts
+++ b/packages/init/src/steps/install-required-dependencies.ts
@@ -2,25 +2,25 @@ import path from "path";
 import { execCommand } from "../utils";
 import { Utils, FileSystem } from "@slicemachine/core";
 
+const {
+  PRISMIC_CLIENT,
+  PRISMIC_DOM_PACKAGE_NAME,
+  PRISMIC_REACT_PACKAGE_NAME,
+  NEXT_SLICEZONE,
+  SM_PACKAGE_NAME,
+} = Utils.CONSTS;
+
 function depsForFramework(framework: Utils.Framework.FrameworkEnum): string {
-  const packages = [];
-
-  if (
-    framework === Utils.Framework.FrameworkEnum.react ||
-    framework === Utils.Framework.FrameworkEnum.next
-  ) {
-    packages.push(Utils.CONSTS.PRISMIC_REACT_PACKAGE_NAME, "@prismicio/client");
+  switch (framework) {
+    case Utils.Framework.FrameworkEnum.react:
+      return `${PRISMIC_REACT_PACKAGE_NAME} ${PRISMIC_CLIENT}`;
+    case Utils.Framework.FrameworkEnum.next:
+      return `${PRISMIC_REACT_PACKAGE_NAME} ${PRISMIC_CLIENT} ${NEXT_SLICEZONE}`;
+    case Utils.Framework.FrameworkEnum.svelte:
+      return `${PRISMIC_DOM_PACKAGE_NAME} ${PRISMIC_CLIENT}`;
+    default:
+      return "";
   }
-
-  if (framework === Utils.Framework.FrameworkEnum.next) {
-    packages.push("next-slicezone");
-  }
-
-  if (framework === Utils.Framework.FrameworkEnum.svelte) {
-    packages.push(Utils.CONSTS.PRISMIC_DOM_PACKAGE_NAME, "@prismicio/client");
-  }
-
-  return packages.join(" ");
 }
 
 export async function installRequiredDependencies(
@@ -37,7 +37,7 @@ export async function installRequiredDependencies(
   spinner.start();
 
   const { stderr } = await execCommand(
-    `${installDevDependencyCommand} ${Utils.CONSTS.SM_PACKAGE_NAME}`
+    `${installDevDependencyCommand} ${SM_PACKAGE_NAME}`
   );
 
   const deps = depsForFramework(framework);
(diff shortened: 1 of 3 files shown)


packages/init/src/steps/install-required-dependencies.ts
```diff
@@ -2,6 +2,27 @@ import path from "path";
...
+    framework === Utils.Framework.FrameworkEnum.next
+  ) {
+    packages.push(Utils.CONSTS.PRISMIC_REACT_PACKAGE_NAME, "@prismicio/client");
+  }
+
+  if (framework === Utils.Framework.FrameworkEnum.next) {
+    packages.push("next-slicezone");
+  }
+
+  if (framework === Utils.Framework.FrameworkEnum.svelte) {
```
Heyrows: can be put with react and next :)

packages/init/src/steps/install-required-dependencies.ts
```diff
@@ -2,6 +2,27 @@ import path from "path";
...
+
+  if (
+    framework === Utils.Framework.FrameworkEnum.react ||
+    framework === Utils.Framework.FrameworkEnum.next
+  ) {
+    packages.push(Utils.CONSTS.PRISMIC_REACT_PACKAGE_NAME, "@prismicio/client");
+  }
+
+  if (framework === Utils.Framework.FrameworkEnum.next) {
+    packages.push("next-slicezone");
```
Heyrows: OPT: the packages could be all Consts.

packages/init/src/steps/install-required-dependencies.ts
```diff
@@ -2,6 +2,27 @@ import path from "path";
...
 import { Utils, FileSystem } from "@slicemachine/core";
 
+function depsForFramework(framework: Utils.Framework.FrameworkEnum): string {
+  const packages = [];
+
+  if (
+    framework === Utils.Framework.FrameworkEnum.react ||
+    framework === Utils.Framework.FrameworkEnum.next
+  ) {
+    packages.push(Utils.CONSTS.PRISMIC_REACT_PACKAGE_NAME, "@prismicio/client");
```
Miniplop: can you create const for all the deps ?

packages/init/src/steps/install-required-dependencies.ts
```diff
@@ -2,6 +2,27 @@ import path from "path";
 import { execCommand } from "../utils";
 import { Utils, FileSystem } from "@slicemachine/core";
 
+function depsForFramework(framework: Utils.Framework.FrameworkEnum): string {
+  const packages = [];
```
Heyrows: OPT: I think a switch with the frameworks would be better, to have a clearer Code.

```
switch(framework) {
  case(Framework.next): ([SM_PACKAGE_UI, PRISMIC_PRISMIC_PACKAGE_REACT, PRISMIC_PACKAGE_CLIENT]).join(" ")
}
```

Or something similar 😄

packages/init/src/steps/install-required-dependencies.ts
```diff
@@ -19,19 +40,8 @@ export async function installRequiredDependencies(
...
-      );
-      break;
-    case Utils.Framework.FrameworkEnum.svelte:
-      await execCommand(
-        `${installDependencyCommand} ${Utils.CONSTS.PRISMIC_DOM_PACKAGE_NAME}`
-      );
-      break;
-  }
+  const deps = depsForFramework(framework);
+  if (deps) await execCommand(`${installDependencyCommand} ${deps}`);
```
Heyrows, line 44: I'm wondering if we should install with the `@latest` tag all the time 🤷‍♂️

Heyrows, line 44: I know this wasn't in the spec, but do you think we could/should put a spinner or a message to let the user know an action is on going.

packages/init/src/steps/install-required-dependencies.ts
```diff
@@ -10,7 +31,7 @@ export async function installRequiredDependencies(
   const installDevDependencyCommand = yarnLock
     ? "yarn add -D"
     : "npm install --save-dev";
-  const installDependencyCommand = yarnLock ? "yarn add" : "npm install";
+  const installDependencyCommand = yarnLock ? "yarn add" : "npm install --save";
 
   const spinner = Utils.spinner("Downloading Prismic Visual Builder");
```
Heyrows, line 36: OPT: I think it could be good to split this function into two
- installDevDeps
- installDeps
- installRequiredDeps (that call the two above)

This would declare two functions clear that each display their owner spinner or create their own command based on the packageManager passed as parameter.
I think it would make the code easier to test.
But I acknowledge that this would require work on the test and all.
That's why my comment is optional 😄

packages/init/src/steps/install-required-dependencies.ts
MarcMcIntosh: switches fall through, i did try them. I guess it can have unique cases that return.

packages/init/src/steps/install-required-dependencies.ts
MarcMcIntosh: not any more :)

packages/init/src/steps/install-required-dependencies.ts
MarcMcIntosh: Maybe for QA it would be nice for them to have a flag to state tagged versions

MarcMcIntosh: There's already a spinner, 
if we use execSync the output from the command can be piped to stdout

packages/init/src/steps/install-required-dependencies.ts
Miniplop: I think the switch makes clear what are the dependencies by framework but create some redundant code. What is the most to you @MarcMcIntosh @Heyrows ?

Heyrows (approved):

PR #247, opened 2021-12-02
support tracking for authenticated and anonymous user

arnaudlewis:


packages/core/src/core/auth.ts
```diff
@@ -1,7 +1,8 @@
 import * as hapi from "@hapi/hapi";
 import open from "open";
 import { CONSTS, bold, underline, spinner, writeError } from "../utils";
-import { setAuthConfig } from "../filesystem";
+import { PrismicSharedConfigManager } from "../filesystem/PrismicSharedConfig";
```
Miniplop, line 4: Why changing the name of this services ? As this file handle only the auth information, I found that sharedConfig doesn't help to understand what is inside the file.

packages/core/src/core/auth.ts
```diff
@@ -131,7 +132,10 @@ export async function startServerAndOpenBrowser(
 
   function onSuccess(data: HandlerData) {
     s.succeed(`Logged in as ${bold(data.email)}`).stop();
-    setAuthConfig(data.cookies, base);
+    PrismicSharedConfigManager.setProps({
```
Miniplop: setProps is not clear, especially in a React Ecosystem "props" mean other thing. Maybe something like setProperties, setContent.

packages/slice-machine/lib/models/common/tracker.ts
```diff
@@ -1,38 +1,93 @@
 import Analytics from "analytics-node";
```
Miniplop: I don't think this file should be in the lib as it's only a backend service, we should put it in the server/src/api/services instead. Only stuff shared between the front and the back should be added to this folder.

packages/slice-machine/lib/models/common/tracker.ts
```diff
@@ -1,38 +1,93 @@
...
 import { LibraryUI } from "./LibraryUI";
 
-export default class Tracker {
+export interface TrackerOptions {
```
Miniplop: Analytics-node doesn't have a type for that ?

packages/slice-machine/lib/models/common/tracker.ts
```diff
@@ -1,38 +1,93 @@
...
+    console.log({ credentials });
 
-    this.analytics = new Analytics(key);
+    this.analytics = new Analytics(key, this.options);
     this.slicemachineVersion = smVersion || "default"; //default will be logged if we can't parse package.json
   }
 
-  Repository = (repo: string | void) => {
-    if (!repo) return; // we can't track as long as we're not tied to a repo.
+  build: () => Tracker = () => {
```
Miniplop: As we have a class why don't expose directly the functions `resolveUser` and `libraries` ? Instead of retruning a "fake tracker" w can throw error or early return on every functions no ?

packages/slice-machine/lib/models/common/tracker.ts
```diff
@@ -1,38 +1,93 @@
...
+          (err: any, data?: unknown) => {
+            console.log({ err, data: JSON.stringify(data) });
+          }
+        );
       },
     };
   };
 }
+
+export interface Tracker {
```
Miniplop: I think this interface is useless if we don't use a build function.

packages/slice-machine/server/src/api/http/common.ts
```diff
@@ -0,0 +1,52 @@
...
+      {
+        ...(userId ? { userId } : {}),
+        ...(!userId && anonymousId ? { anonymousId } : {}),
+      },
+      env.repo,
+      {
+        flushAt: 20,
+        flushInterval: 300,
+      }
+    ).build();
```
Miniplop, line 40: this is redundant with the constructor call, I think the constructor can handle the logic behind the build

packages/slice-machine/server/src/api/state.ts
```diff
@@ -68,7 +68,7 @@ export default async function handler() {
         Math.floor(newTokenResponse.status / 100) === 2
       ) {
         const newtToken = await newTokenResponse.text();
```
Miniplop, line 70: little typo here

packages/slice-machine/server/src/api/state.ts
```diff
@@ -48,8 +48,8 @@ export async function createWarnings(
   ) as ReadonlyArray<Warning>;
 }
 
-export default async function handler() {
-  const { env, errors: configErrors } = await getEnv();
+export default async function handler(req: RequestWithEnv) {
+  const { errors: configErrors, env } = req;
```
Miniplop, line 52: Most of this function has changed in the release v0.1.3 branch, so I think we will have synchronization to do in this part :p

packages/slice-machine/server/src/api/http/common.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    const reqWithEnv = (() => {
```
Miniplop, line 42: I think we mush avoid modifying the request sent to pass some tool to the backend, the request object is not supposed to do that. We can create a context and pass the context to every controller that need it.

packages/core/src/core/auth.ts
arnaudlewis: because actually it does more than this,  it handles the base which is not specifically auth oriented and same for the userId

packages/core/src/core/auth.ts
arnaudlewis: ok

packages/slice-machine/lib/models/common/tracker.ts
arnaudlewis: I still want a tracker anyway because at some point we'll track even if we're not repo related but I don't want to throw an error if something is missing. So in the meantime I did it this way but it's not ideal. And I don't want conditions on params on every single tracking function. It definitely needs some refinement

packages/slice-machine/lib/models/common/tracker.ts
arnaudlewis: It does but I don't want to expose all of em

packages/slice-machine/lib/models/common/tracker.ts
arnaudlewis: true if no build, let's discuss it above

packages/slice-machine/server/src/api/state.ts
arnaudlewis: ? it's not even a change, I don't see the typo xD

packages/slice-machine/server/src/api/state.ts
arnaudlewis: no problem :)

packages/slice-machine/server/src/api/http/common.ts
arnaudlewis: changing the request is definitely the node approach but instead of making a mutation in a middleware we simply defined a subType of request to explicitly expose more to the handler.

arnaudlewis: With regular handler, you don't get an extra context and using the request to enrich your handler seems to be the current way of doing it in node.

packages/slice-machine/lib/models/common/tracker.ts
Miniplop: So maybe we can use Pick from Typescript ;)

arnaudlewis: nope Pick is used on a specific type, here it's defined this way:

arnaudlewis:
```
class Analytics {
  constructor(writeKey: string, opts?: {
    flushAt?: number | undefined,
    flushInterval?: number | undefined,
    host?: string | undefined,
    enable?: boolean | undefined,
    timeout?: number | string | undefined,
  });
```

PR #259, opened 2021-12-09
[Draft] Download lib with the init

Miniplop:
## Types of changes

## Checklist:


packages/init/src/index.ts
```diff
@@ -49,8 +52,14 @@ async function init() {
   // install the required dependencies in the project.
   await installRequiredDependencies(cwd, frameworkResult.value);
 
+  const sliceLibPath = await (() => {
```
Miniplop: We can change this notation by 
 
`
const sliceLibPath = lib ? await installLib(cwd, lib, branch) : null
`

The default case is explicit and we avoid the IIFE overload

packages/init/src/steps/install-lib.ts
```diff
@@ -0,0 +1,110 @@
...
+    const pkgJson = PackageJsonHelper.fromPath(
+      path.join(projectPath, "package.json")
+    );
+    if (pkgJson instanceof Error) throw pkgJson;
+
+    const dependencies = Dependencies.fromPkgFormat(pkgJson.dependencies);
+    if (dependencies) await pkgManager.install(dependencies);
+
+    // generate meta file
+    Files.write(path.join(libDestinationFolder, "meta.json"), {
```
Miniplop, line 79: what is this file for ? What happen if the user doesn't have this file or delete it ?

packages/init/src/steps/install-lib.ts
```diff
@@ -0,0 +1,110 @@
...
+      ({ path }) => {
+        return `~/${name}/${path.replace("src/", "")}`;
+      }
+    );
+
+    spinner.succeed(
+      `Slice library "${libGithubPath}" was installed successfully`
+    );
+
+    // If the libs have multiple slice libs here we are grouping them into one
```
Miniplop: not true anymore

packages/init/src/steps/install-lib.ts
arnaudlewis: Used to identify the lib that was downloaded so we can track it on amplitude later on, on SM

packages/init/src/steps/install-lib.ts
arnaudlewis: true

packages/init/src/index.ts
arnaudlewis: it's undefined not null here but yeah if you prefer

PR #271, opened 2021-12-14
Fix: Puppeteer waits for root element to display and not for network …

Heyrows:
…idle

## Types of changes

## Description

Issue: We can't take screenshot with a Nuxt project.

After adding some logs and testing, it appears that we used to wait non active network (all the connections are done and closed), while Nuxt keeps a socket open for webpack's hotreload.
We choose to wait until a maximum of two connections are left open (an option of puppeteer).

On top of that, while tested we noticed that after two screenshot, it stopped working.
We realised that the puppeteer pages were never closed which means they remained open and with them the sockets.
We now close the page after each screenshot, preventing an endless amount of pages to be opened.

Extra:
- On slice creation, the screenshot was not generated because the library state was generated after trying to take the actual screenshot.
 -> We generate the state before now.
- On slice save, we could generate the screenshots but they weren't displayed until the page reloads.
 -> We fixed the front-end state when a save is performed, it was still adapted to the old models.

## Checklist:

### DIFF
--- a/packages/slice-machine/server/src/api/screenshots/puppeteer.ts
+++ b/packages/slice-machine/server/src/api/screenshots/puppeteer.ts
@@ -44,8 +44,12 @@ const generateScreenshot = async (
   try {
     Files.mkdir(path.dirname(pathToFile), { recursive: true });
     const page = await browser.newPage();
+
+    /* We use the waitUntil option in order for the component to be rendered properly.
+     ** The value networkidle2 is required because Nuxt has an open socket with Webpack.
+     */
     await page.goto(screenshotUrl, {
-      waitUntil: "networkidle2", // websocket open for webpack in nuxt.
+      waitUntil: "networkidle2",
     });
 
     await page.waitForSelector("#root", { timeout: 2000 });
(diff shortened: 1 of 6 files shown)


Miniplop (approved):

hypervillain:
- "Nuxt has a websocket so we are more flexible on the network now": I have zero clue what this means
- "the library state is now generated after trying to take the screenshots" does not explain why
- "to avoid having a lot of pages opens on the puppeteer browser" is probably related to websockets but how?

packages/slice-machine/server/src/api/screenshots/puppeteer.ts
```diff
@@ -44,15 +44,14 @@ const generateScreenshot = async (
   try {
     Files.mkdir(path.dirname(pathToFile), { recursive: true });
     const page = await browser.newPage();
-    await page.goto(screenshotUrl);
-
-    await page.waitForNavigation({
-      waitUntil: "networkidle0",
+    await page.goto(screenshotUrl, {
+      waitUntil: "networkidle2", // websocket open for webpack in nuxt.
```
hypervillain: This comment does not really give information on why it's that way now

hypervillain (changes requested):

hypervillain (approved):
PR description was really helpful, thanks!

PR #272, opened 2021-12-14
Check that the preview is correctly setup

Miniplop:
## Types of changes

## Checklist:


Miniplop:
We should change the condition to trigger the reverse proxy to avoid conflict with local user ENV variable

packages/slice-machine/lib/builders/SliceBuilder/SideBar/index.tsx
```diff
@@ -112,7 +62,7 @@ const SideBar: React.FunctionComponent<SideBarProps> = ({
           imageLoading={imageLoading}
           onScreenshot={onScreenshot}
           onHandleFile={onHandleFile}
-          preventScreenshot={!isPreviewSetup}
+          preventScreenshot={false}
```
Miniplop: maybe we can remove this props

packages/slice-machine/src/modules/loading/index.ts
```diff
@@ -50,3 +51,20 @@ export const loadingReducer: Reducer<LoadingStoreType, LoadingActions> = (
...
   }
 };
+
+// Saga decorator
+export const withLoader = (saga: any, loadingKey: LoadingKeysEnum) =>
+  function* (...args: any[]) {
+    yield put(startLoadingActionCreator({ loadingKey }));
+    yield call(
+      saga,
+      args[0],
```
Miniplop, line 62: maybe we can use ...args here

packages/slice-machine/src/modules/preview/index.ts
```diff
@@ -0,0 +1,277 @@
...
+          isOpen: false,
+        },
+      };
+    default:
+      return state;
+  }
+};
+
+// Sagas
+export function* checkSetupSaga(
```
Miniplop: export useless here

packages/slice-machine/src/modules/preview/index.ts
```diff
@@ -0,0 +1,277 @@
...
+
+    // All the backend checks are ok ask for the frontend Iframe check
+    if ("ok" === setupStatus.manifest && "ok" === setupStatus.dependencies) {
+      yield put(
+        checkPreviewSetupCreator.success({
+          setupStatus: { iframe: null, ...setupStatus },
+        })
+      );
+      yield put(connectToPreviewIframeCreator.request());
+      const { timeout, iFrameCheckKO, iFrameCheckOk } = yield race({
```
Miniplop: typo iframe

packages/slice-machine/components/Preview/components/IframeRenderer/index.tsx
```diff
@@ -49,31 +50,50 @@ type IframeRendererProps = {
...
+      });
   }, [client, size, sliceView]);
 
   return (
-    <Box sx={{ flex: "1", bg: "grey01" }}>
+    <Box
+      sx={{
+        flex: "1",
+        bg: "grey01",
+        ...(dryRun ? { visibility: "hidden" } : {}),
```
Heyrows, line 97: if we want to render it without showing it, I think more than the viisibility, the size should be changed too.

packages/slice-machine/components/Preview/index.tsx
```diff
@@ -25,13 +31,10 @@ export default function Preview() {
     return <div />;
   }
 
-  const sliceView: SliceView = [
-    { sliceID: Model.infos.model.id, variationID: variation.id },
-  ];
-
-  const canvasUrl = `http://localhost:${
```
Heyrows, line 52: good catch

packages/slice-machine/lib/builders/SliceBuilder/SetupDrawer/components/StepSection.tsx
```diff
@@ -16,13 +18,22 @@ const StepSection: React.FunctionComponent<StepSectionProps> = ({
...
+  status = null,
 }) => {
   const stepSectionContainer = useRef<HTMLDivElement>(null);
   const contentHeight: number =
     !!stepSectionContainer && !!stepSectionContainer.current
       ? stepSectionContainer.current.scrollHeight
       : 0;
 
+  const additionalStepTitleStyle: ThemeUIStyleObject =
+    status === "ok"
```
Heyrows, line 30: lots of time the statuses are plain string, this could be typed.

packages/slice-machine/lib/builders/SliceBuilder/SetupDrawer/NuxtSetupSteps.tsx
```diff
@@ -4,101 +4,172 @@ import { Flex, Text } from "theme-ui";
...
+    userHasAtLeastOneStepMissing,
+    userHasConfiguredAllSteps,
+    isCheckingSetup,
+  } = useSelector((state: SliceMachineStoreType) => ({
+    openedStep: selectOpenedStep(state),
+    setupStatus: selectSetupStatus(state),
+    isCheckingSetup: isLoading(state, LoadingKeysEnum.CHECK_PREVIEW),
+    userHasAtLeastOneStepMissing: selectUserHasAtLeastOneStepMissing(state),
+    userHasConfiguredAllSteps: selectUserHasConfiguredAllSteps(state),
+  }));
```
Heyrows: lots of redundancy between next and nuxt, do you think there is somethng to do or the two drawers are too different ?

packages/slice-machine/lib/builders/SliceBuilder/SetupDrawer/NuxtSetupSteps.tsx
```diff
@@ -4,101 +4,172 @@ import { Flex, Text } from "theme-ui";
...
+import SuccessSection from "./components/SuccessSection";
 
-type NuxtSetupStepProps = {
-  activeStep: number;
-  onOpenStep: (stepNumber: number) => () => void;
-};
+const NuxtSetupSteps: React.FunctionComponent = () => {
+  const { toggleSetupDrawerStep, checkPreviewSetup } = useSliceMachineActions();
+
+  const {
```
Heyrows: feels like all these states are too complicated, i was expecting something simple like an object with stepId / name and the status.

packages/slice-machine/lib/builders/SliceBuilder/SideBar/index.tsx
Heyrows: ➕ yes we can, it was used to disable the button.

packages/slice-machine/server/src/api/preview.ts
```diff
@@ -0,0 +1,65 @@
...
+import { previewIsSupported } from "@lib/utils";
+import { RequestWithEnv } from "./http/common";
+import { Frameworks } from "@slicemachine/core/build/src/models/Framework";
+import { PreviewCheckResponse } from "@models/common/Preview";
+
+export default async function handler(
+  req: RequestWithEnv
+): Promise<PreviewCheckResponse> {
+  const cwd = process.env.CWD || process.cwd();
+  let response: PreviewCheckResponse = {
```
Heyrows, line 11: no let please 🙏

packages/slice-machine/src/modules/environment/index.ts
```diff
@@ -29,9 +29,18 @@ export const getEnvironment = (
   store: SliceMachineStoreType
 ): FrontEndEnvironment | null => store.environment.env;
 
-export const selectIsThePreviewSetUp = (
-  store: SliceMachineStoreType
-): boolean => !!store.environment.env?.manifest.localSliceCanvasURL;
+export const selectCanvasUrl = (store: SliceMachineStoreType): string => {
+  const defaultCanvasUrl = `http://localhost:${
```
Heyrows: I think we shouldn't have a default url.
It should just take what's coming from the manifest.

packages/slice-machine/src/modules/preview/index.ts
```diff
@@ -0,0 +1,277 @@
...
+  setupStatus: {
+    manifest: null,
+    iframe: null,
+    dependencies: null,
+  },
+  setupDrawer: {
+    isOpen: false,
+    openedStep: NoStepSelected,
+  },
+  isWaitingForIframeCheck: false,
```
Heyrows, line 45: not sure to see what this is, it seems like a second "isLoading" state to prevent multiple checking on the iframe at the same time.

packages/slice-machine/src/modules/preview/index.ts
```diff
@@ -0,0 +1,277 @@
...
+          isOpen: true,
+          ...(!!action.payload.stepToOpen
+            ? { openedStep: action.payload.stepToOpen }
+            : null),
+        },
+      };
+    case getType(connectToPreviewIframeCreator.request):
+      return {
+        ...state,
+        isWaitingForIframeCheck: true,
```
Heyrows, line 145: seeing how it's used, i'd say it's the same as the loading state / should be a loading state if it doesn't exist yet, should not be there.

packages/slice-machine/src/modules/preview/types.ts
```diff
@@ -0,0 +1,14 @@
+export type PreviewStoreType = {
+  setupStatus: SetupStatus;
+  setupDrawer: {
+    isOpen: boolean;
+    openedStep: number;
```
Heyrows, line 5: does the "isOpen" and "openedStep" really need to be global ?
The setup drawer is only displayed on the slice page so far.

Moreover, we already have OpenState for modal i think, should it be there or with the "open" states ?

packages/slice-machine/src/redux/saga.ts
```diff
@@ -0,0 +1,8 @@
+import { fork } from "redux-saga/effects";
+
+import { watchPreviewSagas } from "@src/modules/preview";
+
+// Single entry point to start all Sagas at once
+export default function* rootSaga() {
+  yield fork(watchPreviewSagas);
```
Heyrows, line 7: why do you need to fork the watchPreviewSage which is already a fork of something ?

Heyrows (commented):
Functional
- Actual steps 2 and 3 should be switched as Step 2 depends on Step 3
- I think the step 2 text might not be clear enough, it's kinda blurry between the error or "project running" and "page has been created" (Hugo noticed).
- The canvas url should be used when saying (your project is not running) (Hugo noticed)
- Missing documentation link in actual step 2 (learn more)
- Add a loader next to the last step when there is a check via the button so we know what is being checked ?
- Finally, I think we should add a button to redirect to the preview one the checked are done on the drawer, so people won't have to close the drawer manually and then click again on the button.

Technical
- Redux: I think we should’ve first finish migrating the remaining providers before starting something new about Redux environment.

packages/slice-machine/src/modules/preview/index.ts
hypervillain: it's used to only render the iframe renderer after click

packages/slice-machine/src/modules/environment/index.ts
hypervillain: removed

hypervillain:
What's left to be done:
- small styling issue on the check preview button
- missing links to documentation
- texts should be updated by education team (@phillysnow )

phillysnow:
@hypervillain Louis and I have verified the text in figma.

packages/slice-machine/src/modules/preview/types.ts
Miniplop: As we want to trigger the setup drawer opening during a saga, for exemple if the check 1 or 3 are KOs we have 2 options : 
- make a local state in the setup drawer components and control this local state with a variable in the redux state 

```
const [ isOpen, setIsOpen ] = useState<boolean>(false);

 const { drawerOpeningStatus } = useSelector(
    (state: SliceMachineStoreType) => ({
      drawerOpeningStatus: getDrawerOpeningStatus(state),
    })
  );

useEffect (() => {
  setIsOpen(drawerOpeningStatus)
}, [drawerOpeningStatus])
```
I this case we have to make sure the two state are always sync, that can be difficult to maintain.

- Have a state in the redux state that represents the drawer state. With this we have no  duplication. 

Is it clearer ?

packages/slice-machine/lib/builders/SliceBuilder/SetupDrawer/components/StepSection.tsx
Miniplop: The status values are typed with a union "ok" | "ko" | null, try to put something else and you will have a typescript error on the compilation. 

I choose a union instead of a enum because : 
- that avoid importing the enum everytime we want to make a assertion
- each step status value remain independent, if one day we want to have a check with a new status or less status we can do it easily without creating sub enum or specific enum

To sum-up with this approach we ensure we will always use the right value (type checked by typescript) without adding any complexity and that add a little bit of flexibility for later on if we need it.

packages/slice-machine/src/redux/saga.ts
Miniplop: Having one entry point for all the saga + having a entry point per modules allow : 
- when you create a new saga in a module that has already some sagas you need to add it only on the module saga export function.
- when you create a saga on a module that doesn't have any sagas you need to add it on the module saga export function + on the root saga export function.

With this the file store.ts is only made for config not implem + we reduce the saga boilerplate when creating new sagas on modules that use sagas :)

Miniplop:
> * Step 2

Miniplop:
> Functional
> 
> * Actual steps 2 and 3 should be switched as Step 2 depends on Step 3
> * I think the step 2 text might not be clear enough, it's kinda blurry between the error or "project running" and "page has been created" (Hugo noticed).
> * The canvas url should be used when saying (your project is not running) (Hugo noticed)
> * Missing documentation link in actual step 2 (learn more)
> * Add a loader next to the last step when there is a check via the button so we know what is being checked ?
> * Finally, I think we should add a button to redirect to the preview one the checked are done on the drawer, so people won't have to close the drawer manually and then click again on the button.
> 
> Technical
> 
> * Redux: I think we should’ve first finish migrating the remaining providers before starting something new about Redux environment.

@Heyrows here the answers discussed with Côme.

Functional

- The current order is the implementation order, as a dev it looks logical to create a page then reference it. I agree that the check up order can be a little bit disturbing. For now, we will try with this and change it if we have feedbacks on it. 
- The text is changed to a generic text and  with a link to redirect to a troubleshooting page, as we don't want to handle all the specific case in the checks.
- Already done in the PR
- (Answered in the question above)
- Already done in the PR
- Ok, we will provide a design for adding a button

Tech 
- Redux saga don't replace redux it extend it, so it doesn't slow us down to migrate the remaining state into it. This lib allow us to push the migration further if we want (but we are not force to do all the stuff together)

Miniplop:
The step number displayed on the check configuration is not the good one for Nuxt. No warning on nuxt on the step 2 on iframe ko

Miniplop:
> The step number displayed on the check configuration is not the good one for Nuxt. No warning on nuxt on the step 2 on iframe ko

Solved

packages/slice-machine/lib/builders/SliceBuilder/SetupDrawer/index.tsx
```diff
@@ -69,8 +69,8 @@ const SetupDrawer: React.FunctionComponent = ({}) => {
             flex: 1,
             overflow: "auto",
             flexDirection: "column",
-            pl: 4,
-            pr: 4,
+            pl: 24,
```
hypervillain: Can you write `paddingLeft: "24px"` or `pl: "24px"`? It helps understand that the value is not a theme-ui value

hypervillain: (it is _ok_ although not ideal when the left-hand key is not theme-ui specific eg. `width` but even there I would prefer to be less ambiguous)

hypervillain (changes requested):

packages/slice-machine/lib/builders/SliceBuilder/SetupDrawer/Stepper.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { LoadingKeysEnum } from "@src/modules/loading/types";
+import { Frameworks } from "@slicemachine/core/build/src/models";
+import { getFramework, selectPreviewUrl } from "@src/modules/environment";
+
+interface Props {
+  framework: Frameworks | undefined;
+}
+
+export default function Stepper({ framework }: Props): React.ReactElement {
+  if (framework !== Frameworks.next && framework !== Frameworks.nuxt) {
```
hypervillain: Actually we can use previewIsSupported here

packages/slice-machine/src/modules/preview/index.ts
```diff
@@ -0,0 +1,284 @@
...
+    yield call(failCheckSetupSaga);
+  } catch (error) {
+    yield put(checkPreviewSetupCreator.failure(error));
+  }
+}
+
+function* failCheckSetupSaga() {
+  const framework: Frameworks = yield select(getFramework);
+
+  if (![Frameworks.next, Frameworks.nuxt].includes(framework)) {
```
hypervillain: Same here, we can use `previewIsSupported`

packages/slice-machine/lib/builders/SliceBuilder/SetupDrawer/Stepper.tsx
```diff
@@ -0,0 +1,87 @@
...
+    openedStep: selectOpenedStep(state),
+    isCheckingSetup: isLoading(state, LoadingKeysEnum.CHECK_PREVIEW),
+    setupStatus: selectSetupStatus(state),
+    previewUrl: selectPreviewUrl(state),
+    framework: getFramework(state),
+    userHasAtLeastOneStepMissing: selectUserHasAtLeastOneStepMissing(state),
+    userHasConfiguredAllSteps: selectUserHasConfiguredAllSteps(state),
+  }));
+
+  const stepNumberWithErrors = [
```
hypervillain: Although this is temporary, I think we should create something like a selector that returns an array per framework. It could be stored next to steps

hypervillain (changes requested):

packages/slice-machine/lib/builders/SliceBuilder/SetupDrawer/index.tsx
Miniplop: Done :)

packages/slice-machine/lib/builders/SliceBuilder/SetupDrawer/Stepper.tsx
Miniplop: done

packages/slice-machine/src/modules/preview/index.ts
Miniplop: Done :)

packages/slice-machine/lib/builders/SliceBuilder/SetupDrawer/Stepper.tsx
Miniplop: As for now this information is only used here I don't think a selector is needed. Selector are used for centralizing logic around store so we can create a selector when we will reuse somewhere else this logic. 

What do you think about it :)

hypervillain: Selector was not the right term. I had something like this in mind

```javascript
export {
   steps: [/* ... */],
   withErrorSteps(setupStatus) {
       return [/* ... */]
   }
```

And 

```javascript
    const { steps: currentSteps, withErrorSteps } = steps[framework]
    const stepsWithErrors = withErrorSteps(setupStatus)
```

But it's fine

hypervillain (approved):

PR #274, opened 2021-12-15
use the user service to retrieve user profile or user repo list

arnaudlewis:


arnaudlewis:
WIP: the service is not up yet and we still need to add the role on each repo inside the user service to be able to use there

arnaudlewis:
UPDATE: service UP + tests good, we just miss the `roles` coming from the `user service`

packages/core/src/core/communication.ts
```diff
@@ -86,13 +87,17 @@ export async function validateSession(
     });
 }
```
Miniplop, line 89: Maybe we should remove the repositories parsing in the `validateSession` function no ?

Miniplop: Also we can check that this function is still use in the code as we remove one of the usage :)

packages/core/src/core/index.ts
```diff
@@ -26,7 +27,7 @@ export interface Core {
   },*/
 
   Repository: {
-    list: (token: string) => Promise<Communication.RepoData>;
```
Miniplop, line 29: Is the `Communication.RepoData` variable is  still used ? If no you can remove it :)

packages/core/src/core/communication.ts
arnaudlewis: It's still use to check actual user authentication so I kept it and repositories still exist in that context, it just work differently so for now I didn't change that behavior but like I told you I want to get rid of this validate endpoint in a way or another. It'll obviously mean some update on the auth server since SM still need it for other reasons

packages/core/src/core/index.ts
arnaudlewis: same here, validateSession is still used so RepoData as well

PR #276, opened 2021-12-17
Fix: libraries states.

Heyrows:
## Description

This PR fixes 2 issues:
- On the init script, we define the default `slices` but don't create the folder that goes with it. It causes Slice Machine to consider it to be an invalid library.

- On Slice Machine, we don't make a good distinction between the two following cases
  - the libraries key is not defined in the SM.Json file
  - the libraries are invalid (the folder don't exists)

The fixes:
- On the init script, create the slices folder if it doesn't exist when we configure the project.

- On Slice Machine, the libraries should be considered undefined if the key is not defined in the SM.json, and should be an empty array if no libraries are valid but the key is defined.

Extra:
- The empty state when libraries are not defined was put in the _app components, as custom types removes not found slices.
- When there no custom types / buttons, we don't display the big plus buttons, we only display the empty state that allows to create ressources and gives link to the documentation.


packages/slice-machine/components/App/index.tsx
```diff
@@ -45,33 +46,33 @@ const SliceMachineApp: React.FunctionComponent<AppProps> = ({
       <BaseStyles>
         <RemoveDarkMode>
           {!serverState ? (
-            <>{children}</>
+            children
```
MarcMcIntosh, line 49: Weird, I made another comment in a different PR asking about the. extra `<></>` .
This works correct when children is array like?

packages/slice-machine/components/MissingLibraries/index.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export const MissingLibraries: React.FC = () => (
+  <Box sx={{ marginLeft: "20px" }}>
+    <p>
+      We could not find any local library in your project.
+      <br />
+      Please update your `sm.json` file with a path to slices, eg:
+    </p>
+    <p>
+      <pre>{`{ "libraries": ["@/slices"] }`}</pre>
```
MarcMcIntosh, line 12: Is the prefix `@/` for all frameworks?

packages/slice-machine/lib/models/server/ServerState.ts
```diff
@@ -12,11 +12,11 @@ export interface ConfigErrors {
 
 export default interface ServerState {
   env: FrontEndEnvironment;
-  libraries?: ReadonlyArray<LibraryUI>;
-  customTypes?: ReadonlyArray<CustomType<ObjectTabs>>;
-  remoteCustomTypes?: ReadonlyArray<CustomType<ObjectTabs>>;
-  remoteSlices?: ReadonlyArray<Models.SliceAsObject>;
-  clientError?: ErrorWithStatus;
+  libraries: ReadonlyArray<LibraryUI> | undefined;
```
MarcMcIntosh, line 15: aren't `?` and `| undefined the same?

packages/slice-machine/pages/index.tsx
```diff
@@ -134,19 +134,21 @@ const CustomTypes: React.FunctionComponent = () => {
...
-              justifyContent: "center",
-              alignItems: "center",
-              borderRadius: "50%",
-              height: "48px",
-              width: "48px",
-            }}
-          >
-            <GoPlus size={"2em"} />
-          </Button>
+          customTypes.length ? (
```
MarcMcIntosh, line 137: `{ customtypes.length && (<Button>....</Button> )}` would also work here ;)

packages/slice-machine/pages/slices.tsx
```diff
@@ -77,34 +77,19 @@ const SlicesIndex: React.FunctionComponent = () => {
...
+    (l) => l.isLocal
+  );
 
-  const sliceCount =
-    libraries && libraries.length
-      ? libraries.reduce((count, lib) => {
-          if (!lib) {
-            return count;
-          }
+  const sliceCount = libraries?.length
```
MarcMcIntosh, line 84: the condition on `libraries.length` isn't needed here as reduce on an empty array would also be 0,
`[].reduce((acc, curr) => acc + curr, 0) === 0`

packages/slice-machine/pages/slices.tsx
```diff
@@ -233,11 +203,11 @@ const SlicesIndex: React.FunctionComponent = () => {
           )}
         </Box>
       </Container>
-      {!!configLocalLibs.length && (
+      {localLibraries?.length ? (
```
MarcMcIntosh, line 206: is there a reason we're adding ternary operators?

packages/slice-machine/server/src/api/custom-types/push.ts
```diff
@@ -29,6 +29,17 @@ export default async function handler(req: RequestWithEnv) {
 
   const state = await getBackendState(req.errors, req.env);
 
+  if (!state.libraries) {
+    const code = 400;
+    const message = `Error ${code}: Slice libraries needs to be define in your sm.json file.`;
```
MarcMcIntosh, line 34: I guess it is a `bad request`...

packages/slice-machine/server/src/api/libraries.ts
```diff
@@ -6,11 +6,15 @@ import ErrorWithStatus from "@lib/models/common/ErrorWithStatus";
 
 import { LibraryUI } from "@lib/models/common/LibraryUI";
 
-export async function getLibrariesWithFlags(env: BackendEnvironment): Promise<{
```
MarcMcIntosh, line 9: was this function not used any where?

packages/slice-machine/server/src/api/libraries.ts
```diff
@@ -6,11 +6,15 @@ import ErrorWithStatus from "@lib/models/common/ErrorWithStatus";
...
+interface LibrariesResult {
   remoteSlices: ReadonlyArray<Models.SliceAsObject>;
   clientError: ErrorWithStatus | undefined;
-  libraries: ReadonlyArray<LibraryUI>;
-}> {
+  libraries: ReadonlyArray<LibraryUI> | undefined;
```
MarcMcIntosh, line 12: `libraries?` then later on it `libraries: undefined` won't be needed

packages/slice-machine/server/src/api/libraries.ts
```diff
@@ -27,10 +31,10 @@ export async function getLibrariesWithFlags(env: BackendEnvironment): Promise<{
...
     })();
 
-    const libraries = Libraries.libraries(
-      env.cwd,
-      env.manifest.libraries || []
-    );
+    if (!env.manifest.libraries)
+      return { remoteSlices, libraries: undefined, clientError };
+
+    const libraries = Libraries.libraries(env.cwd, env.manifest.libraries);
```
MarcMcIntosh, line 37: There's a lot things called `libraries`, maybe we should rename `Libraries.libraries` to `getLibraries` ?

packages/slice-machine/src/models/libraries/context.tsx
```diff
@@ -9,13 +9,13 @@ import { LibraryUI } from "lib/models/common/LibraryUI";
 import { SliceMockConfig } from "lib/models/common/MockConfig";
 
 export const LibrariesContext = React.createContext<
-  Partial<ReadonlyArray<LibraryState>>
->([]);
+  ReadonlyArray<LibraryState> | undefined
+>(undefined);
```
MarcMcIntosh, line 13: why `| undefined`?

packages/slice-machine/src/models/libraries/context.tsx
```diff
@@ -9,13 +9,13 @@ import { LibraryUI } from "lib/models/common/LibraryUI";
...
 
 type LibraryHandlerProps = {
-  libraries: ReadonlyArray<LibraryUI>;
+  libraries: ReadonlyArray<LibraryUI> | undefined;
```
MarcMcIntosh, line 16: same

packages/slice-machine/src/models/libraries/context.tsx
```diff
@@ -24,23 +24,25 @@ const LibraryHandler: React.FunctionComponent<LibraryHandlerProps> = ({
...
-            env.mockConfig,
-            lib.name,
-            component.infos.sliceName
-          ),
-          remoteSlice: remoteSlices?.find((e) => e.id === component.model.id),
-        })
-      ),
-    };
-  });
+  const models: ReadonlyArray<LibraryState> | undefined = libraries?.map(
```
MarcMcIntosh, line 27: same

MarcMcIntosh (changes requested):

packages/slice-machine/pages/index.tsx
Heyrows: it displays a 0 when the value is false :)

packages/slice-machine/components/App/index.tsx
Heyrows: That's a good question, children being a ReactNode i'm not sure.
Do you think I should add them back ?

To me `<></>` or react fragment are used when you have like 2 element but not a single main element to return.

packages/slice-machine/components/MissingLibraries/index.tsx
Heyrows: it's totally independant of the framework, it's just a prefix to state that the library is local.

packages/slice-machine/lib/models/server/ServerState.ts
Heyrows: No, if you put `?` it means that the parameter doesn't need to be specified.
If you put `| undefined` it means that the parameter needs to be specified but it's value can be undefined :)

packages/slice-machine/pages/index.tsx
Heyrows: That's actually what I did at first but then I had a huge 0 on my page haha

packages/slice-machine/pages/slices.tsx
Heyrows: good call

packages/slice-machine/pages/slices.tsx
Heyrows: Personal preference and the reason I stated on the other comment about the button :)

packages/slice-machine/server/src/api/custom-types/push.ts
Heyrows: It should literally never happened because people would not access the page haha but just in case.
Is there something you wanted me to change ?

packages/slice-machine/server/src/api/libraries.ts
Heyrows: I couldn't find any trace of it

packages/slice-machine/server/src/api/libraries.ts
Heyrows: the libraries key is always defined by the backend, it's value is undefined if there is no "libraries" key in the SM.json, that's the goal of this PR.

packages/slice-machine/src/models/libraries/context.tsx
Heyrows: That's the goal of this PR, it's in the description,

There are 2 cases:
- the "libraries" key from SM.json is missing and we display a special screen
- the "libraries" key from SM.Json is empty or contains only invalid names, and we display a special screen too.

So here, having it undefined or defined means that we differentiate the two cases.

packages/slice-machine/src/models/libraries/context.tsx
Heyrows: If the libraries returned are not defined, this "models" which will be in the provider should be undefined as well.

packages/slice-machine/components/MissingLibraries/index.tsx
MarcMcIntosh: is framework independant

packages/slice-machine/components/App/index.tsx
MarcMcIntosh: doesn't matter

packages/slice-machine/lib/models/server/ServerState.ts
MarcMcIntosh: For when libraries is not defined in sm.json

packages/slice-machine/server/src/api/libraries.ts
MarcMcIntosh: created a ticket for it 
https://linear.app/prismic/issue/SM-395/[technical-debt]-core-rename-libarieslibraries

MarcMcIntosh (approved):

PR #278, opened 2021-12-20
[Draft] Tracking through segment rework

arnaudlewis:


packages/init/src/utils/tracker.ts
```diff
@@ -0,0 +1,54 @@
+/*
+ * DUPLICATION
+ * this file is a duplication of ServerTracker in sm-ui
+ * this duplication has been done on purpose to simplify the merge if init becomes part of sm-ui
+ * or if we decide to share the tracker implementation from a lib, but it'll need some refactoring
+ * Duplication means no entanglement so don't merge this with the other implem and keep the structure.
```
MarcMcIntosh, line 6: `if init becomes part of sm-ui` .... was that a discussion i missed out on?

packages/slice-machine/server/src/api/index.ts
```diff
@@ -238,16 +219,30 @@ router.post(
...
+    const payload = await statusAuth(req);
     if (payload.status === "error") {
       return res.status(500).json(payload);
     }
     return res.status(200).json(payload);
-  }
+  })
+);
+
+router.get(
```
Heyrows, line 234: this one is doubled, there is already a state route in this file.

packages/slice-machine/server/src/api/state.ts
```diff
@@ -105,7 +107,21 @@ export default async function handler(
...
-  if (serverState.libraries) req.tracker.libraries(serverState.libraries);
+  /*
+   * TEMPORARY
+   * Today, we need to have access to the cookie to get the anonymousID from segment but
+   * we want to log that event only at start time. Because of the necessity to access the cookies,
+   * we can't put that code on the start script just yet.
+   * We need to define a potential alternative for this event.
+   * In the meantime, this flag prevent the event flood
+   */
+  if (temp_first_start_flag && serverState.libraries) {
```
Heyrows, line 118: If this is meant to be send only once, shouldn't it be send on the front-end ?
I didn't saw yet where the anonymousId is define but I think this would fit on a useEffect when the state is define for the first time.

packages/slice-machine/src/utils/tracker.ts
```diff
@@ -0,0 +1,57 @@
+import { createContext } from "react";
```
Heyrows, line 1: I wondering shouldn't we have a common Tracker file in the Lib folder ?
The difference I see is that front-end event are not tracked by user as we don't seem to know who this is.
And back-end events can be tracked by users.

I thought the point of the anonymousId (which I don't see here, so i'm guessing it's sent by segment when we track and event) was to be anonymous until we actually login once.

But here it seems that the events are tracked anonymously even if we are logged in already.

packages/slice-machine/server/src/api/state.ts
arnaudlewis: We need to do it on server side to be able to parse the manifest and enrich the event.
The best way would be to do it in the start script but then we wouldn't have access to the anonymousID stored in the cookies

packages/slice-machine/src/utils/tracker.ts
arnaudlewis: They are because segment send a cookie with the Id along with the request on client side.
Look for 'resolveUser' that will attach a userId to the anonymousID considered as the current sessionId basically

packages/slice-machine/server/src/api/index.ts
arnaudlewis: Oh true, I may have copy paste for some reason and forgot to delete it. Nice catch

packages/init/src/utils/tracker.ts
arnaudlewis: It's an information I got from Hugo that made me do it this way since I don't have much info on the next steps. I think that's a discussion you should bring to him

hypervillain: @MarcMcIntosh a good amount of code was added to the core because it had to be shared between sm-ui and init. But when we started defining the core more precisely, we realized the init simply is a dependent of sm-ui. To say it differently: the init is the init of sm-ui. This explains why they share common code, so the init could be a script (like start) in sm-ui. No need to rush things as everything works as expected right now

MarcMcIntosh (commented):
I have a few questions about how do we disable tracking?
Use case would be:
+ during tests so user data isn't polluted with data from the tests
+ and if user's don't want to be tracked, (i think we need to ask for consent if we're using cookies)

PR #282, opened 2022-01-03
AAuser, In SM UI, I can create a link field that works for both web, media, internal document

Miniplop:
…eams to precise it on contribution

## Types of changes

## Checklist:


MarcMcIntosh (approved):
Works, I was wondering about media links and links to documents but those would be content rather than mocks :)

PR #283, opened 2022-01-03
Migrate all the existing event to the new tracking SDK

Miniplop:
Migrate all the existing event to the new tracking SDK + fix the preview event bug


Miniplop:
We found a bug with the tracking of the video status on the onboarding. I'm discussing right now with the rest of the team about it

packages/slice-machine/src/tracker.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export enum ContinueOnboardingType {
+  OnboardingContinueIntro = "SliceMachine Onboarding Continue Screen Intro",
+  OnboardingContinueScreen1 = "SliceMachine Onboarding Continue Screen 1",
+  OnboardingContinueScreen2 = "SliceMachine Onboarding Continue Screen 2",
+  OnboardingContinueScreen3 = "SliceMachine Onboarding Continue Screen 3",
+}
+
+let _client: ClientAnalytics | null = null;
```
arnaudlewis: I don't think it's a good idea to have a mutable singleton in a module instead of a proper class. It's less testable and more error prone.

Miniplop:
The video status data is removed from the onboarding events because this data was not used anymore

Miniplop:
This PR fix also a problem with the event "SliceMachine Slice Preview Setup" event, to trigger this event only when the drawer open :)

packages/slice-machine/src/tracker.ts
```diff
@@ -0,0 +1,176 @@
...
+const identifyUser = (userId: string): void => {
+  _identify(userId);
+};
+
+const groupLibraries = (libs: readonly LibraryUI[], version: string): void => {
+  if (!_temp_first_start_flag) {
+    return;
+  }
+
+  _temp_first_start_flag = false;
```
arnaudlewis: it's not up to the tracker to determine if it should do that once only, the flag should be handled outside

packages/slice-machine/src/tracker.ts
```diff
@@ -0,0 +1,176 @@
...
+}
+
+export enum ContinueOnboardingType {
+  OnboardingContinueIntro = "SliceMachine Onboarding Continue Screen Intro",
+  OnboardingContinueScreen1 = "SliceMachine Onboarding Continue Screen 1",
+  OnboardingContinueScreen2 = "SliceMachine Onboarding Continue Screen 2",
+  OnboardingContinueScreen3 = "SliceMachine Onboarding Continue Screen 3",
+}
+
+let _client: ClientAnalytics | null = null;
```
arnaudlewis: if you really want to keep a singleton, you can't let this client out there like this, it should be properly scoped.
That's kind of a standard pattern in js.
```
// the closure ensure that you scope your client inside and is accessible when calling public methods
const Tracker = (() => {
  let client

  // this part is actually the setup of the tracker itself
	const init = () => ({
		trackSomething() {
      console.log('track somethign here')
		}
	})

  // this part handles the singleton mecanism
	return () => ({
		get() {
			if (!client)  client = init();
      return client;
		}
	})
})();

// How it's supposed to be used
Tracker.get().trackSomething()
```

Closures enforce this.

packages/slice-machine/src/tracker.ts
```diff
@@ -0,0 +1,176 @@
...
+let _isTrackingActive = true;
+let _repoName: string | null = null;
```
arnaudlewis: same for this, it should be contained in the closure. But ideally it should be a class, if you have all these variables it means that the tracker is stateful here.

packages/slice-machine/src/tracker.ts
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  try {
+    _client.group(_repoName, attributes);
+  } catch {
+    console.warn(`Couldn't report group: Tracking error`);
+  }
+};
+
+const _isTrackingPossible = (
```
arnaudlewis: if you need such helper, which seems weirder than just expecting that the tracker could be null BTW, you should put it next to the get function in the singleton

packages/slice-machine/src/tracker.ts
```diff
@@ -0,0 +1,176 @@
...
+let _temp_first_start_flag = true;
+
+/** Private methods **/
+
+// Native client event method (don't expose these functions)
+const _trackEvent = (
+  eventType: AllSliceMachineEventType,
+  attributes: Record<string, unknown> = {}
+): void => {
+  if (!_isTrackingPossible(_client)) {
```
arnaudlewis: you cannot make this check everywhere, either the tracker is possibly null or you wrap the whole thing that will process each call to the tracker and drop them if the client is null or execute it but you can't make that check for every single track

arnaudlewis (changes requested):

packages/slice-machine/src/tracker.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+}
+
+const Tracker = (() => {
+  let smTrackerInstance: SMTracker;
+
+  const init = () => new SMTracker();
+
+  return {
+    get() {
```
arnaudlewis, line 170: By seeing this implem, it looks good but maybe the init part we could simplify it maybe this way: 
```
return {
  get() {
    return smTrackerInstance
  },
  init(param1, param2, param3) {
    if (!instance) new Tracker(param1, param2, param3)
  },
}
```

so we could remove the initialize function, move everything in the constructor and directly call Tracker.init()
The benefit would be a simpler syntax but also setup the class with the right params right away instead of default null values to wait for the initialize to be executed. WDYT?

packages/slice-machine/src/tracker.ts
Miniplop: Done

packages/slice-machine/src/tracker.ts
Miniplop: Done

packages/slice-machine/src/tracker.ts
Miniplop: Done

packages/slice-machine/src/tracker.ts
Miniplop: Done

Miniplop: The benefit of the singleton is to abstract the instance creation, so if get return null or SMTracker we loose this benefit. That also mean that in the component we should listen for tracker changes to get the moment when the Tracker will be initialize. 

````
useEffect(() => {
    if (tracker) {
      tracker.Track.SlicePreview({ framework, version });    }
  }, [tracker]);
`

PR #286, opened 2022-01-05
Kje/use short id for tracking

Miniplop:
This PR is for using the shortId returned by the /profile route :)


Heyrows (approved):
No PR description,
simple PR with not much change, I still think that
- long IDs should be only manipulated by services that goes in the DB.
- short IDs should be manipulated outside, for services that don't check the DB or shouldn't know what are the real Ids or users in the DB.

PS: i'm approving because that's what the team decided, but i still think it's wrong 😄

Miniplop:
As seen with all the team : 
- We don't want to store the userId anymore (as it's not used)
- We keep the name "shortId" for the attribute that contains the intercom id in the /profile route

PR #287, opened 2022-01-06
Mm/track preview

MarcMcIntosh:
Adds two tracking events to slice-machihne.

It was going to be three but the iframe will needed to be modified to handle these.

## Types of changes

## Checklist:


packages/slice-machine/lib/builders/SliceBuilder/SideBar/index.tsx
```diff
@@ -171,7 +172,7 @@ const StoryBookOrPreview: React.FC<{
           mt: 3,
           color: "textGray",
           "::first-letter": {
-            "text-transform": "uppercase",
+            textTransform: "uppercase",
```
Miniplop, line 175: good catch

packages/slice-machine/src/modules/environment/index.ts
```diff
@@ -49,8 +49,19 @@ export const getUpdateVersionInfo = (
...
+};
+
+export const getStorybookUrl = (
+  store: SliceMachineStoreType
+): string | null => {
+  return store.environment.env.manifest.storybook || null;
+};
+
+export const getUserID = (store: SliceMachineStoreType): string | null => {
+  return store.environment.env.userId || null;
```
Miniplop: I think you can return 'store.environment.env.userId' directly as this is already null | string :)

Miniplop (commented):
I think you can remove all the logic to send the userID to segment, because this will be change when we will make the reconciliation on login. 

I let you also some other comments :)

packages/slice-machine/src/modules/environment/index.ts
Miniplop: Ignore this comment because you will remove this selector :)

MarcMcIntosh: done

Miniplop:
LGTM

Miniplop (approved):

PR #292, opened 2022-01-11
Fix the start script

Miniplop:
## Description

This PR is to solve the regression on the start script where an typescript error occurs : 

Before : 
image

After : 
image

### DIFF
--- a/packages/slice-machine/start/index.js
+++ b/packages/slice-machine/start/index.js
@@ -77,11 +77,8 @@ function start({ cwd, port }, callback) {
   });
 
   smServer.stdout.on("data", function (data) {
-    const lns = data.toString().split("=");
+    const lns = data.toString().split("Server running");
     if (lns.length === 2) {
-      // server was launched
-      if (callback) {
-      }
       callback(lns[1].replace(/\\n/, "").trim());
     } else {
       console.log(data.toString());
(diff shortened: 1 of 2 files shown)


packages/slice-machine/start/index.js
```diff
@@ -77,11 +77,8 @@ function start({ cwd, port }, callback) {
   });
 
   smServer.stdout.on("data", function (data) {
-    const lns = data.toString().split("=");
+    const lns = data.toString().split("Server running");
```
Heyrows, line 80: The fix that you made by reading the function SMConfig is good, so why did you change this split ?
What does it change ?

Miniplop: This line of code is here to replace "server running : localhost...." by the green info box that we know

Miniplop: Before the dev server write "p=9999", now the server write "Server Running: localhost:9999"

Miniplop: The regression was introduced when we change the Express message in the setup drawer PR

Heyrows: ah ok i see. I'll trust your screenshots and the validation part to detect if there is an issue but i'm sure there won't be :)

Heyrows (approved):

PR #293, opened 2022-01-11
use new URL to create sb slice href

hypervillain:
Fix a bug that created wrong Storybook Slice URLs when manifest `storybook` property ended with a slash


Miniplop (approved):

Miniplop:
As the storybook url generation is only used on the frontend side we can move this file to the src folder. 

In the lib folder I put only stuff shared between the back and the frontend part. What do you think about that  ?

PR #296, opened 2022-01-13
Fix linter issues

Heyrows:
## Description

This PR is meant to simply fix some linter issues / type some components without modifying any behavior or breaking stuff


packages/slice-machine/src/theme.ts
```diff
@@ -262,7 +262,7 @@ const AppTheme = (): Theme =>
         fontWeight: "body",
         fontSize: "1",
         color: "textClear",
-        border: (t) => `1px solid ${t?.colors?.borders}`,
+        border: (t) => `1px solid ${String(t?.colors?.borders)}`,
```
MarcMcIntosh, line 265: theme ui is weird with typescript

packages/slice-machine/lib/models/common/CustomType/fields.ts
```diff
@@ -30,5 +30,5 @@ export const SimpleField = {
 export interface Field {
   type: FieldType | SliceZoneType;
   fieldset?: string;
-  config: {};
+  config: any;
```
MarcMcIntosh, line 33: was there not an extra error about any type?
.also .. do we have types for the configs?

packages/slice-machine/package.json
```diff
@@ -29,7 +29,7 @@
     "export-build": "next export && npm run clean-build",
     "bundle": "npm run build && npm run test && npm run export-build",
     "slicemachine": "start-slicemachine --port 9999",
-    "lint": "eslint --max-warnings 1049 --ext .ts,.tsx .",
+    "lint": "eslint --max-warnings 976 --ext .ts,.tsx .",
```
MarcMcIntosh, line 32: nice

packages/slice-machine/components/Preview/components/Header/index.tsx
```diff
@@ -23,7 +23,7 @@ const redirect = (
   isPreview?: boolean
 ): void => {
   if (!variation) {
-    router.push(`/${model.href}/${model.infos.sliceName}`);
+    void router.push(`/${model.href}/${model.infos.sliceName}`);
```
MarcMcIntosh, line 26: i think `router.push` is async, which might explain the need for void

packages/slice-machine/components/ListItem/index.tsx
```diff
@@ -12,27 +12,49 @@ import { AiOutlineEdit } from "react-icons/ai";
 import { BsThreeDotsVertical } from "react-icons/bs";
 
 import { FaBars } from "react-icons/fa";
+import { Widget } from "@lib/models/common/widgets/Widget";
+import { Field } from "@lib/models/common/CustomType/fields";
+import { AnyObjectSchema } from "yup";
```
MarcMcIntosh, line 17: What's this do?

packages/slice-machine/components/Grid/index.tsx
```diff
@@ -1,30 +1,35 @@
...
         pt: 2,
         mb: 3,
       }}
     >
-      {elems.map((elem: any, i: number) => (
-        <span key={`list-item-${i + 1}`}>{renderElem(elem, i)}</span>
-      ))}
+      {elems.map((elem: T | undefined, i: number) =>
+        elem ? (
+          <span key={`list-item-${i + 1}`}>{renderElem(elem, i)}</span>
```
MarcMcIntosh: pretty sure when removing / sorting items, these indexes can get mixed up. I've got no idea how to handle it when dealing with generic list items.

packages/slice-machine/components/Forms/components/SelectRepeatable.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <FlexCard selected={field.value} onClick={() => helpers.setValue(true)}>
+        <Radio checked={field.value} />
+        <Box
+          sx={{
+            marginLeft: 2,
+          }}
+        >
+          Repeatable type
+          <Box as="p" sx={{ fontSize: "12px", color: "textClear", mt: 1 }}>
+            Best for multiple instances like blog posts, authors, products...
```
MarcMcIntosh, line 19: Check the wording with a designer

packages/slice-machine/components/Forms/CreateSliceModal.tsx
```diff
@@ -82,35 +57,23 @@ const CreateSliceModal: React.FunctionComponent<CreateSliceModalProps> = ({
...
         </Label>
         <Select
           name="origin"
           options={libraries.map((v) => ({ value: v.name, label: v.name }))}
-          onChange={(v: { label: string; value: string } | null) => {
-            if (v) {
-              setFieldValue("from", v.value);
-            }
-          }}
+          onChange={(v: { label: string; value: string } | null) =>
```
MarcMcIntosh, line 74: what happens when deleting to an empty string?

packages/slice-machine/components/Flex/index.tsx
```diff
@@ -22,10 +20,11 @@ export const Flex = ({
...
-  children,
-  cols = 2,
-}: {
+interface ColProps {
   children: React.ReactNode;
   cols?: number;
-}) => <Box sx={{ flex: `0 ${100 / cols - 1}%`, mb: 1 }}>{children}</Box>;
+}
+
+export const Col: React.FC<ColProps> = ({ children, cols = 2 }) => (
```
MarcMcIntosh, line 28: You know you can pass the propTypes like 
```
FC<{{ children: React.ReactNode; cols?: number;}>
```

packages/slice-machine/components/AppLayout/Navigation/Menu/Navigation/List.tsx
```diff
@@ -1,17 +1,21 @@
...
-        {links.map((link: LinkProps, i) => (
-          <Item key={`${link} - ${i}`} link={link} />
-        ))}
-      </Box>
+interface ItemsListProps {
+  links: LinkProps[];
+  mt?: number;
+}
+
+const ItemsList: React.FC<ItemsListProps> = ({ mt, links }) => (
```
MarcMcIntosh, line 11: you know you can pass the types like
```
 React.FC<{ links: LinkProps[]; mt?: number; }>
 ```

File: packages/slice-machine/components/AppLayout/Navigation/Menu/Navigation/Item.tsx [hunk 11]
```diff
@@ -1,9 +1,14 @@
...
 import { useRouter } from "next/router";
 import { Box, Link as ThemeLink } from "theme-ui";
 import { LinkProps } from "components/AppLayout/Navigation";
 
-const Item = ({ link }: { link: LinkProps }) => {
+interface itemProps {
+  link: LinkProps;
+}
+
+const Item: React.FC<itemProps> = ({ link }) => {
```
MarcMcIntosh, line 11: You know you can pass the types like
```
React.FC<{link: LinkProps;}> 
```

packages/slice-machine/pages/index.tsx
```diff
@@ -181,11 +182,11 @@ const CustomTypes: React.FunctionComponent = () => {
...
         />
       ) : (
-        <Grid
+        <Grid // The CtPayload should be used here once the Context has been changed.
           elems={customTypes}
-          renderElem={(ct: CtPayload) => (
+          renderElem={(ct: CustomType<ObjectTabs>) => (
             <Link passHref href={`/cts/${ct.id}`} key={ct.id}>
-              <Card ct={ct} />
+              <Card ct={ct as CtPayload} />
```
MarcMcIntosh, line 190: weird, how come the type had to be cast here?

MarcMcIntosh (changes requested):

packages/slice-machine/src/theme.ts
Heyrows: haha, it's what they recommand so that the value that can be undefined is a string

packages/slice-machine/lib/models/common/CustomType/fields.ts
Heyrows: Hello, this one had me struggle for a good hour or two.
Actually we don't have types about it yet.
In the Core it's unknown, here it was an empty object.

Problem is as some point we try to access some properties from it.
So I couldn't let an empty object unfortunately and after struggling for a while, I put any until we get better types.
I wanted to add it to the technical debt board but it's already there.

packages/slice-machine/components/Preview/components/Header/index.tsx
Heyrows: yes that's why :)

packages/slice-machine/components/ListItem/index.tsx
Heyrows: I took it from the definition of a widget, i think it's just a schema to validate any object, it must have been something prior to io-ts.
It will be resolved once the Field / Widget types are all put in common.

packages/slice-machine/components/Grid/index.tsx
Heyrows: i'm not sure to understand.

Heyrows: ah yes i see what you mean, having the key made with the element and the index so that it's really unique and will re-render if the items changed position. Good catch 👍

packages/slice-machine/components/Flex/index.tsx
Heyrows: Yes :) it's just that i don't like this way.
Because If i need the type of those properties elsewhere then it's not re-usable.
That's a personal preference that I consider best practice for code re-usability 😄 .

packages/slice-machine/components/AppLayout/Navigation/Menu/Navigation/List.tsx
Heyrows: same

packages/slice-machine/components/AppLayout/Navigation/Menu/Navigation/Item.tsx
Heyrows: same

packages/slice-machine/pages/index.tsx
Heyrows: hey, I put a comment about that a couple of lines before.
The Grid components takes elements and then a function to transform an element.
The entry of this Grid element was `CustomType<ObjectTabs>` but in the function to render the element itself it was casted as `CtPayload`.
Since I typed the Grid Components, we now know that ct is a `CustomType<ObjectTabs>` but I had to cast it for Card to work, not knowing which of the two types was suppose to be the right one.

packages/slice-machine/components/Flex/index.tsx
MarcMcIntosh: harder to check in vscode, it'll just show "ColProps" as the type signature

Heyrows: True

MarcMcIntosh (approved):

PR #298, opened 2022-01-13
chore(init): use short id for tracking

MarcMcIntosh:
Use the shortId in the tracking service

## Types of changes

## Checklist:


packages/core/src/core/index.ts
```diff
@@ -121,4 +121,26 @@ export const Auth = {
       () => null
     );
   },
+
+  validateSessionAndGetProfile: async function (
```
Heyrows: maybe we should have a single function to do everything.
If only the getProfile fails, we might not want to fails everything to track the user.
Without a single function it becomes easier to handle such a case.

packages/init/__tests__/logInOrBypass.test.ts
```diff
@@ -0,0 +1,82 @@
+import {
```
Heyrows, line 1: well done 👍

packages/core/src/core/communication.ts
```diff
@@ -205,3 +205,50 @@ export async function createRepository(
       throw error;
     });
 }
+
+const UserProfile = t.exact(
```
Heyrows: We already have that logic in SM-UI.
I don't think this should be in the Core (debate ongoing on slack).
Either
- we put it in the Core and make SM-UI use it.
- we duplicate the code from SM-UI to the init

Heyrows:
The shortID also needs to be put in the .prismic file so that it can be reused elsewhere.

packages/init/src/index.ts
```diff
@@ -57,7 +57,7 @@ async function init() {
     "JfTfmHaATChc4xueS7RcCBsixI71dJIJ",
     name,
     {
-      userId: user.userId,
+      userId: user.profile?.shortId || user.info.userId,
```
Miniplop: why you changed what you did before ? The userId should be always the shortId

packages/init/src/index.ts
```diff
@@ -57,7 +57,7 @@ async function init() {
     "JfTfmHaATChc4xueS7RcCBsixI71dJIJ",
     name,
     {
-      userId: user.shortId,
+      userId: user.profile?.shortId || user.info.userId,
```
Heyrows: I wouldn't build the tracker at all if I don't have the short ID to identify the user.
We can't use the userID for sure, or we would end up with events attributed to the wrong user.

packages/init/src/utils/communication.ts
```diff
@@ -0,0 +1,76 @@
...
+  info: Communication.UserInfo;
+  profile: UserProfile | null;
+} | null> {
+  const config = FileSystem.PrismicSharedConfigManager.get();
+
+  if (!config.cookies.length) return Promise.resolve(null); // default config, logged out.
+  if (base != config.base) return Promise.resolve(null); // not the same base so it doesn't
+
+  try {
+    const info = await Communication.validateSession(config.cookies, base);
```
Heyrows, line 63: if this fail, will it fall in the catch or continue and call the profile ?
Because it shouldn't call the profile if the info calls fail.

-> because we know the profile call will fail if the info call fails.

MarcMcIntosh: it'll be done sequentially

packages/init/src/index.ts
MarcMcIntosh: okay, so tracker is going to have to be optional later on

MarcMcIntosh: done

packages/init/src/utils/communication.ts
MarcMcIntosh: so if validateSession fails so will getProfile,

Miniplop (approved):

PR #310, opened 2022-01-21
[WIP] Mm/wroom login

MarcMcIntosh:
## Types of changes

## Checklist:


packages/slice-machine/lib/models/common/http/DefaultClient.ts
```diff
@@ -129,21 +130,23 @@ export default class DefaultClient {
...
+      const result = await axios.get<UserProfile>(
+        `${createApiUrl(base, UserService)}profile`,
+        {
+          headers: {
+            Authorization: `Bearer ${auth}`,
+          },
+        }
+      );
 
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
Miniplop: can you removed this line as the code line before is now removed ?

packages/slice-machine/lib/models/common/http/DefaultClient.ts
```diff
@@ -129,21 +130,23 @@ export default class DefaultClient {
...
 
       return getOrElseW(
         // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-        () => new Error(`Unable to parse profile: ${jsResult}`)
-      )(UserProfile.decode(jsResult));
+        () => new Error(`Unable to parse profile: ${result.data}`)
+      )(UserProfile.decode(result.data));
     } catch (e) {
+      // TODO: why not throw?
       return e as Error;
```
Miniplop: I think we have to discuss on what is our Error management strategy, for now we had 2 strategies in the code : 
- Create Error object in the lower-level functions and add some condition on the top-level functions
- Throw error on lower-level functions and try catch on top-level functions

packages/slice-machine/tests/server/utils.test.ts
```diff
@@ -0,0 +1,44 @@
+import { preferWroomBase, delay } from "../../server/src/api/common/utils";
```
Miniplop: Nice 🤟

packages/slice-machine/server/src/api/state.ts
```diff
@@ -109,7 +115,10 @@ export default async function handler(
     ...frontEnv,
     sliceMachineAPIUrl: baseUrl,
     shortId: prismicData.shortId,
-    prismicAPIUrl: prismicData.base,
+    prismicAPIUrl: preferWroomBase(
```
Miniplop: why not using base variable here instead of calling again the function ?

Miniplop (commented):
Nice, what about all the other calls to the other API (custom types API, Auth API, ACL provider API) ? They should also be redirected to the env corresponding the repo URL :)

packages/slice-machine/lib/models/common/http/DefaultClient.ts
MarcMcIntosh: yeah the return type here doesn't make much sense. `Promise<UserProfile | Error>` means that the response had to be cast later on

packages/slice-machine/server/src/api/state.ts
MarcMcIntosh: I can change that back and it should still work now

MarcMcIntosh: but who know's where it'll be incorrectly used

packages/slice-machine/lib/models/common/http/DefaultClient.ts
MarcMcIntosh: I would throw, 
by the looks of this we'd need to have some kind of pattern matching to figure out there was an error, which at that point we may as well be using scala.js :/

packages/slice-machine/components/LoginModal/index.tsx
```diff
@@ -25,15 +25,16 @@ import { ModalKeysEnum } from "@src/modules/modal/types";
 import { getEnvironment } from "@src/modules/environment";
 import useSliceMachineActions from "@src/modules/useSliceMachineActions";
 import Tracker from "@src/tracker";
-
-Modal.setAppElement("#__next");
+import { preferWroomBase } from "../../server/src/api/common/utils";
```
Miniplop: I think as this fonction is shared between the front and the backend it should be move to lib folder.

MarcMcIntosh: it's been moved

MarcMcIntosh: i think I moved it to the src folder :/

packages/slice-machine/lib/models/common/http/DefaultClient.ts
Heyrows: The response doesn't have to be cast because you can use `instanceof error` to determine if the result is an error or not.

I personally prefer to return an Error because it's a "known" issue, we expect this to happen for some case so our code has to cover this possibility.
To me try / catch are more for unexpected errors (no network, unreachable services) or something else that were not planned and we need a "default error managment" for those.

packages/slice-machine/lib/models/common/http/DefaultClient.ts
```diff
@@ -129,21 +130,21 @@ export default class DefaultClient {
...
-      const result = await fetch(`${createApiUrl(base, UserService)}profile`, {
-        method: "GET",
-        headers: {
-          Authorization: `Bearer ${auth}`,
-        },
-      });
-
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-      const jsResult = await result.json();
+      const result = await axios.get<UserProfile>(
```
Heyrows: The point of using trying to decode the user profile given at line 145 is that until that point we are not sure what's the shape of the data.

You can use axios, I would understand that you prefer it to fetch but you can't cast with it to a UserProfile.
That's the job of the decode later on, to validate this data.


I would type it to unknown (common type used by io-ts before validation.
```suggestion
      const result = await axios.get<unknown>(
```

Heyrows (commented):
I would not use the `.prismic` file as default for the base.
The project should be running based on the `SM.json` and not the `.prismic` which is not project related.

What I would do
- extract the base from SM.json and verify that the API endpoint has the right format (regex ?).
- test if the base is included in the known bases ([‘wroom.io’, ‘prismic.io’, ‘wroom.test’])
  - YES -> use this
  - NO -> either take `prismic.io` as default value or display something to the user to tell him that his API endpoint is wrong.

I think this should be discussed with Come to know what would be the functional best option.

packages/slice-machine/lib/models/common/http/DefaultClient.ts
MarcMcIntosh: using global fetch doesn't work in the tests, axios will because it's isomorphic (works on client and server)

Also there maybe two error return types, AxiosError and another Error,

I'm not sure getOrElseW is being used correctly here as it should be an Either<good, bad> return type, and also the error message is a bit naff

packages/slice-machine/lib/utils/preferWroomBase.ts
```diff
@@ -0,0 +1,15 @@
+export default function preferWroomBase(smApiUrl: string): string {
+  try {
```
Heyrows, line 2: I think it would be better to have list of valid bases and see if one of them is included in the smApiUrl.
It would be easier to add a valid base and handle cases.

packages/slice-machine/lib/models/common/http/DefaultClient.ts
MarcMcIntosh: Done

packages/slice-machine/lib/utils/preferWroomBase.ts
MarcMcIntosh: I don't think that will work during run time

MarcMcIntosh: It would have to be validate when it's loaded,

packages/core/src/models/Manifest.ts
```diff
@@ -1,9 +1,34 @@
 import * as t from "io-ts";
 import { FrameworksC } from "./Framework";
 
+const apiEndpoint = new t.Type<string, string, unknown>(
```
Heyrows: I wasn't thinking of going that far but that's amazin !

Heyrows: to go a little further, we could even make it a type (based on what you did) and use it where we need.

Heyrows: I wonder what it would look like :)

Heyrows: in the type I mean, is it a string or something more precise

packages/slice-machine/lib/models/common/http/DefaultClient.ts
```diff
@@ -144,7 +144,7 @@ export default class DefaultClient {
         () => new Error(`Unable to parse profile: ${result.data}`)
       )(UserProfile.decode(result.data));
     } catch (e) {
-      // TODO: why not throw?
+      if (axios.isAxiosError(e)) throw e;
```
Heyrows: I wouldn't throw, i don't like to throw if the function can return an error.
Here we need to choose
- throw and adapt when the function is called.
- return an error and adapt when it's called.

I don't think it's appropriate to have both.

packages/core/src/models/Manifest.ts
MarcMcIntosh: I could see away for io-ts to validate strings.... I don't think it's made to do this

packages/core/src/models/Manifest.ts
```diff
@@ -1,30 +1,32 @@
...
-        return t.success(input);
-      }
-      return t.failure(input, context);
-    } catch {
-      return t.failure(input, context);
-    }
-  },
-  t.identity
-);
+class ApiEndPointType extends t.Type<string> {
```
Heyrows, line 4: I think the previous way you defined API Endpoint was better because you didn't had to instantiate it.
I think it just needs to be a type or a function that takes a string and validate that it is an API Endpoint.

Why did you wanted / needed to make it a class ?

packages/slice-machine/lib/models/common/http/DefaultClient.ts
```diff
@@ -125,28 +125,17 @@ export default class DefaultClient {
...
-    try {
-      const result = await axios.get(
-        `${createApiUrl(base, UserService)}profile`,
-        {
-          headers: {
-            Authorization: `Bearer ${auth}`,
-          },
-        }
-      );
+  static async profile(base: string, auth: string): Promise<UserProfile> {
```
Heyrows, line 128: you were rgiht, it looks definitely better without try catch :) and just let there be throw !!

packages/slice-machine/server/src/api/auth/status.ts
```diff
@@ -13,10 +13,10 @@ export default async function handler(
     }
 
     const profile = await setShortId(req.env, authToken);
-    if (profile instanceof Error) return { status: "error" };
 
     return { status: "ok", userId: profile.userId };
   } catch (e) {
+    console.error(e);
```
Heyrows, line 19: just add a short message please like "Error while checking the authentication status" or something, so that the users that don't check the network has a better idea of whats happening

packages/slice-machine/server/src/api/screenshots/generate.ts
```diff
@@ -59,37 +59,40 @@ export async function generateScreenshot(
...
-        screenshots: {
-          ...acc.screenshots,
-          [variationIds[index]]: screenshotOrError,
-        },
-        failure: acc.failure,
-      };
+      const key = variationIds[index];
+
+      const screenshots =
+        result.status === "fulfilled"
```
Heyrows, line 78: Interesting type the PromiseSettledResult, I'm a bit sad we can't match on `GoodResult.status` instead of a string.

packages/slice-machine/server/src/api/screenshots/generate.ts
```diff
@@ -59,37 +59,40 @@ export async function generateScreenshot(
...
+              ...acc.screenshots,
+              [key]: result.value,
+            }
+          : acc.screenshots;
+
+      const failure =
+        result.status === "rejected"
+          ? [
+              ...acc.failure,
+              { error: result.reason as Error, variationId: key },
```
Heyrows, line 89: One thing i'm unsure of, what the type of reason, at first it's `any` so how are you sure it's an error ?
Maybe we should verify it's an error and if not instantiate one ?

That one is good example of "expected failure" haha, but letting it throw to handle it afterward was a good call 👍

packages/slice-machine/server/src/index.ts
```diff
@@ -95,6 +96,21 @@ app.use("/onboarding", (_, res) => {
...
+app.use(
+  (
+    err: Error,
+    _req: express.Request,
+    res: express.Response,
+    // This is need becuase express middleware uses arguments.length
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    _next: express.NextFunction
+  ) => {
+    console.error(err);
```
Heyrows, line 109: Just asking because I don't know but will that display the error twice ?

packages/core/src/models/Manifest.ts
MarcMcIntosh: This way the type show up correctly as `apiEndpoint` and not `Type<string, string, unkown>`

packages/slice-machine/server/src/api/screenshots/generate.ts
MarcMcIntosh: Yeah, it's a bit weird

packages/slice-machine/server/src/api/screenshots/generate.ts
MarcMcIntosh: throw/catch has the signature `catch (e: any)`, I'm pretty sure it's an error from the previous code that was there

packages/slice-machine/server/src/index.ts
MarcMcIntosh: It's there to handle unhandled errors/throws,
but if an error is logged and thrown again, then yes.

Heyrows: Ok i see, then yes i guess if it catches an unhandled error then it would not have been logged before, good.

packages/slice-machine/server/src/api/screenshots/generate.ts
Heyrows: As long as we don't throw ourself then i guess you're right :)

packages/core/src/models/Manifest.ts
Heyrows: I see, you can do that with the Io-ts you had by adding the following line to declare the type.
```
export type ApiEndPointType = t.TypeOf<typeof ApiEndPointType>;
```

with the typeof ApiEndPointType being the validator.

MarcMcIntosh: I'm just doing it the way it's done in io-ts.

MarcMcIntosh: https://github.com/gcanti/io-ts/blob/master/src/index.ts#L845-L869

Heyrows (approved):

PR #309, opened 2022-01-20
QA Process review

Miniplop:
This PR : 
- move some dependencies to devDependencies to avoid to install them in the user project
- add a try catch around the screenshoot controller to serve a proper error in case of internal errors
- fix a regression on the create slice modal where the wrong slice name format message disappear

PR #313, opened 2022-01-24
Add express-http-reverse-proxy dep as prod dependencies

Miniplop:
- Add "http-reverse-proxy" as a dependency, to make the start script work
- Migrate some frontend deps into the dev deps


Miniplop:
Currently the SM is not starting on staging : 

image

In [this previous PR](https://github.com/prismicio/slice-machine/pull/309) we moved the `http-reverse-proxy-dep` dep into dev-dependency

Heyrows (approved):

PR #319, opened 2022-01-25
Technical debt: Pass start script to TS.

Heyrows:
## Description

PR to pass the Start script of slice-machine to typescript in order to prevent unwanted breaking changes.


packages/slice-machine/package.json
```diff
@@ -35,7 +36,7 @@
     "audit-fix": "npm audit fix"
   },
   "bin": {
-    "start-slicemachine": "start/index.js"
+    "start-slicemachine": "bins/start.js"
```
MarcMcIntosh: traditionally it's called `bin/`

packages/slice-machine/server/tsconfig.json
```diff
@@ -23,5 +24,6 @@
       "@models/*": ["../lib/models/*"]
     }
   },
-  "include": ["./**/*.ts", "./**/*.js", "../lib/**/*.ts", "../lib/**/*.js"]
+  "exclude": ["../build"],
```
MarcMcIntosh: does adding build change anything?

MarcMcIntosh (commented):
Interesting thing that could be done is to export the module from `start/index.ys` and call the function in `bin/start.js`

I guess it's out of the scope of this ticket

packages/slice-machine/server/tsconfig.json
Heyrows: If a file from the build is imported directly then it avoid to try to transpile it. (we had this for the last start script).

packages/slice-machine/package.json
Heyrows: Ok, will change that

Heyrows:
@MarcMcIntosh To answer to your proposition, we wanted to do it that way, but both would have worked 😄

packages/core/src/core/communication.ts
```diff
@@ -37,38 +41,17 @@ export async function refreshSession(
...
-      fold<t.Errors, RepoData, RepoData>(
-        () => {
-          throw new Error("Can't parse repo data");
-        },
-        (f: RepoData) => {
-          return f;
-        }
-      )
-    );
+    return fold<t.Errors, RepositoriesRoles, RepositoriesRoles>(
```
Miniplop, line 49: I like that, when we remove extra complexity for improving the clarity :)

packages/slice-machine/server/src/resolveAliases.ts
```diff
@@ -0,0 +1,18 @@
+import path from "path";
```
Miniplop: As this file is used by script and node server, we can put this file into the lib folder na ?

packages/slice-machine/tests/scripts/common/findArgument.test.ts
```diff
@@ -0,0 +1,35 @@
+import { findArgument } from "../../../scripts/common/findArgument";
```
Miniplop, line 1: GG

Miniplop (commented):
I think you commited too many lock files :p, plus I would expect a little unit test on the resolve alias function.

packages/slice-machine/server/src/resolveAliases.ts
Heyrows: good point, i didn't know where to put it.

packages/core/src/core/communication.ts
```diff
@@ -37,38 +41,17 @@ export async function refreshSession(
...
-);
-export type UserInfo = {
-  userId: string;
-  email: string;
-  type: string;
-  repositories: RepoData;
-};
-
-export function maybeParseRepoData(repos?: string | RepoData): RepoData {
+export function maybeParseRepoData(
```
MarcMcIntosh, line 44: I can't remember where this function was used, but if it's parameters and return type have changed so should the name,

MarcMcIntosh (approved):

PR #320, opened 2022-01-25
Save custom type on creation

Miniplop:
Create and save the custom types when the user click on create custom types


packages/slice-machine/components/Button/index.tsx
```diff
@@ -4,23 +4,32 @@ import { Button as ThemeUIButton, Spinner } from "theme-ui";
...
+  disabled,
+  sx = {},
 }) => {
   return (
     <ThemeUIButton
       sx={{
         alignSelf: "flex-start",
         ...sx,
       }}
+      form={form}
```
MarcMcIntosh, line 30: you could use `({...props})  => <Button {..props} />` it saves some typing

packages/slice-machine/tests/project/cypress/integration/customTypes/00-create.spec.js
```diff
@@ -16,5 +16,7 @@ describe("Custom Types specs", () => {
     cy.get('input[data-cy=ct-id-input]').type('my-custom-type')
     cy.get('[data-cy=create-ct-modal]').submit()
     cy.location('pathname', {timeout: 5000}).should('eq', '/cts/my-custom-type')
+    cy.visit('/cts/my-custom-type')
```
MarcMcIntosh, line 19: refreshing the page?

packages/slice-machine/src/redux/store.ts
```diff
@@ -23,7 +25,7 @@ declare const window: {
 export default function configureStore(
   preloadedState: Partial<SliceMachineStoreType> = {}
 ): { store: Store; persistor: Persistor } {
-  const middlewares = [sagaMiddleware];
+  const middlewares = [sagaMiddleware, routerMiddleware];
```
MarcMcIntosh, line 28: willl the routes be in the state now?

packages/slice-machine/src/modules/customTypes/index.ts
```diff
@@ -0,0 +1,118 @@
...
+}: ReturnType<typeof createCustomTypesCreator.request>) {
+  const newCustomType = createCustomType(
+    payload.id,
+    payload.label,
+    payload.repeatable
+  );
+  yield call(saveCustomType, newCustomType, {});
+  yield put(createCustomTypesCreator.success({ newCustomType }));
+  yield put(modalCloseCreator({ modalKey: ModalKeysEnum.CREATE_CUSTOM_TYPE }));
+  yield put(push(`/cts/${payload.id}`));
```
MarcMcIntosh, line 104: what?

packages/slice-machine/tests/components/UpdateVersionModal.tsx
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+import "@testing-library/jest-dom";
+import React from "react";
+import { useSelector, useDispatch } from "react-redux";
+import { render } from "@testing-library/react";
+import UpdateVersionModal from "../../components/UpdateVersionModal";
+
+const mockDispatch = jest.fn();
+jest.mock("react-redux", () => ({
```
MarcMcIntosh, line 12: are you sure we should be mocking redux?

packages/slice-machine/src/apiClient.ts
```diff
@@ -10,6 +16,20 @@ const defaultAxiosConfig = {
...
+export const saveCustomType = (
+  customType: CustomType<ObjectTabs>,
+  mockConfig: CustomTypeMockConfig
+): Promise<AxiosResponse<Record<string, never>>> => {
+  const requestBody: SaveCustomTypeBody = {
+    model: customType,
+    mockConfig: mockConfig,
+  };
+
+  return axios.post("/api/custom-types/save", requestBody, defaultAxiosConfig);
```
MarcMcIntosh, line 30: you know you can pass types to axios.
```
const foo: Promise<{bar: string}> = await axios.post<{bar: string}>("/url").then(res => res.data)
```

packages/slice-machine/pages/index.tsx
```diff
@@ -191,17 +179,12 @@ const CustomTypes: React.FunctionComponent = () => {
           defineElementKey={(ct: CustomType<ObjectTabs>) => ct.id}
           renderElem={(ct: CustomType<ObjectTabs>) => (
             <Link passHref href={`/cts/${ct.id}`} key={ct.id}>
-              <Card ct={ct as CtPayload} />
```
MarcMcIntosh, line 194: nice

packages/slice-machine/pages/cts/[ct].tsx
```diff
@@ -40,25 +42,29 @@ const CustomTypeBuilderWithProvider: React.FunctionComponent<CustomTypeBuilderWi
...
 
-  const customType = customTypes?.find((e) => e && e.id === router.query.ct);
-  const remoteCustomType = remoteCustomTypes?.find(
+  const customType = customTypes.find((e) => e && e.id === router.query.ct);
+  const remoteCustomType = remoteCustomTypes.find(
     (e) => e && e.id === router.query.ct
   );
+
   if (!customType) {
-    // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
Comment (MarcMcIntosh) [inline packages/slice-machine/pages/cts/[ct].tsx:51]: nice

packages/slice-machine/pages/cts/[ct].tsx
```diff
@@ -40,25 +42,29 @@ const CustomTypeBuilderWithProvider: React.FunctionComponent<CustomTypeBuilderWi
...
-    router.replace("/");
+    void router.replace("/");
     return null;
   }
 
   return (
     <CustomTypeBuilderWithProvider
       customType={customType}
       remoteCustomType={remoteCustomType}
-      // eslint-disable-next-line @typescript-eslint/no-empty-function
```
Comment (MarcMcIntosh) [inline packages/slice-machine/pages/cts/[ct].tsx:60]: nice

packages/slice-machine/pages/_app.tsx
```diff
@@ -96,6 +97,10 @@ function MyApp({ Component, pageProps }: AppContext & AppInitialProps) {
     if (!storeInitiated) {
       const { store, persistor } = configureStore({
         environment: { env: serverState.env, warnings: [], configErrors: {} },
+        customTypes: {
```
MarcMcIntosh, line 100: are updates handled correctly?

packages/slice-machine/package.json
```diff
@@ -163,6 +164,7 @@
     "react-tooltip": "^4.2.8",
     "redux-persist": "^6.0.0",
     "redux-saga": "^1.1.3",
+    "redux-saga-test-plan": "^4.0.4",
```
MarcMcIntosh, line 167: instresting

packages/slice-machine/lib/models/common/widgets/ContentRelationship/Form.js
```diff
@@ -28,9 +27,9 @@ const WidgetForm = ({
   fields,
   setFieldValue,
 }) => {
-  const { customTypes } = useContext(CustomTypesContext);
-
-  // const { errors, Model, fieldType } = rest
+  const { customTypes } = useSelector((store) => ({
```
MarcMcIntosh: is an object t needed here?

packages/slice-machine/components/UpdateVersionModal/index.tsx
```diff
@@ -31,7 +31,7 @@ const UpdateVersionModal: React.FC = () => {
     useSliceMachineActions();
 
   React.useEffect(() => {
-    if (updateVersionInfo.updateAvailable) openUpdateVersionModal();
+    if (updateVersionInfo.updateAvailable && !isOpen) openUpdateVersionModal();
```
MarcMcIntosh, line 34: pretty sure isOpen got removed and added back in a couple of time

packages/slice-machine/components/ModalFormCard/index.tsx
```diff
@@ -1,140 +1,151 @@
...
 import SliceMachineModal from "@components/SliceMachineModal";
-import { Formik, Form } from "formik";
-import { Flex, Heading, Close, Box, Button } from "theme-ui";
+import { Formik, Form, FormikErrors, FormikTouched } from "formik";
+import { Flex, Heading, Close, Box, Button as ThemeButton } from "theme-ui";
+import Button from "@components/Button";
 
 import Card from "../Card";
 
 Modal.setAppElement("#__next");
```
MarcMcIntosh, line 9: If testing this you'll need to create this element before rendering.
```
const div = document.createElement("div")
div.setAttribute("id", "__next")
document.body.appendChild(div)
render(<ModalCard />)
```

packages/slice-machine/components/Forms/CreateCustomTypeModal.tsx
```diff
@@ -1,73 +1,99 @@
...
+          values.repeatable
+        );
+      }}
+      isLoading={isCreatingCustomType}
+      initialValues={{
+        repeatable: true,
+        id: "",
+        label: "",
+      }}
+      validate={({ id, label }) => {
```
MarcMcIntosh, line 52: remind to check how this validation works

packages/slice-machine/components/Button/index.tsx
```diff
@@ -4,23 +4,32 @@ import { Button as ThemeUIButton, Spinner } from "theme-ui";
 import { ThemeUIStyleObject } from "@theme-ui/css";
 
 export type SliceMachineButtonProps = {
+  form?: string;
+  type?: "submit" | "reset" | "button" | undefined;
```
MarcMcIntosh: if it's optional doesn't that imply `undefined`?

MarcMcIntosh (commented):
https://redux.js.org/usage/writing-tests#connected-components
Suggests injecting the state via render.

Are you sure we should be remove `test-utils` and mocking redux?

packages/slice-machine/src/redux/store.ts
Miniplop: yes, and you can emit actions (push , replace, etc ...) to trigger router changes

packages/slice-machine/tests/project/cypress/integration/customTypes/00-create.spec.js
Miniplop: to see if the custom type is well saved on the fs

packages/slice-machine/components/Button/index.tsx
Miniplop: I prefer loosing 5s one time on writing every props int the component instead of loosing 10s guessing what props is given to the component every time I read it :p

packages/slice-machine/src/modules/customTypes/index.ts
Miniplop: It's an action to push on the nuxt router a new route

packages/slice-machine/tests/components/UpdateVersionModal.tsx
Miniplop: I think the best way of testing components is to do integration tests (as the redux documentation suggested). 

As we discussed on the tests strategy meeting, our priority number one is to spend all our effort on writing unit tests, because they are quicker to write and give quicker feedbacks.

So we have to stop spending time to test react components for the moment, the question is what we want to do for this test ? 
1) We can remove it 
2) We can reduce his scope by transforming it into a unit tests 
3) Other options ?

But for sure, we cannot keep a test that requires to be updated every time we changed the SM state just for testing a minor feature (the cost is heavy for a tiny gain).

What do you suggest ?

packages/slice-machine/pages/index.tsx
Miniplop: 😅

packages/slice-machine/pages/cts/[ct].tsx
Comment (Miniplop) [reply packages/slice-machine/pages/cts/[ct].tsx:51]: oh yeah 😎

packages/slice-machine/pages/cts/[ct].tsx
Comment (Miniplop) [reply packages/slice-machine/pages/cts/[ct].tsx:60]: oh yeah 😎

packages/slice-machine/pages/_app.tsx
Miniplop: Yes take a look at the useServerState file

packages/slice-machine/package.json
Miniplop: I added unit tests on 2/3 sagas

packages/slice-machine/lib/models/common/widgets/ContentRelationship/Form.js
Miniplop: I think we can change it

packages/slice-machine/components/UpdateVersionModal/index.tsx
Miniplop: I will investigate 🧐

packages/slice-machine/components/ModalFormCard/index.tsx
Miniplop: As we discussed in the test strategy, react component tests is not the focus yet. So not a big deal for now

packages/slice-machine/components/Button/index.tsx
Miniplop: I will check

packages/slice-machine/components/UpdateVersionModal/index.tsx
Miniplop: I took a look to the Github history and there is no back-and-forth with this isOpen status. I think we you are confusing with the fact that the updateVersionModal popped up too many times. This is not related to that, it's just to avoid emitting actions when the modal is already open :)

packages/slice-machine/components/Button/index.tsx
Miniplop: you are right :)

packages/slice-machine/src/apiClient.ts
Miniplop: The post method accept 3 types : 
- the type of the response
- a type to override the response type
- the data type for the body

here the signature : 

```
post<T = any, R = AxiosResponse<T>, D = any>(url: string, data?: D, config?: AxiosRequestConfig<D>): Promise<R>;
```

When I do : 

```
  return axios.post<Record<string, never>, AxiosResponse<Record<string, never>>, SaveCustomTypeBody>("/api/custom-types/save", requestBody, defaultAxiosConfig);

```

I can put whatever I want into the data attribute, do you know a way of making this work correctly ?

packages/slice-machine/components/UpdateVersionModal/index.tsx
MarcMcIntosh: k

MarcMcIntosh (approved):

PR #322, opened 2022-01-26
Improve framework helpers

Heyrows:
## Description

- Improve framework helpers, modify helper to find frameworks based on the content of a package json for the editor-team.
- Improve init build with typescript check


packages/core/__tests__/src/utils/framework.test.ts
```diff
@@ -16,10 +16,7 @@ describe("framework.detectFramework", () => {
     mockedFs.lstatSync.mockReturnValue({ dev: 1 } as fs.Stats);
     mockedFs.readFileSync.mockReturnValue(JSON.stringify({ dependencies: {} }));
```
Miniplop: I think we can split a little bit this test into test for the detectFramework functions and tests for the defineFrameworks functions, and use a different describe for them

packages/core/__tests__/src/utils/framework.test.ts
```diff
@@ -60,17 +59,12 @@ describe("framework.detectFramework", () => {
...
 
-      const fallback =
-        value === Frameworks.gatsby ? Frameworks.gatsby : Frameworks.vanillajs;
-
-      const wanted = FrameworkUtils.isValidFramework(value as Frameworks)
+      // if framework is not supported, vanillajs should always be sent back.
+      const wanted = FrameworkUtils.isFrameworkSupported(value as Frameworks)
         ? value
-        : fallback;
+        : Frameworks.vanillajs;
```
Miniplop: it is not a good practice to write conditions in the tests because that means we need a tests to test it :p. 

The solution here : splitting more the test into different tests for define frameworks / isFrameworkSupported.

packages/core/src/utils/framework.ts
```diff
@@ -1,15 +1,16 @@
-import { retrieveJsonPackage } from "../filesystem";
+import { retrieveJsonPackage, JsonPackage } from "../filesystem";
 import { Frameworks, SupportedFrameworks } from "../models/Framework";
 import { Manifest } from "../models/Manifest";
 
 export { Frameworks } from "../models/Framework";
 
 export const UnsupportedFrameWorks = Object.values(Frameworks).filter(
```
Miniplop, line 7: No unit tests around this function ?

Miniplop (commented):
Nice I let some comments, plus can you remove the slice-machine-ui package-lock.json from this PR, you didn't make any changes in the package.json

Miniplop (approved):

PR #325, opened 2022-01-27
Track init script usage

Miniplop:
Add events on the init script + handle the anonymous mode in the init


packages/init/src/index.ts
```diff
@@ -19,7 +20,14 @@ async function init() {
   const lib: string | undefined = findArgument(process.argv, "library");
   const branch: string | undefined = findArgument(process.argv, "branch");
   const isTrackingAvailable =
-    findArgument(process.argv, "tracking") === "false" ? false : true;
+    findArgument(process.argv, "tracking") !== "false";
+
+  Tracker.get().initialize(
```
Heyrows, line 25: Does the tracker.get still make sense here ?
Just a challenge, we don't need to modify it but i'm wondering since it's a script does it make sense to secure the tracker and all knowing that we could instantiate it in another file and just call the helpers here.
Like instantiate it in memory and then see if we want to call the helpers or not

packages/init/src/index.ts
```diff
@@ -41,46 +49,39 @@ async function init() {
...
   );
 
   if (!existing) {
-    await createRepository(name, frameworkResult.value, config);
+    await createRepository(repository, frameworkResult.value, config);
+  }
+
+  // If we get the info from the profile we want to identify all the previous events sent or continue in anonymous mode
+  if (user.profile) {
+    Tracker.get().identifyUser(user.profile.userId);
```
Heyrows: - Shouldn't we use the shortID here ?
- I think this should be done right after getting the user.
- I personally don't see the point to identify the user for one event (start) :x, I would rather have had only this event when the user logs in and put the login at the first position in the steps.
- I think profile should never be undefined, what I mean is if you're identified, the profile call should not fail. Maybe the profile call could be what we use to verify that the user is logged in.

packages/init/src/index.ts
```diff
@@ -41,46 +49,39 @@ async function init() {
...
-    : undefined;
-
-  // install the required dependencies in the project.
+  // Install the required dependencies in the project.
   await installRequiredDependencies(cwd, frameworkResult.value);
 
-  const sliceLibPath = lib
-    ? await installLib(tracker, cwd, lib, branch)
-    : undefined;
+  const sliceLibPath = lib ? await installLib(cwd, lib, branch) : undefined;
```
Heyrows, line 70: opt
```suggestion
  const sliceLibPath = lib && await installLib(cwd, lib, branch);
```

packages/init/src/steps/install-lib.ts
```diff
@@ -103,7 +102,7 @@ export async function installLib(
       `Slice library "${libGithubPath}" was installed successfully`
     );
 
-    tracker?.Track.downloadLibrary(libGithubPath);
+    Tracker.get().trackDownloadLibrary(libGithubPath);
```
Heyrows, line 105: should this be catch to not make the rest fails or does the fact that there is no await covers it ?

packages/init/src/utils/communication.ts
```diff
@@ -20,10 +20,13 @@ export async function getUserProfile(
   cookies: string,
   base = Utils.CONSTS.DEFAULT_BASE
 ): Promise<UserProfile> {
-  // note the auth server also provides a userId
+  const userServiceBase =
```
Heyrows, line 23: note If the base was like `prismic.io | wroom.io`, it would make that easier to define and we could test it on other bases. 😄

packages/init/src/utils/communication.ts
```diff
@@ -32,7 +35,7 @@ export async function getUserProfile(
   return axios
     .get<UserProfile>(endpoint, {
       headers: {
-        Authorization: `Bearer Token ${token}`,
+        Authorization: `Bearer ${token}`,
```
Heyrows, line 38: that does seems better but are you sure it won't break things ?

packages/init/src/utils/tracker.ts
```diff
@@ -1,57 +1,115 @@
...
+  initialize(segmentKey: string, isTrackingActive = true): void {
     try {
-      if (!repo) return;
-      const analytics = new ServerAnalytics(writeKey);
-      return new Tracker(analytics, repo, identifier, tracking);
+      // The initialize method cannot be used for changing some class properties
+      if (!!this.#client) return;
+
+      this.#isTrackingActive = isTrackingActive;
+      this.#anonymousId = uuidv4();
```
Heyrows, line 25: Note: are we forced to define it ?
Can't the lib take the deviceID by default like the frontend lib does ?

packages/init/src/index.ts
Heyrows: Furthermore after looking at the code, i'm more convince about point 3.
The login being the first step and a mandatory step, if the user doesn't login it means he didn't start the script so I would definitely
- let the user login
- create the tracker with profile data
- launch the start event right after the initialisation.

Heyrows (approved):

PR #329, opened 2022-01-28
fix: suppport apiendpoints using the cdn

MarcMcIntosh:
## Types of changes

## Checklist:

### DIFF
--- a/packages/core/__tests__/src/models.test.ts
+++ b/packages/core/__tests__/src/models.test.ts
@@ -86,4 +86,21 @@ describe("Manifest", () => {
     );
     expect(decoded).toBeNull();
   });
+
+  test("apiEnpoint: https://foo-bar.cdn.prismic.io/api/v2", () => {
+    const input = { apiEndpoint: "https://foo-bar.cdn.prismic.io/api/v2" };
+    const result = Manifest.decode(input);
+    expect(isRight(result)).toBeTruthy();
+    const decoded = pipe(
+      result,
+      fold(
+        (e) => {
+          console.log(e);
+          return null;
+        },
+        (d: Manifest) => d
+      )
+    );
+    expect(decoded?.apiEndpoint).toEqual(input.apiEndpoint);
+  });
 });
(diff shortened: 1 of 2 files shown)


packages/core/__tests__/src/models.test.ts
```diff
@@ -86,4 +86,21 @@ describe("Manifest", () => {
     );
     expect(decoded).toBeNull();
   });
+
+  test("apiEnpoint: https://foo-bar.cdn.prismic.io/api/v2", () => {
+    const input = { apiEndpoint: "https://foo-bar.cdn.prismic.io/api/v2" };
+    const result = Manifest.decode(input);
+    expect(isRight(result)).toBeTruthy();
+    const decoded = pipe(
```
Heyrows, line 94: If the test before says "isRight" = true, I think you can do result.right.
Like the either in scala

packages/core/src/models/Manifest.ts
```diff
@@ -9,8 +9,8 @@ class ApiEndPointType extends t.Type<string> {
       (input, context) => {
         if (typeof input !== "string") return t.failure(input, context);
         const regx = new RegExp(
-          "^https?://[a-z0-9][a-z0-9-]{2,}[a-z0-9].(prismic.io|wroom.io|wroom.test)/api/v2/?$",
-          "i"
+          "^https?://[a-z0-9][a-z0-9-]{2,}[a-z0-9](.cdn)?.(prismic.io|wroom.io|wroom.test)/api/v2/?$",
```
Heyrows, line 12: I've tested the Regexp online and it seems that the regexp now returns 2 groups: `cdn` and `prismic.io`.
How do you know that your function will return the second parameter now ?

MarcMcIntosh: it users the the `test` method, it only returns true or false

packages/core/__tests__/src/models.test.ts
MarcMcIntosh: That's not how io-ts works,

it returns an object with a _tag prop that's either "right" or "left"

MarcMcIntosh: https://github.com/gcanti/fp-ts/blob/master/src/internal.ts#L33

Heyrows (approved):

PR #330, opened 2022-01-28
[WIP] push slice to prismic spinners

MarcMcIntosh:
Found some issues when converting old code to type script, the spinners them self are easy to sort out. the types are more difficult


MarcMcIntosh:
Note: ideally both loadiing and image-loading wouldn't be set at the same time, but this require further refactoring off `packages/slice-machine/lib/models/ui/Store.ts` and everything that uses it. I've added refactoring this to the technical debt board

packages/slice-machine/lib/builders/SliceBuilder/Header/index.tsx
```diff
@@ -55,91 +70,62 @@ const Header = ({ Model, store, variation, onSave, onPush, isLoading }) => {
...
-                    {
-                      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-                      Model.infos.sliceName
-                    }{" "}
-                    /{" "}
-                    {
-                      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-                      variation.name
-                    }
+                    {`/ ${Model.infos.sliceName} / ${variation.name}`}
```
Heyrows, line 74: why do you need string interpolation ? wasn't it good before ?

packages/slice-machine/lib/builders/SliceBuilder/Header/index.tsx
```diff
@@ -55,91 +70,62 @@ const Header = ({ Model, store, variation, onSave, onPush, isLoading }) => {
...
           />
           <VariationModal
             isOpen={showVariationModal}
             onClose={() => setShowVariationModal(false)}
             onSubmit={(id, name, copiedVariation) => {
-              // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
               store.copyVariation(id, name, copiedVariation);
-              // eslint-disable-next-line @typescript-eslint/no-floating-promises
-              router.push(
+              return router.push(
```
Heyrows, line 122: Note: if you don't want to return it, I think you can do 
```suggestion
              void router.push(
```

packages/slice-machine/lib/builders/SliceBuilder/SideBar/components/FooterButton.tsx
```diff
@@ -1,9 +1,15 @@
-import { Button, Flex, Spinner } from "theme-ui";
```
Heyrows: Remove the file if it's dead code, no need to transform it to typescript

packages/slice-machine/lib/builders/SliceBuilder/Header/SaveButton.tsx
```diff
@@ -0,0 +1,52 @@
+import React from "react";
+import { Spinner, Button } from "theme-ui";
+
+const SaveButton: React.FC<{
+  __status: string;
+  isTouched?: boolean;
+  onSave: () => void;
+  onPush: () => void;
+  isLoading: boolean;
+  imageLoading: boolean;
```
Heyrows: call it disabled please, it's a button and it's used that way.

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -110,6 +110,7 @@ const SliceBuilder: React.FunctionComponent = () => {
         // @ts-expect-error
         onSave={() => store.save(Model, setData)}
         isLoading={data.loading}
+        imageLoading={data.imageLoading}
```
Heyrows, line 113: I understand why you can't change the shape of data, that would hard.
But please change Header parameter's name, we don't want to pass `imageLoading` everywhere.

Heyrows (changes requested):

packages/slice-machine/lib/builders/SliceBuilder/Header/index.tsx
MarcMcIntosh: same thing, just easier to read

packages/slice-machine/lib/builders/SliceBuilder/Header/index.tsx
MarcMcIntosh: I think return works better as the process is async

packages/slice-machine/lib/builders/SliceBuilder/Header/index.tsx
Heyrows: ok 👍 , did you know that it's prettier that adds the `{" "}` in the tsx XD ?

packages/slice-machine/lib/builders/SliceBuilder/Header/index.tsx
Heyrows: Yeah but we don't need what the function returns.
I'm supposing you put `return` for eslint.
The good way to do it for a promise on which you don't need the result is void.

But that's a `Note`, so don't change it if you don't want to. That was just me trying to spread what i consider best practices 😃

packages/slice-machine/lib/builders/SliceBuilder/Header/index.tsx
MarcMcIntosh: nope, I didn't know that

Heyrows (approved):
I didn't require so much extraction out of the button, just the rename of the disable.
But it's fine and it does looks nicer than before.

PR #332, opened 2022-01-31
Update deploy-alpha.yml

MarcMcIntosh:
force success condition on delpoy alpha
I noticed when merging a PR from the community that the workflow ran although the tests failed, this will make it require that the test work flow run and has a positive result

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/deploy-alpha.yml
+++ b/.github/workflows/deploy-alpha.yml
@@ -12,6 +12,7 @@ on:
 jobs:
   deployment:
     runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
     environment: alpha
     steps: 
       - uses: actions/checkout@v2


.github/workflows/deploy-alpha.yml
```diff
@@ -12,6 +12,7 @@ on:
 jobs:
   deployment:
     runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
Heyrows, line 15: You should use needs instead.
CF a quick look at the doc : https://docs.github.com/en/actions/using-workflows/advanced-workflow-features#creating-dependent-jobs 😉 

Maybe we could redefine each step of the CI using this.
I don't know 🤷‍♂️

Heyrows (changes requested):

.github/workflows/deploy-alpha.yml
MarcMcIntosh: needs is for jobs in the same file :/

PR #334, opened 2022-02-01
Rename the builder to Slice Machine everywhere

Miniplop:
Rename the builder to Slice Machine everywhere


MarcMcIntosh (approved):
`grep -irn --exclude-dir=node_modules --exclude-dir=build --exclude-dir=out --exclude-dir=coverage "builder" packages/` still shows a feew uses of builder, but it's mostly in the package description and in the change log, and also one in `packages//slice-machine/pages/onboarding.tsx:86:    <SubHeader>Group your Slices as page builders</SubHeader>`

PR #337, opened 2022-02-02
Improve send snapshot code

Miniplop:
Refacto the push slice route + add unit test

### DIFF
--- a/packages/slice-machine/lib/migrate.js
+++ b/packages/slice-machine/lib/migrate.js
@@ -37,9 +37,9 @@ const migrate = (model, info, env, write = true) => {
   if (write) {
     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-member-access
     const modelPath = CustomPaths(env.cwd)
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-member-access
       .library(info.from)
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-member-access
       .slice(info.sliceName)
       .model();
     // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
--- a/packages/slice-machine/lib/models/common/Environment.ts
+++ b/packages/slice-machine/lib/models/common/Environment.ts
@@ -21,6 +21,8 @@ export interface BackendEnvironment {
   mockConfig: any;
   framework: Models.Frameworks;
   baseUrl: string;
+  // Here to replace the fake client
+  isUserLoggedIn: boolean;
   client: DefaultClient | FakeClient;
 }
 
--- a/packages/slice-machine/lib/models/common/http/DefaultClient.ts
+++ b/packages/slice-machine/lib/models/common/http/DefaultClient.ts
@@ -149,10 +149,9 @@ export default class DefaultClient {
         return {};
       }
       try {
-        return Files.readJson(path.join(cwd, "sm.dev.json")) as Record<
-          string,
-          DevConfig
-        >;
+        return Files.readJson<Record<string, DevConfig>>(
+          path.join(cwd, "sm.dev.json")
+        );
       } catch (e) {
         return {};
       }
@@ -174,10 +173,6 @@ export default class DefaultClient {
       repo,
       auth
     );
-  }
-
-  isFake(): boolean {
-    return false;
   }
 
   async getSlice(): Promise<Response> {
--- a/packages/slice-machine/lib/models/common/http/FakeClient.ts
+++ b/packages/slice-machine/lib/models/common/http/FakeClient.ts
@@ -94,8 +94,4 @@ export default class FakeClient {
       return { status: 200 };
     },
   };
-
-  isFake() {
-    return true;
-  }
 }
--- a/packages/slice-machine/lib/utils/files.ts
+++ b/packages/slice-machine/lib/utils/files.ts
@@ -47,9 +47,9 @@ const Files = {
     }
   },
 
-  readJson(pathToFile: string) {
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
-    return JSON.parse(this.readString(pathToFile));
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  readJson<Return = any>(pathToFile: string): Return {
+    return JSON.parse(this.readString(pathToFile)) as Return;
   },
   safeReadJson(pathToFile: string) {
     try {
--- a/packages/slice-machine/server/src/api/custom-types/index.ts
+++ b/packages/slice-machine/server/src/api/custom-types/index.ts
@@ -36,7 +36,7 @@ const handleMatch = (matches: string[], env: BackendEnvironment) => {
 const fetchRemoteCustomTypes = async (
   env: BackendEnvironment
 ): Promise<{ remoteCustomTypes: CustomTypeJsonModel[]; isFake?: boolean }> => {
-  if (env.client.isFake()) {
+  if (!env.isUserLoggedIn) {
     return { remoteCustomTypes: [], isFake: true };
   }
   try {
--- a/packages/slice-machine/server/src/api/custom-types/index.ts
+++ b/packages/slice-machine/server/src/api/custom-types/index.ts
@@ -35,10 +35,9 @@ const handleMatch = (matches: string[], env: BackendEnvironment) => {
 
 const fetchRemoteCustomTypes = async (
   env: BackendEnvironment
-): Promise<{ remoteCustomTypes: CustomTypeJsonModel[]; isFake?: boolean }> => {
-  if (!env.isUserLoggedIn) {
-    return { remoteCustomTypes: [], isFake: true };
-  }
+): Promise<{ remoteCustomTypes: CustomTypeJsonModel[] }> => {
+  if (!env.isUserLoggedIn) return { remoteCustomTypes: [] };
+
   try {
     const res = await env.client.getCustomTypes();
     const { remoteCustomTypes } = await (async (): Promise<{
@@ -71,7 +70,6 @@ const saveCustomTypes = (
 };
 
 export default async function handler(env: BackendEnvironment): Promise<{
-  isFake: boolean;
   customTypes: ReadonlyArray<CustomType<ObjectTabs>>;
   remoteCustomTypes: ReadonlyArray<CustomType<ObjectTabs>>;
 }> {
@@ -81,7 +79,7 @@ export default async function handler(env: BackendEnvironment): Promise<{
   // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
   const folderExists = Files.exists(pathToCustomTypes);
 
-  const { remoteCustomTypes, isFake } = await fetchRemoteCustomTypes(env);
+  const { remoteCustomTypes } = await fetchRemoteCustomTypes(env);
 
   if (!folderExists) {
     saveCustomTypes(remoteCustomTypes, cwd);
@@ -89,7 +87,6 @@ export default async function handler(env: BackendEnvironment): Promise<{
   // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
   const matches = glob.sync(`${pathToCustomTypes}/**/index.json`);
   return {
-    isFake: !!isFake,
     customTypes: handleMatch(matches, env),
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     remoteCustomTypes: remoteCustomTypes.map((ct: any) =>
--- a/packages/slice-machine/server/src/api/custom-types/push.ts
+++ b/packages/slice-machine/server/src/api/custom-types/push.ts
@@ -7,6 +7,7 @@ import Files from "@lib/utils/files";
 import { CustomTypesPaths } from "@lib/models/paths";
 import DefaultClient from "@lib/models/common/http/DefaultClient";
 import FakeClient from "@lib/models/common/http/FakeClient";
+import { ApiResult } from "@lib/models/server/ApiResult";
 
 import { ComponentUI } from "@lib/models/common/ComponentUI";
 import { Tab } from "@lib/models/common/CustomType/tab";
@@ -28,9 +29,7 @@ const createOrUpdate = (
   return client.insertCustomType(model);
 };
 
-export default async function handler(
-  req: RequestWithEnv
-): Promise<ReturnType<typeof onError> | Record<string, never>> {
+export default async function handler(req: RequestWithEnv): Promise<ApiResult> {
   const { id } = req.query;
 
   const state = await getBackendState(req.errors, req.env);
--- a/packages/slice-machine/server/src/api/custom-types/push.ts
+++ b/packages/slice-machine/server/src/api/custom-types/push.ts
@@ -46,15 +46,16 @@ export default async function handler(
     };
   }
 
-  if (state.clientError || state.isFake) {
+  if (state.clientError || !state.env.isUserLoggedIn) {
     const isAnAuthenticationError =
-      state.isFake || (state.clientError && state.clientError.status === 403);
+      !state.env.isUserLoggedIn ||
+      (state.clientError && state.clientError.status === 403);
     const errorExplanation = isAnAuthenticationError
       ? "Please log in to Prismic!"
       : // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
         `You don\'t have access to the repo \"${state.env.repo}\"`;
 
-    const errorCode = state.isFake
+    const errorCode = !state.env.isUserLoggedIn
       ? 403
       : state.clientError
       ? state.clientError.status
--- a/packages/slice-machine/server/src/api/custom-types/push.ts
+++ b/packages/slice-machine/server/src/api/custom-types/push.ts
@@ -1,12 +1,12 @@
 import { getBackendState } from "../state";
-import { handler as pushSlice } from "../slices/push";
+import { pushSlice } from "../slices/push";
 import { handler as saveSlice } from "../slices/save";
 
 import { onError } from "../common/error";
 import Files from "@lib/utils/files";
 import { CustomTypesPaths } from "@lib/models/paths";
 import DefaultClient from "@lib/models/common/http/DefaultClient";
-import FakeClient, { FakeResponse } from "@lib/models/common/http/FakeClient";
+import FakeClient from "@lib/models/common/http/FakeClient";
 
 import { ComponentUI } from "@lib/models/common/ComponentUI";
 import { Tab } from "@lib/models/common/CustomType/tab";
@@ -28,11 +28,9 @@ const createOrUpdate = (
   return client.insertCustomType(model);
 };
 
-export default async function handler(req: RequestWithEnv): Promise<{
-  err: Response | FakeResponse | Error | null;
-  reason: string | null;
-  status: number;
-}> {
+export default async function handler(
+  req: RequestWithEnv
+): Promise<ReturnType<typeof onError> | Record<string, never>> {
   const { id } = req.query;
 
   const state = await getBackendState(req.errors, req.env);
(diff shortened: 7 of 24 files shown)


packages/slice-machine/lib/migrate.js
```diff
@@ -37,9 +37,9 @@ const migrate = (model, info, env, write = true) => {
   if (write) {
     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-member-access
     const modelPath = CustomPaths(env.cwd)
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-member-access
```
MarcMcIntosh, line 40: migrating to typescript should fix these issues

packages/slice-machine/lib/models/common/Environment.ts
```diff
@@ -21,6 +21,8 @@ export interface BackendEnvironment {
   mockConfig: any;
   framework: Models.Frameworks;
   baseUrl: string;
+  // Here to replace the fake client
```
MarcMcIntosh, line 24: Why do we have fake client in the first palce?

packages/slice-machine/lib/models/common/http/DefaultClient.ts
```diff
@@ -176,10 +175,6 @@ export default class DefaultClient {
     );
   }
 
-  isFake(): boolean {
```
MarcMcIntosh, line 179: dead code?

packages/slice-machine/lib/models/common/http/FakeClient.ts
```diff
@@ -94,8 +94,4 @@ export default class FakeClient {
       return { status: 200 };
     },
   };
-
-  isFake() {
```
MarcMcIntosh, line 98: dead code?

packages/slice-machine/lib/utils/files.ts
```diff
@@ -47,9 +47,9 @@ const Files = {
     }
   },
 
-  readJson(pathToFile: string) {
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
-    return JSON.parse(this.readString(pathToFile));
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  readJson<Return = any>(pathToFile: string): Return {
+    return JSON.parse(this.readString(pathToFile)) as Return;
```
MarcMcIntosh, line 52: Is there a JSON type any where?

MarcMcIntosh: or should this be a generic function?

packages/slice-machine/server/src/api/custom-types/index.ts
```diff
@@ -36,7 +36,7 @@ const handleMatch = (matches: string[], env: BackendEnvironment) => {
 const fetchRemoteCustomTypes = async (
   env: BackendEnvironment
 ): Promise<{ remoteCustomTypes: CustomTypeJsonModel[]; isFake?: boolean }> => {
-  if (env.client.isFake()) {
+  if (!env.isUserLoggedIn) {
     return { remoteCustomTypes: [], isFake: true };
```
MarcMcIntosh: isFake in the response?

packages/slice-machine/server/src/api/index.ts
```diff
@@ -174,13 +175,13 @@ router.get(
     >,
     res: express.Response
   ): Promise<Express.Response> {
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call
     const payload = await pushSlice(req.query);
```
MarcMcIntosh, line 178: can this throw?

packages/slice-machine/server/src/api/index.ts
```diff
@@ -209,11 +210,11 @@ router.get(
   ): Promise<Express.Response> {
     // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call
     const payload = await pushCustomType(req);
```
MarcMcIntosh, line 212: same

packages/slice-machine/server/src/api/services/uploadScreenshotClient.ts
```diff
@@ -57,7 +50,7 @@ export async function upload(
     console.error(msg);
     console.error(`Full error: ${JSON.stringify(aclResponse)}`);
     // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-argument
-    return onError(aclResponse, msg);
+    return { err: onError(aclResponse, msg) };
```
MarcMcIntosh, line 54: what does onError do?

packages/slice-machine/server/src/api/services/sliceService.ts
```diff
@@ -0,0 +1,71 @@
...
+  from: string
+): Promise<{ [variationId: string]: string | null }> {
+  const variationIds: string[] = sliceModel.variations.map(
+    (v: Models.VariationAsObject) => v.id
+  );
+
+  const imageUrlsByVariation: { [variationId: string]: string | null } = {};
+
+  for (let i = 0; i < variationIds.length; i += 1) {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
```
MarcMcIntosh: how is this unsafe?

packages/slice-machine/server/src/api/services/sliceService.ts
```diff
@@ -0,0 +1,71 @@
...
+    const variationId = variationIds[i];
+
+    const screenshot = resolvePathsToScreenshot({
+      paths: [env.cwd, path.join(env.cwd, ".slicemachine/assets")],
+      from,
+      sliceName,
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
MarcMcIntosh: same

packages/slice-machine/server/src/api/services/sliceService.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (!!screenshot) {
+      const {
+        err,
+        s3ImageUrl,
+      }: { err?: ReturnType<typeof onError>; s3ImageUrl?: string } =
+        await upload(env, sliceName, variationId, screenshot.path);
+
+      if (err) throw new Error(err.reason);
+
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
MarcMcIntosh: safe

packages/slice-machine/server/src/api/services/sliceService.ts
```diff
@@ -0,0 +1,71 @@
...
+      imageUrlsByVariation[variationId] = !!s3ImageUrl ? s3ImageUrl : null;
+    } else {
+      console.error(
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/restrict-template-expressions
```
MarcMcIntosh: same

packages/slice-machine/server/src/api/services/sliceService.ts
```diff
@@ -0,0 +1,71 @@
...
+        `--- Unable to find a screenshot for slice ${sliceName} | variation ${variationId}`
+      );
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
MarcMcIntosh: same

MarcMcIntosh (commented):
Some questions

packages/slice-machine/lib/migrate.js
Heyrows: done in the start script migration :)

packages/slice-machine/lib/models/common/Environment.ts
Heyrows: We don't know why it was created at the beginning, but it seems it is used when you're not logged in instead of the client.
That's a bad practice that we should at some point remove.

packages/slice-machine/lib/utils/files.ts
Heyrows: It's casting so i guess it's normal to not have general type.
The problem is those functions are doubled with what we have in the core :x

packages/slice-machine/server/src/api/custom-types/index.ts
Heyrows: good point

packages/slice-machine/server/src/api/custom-types/push.ts
```diff
@@ -28,11 +28,9 @@ const createOrUpdate = (
...
 };
 
-export default async function handler(req: RequestWithEnv): Promise<{
-  err: Response | FakeResponse | Error | null;
-  reason: string | null;
-  status: number;
-}> {
+export default async function handler(
+  req: RequestWithEnv
+): Promise<ReturnType<typeof onError> | Record<string, never>> {
```
Heyrows: i think the type was clearer before.

packages/slice-machine/server/src/api/custom-types/push.ts
```diff
@@ -155,9 +153,5 @@ export default async function handler(req: RequestWithEnv): Promise<{
 
   // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
   console.log(`[custom-types/push] Custom Type ${id} was pushed!`);
-  return {
-    err: null,
-    reason: null,
-    status: 200,
-  };
+  return {};
```
Heyrows, line 156: doesn't return anything anymore, no error or anything.

packages/slice-machine/server/src/api/index.ts
```diff
@@ -226,9 +227,6 @@ router.get(
     res: express.Response
   ): Promise<Express.Response> {
     const payload = await validateAuth();
-    if (payload.err) {
```
Heyrows, line 229: are we sure about this :x ?

packages/slice-machine/server/src/api/index.ts
```diff
@@ -305,4 +303,14 @@ router.use("*", async function (req: express.Request, res: express.Response) {
   });
 });
 
+function isPayloadHasError<Request extends Record<string, unknown>>(
+  payload: ReturnType<typeof onError> | Request
+): payload is ReturnType<typeof onError> {
```
Heyrows: seeing how much ReturnType<typeof onError> is used so much, i would make it a type with a name

packages/slice-machine/server/src/api/libraries.ts
```diff
@@ -25,7 +25,7 @@ export default async function handler(
           clientError: new ErrorWithStatus(res.statusText, res.status),
         };
       }
-      if (env.client.isFake()) {
+      if (!env.isUserLoggedIn) {
         return { remoteSlices: [] };
```
Heyrows, line 29: this one does not return isFake like the remoteCustomType routes, maybe isFake is dead code on the other

packages/slice-machine/server/src/api/services/sliceService.ts
```diff
@@ -0,0 +1,71 @@
...
+import FakeClient from "@models/common/http/FakeClient";
+import { snakelize } from "@lib/utils/str";
+
+export const createOrUpdate = async (
+  slices: ReadonlyArray<Models.SliceAsObject>,
+  sliceName: string,
+  model: Models.SliceAsObject,
+  client: DefaultClient | FakeClient
+) => {
+  if (slices.find((e) => e.id === snakelize(sliceName))) {
```
Heyrows, line 18: I think we should use the model rather than the snakelize.
This could be a source of bugs if we decide one day to create slice IDs differently.
This condition would fuck up older slices potentially

packages/slice-machine/server/src/api/services/sliceService.ts
```diff
@@ -0,0 +1,71 @@
...
+    return await client.insertSlice(model);
+  }
+};
+
+export async function uploadScreenshots(
+  env: BackendEnvironment,
+  sliceModel: Models.SliceAsObject,
+  sliceName: string,
+  from: string
+): Promise<{ [variationId: string]: string | null }> {
```
Heyrows, line 30: We had a full module with everything related to screenshots in a folder so i'm a bit perplex about this.

Heyrows: Guess we should decide to make an api or modules and then we'll stick to it and adapt what we have today.

Heyrows: In any case it's still a really good thing to have exported that.

packages/slice-machine/server/src/api/index.ts
Heyrows: It's design to send specific errors and if it throws then it would go into the 500 catch that you added last time :)

packages/slice-machine/server/src/api/index.ts
```diff
@@ -305,4 +303,14 @@ router.use("*", async function (req: express.Request, res: express.Response) {
...
+  return (
+    typeof payload === "object" &&
+    payload !== null &&
+    payload.hasOwnProperty("err")
```
Heyrows: this part conflict a bit with one things you did in this PR -> `{ err: OnError(..., ...) }`.
I don't know if that's intentional.

packages/slice-machine/server/src/api/services/getEnv.ts
```diff
@@ -116,6 +116,7 @@ export default async function getEnv(
         supportedFrameworks: Models.SupportedFrameworks,
         manifest: manifestInfo.content,
       }),
+      isUserLoggedIn: !!prismicData.value.auth && !!repo,
```
Heyrows, line 119: This is good but maybe it hold too many responsabilities.
- if the user is not logged in
- if the user is not in the repo
- if the repo doesn't exists

packages/slice-machine/server/src/api/slices/push.ts
```diff
@@ -103,17 +42,20 @@ export async function handler(
...
       sliceName,
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-      model: {
-        ...jsonModel,
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-        variations,
-      },
-      client: env.client,
-    });
+      // This type is wrong because there is the property imageUrl inside it.
```
Heyrows: in that case maybe we can destruct the object before using it.

packages/slice-machine/tests/server/services/sliceService.test.ts
```diff
@@ -0,0 +1,116 @@
...
+jest.mock("../../../lib/models/common/http/DefaultClient", () => {
+  return jest.fn().mockImplementation(() => {
+    return {
+      updateSlice: mockUpdateSlice,
+      insertSlice: mockInsertSlice,
+    };
+  });
+});
+
+jest.mock("@slicemachine/core/build/src/libraries/screenshot", () => {
```
Heyrows, line 25: funny, didn't know you could target specific file to prevent having to rebuild the Core with mocks.

packages/slice-machine/tests/server/services/sliceService.test.ts
```diff
@@ -0,0 +1,116 @@
...
+  };
+});
+
+jest.mock("../../../server/src/api/services/uploadScreenshotClient", () => {
+  return {
+    upload: jest.fn(),
+  };
+});
+
+describe("createOrUpdate", () => {
```
Heyrows: name should be SliceService i guess

packages/slice-machine/server/src/api/slices/push.ts
```diff
@@ -1,97 +1,36 @@
...
-}) => {
-  if (slices.find((e) => e.id === snakelize(sliceName))) {
-    return await client.updateSlice(model);
-  } else {
-    return await client.insertSlice(model);
-  }
-};
-
-export async function handler(
+export async function pushSlice(
```
Heyrows, line 15: This could be in the SliceService as well.

packages/slice-machine/lib/models/common/http/FakeClient.ts
Heyrows: yes, replace by isUserLoggedIn

packages/slice-machine/server/src/api/custom-types/index.ts
Heyrows: I removed all the isFake from slicemachine 👍

packages/slice-machine/server/src/api/index.ts
Heyrows: There's a try catch inside the pushSlice (the one on this line is a default export) that is used at multiple places, it returns an "OnError".

packages/slice-machine/server/src/api/services/uploadScreenshotClient.ts
Heyrows: onError returns the usual object that we expect { err, reason, status }

packages/slice-machine/server/src/api/index.ts
Heyrows: Same as slices, a big try catch inside the function itself

packages/slice-machine/server/src/api/services/sliceService.ts
Heyrows: it was not good catch, i think the unsafe warning came from the core not being rebuild

packages/slice-machine/server/src/api/state.ts
```diff
@@ -54,17 +54,15 @@ export const getBackendState = async (
...
-  );
+  const { customTypes, remoteCustomTypes } = await fetchCustomTypes(env);
 
   const base = preferWroomBase(env.manifest.apiEndpoint);
   if (base !== env.prismicData.base) {
     FileSystem.PrismicSharedConfigManager.setProperties({ base });
   }
 
   // Refresh auth
-  if (!isFake && env.prismicData.auth) {
```
Miniplop, line 67: As env.isUserLoggedIn === !isFake 

=> We should rewrite this condition like this 

```
  if (env.isUserLoggedIn && env.prismicData.auth)
```

packages/slice-machine/server/src/api/custom-types/push.ts
Miniplop: Yes, I did this to match the push slice route behavior => return an empty object on success

But we can change the push slice route to return a fake object like this one if we want to enforce the return type to some key (err, reason) etc ...

packages/slice-machine/server/src/api/index.ts
Miniplop: valide auth cannot return the key "err" in the payload so this condition is useless

Miniplop: That also show that this route is not protected at all against unexpected error

packages/slice-machine/server/src/api/index.ts
Miniplop: We can

packages/slice-machine/server/src/api/index.ts
Miniplop: Why ?

packages/slice-machine/server/src/api/libraries.ts
Miniplop: yeah, good catch we can also improve this

packages/slice-machine/server/src/api/services/getEnv.ts
Miniplop: I copy this condition from the initClient fonction that I replaced. Maybe the condition can be improved but I think that will introduce some regressions. 

So if we want to improve this condition, we have to have a dedicated task to clean the auth on the server side and ensure that we don't introduce any regressions. Probably something we want to do when removing the fake client/Default client

packages/slice-machine/server/src/api/services/sliceService.ts
Miniplop: I don't get your point what do you want to do ?

packages/slice-machine/server/src/api/services/sliceService.ts
Miniplop: Sure we can move this piece of code to the screenshot service

packages/slice-machine/server/src/api/slices/push.ts
Miniplop: Nop we cannot because the function createOrUpdate use Models.SliceAsObject and this model doesn't have the imageUrl property, so there is 2 options : 
- This property `imageUrl` is not used by the custom Type API and we can remove the extra logic on the function
- This property `imageUrl` is used by the custom API and we can update the Models.SliceAsObject type to add this property (probably a huge rework)

I will check the Custom Type API, to check what is the good option

Heyrows (approved):

PR #339, opened 2022-02-03
[WIP] Mm/summary of new releases

MarcMcIntosh:
Screenshot 2022-02-07 at 16 31 57

## Types of changes

## Checklist:


packages/slice-machine/lib/env/npmApi.ts
```diff
@@ -0,0 +1,58 @@
+import npmFetch from "npm-registry-fetch";
```
Miniplop, line 1: This file is only used by the server, so we can put it in the server folder :)

packages/slice-machine/lib/env/npmApi.ts
```diff
@@ -0,0 +1,58 @@
...
+  const majorVersion = current.replace(/^(\d+).*/, "$1");
+
+  const top = versions.reduce((acc, version) => {
+    if (/^\d+\.\d+\.\d+$/.test(version) === false) return acc;
+    if (semver.gt(acc, version)) return acc;
+    if (kind === "patch" && version.startsWith(minorVersion)) return version;
+    if (
+      kind === "minor" &&
+      version.startsWith(majorVersion) &&
+      version.startsWith(minorVersion) === false
```
Miniplop: Can you simplify this condition with `!version.startsWith(minorVersion)`

packages/slice-machine/lib/env/npmApi.ts
```diff
@@ -0,0 +1,58 @@
...
+    )
+      return version;
+    if (kind === "major" && version.startsWith(majorVersion) === false)
```
Miniplop: Can you simplify this condition with `!version.startsWith(majorVersion)`

packages/slice-machine/tests/lib/npmApi.test.ts
```diff
@@ -0,0 +1,151 @@
+import {
```
Miniplop, line 1: Cool

Miniplop (commented):
Can you also remove the updateVersionModal ?

Heyrows:
@MarcMcIntosh The text / space between doesn't seem to be the same.

I prefer you text tho, but
- we need two different messages (depending if major/minor updates are available or just one of them)
- the text seems to be justify compared to the design.

image

packages/slice-machine/lib/env/npmApi.ts
```diff
@@ -0,0 +1,58 @@
...
+  return npmFetch
+    .json(name)
+    .then((json) => {
+      const versions = (json.versions || {}) as Record<string, unknown>;
+      return Object.keys(versions);
+    })
+    .catch(() => []);
+}
+
+const highest = (
```
Heyrows: the name is not very intuitive.
Since we're searching on an array, I would call it `findHighestUpdate` or something like that.

When a function has the name find, it implies that it's return can be undefined.

packages/slice-machine/lib/models/common/Environment.ts
```diff
@@ -9,6 +9,11 @@ export interface UpdateVersionInfo {
   packageManager: "npm" | "yarn";
   updateCommand: string;
   updateAvailable: boolean;
+  availableVersions: {
```
Heyrows, line 9: It should be called availableUpdates.
AvailableVersions is too global for the data we're trying to form.
Also, each key should be potentially undefined.

If there is no patch, minor or major then this object should be empty.

packages/slice-machine/lib/env/semver.ts
```diff
@@ -65,12 +73,18 @@ export const createComparator =
...
 
+      const versions = await getAvailableVersionInfo(
+        manifest.name,
+        manifest.version
+      );
+
       return {
         currentVersion: manifest.version,
         onlinePackage,
         updateAvailable,
```
Heyrows, line 85: updateAvailable should come from the fact that there is nothing in the versions object.
```suggestion
        updateAvailable: availableUpdates.patch || availableUpdates.minor || availableUpdates.major
```

packages/slice-machine/lib/env/npmApi.ts
```diff
@@ -0,0 +1,58 @@
...
+      return version;
+    if (kind === "major" && version.startsWith(majorVersion) === false)
+      return version;
+    return acc;
+  }, current);
+
+  return top === current ? "" : top;
+};
+
+export function whatSortOfVersions(
```
Heyrows: This name is not intuitive at all.

Heyrows: A name is the first step to having code understandable easily by people.

Heyrows (commented):
After talking a bit with Come, the Path should be displayed as well.
The design & colors will be the same for patch & minor, just the text would be different (patch instead of minor).

packages/slice-machine/lib/env/npmApi.ts
MarcMcIntosh: it's also only used here `packages/slice-machine/lib/env/semver.ts` make more sense to keep it next to the file where it's used?

packages/slice-machine/lib/env/npmApi.ts
MarcMcIntosh: sure, I think `==false` is more readable, but I'll do it

MarcMcIntosh:
Screenshot 2022-02-08 at 10 08 42

MarcMcIntosh:
Screenshot 2022-02-08 at 10 46 05
Screenshot 2022-02-08 at 10 46 13
Screenshot 2022-02-08 at 10 46 20

packages/slice-machine/lib/models/common/Environment.ts
MarcMcIntosh: `UpdateVersionInfo.availableVersions` is too global?

Miniplop (approved):
LGTM

PR #340, opened 2022-02-04
fix(env.client): take the base from the api endpoint in sm.json

MarcMcIntosh:
## Types of changes

## Checklist:


packages/slice-machine/server/src/api/services/getEnv.ts
```diff
@@ -88,12 +89,12 @@ export default async function getEnv(
...
   const mockConfig = getMockConfig(cwd);
 
-  const client = initClient(
-    cwd,
-    prismicData.value.base,
-    repo,
-    prismicData.value.auth
-  );
+  const base = preferWroomBase(manifestInfo.content.apiEndpoint);
+  const auth = prismicData.value.base.startsWith(base)
```
Heyrows: Good, but I would have done that in the function that retrieve the `prismicData`.
By passing the manifest to it.
That way we ensure that whenever we need the `.prismic` auth informations, we get the manifest first and it is used to know if we're logged in or not.

packages/slice-machine/package.json
```diff
@@ -28,7 +28,7 @@
     "build-server": "tsc --project server/tsconfig.json",
     "build-next-app": "cross-env NODE_OPTIONS=--max_old_space_size=8192 next build",
     "export-build": "next export && npm run clean-build",
-    "bundle": "npm run test && npm run build && npm run export-build",
```
MarcMcIntosh, line 31: are you sure about doing that?

packages/slice-machine/server/src/api/services/getPrismicData.ts
```diff
@@ -17,13 +25,9 @@ export default function getPrismicData(): Result<PrismicData, ErrorWithStatus> {
...
     }
 
-    const authResult = Utils.Cookie.parsePrismicAuthToken(
-      prismicConfig.cookies
-    );
-
     return ok({
       ...prismicData,
-      auth: authResult,
+      auth: Utils.Cookie.parsePrismicAuthToken(prismicConfig.cookies),
```
MarcMcIntosh, line 30: I'm assuming when it throws it'll go to the catch block

MarcMcIntosh (commented):
nice

packages/slice-machine/server/src/api/services/getPrismicData.ts
Heyrows: I didn't actually change the code here, just avoided to declare the variable :)

packages/slice-machine/package.json
Heyrows: I just removed the test part.
I've been thinking about it for a while, running the test or just part of the test when doing is the bundle takes more time for what it provide.
I think we used to do that before having CI.

PR #342, opened 2022-02-07
[WIP]: plugin architecture

MarcMcIntosh:
Very basic implementation of a plugin system so more implementation details can be discussed.

PR contains a plugin container/runner, some methods and a dummy plugin.

container should be able to handle plugins from node_modules and a local file (for development)

#### Properties a plugin should have.
So far I've thought of, creating a slice, creating a story, creating the index file and the hints.

#### Hints: 
Will have to be sent from the server to the client, as the plugin is loaded server-side.... not sure if there's another way.

#### Simulator
We could add the simulator to the plugin, that way all the user has to do is `export simulator from "plugin-foo/simulator"

#### SliceZone
We could also add the sliceZone to the plugin, `import SliceZone from "plugin-foo/slice-zone" at least this way everything for the frame work is in one place?

#### Migrating the current architecture.
TBD: That's where a lot of the heavy lifting will have to happen as we have, next, nuxt, vue, react, vanilla, none, and svelte to consider.

#### notion discussion (I did not know about this)
https://www.notion.so/prismic/Proposed-Slice-Machine-Architecture-d1733c840b28485f8d9cc5c37376a9c1#8eb364129ad6418f811c1480cb5c9cae

# TODO:
Test this out using a deicated build, or a an npm proxy

loading modules from the cocnfig is tricky when links are involved.

[ ] hints in for the custom-type builder
[ ] update frameworks or plugins
[ ] add some tests


MarcMcIntosh:
https://www.notion.so/prismic/Proposed-Slice-Machine-Architecture-d1733c840b28485f8d9cc5c37376a9c1#17c71d6890144ca8a57065a6c29b2b72

Interesting, some frameworks require custom mocks, so an optional mock method should be added

MarcMcIntosh:
Tests for server/src/api/slces/create.ts slice result in this 
```
 FAIL  tests/server/create-slice.test.js
  ● Test suite failed to run

    TypeError: Cannot read properties of undefined (reading 'bind')

      at Object.<anonymous> (node_modules/puppeteer/src/node/BrowserFetcher.ts:127:43)
      at Object.<anonymous> (node_modules/puppeteer/src/node/Puppeteer.ts:22:1)
```

MarcMcIntosh:
There some edge cases with the hints where the widgets types on the server side are a subset of the widgets on the client side.

This will require some investigation.

PR #344, opened 2022-02-07
Fix Status on slice list page

Miniplop:
This PR fix 3 bugs : 
- When you go on a slice detail page, the status is always set to modified even if you don't do any modifications
- After pushing a slice to prismic the status was not changed to synced on the client side
- A slice synced with the remote slice had the "modified" status instead of "synced"

### DIFF
--- a/packages/slice-machine/src/models/slice/reducer.ts
+++ b/packages/slice-machine/src/models/slice/reducer.ts
@@ -203,11 +203,7 @@ export function reducer(
   return {
     ...result,
     isTouched: (() => {
-      // If the model is not pushed the model cannot be touched
-      if (result.__status === LibStatus.NewSlice) {
-        return false;
-      }
-
+      // Should be the slice be saved ?
       return (
         !equal(result.initialVariations, result.variations) ||
         !equal(result.initialMockConfig, result.mockConfig)
@@ -223,12 +219,14 @@ export function reducer(
         result.screenshotUrls,
         result.initialScreenshotUrls
       );
+
+      // Should be the slice be pushed ?
       const isModelModified = !compareVariations(
         result.remoteVariations,
         result.initialVariations
       );
 
-      return isScreenshotModified || isModelModified
+      return isModelModified || isScreenshotModified
         ? LibStatus.Modified
         : result.__status;
     })(),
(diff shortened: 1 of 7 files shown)


packages/slice-machine/src/models/slice/reducer.ts
```diff
@@ -202,16 +203,34 @@ export function reducer(
   return {
     ...result,
     isTouched: (() => {
+      // If the model is not pushed the model cannot be touched
+      if (result.__status === LibStatus.NewSlice) {
```
Miniplop: Mmm I think this introduce a regression I need to check :)

packages/slice-machine/src/models/slice/reducer.ts
```diff
@@ -202,16 +203,34 @@ export function reducer(
   return {
     ...result,
     isTouched: (() => {
```
Heyrows, line 205: Here's what I don't get, if we have a version to compareVersions
- 1, it must be a bit heavy so I would do it only once.
- 2, why isn't "isTouched" a property instead of a function

packages/slice-machine/src/models/slice/reducer.ts
```diff
@@ -202,16 +203,34 @@ export function reducer(
...
       return (
         !equal(result.initialVariations, result.variations) ||
         !equal(result.initialMockConfig, result.mockConfig)
       );
     })(),
     __status: (() => {
-      return !equal(result.screenshotUrls, result.initialScreenshotUrls) ||
-        !compareVariations(result.remoteVariations, result.initialVariations)
+      // If the model is not pushed the only status possible is new slice
+      if (result.__status === LibStatus.NewSlice) {
```
Heyrows, line 214: On the ComponentUI there is a function to compute the status, we should reuse it here to keep the logic only once.

packages/slice-machine/src/models/slice/reducer.ts
```diff
@@ -36,6 +36,7 @@ export function reducer(
       case SliceActions.Push:
         return {
           ...prevState,
+          __status: LibStatus.Synced,
```
Heyrows, line 39: I'm not sure this is the right move.
In a sense that all the other case don't have a status set.

Heyrows: It should be set only once by the `__status: (() => {})` below based on row data.

Heyrows (approved):

PR #350, opened 2022-02-09
Rename default variationID

Heyrows:
## Description

Rename the default variation ID of a slice from `default-slice` to `default`


Miniplop (approved):
LGTM

PR #351, opened 2022-02-09
Create the slice module + connect the create slice modal to it

Miniplop:
Clean the slice creation + send all the logic into saga. 

For now we cannot avoid reloading the whole page due to the libraryContext not updated when adding a new Slice. 

For doing that we need : 
- A clearer component Model (without any data duplicated or dead code)
- Returning the slice created on the slice/create route to injected in the store
- A way of migrating the slice list page into redux without removing the LibContext (used for creating the slice context)

These steps are too long for one task, and we need to evaluate migrations options before starting it :)


packages/slice-machine/components/ModalFormCard/index.tsx
```diff
@@ -25,6 +25,7 @@ type ModalCardProps<T> = {
   close: () => void;
   isOpen: boolean;
   formId: string;
+  buttonLabel?: string;
```
MarcMcIntosh, line 28: was children already taken?

packages/slice-machine/server/src/api/slices/create/index.ts
```diff
@@ -75,39 +64,23 @@ const fromTemplate = async (
   return copyTemplate(env, templatePath, from, sliceName);
 };
 
-export default async function handler({
-  sliceName,
-  from,
-  values,
-}: SliceCreateBody) {
+export default async function handler({ sliceName, from }: SliceBody) {
```
MarcMcIntosh, line 67: I tried to unravel this function last night (side project): I think I'll have to make snap-shot with memfs to avoid breaking it :/

packages/slice-machine/src/apiClient.ts
```diff
@@ -30,15 +31,31 @@ export const saveCustomType = (
...
+};
+
 /** Auth Routes **/
 
 export const startAuth = (): Promise<AxiosResponse<Record<string, never>>> =>
   axios.post("/api/auth/start", {}, defaultAxiosConfig);
 
 export const checkAuthStatus = (): Promise<CheckAuthStatusResponse> =>
   axios
-    .post<CheckAuthStatusResponse>("/api/auth/status", {}, defaultAxiosConfig)
```
MarcMcIntosh, line 40: Why change this?
the types are equivalent?

MarcMcIntosh (commented):
A few questions :)

packages/slice-machine/components/ModalFormCard/index.tsx
Miniplop: Yes they are, but the children are just for the form content not the submit button which is inserted in the DOm by this component

packages/slice-machine/server/src/api/slices/create/index.ts
Miniplop: I don't get it, what to mean by unravel ? You are refactoring this function ?

packages/slice-machine/src/apiClient.ts
Miniplop: I think that for using the same method for both createSlice and checkAuthStatus. I will check why I used this method, don't remember :)

packages/slice-machine/server/src/api/slices/create/index.ts
MarcMcIntosh: Yes some refactoring for supporting plugins, 

by unravel.... I mean work out what's all going on in it (I'd probably need more displays for that)

I eventually gave up trying to test it because of weird errors caused by puppeter

packages/slice-machine/src/apiClient.ts
Miniplop: Typescript doesn't infers the type in the `then` function  if we put the type on the `post` generic

MarcMcIntosh (approved):

PR #352, opened 2022-02-09
Remove prismicData base

Heyrows:
## Description

Remove totally the use & set of .prismic base.


packages/slice-machine/tests/server/env.test.js
```diff
@@ -211,59 +211,4 @@ describe("getEnv", () => {
     const { env } = await getEnv(TMP);
     expect(env.framework).toEqual("vanillajs");
   });
-
-  test("it should not consider the user logged in if the base from .prismic is different than the one in SM.json", async () => {
```
MarcMcIntosh, line 215: probably worth keeping one of these to make sure the cookie is loaded (not sure if that's covered else where)

packages/slice-machine/tests/server/services/getPrismicData.test.ts
```diff
@@ -23,42 +22,21 @@ describe("getPrismicData", () => {
       os.homedir()
     );
 
-    const result = await getPrismicData(base);
+    const result = getPrismicData();
     expect(result.isOk()).toBeTruthy();
-    expect(result.isOk() && result.value.base).toEqual(base);
```
MarcMcIntosh, line 28: what's base going to be now?

MarcMcIntosh (commented):
Nice to see a lot of removals :)

packages/slice-machine/tests/server/services/getPrismicData.test.ts
Heyrows: getPrismicData doesn't return a base anymore.

Heyrows: So the base has to be retrieved from the SM.json like you did

MarcMcIntosh (approved):

PR #359, opened 2022-02-15
Update the toaster lib + design

Miniplop:
Update the toaster lib + design


Miniplop:
Error toaster 

image

Success toaster

Capture d’écran 2022-02-16 à 18 07 01

Info toaster 

Capture d’écran 2022-02-16 à 18 10 39

Warning toaster

image

packages/slice-machine/pages/_app.tsx
```diff
@@ -15,9 +15,11 @@ import Tracker from "@src/tracker";
 import "react-tabs/style/react-tabs.css";
 import "rc-drawer/assets/index.css";
 import "react-datepicker/dist/react-datepicker.css";
+import "react-toastify/dist/ReactToastify.css";
```
Heyrows, line 18: already imported in packages/slice-machine/components/ToasterContainer/index.tsx

packages/slice-machine/src/modules/customTypes/index.ts
```diff
@@ -91,15 +92,32 @@ export const customTypesReducer: Reducer<
...
+    yield put(push(`/cts/${payload.id}`));
+    yield put(
+      openToasterCreator({
+        message: "Custom type saved",
+        type: ToasterType.SUCCESS,
+      })
+    );
+  } catch (e) {
+    yield put(
+      openToasterCreator({
```
Heyrows, line 115: Question, if Events can be used that way, why do we use dispatch in the frontend usually in the actions.

Or the other way around why don't we need it here ?

packages/slice-machine/tests/components/LoginModal.spec.tsx
```diff
@@ -31,29 +30,7 @@ jest.mock("react-redux", () => ({
...
-    useToasts: () => ({ addToast: jest.fn() }),
-  };
-});
-
-const App = () => (
-  <ToastProvider>
-    <LoginModal />
-  </ToastProvider>
-);
+const App = () => <LoginModal />;
```
Heyrows, line 33: feels good to see that

packages/slice-machine/tests/src/modules/customTypes/index.test.ts
```diff
@@ -101,6 +102,33 @@ describe("[Custom types module]", () => {
...
+        actionPayload.label,
+        actionPayload.repeatable
+      );
+      const saga = testSaga(
+        createCustomTypeSaga,
+        createCustomTypeCreator.request(actionPayload)
+      );
+
+      saga.next().call(saveCustomType, customTypeCreated, {});
+      saga.throw(new Error()).put(
```
Heyrows, line 126: not sure to understand those tests because I don't see any expect.

packages/slice-machine/src/modules/toaster/utils.tsx
```diff
@@ -0,0 +1,29 @@
...
+export interface ToastPayload {
+  loading: boolean;
+  done: boolean;
+  warning: boolean;
+  message: string;
+  status: number;
+  error: Error | null;
+}
+
+export const handleRemoteResponse =
```
Heyrows, line 12: I'm not sure what this is suppose to do.

packages/slice-machine/pages/_app.tsx
Miniplop: oupsi something forgotten during the rebase :/

packages/slice-machine/tests/components/LoginModal.spec.tsx
Miniplop: As your service

packages/slice-machine/tests/src/modules/customTypes/index.test.ts
Miniplop: Yeah this test use a lib redux-saga-test-plan, a lib that is very useful for unit testing generator functions, each time I call next() the generator is jumping in the next step. Then the lib provide helpers 'call', 'throw', 'put' and others to expect specific side effects from the saga. 

I will do a specific formation for that

packages/slice-machine/pages/_app.tsx
Miniplop: fixed

Heyrows (approved):
Looks good to me, maybe the design needs to be challenge when there is a lot of text.

PR #363, opened 2022-02-18
Add components to library index

Heyrows:
## Description

The goal of this PR is to provide a components object in the format of
```
 { Slice_API_ID : SliceComponent }
```

So that it can be used by newer kits.

extra: as we now export our own object components as we would export a slice, the words `components` needs to be reserved so that the user can't create a slice called `components`


packages/slice-machine/server/src/api/common/hooks/onSaveSlice.ts
```diff
@@ -7,13 +7,35 @@ import { BackendEnvironment } from "@lib/models/common/Environment";
...
+  const { imports, exportList, componentsProperties } = lib.components.reduce(
+    (
+      acc: {
+        imports: string;
+        exportList: string;
+        componentsProperties: string;
+      },
+      component
+    ) => {
+      const i =
```
Miniplop: can we found a better name for this variable, like importLine

packages/slice-machine/server/src/api/common/hooks/onSaveSlice.ts
```diff
@@ -7,13 +7,35 @@ import { BackendEnvironment } from "@lib/models/common/Environment";
...
+        acc.imports +
+        `import ${component.infos.sliceName} from '${component.infos.sliceName}';\n`;
+      const e = acc.exportList + `\t${component.infos.sliceName},\n`;
```
Miniplop: Same here

packages/slice-machine/server/src/api/common/hooks/onSaveSlice.ts
```diff
@@ -7,13 +7,35 @@ import { BackendEnvironment } from "@lib/models/common/Environment";
...
+      const c =
```
Miniplop: Same here

Miniplop (commented):
Nice, I think that's also a good occasion to add some unit tests on the function : `createIndexFile` to make sure that this critical function. is battle tested :)

Miniplop (approved):
LGTM

PR #367, opened 2022-02-21
Mm/add legacy react

MarcMcIntosh:
## Types of changes

## Checklist:


packages/slice-machine/server/src/api/simulator.ts
```diff
@@ -38,20 +65,8 @@ export default async function handler(
...
-          Utils.CONSTS.NEXT_SLICEZONE,
-          Utils.CONSTS.PRISMIC_REACT_PACKAGE_NAME,
-        ]
-      : [
-          Utils.CONSTS.SLICE_SIMULATOR_VUE,
-          Utils.CONSTS.NUXT_SM,
-          Utils.CONSTS.VUE_SLICEZONE,
-          Utils.CONSTS.NUXT_PRISMIC,
-        ];
+  // TODO: fix this for legacy frameworks, and svelte?
```
Miniplop: don't get it, svelte and other frameworks doesn't support the simulator why we should handle them at this point ?

Miniplop (changes requested):
Two questions : 
-  Why creating the "previous-react" framework ? Do we want to be retro compatible on this framework ? 
- I think we avoid introducing previous and legacy word for representing the same thing, so I think we should rename all the "legacy" word into "previous". It will help us to maintain and remove easily this feature.

packages/slice-machine/server/src/api/simulator.ts
MarcMcIntosh: good point, what about for legacy nuxt?

Miniplop: It's in another task to support it :)

packages/slice-machine/lib/builders/common/Zone/Card/components/Hints/Renderers/index.tsx
```diff
@@ -1,13 +1,15 @@
 import toVue from "./vue";
 import toReact from "./react";
 import toSvelte from "./svelte";
+import toPreviousReact from "./previous-react";
 
 export const svelte = toSvelte;
 
 export const vue = toVue;
-export const react = toReact;
+export const react = toPreviousReact;
```
Miniplop: I think here you want to use to React as the new user will use the helpers lib :)

packages/core/src/utils/const.ts
```diff
@@ -3,9 +3,11 @@ export const SCRIPT_VALUE = "start-slicemachine";
 export const INIT_COMMAND = "npx sm init";
 export const SM_PACKAGE_NAME = "slice-machine-ui";
 
-export const PRISMIC_REACT_PACKAGE_NAME = "prismic-reactjs";
+export const PREVIOUS_REACT_PACKAGE_NAME = "prismic-reactjs";
```
Miniplop, line 6: Is this variable still used ?

MarcMcIntosh: yes

packages/slice-machine/lib/builders/common/Zone/Card/components/Hints/Renderers/index.tsx
MarcMcIntosh: Then that would be the new version

Miniplop: yes that's what we want, no backward compatibility :)

Miniplop (approved):
LGTM

PR #375, opened 2022-02-22
Page tracking

Miniplop:
## Types of changes

## Checklist:

### DIFF
--- a/packages/slice-machine/pages/_app.tsx
+++ b/packages/slice-machine/pages/_app.tsx
@@ -1,7 +1,6 @@
 import React, { useEffect, useState } from "react";
 import { Provider } from "react-redux";
 import configureStore from "src/redux/store";
-import useSwr from "swr";
 import App, { AppContext } from "next/app";
 import { PersistGate } from "redux-persist/integration/react";
 import { ThemeProvider, BaseStyles, useThemeUI } from "theme-ui";
@@ -10,7 +9,6 @@ import theme from "src/theme";
 
 import LoadingPage from "components/LoadingPage";
 import SliceMachineApp from "components/App";
-import Tracker from "@src/tracker";
 
 import "react-tabs/style/react-tabs.css";
 import "rc-drawer/assets/index.css";
@@ -24,43 +22,13 @@ import "src/css/toaster.css";
 import "highlight.js/styles/atom-one-dark.css";
 
 import ServerState from "lib/models/server/ServerState";
-import { LibraryUI } from "lib/models/common/LibraryUI";
 
 import Head from "next/head";
 import { AppInitialProps } from "next/dist/shared/lib/utils";
 import { Store } from "redux";
 import { Persistor } from "redux-persist/es/types";
 import { ConnectedRouter } from "connected-next-router";
-
-/*
- * TEMPORARY
- * Today, we need to have access to the cookie to get the anonymousID from segment but
- * we want to log that event only at start time. Because of the necessity to access the cookies,
- * we can't put that code on the start script just yet.
- * We need to define a potential alternative for this event.
- * In the meantime, this flag prevents the event flood
- */
-let _temp_first_start_flag = true;
-
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-async function fetcher(url: string): Promise<any> {
-  return fetch(url).then((res) => res.json());
-}
-
-function mapSlices(libraries: ReadonlyArray<LibraryUI> | null) {
-  return (libraries || []).reduce((acc, lib) => {
-    return {
-      ...acc,
-      ...lib.components.reduce(
-        (acc, comp) => ({
-          ...acc,
-          [`${comp.from}:${comp.infos.sliceName}`]: 1,
-        }),
-        {}
-      ),
-    };
-  }, {});
-}
+import { getState } from "@src/apiClient";
 
 const RemoveDarkMode: React.FunctionComponent = ({ children }) => {
   const { setColorMode } = useThemeUI();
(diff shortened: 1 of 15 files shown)


packages/slice-machine/pages/_app.tsx
```diff
@@ -74,77 +42,38 @@ const RemoveDarkMode: React.FunctionComponent = ({ children }) => {
...
-      Object.keys(newSliceMap).forEach((key) => {
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-        if (!sliceMap[key]) {
-          return (window.location.href = `/slices`);
-        }
-      });
-    }
-    setSliceMap(newSliceMap);
-    const { env, configErrors, warnings, libraries } = serverState;
-    console.log("------ SliceMachine log ------");
```
MarcMcIntosh, line 143: we're removing the logs?

packages/slice-machine/server/src/api/index.ts
```diff
@@ -64,11 +64,7 @@ router.get(
   ): Promise<Express.Response> {
     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call
     const payload = await state(req);
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-    if (payload.clientError) {
```
MarcMcIntosh, line 68: how are errors sent to the client now?

packages/slice-machine/src/apiClient.ts
```diff
@@ -17,6 +18,12 @@ const defaultAxiosConfig = {
   },
 };
 
+/** State Routes **/
+
+export const getState = (): Promise<AxiosResponse<ServerState>> => {
+  return axios.get("/api/state", defaultAxiosConfig);
```
MarcMcIntosh: you can pass `ServerState` as a type parameter to `axios.get` not sure if it makes a difference here, as the response object is kept

packages/slice-machine/src/hooks/useSMTracker.tsx
```diff
@@ -0,0 +1,68 @@
...
+  getCurrentVersion,
+  getFramework,
+  getIsTrackingAvailable,
+  getRepoName,
+  getShortId,
+} from "@src/modules/environment";
+import { getLibraries } from "@src/modules/slices";
+import { useRouter } from "next/router";
+
+const useSMTracker = () => {
```
MarcMcIntosh, line 16: neat

packages/slice-machine/src/hooks/useServerState.tsx
```diff
@@ -2,20 +2,23 @@ import { useEffect } from "react";
 
 import ServerState from "@lib/models/server/ServerState";
 import useSliceMachineActions from "@src/modules/useSliceMachineActions";
+import useSwr from "swr";
+import axios, { AxiosResponse } from "axios";
 
-const useServerState = (serverState: ServerState) => {
-  const { getState } = useSliceMachineActions();
+const fetcher = (url: string): Promise<ServerState> =>
+  axios.get(url).then((res: AxiosResponse<ServerState>) => res.data);
```
MarcMcIntosh, line 9: `axios.get<ServerState>` and the res will be inferred for you :)

packages/slice-machine/src/hooks/useServerState.tsx
```diff
@@ -2,20 +2,23 @@ import { useEffect } from "react";
...
-    serverState.configErrors,
-    serverState.customTypes,
-    serverState.remoteCustomTypes,
-    serverState.libraries,
-  ]);
+    if (!serverState) {
+      return;
+    }
+
+    refreshState(serverState);
```
MarcMcIntosh, line 20: what's refresh state do?

packages/slice-machine/src/tracker.ts
```diff
@@ -63,34 +57,41 @@ export class SMTracker {
...
 
-  async #group(attributes: Record<string, unknown> = {}): Promise<void> {
+  async #identify(userId: string): Promise<void> {
     if (!this.#isTrackingPossible(this.#client)) {
       return;
     }
 
-    const repoName = this.#repoName;
+    return this.#client
+      .then((client): void => {
```
MarcMcIntosh, line 77: why void?
plus if you add void before the `client.identify` promise the warning should go away

packages/slice-machine/src/tracker.ts
```diff
@@ -63,34 +57,41 @@ export class SMTracker {
...
+  async #group(attributes: Record<string, unknown> = {}): Promise<void> {
+    if (!this.#isTrackingPossible(this.#client)) {
       return;
     }
 
     return this.#client
       .then((client): void => {
         // eslint-disable-next-line @typescript-eslint/no-floating-promises
-        client.group(repoName, attributes);
+        client.group(attributes);
```
MarcMcIntosh: you can add void here to :)

MarcMcIntosh (commented):
I have a few questions and recommendations :)

packages/slice-machine/server/src/api/index.ts
Miniplop: The error are not sent anymore but in the client side we never use the fact that this call can send a different status code that's why I removed it.

packages/slice-machine/src/hooks/useServerState.tsx
Miniplop: send the data coming from the /api/state into the redux store

packages/slice-machine/pages/_app.tsx
Miniplop: did you use it ? I never used it

packages/slice-machine/src/apiClient.ts
Miniplop: The type we put in the `axios.get` function is only use on `then`, `catch`etc ... functions, but doesn't help typescript to infer the return type of the function, so now I put both :)

packages/slice-machine/src/hooks/useSMTracker.tsx
Miniplop: 😎

packages/slice-machine/src/hooks/useServerState.tsx
Miniplop: Typescript can't inferred the `then` type if I use the generics, don't know why

packages/slice-machine/src/tracker.ts
Miniplop: Because this function return nothing, just calling the segment segment.

Miniplop: I add the void on the promise to remove the eslint comment

MarcMcIntosh (approved):

PR #379, opened 2022-02-24
Migration v0.3.0

Miniplop:
write the installed version inside the latest key in the sm.json + handle the v0.3.0 migration


packages/init/src/steps/configure-project.ts
```diff
@@ -19,13 +23,18 @@ export function configureProject(
...
+    const sliceMachineVersionInstalled =
+      getTheSliceMachineVersionInstalled(packageJson);
 
     const manifestUpdated: Models.Manifest = {
       ...(manifest.exists && manifest.content ? manifest.content : {}),
       apiEndpoint: Utils.Endpoints.buildRepositoryEndpoint(base, repository),
       libraries: ["@/slices", ...sliceLibPath],
       ...(framework.manuallyAdded ? { framework: framework.value } : {}),
       ...(!tracking ? { tracking } : {}),
+      _latest: sliceMachineVersionInstalled,
```
Heyrows: i'm wondering, usually this represent the latest migration that was executed.
Now we're putting the latest version installed.

This should work but i'm thinking we should only set it for new projects.
To not prevent the migrations from happening.

ex:
- the user has 0.1.0 on his sm.json
- he runs the init to update SM
- the 0.3.0 is put in his sm.json
- the latest migration doesn't run because according to the sm.json his code is up to date.

packages/init/src/steps/configure-project.ts
```diff
@@ -47,3 +56,41 @@ export function configureProject(
...
+) => {
+  const sliceMachinePackageInstalled = Object.entries(
+    packageJson.content?.devDependencies || {}
+  ).find((devDependency) => {
+    if (devDependency[0] === Utils.CONSTS.SM_PACKAGE_NAME) {
+      return devDependency;
+    }
+  });
+
+  const defaultSliceMachineVersion = "0.1.0";
```
Heyrows: this is not the version of the first migration :).
It 0.0.41 I believe. It is the default version used in the migration code.

packages/init/src/steps/configure-project.ts
```diff
@@ -47,3 +56,41 @@ export function configureProject(
     process.exit(-1);
   }
 }
+
+const getTheSliceMachineVersionInstalled = (
+  packageJson: FileContent<JsonPackage>
+) => {
+  const sliceMachinePackageInstalled = Object.entries(
```
Heyrows, line 67: This is forcing us to kinda install SM before, or at least make sure the installation is done before to avoid going on default.

Heyrows: OPT: Maybe we could make a test to cover that make sure steps are done in the right order.

packages/init/src/steps/configure-project.ts
Heyrows: Also it could be declared above the function.

packages/init/src/steps/configure-project.ts
Heyrows: I think if we had that, we would never need the default.

packages/slice-machine/scripts/migration/versions/0.3.0/index.ts
```diff
@@ -0,0 +1,37 @@
...
+const migration: Migration = {
+  version: "0.3.0",
+  // eslint-disable-next-line @typescript-eslint/require-await
+  main: async function main({ cwd }): Promise<void> {
+    try {
+      const manifest = FileSystem.retrieveManifest(cwd);
+      if (!manifest.exists || !manifest.content) return;
+
+      const framework = Utils.Framework.defineFramework({ cwd });
+      let previousFramework = null;
```
Heyrows: I think making a big condition
```suggestion
      if (framework === Models.Frameworks.next || framework === Models.Frameworks.nuxt) {}
```
would be better to avoid patching the manifest if we don't need to.

Heyrows: and then 
```suggestion
      frameworkToSet = framework === Models.Frameworsk.next ? Models.Framework.previousNext : Models.Framework.previousNuxt
```

packages/init/src/steps/configure-project.ts
Miniplop: I don't think the user will use the init to do that, but at least we can prevent overriding the _latest if it exist yes

packages/init/src/steps/configure-project.ts
Miniplop: Yes we can change that even if this property is only used if the slicemachine is not installed in the project :p

packages/init/src/steps/configure-project.ts
Miniplop: I don't think we can get ride of the default case, because we cannot trust the previous steps in this step.

Miniplop: If we want to create some dependencies between the steps, the good way of testing that is to do an end to end test and not a unit test

packages/slice-machine/scripts/migration/versions/0.3.0/index.ts
Miniplop: I didi this to avoid writing big condition but as you prefer I can go back on it :)

Miniplop: I change a little bit the code here, waiting for your feedback

Heyrows (approved):

PR #380, opened 2022-02-24
Update page event to add a name to categorized the event 

Miniplop:
Update page event to add a name to categorized the event


packages/slice-machine/src/tracker.ts
```diff
@@ -63,7 +64,7 @@ export class SMTracker {
 
     return this.#client
       .then((client): void => {
-        void client.page(attributes);
+        void client.page(EventType.PageView, attributes);
```
Heyrows, line 67: shouldn't it also track the page the was viewed rather than just saying "SliceMachine", or maybe it is something sent by the SDK automatically ?

Heyrows (approved):

packages/slice-machine/src/tracker.ts
Miniplop: it is :)

PR #385, opened 2022-02-28
Fix mock override

Miniplop:
We fix two problems : 
- The mock configuration can be override 
- The slices are saved in the filesystem before pushing the customTypes (not required anymore as we save the slice in the FS on creation)
-


packages/slice-machine/server/src/api/custom-types/push.ts
```diff
@@ -122,13 +122,6 @@ export default async function handler(req: RequestWithEnv): Promise<ApiResult> {
     const slice = localSlices[sliceKey];
     if (slice) {
       try {
-        console.log("[custom-types/push] Saving slice", sliceKey);
-        await saveSlice(state.env, {
```
MarcMcIntosh, line 126: nice, this should deal with the noise complaints :)

MarcMcIntosh (approved):

PR #387, opened 2022-03-01
Add the getStoriesPatch function with the deprecated comments

Miniplop:
Put the getStoriesPath function back with a deprecated comments as this function will be removed.

### DIFF
--- a/packages/slice-machine/helpers/storybook.js
+++ b/packages/slice-machine/helpers/storybook.js
@@ -1,0 +1,16 @@
+/** Called from user project */
+const glob = require("glob");
+
+/**
+ * @deprecated This function will be removed in future releases, if you need it, copy/paste the following code inside your project.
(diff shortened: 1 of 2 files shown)


packages/slice-machine/helpers/storybook.js
```diff
@@ -0,0 +1,16 @@
+/** Called from user project */
+import glob from "glob";
```
MarcMcIntosh: use require as this will be called as a common js module :)

MarcMcIntosh (changes requested):

packages/slice-machine/helpers/storybook.js
Miniplop: mmm, so I will remove the linter on this file

MarcMcIntosh (approved):

PR #393, opened 2022-03-03
Fix vue-slicezone package

Miniplop:


packages/vue-slicezone/utils/index.js
```diff
@@ -1,10 +0,0 @@
-const camelizeRE = /-(\w)/g;
-export const pascalize = (str) => {
-  if (!str) {
-    return "";
-  }
-  str = str.replace(/_/g, "-").replace(camelizeRE, (_, c) => {
-    return c ? c.toUpperCase() : "";
-  });
-  return str[0].toUpperCase() + str.slice(1);
-};
```
lihbr, line 10: Good catch, the package is since using `pascalize()` from `sm-commons`!

lihbr (approved):
LGTM, I think `./theme.js` can be deleted also: https://diff.intrinsic.com/vue-slicezone/0.1.0/0.1.1#file-theme.js

packages/vue-slicezone/utils/index.js
Miniplop: hehe 😎

PR #394, opened 2022-03-04
Fix staging init

Miniplop:


packages/core/src/utils/const.ts
```diff
@@ -20,5 +20,6 @@ export const MANIFEST_FILE_NAME = "sm.json";
 export const DEFAULT_SERVER_PORT = 5555;
 export const DEFAULT_BASE = "https://prismic.io";
 export const USER_SERVICE_BASE = "https://user.internal-prismic.io";
+export const USER_SERVICE_STAGING_BASE = "https://user.wroom.io";
```
Heyrows, line 23: no internal-wroom.io ?

Heyrows (approved):

packages/core/src/utils/const.ts
Miniplop: nop :)

PR #401, opened 2022-03-08
Update custom type list design 

Miniplop:
image


packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header.tsx
```diff
@@ -77,7 +76,7 @@ const CustomTypeHeader = ({
     <Header
       MainBreadcrumb={
         <Fragment>
-          <FiLayout /> <Text ml={2}>Custom Types</Text>
+          <MdSpaceDashboard /> <Text ml={2}>Custom Types</Text>
```
Heyrows, line 79: I prefer when the two icons are the same,  but on the figma, the icons seems to be different between navigation and this page. (from what I saw on the ticket)

packages/slice-machine/pages/index.tsx
```diff
@@ -1,122 +1,63 @@
...
-          backgroundSize: "contain",
-          backgroundPosition: "50%",
-          backgroundRepeat: "no-repeat",
-          // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-          backgroundImage: "url(" + `${preview?.url}` + ")",
-        }}
-      />
+      <thead>
+        <tr>
+          <Box as={"th"} sx={{ width: "35%" }}>
```
Heyrows: definitely like that, the code looks much simpler.

packages/slice-machine/pages/index.tsx
```diff
@@ -1,122 +1,63 @@
...
+          <Box as={"th"} sx={{ width: "15%" }}>
+            Type
+          </Box>
+        </tr>
+      </thead>
+      <tbody>
+        {customTypes.map((customType) => (
+          <Link passHref href={`/cts/${customType.id}`} key={customType.id}>
+            <tr key={customType.id}>
+              <Box as={"td"} style={{ width: "35%" }}>
```
Heyrows: OPT: extract the box width so that they are being kept between the header and the value from the table.

packages/slice-machine/pages/index.tsx
```diff
@@ -152,7 +93,7 @@ const CustomTypes: React.FunctionComponent = () => {
         }
         MainBreadcrumb={
           <Fragment>
-            <FiLayout /> <Text ml={2}>Custom Types</Text>
+            <MdSpaceDashboard /> <Text ml={2}>Custom Types</Text>
```
Heyrows, line 100: thsi seems to be duplicated with CTBuilder / layout /...

packages/slice-machine/src/theme.ts
```diff
@@ -643,17 +643,38 @@ const AppTheme = (): Theme =>
...
-        "th,td": {
+        "td, th": {
           textAlign: "left",
-          py: "4px",
-          pr: "4px",
-          pl: 0,
           borderColor: "muted",
           borderBottomStyle: "solid",
         },
+        td: {
```
Heyrows, line 653: wouldn't that affect the slice list since they're both grid/table ?

packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header.tsx
Miniplop: The figma preview on the ticket is not up to date, all the icon for custom type list page, navigation and custom type detail page are the same.

packages/slice-machine/pages/index.tsx
Miniplop: Nop it's not exactly the same, in the CTBuilder, there is the same starting name but ont the CTBuilder / layout /... there is the name of the custom types after it. 

I think it's little bit overkill to create a specific component for 2 lines of code. I mean if we had to create a component every time we used 2 components together we will have many components and that will lead on extra overhead going trought all the component tree. What do you want me to do ?

packages/slice-machine/pages/index.tsx
Miniplop: Can be a good idea, I will make the change

packages/slice-machine/pages/index.tsx
Miniplop: cool

packages/slice-machine/src/theme.ts
Miniplop: The slice list is not using the table HTML tags

packages/slice-machine/pages/index.tsx
Miniplop: Done

Heyrows (approved):

PR #403, opened 2022-03-09
generate mocks through the mocks library while supporting legacy mock…

arnaudlewis:
**What has been done:**
- We rely on `@prismicio/mocks` to generate mocks for slices and custom types
- We have a runtime validator for the mock config that will then be translated to the config expected by the library
=> This will disappear once we revamp the UI to build a mock config or just remove the mock config all together

**SM bugs fixed in the process:**
- Timestamp mock format was invalid.
- Mock config would not be saved if in a group or slice repeatable zone (inside the mock config modal)
- Mock config would not be used to generate content inside a slice repeatable zone or a group
- number/range are now properly generated by respecting min max and step from the config and constrained by the Model.
- 
For the sake of the public release, right now the lib generates already API v2 response but in the future it'll output editor data format instead.

# !!! WARNING !!!
`@prismicio/mocks` / `@prismicio/types-internal` / `@prismicio/api-renderer` are getting regular updates and should be upgraded one more time before the merge !

image
image


packages/slice-machine/lib/env/versions.ts
```diff
@@ -13,16 +13,17 @@ export async function getPackageChangelog(
   dependencyCwd: string
 ): Promise<PackageChangelog> {
   const pkg = NodeUtils.retrieveJsonPackage(dependencyCwd);
-  if (!pkg.exists || !pkg.content?.name)
+  const pkgContent = pkg.content as { name?: string; version: string };
```
Heyrows, line 16: between that and resolveAliases, lots of casting, I don't really see why. These piece of codes never gave any linter warnings or anything on our branches.

packages/slice-machine/lib/mock/LegacyMockConfig.ts
```diff
@@ -0,0 +1,482 @@
...
+  Text: {
+    content: null,
+  },
+  UID: null,
+  Timestamp: {
+    content: null,
+  },
+};
+
+export const NonEmptyText = new t.Type<
```
Heyrows, line 58: Are those definitions coming from somewhere or did you recreate everything based on the UI ?

packages/slice-machine/lib/mock/CustomType.ts
```diff
@@ -1,88 +1,54 @@
...
-      const groupFields = groupHandler(value, groupMockConfig);
-      customTypeMock.data[key] = groupFields;
-    });
-
-    if (sliceZone) {
-      sliceZoneHandler();
-    }
-  }
-  return customTypeMock;
+  const prismicModel = CustomTypes.fromSM(model);
```
Heyrows, line 44: interesting that you would have to change the model back to Api model.

Heyrows: totally understandable, but I guess some kind of layer in between SM front / back could be good.

ex:
- SM front send a model through a call (`CustomTypeSM`).
- it goes through a sort of middleware that turns it into `CustomTypePrismic`
- the handler on the server just receives it and use it without ever knowing there is a second type.

packages/slice-machine/lib/mock/CustomType.ts
```diff
@@ -1,88 +1,54 @@
...
+  const documentMockConfig = buildDocumentMockConfig(
+    prismicModel,
+    legacyMockConfig
```
Heyrows, line 47: OPT: I don't think the name `legacyMockConfig` is appropriate as we still generate those datas

packages/slice-machine/lib/mock/Slice.ts
```diff
@@ -1,53 +1,80 @@
...
+    model.primary &&
+    buildFieldsMockConfig(
+      model.primary,
+      legacyVariationMockConfig?.["primary"]
+    );
+  const itemFields =
+    model.items &&
+    buildFieldsMockConfig(model.items, legacyVariationMockConfig?.["items"]);
+  return {
+    nbItemsBlocks: 1,
```
Heyrows, line 29: interesting, is this the number of times a block is repeated ?

packages/slice-machine/lib/mock/Slice.ts
```diff
@@ -1,53 +1,80 @@
...
+  legacyMockConfig: Record<string, Record<string, Record<string, unknown>>> // not sure about this one.
+): unknown[] {
+  const model = Slices.fromSM(smModel);
+  const sliceMockConfig = buildSliceMockConfig(model, legacyMockConfig);
+  const sliceModel: SharedSlice = {
+    ...model,
+    variations: model.variations.map((v) => {
+      return {
+        ...v,
+        imageUrl: "",
```
Heyrows, line 72: what if variations already have a screenshot uploaded ?

packages/slice-machine/lib/mock/CustomType.ts
```diff
@@ -1,88 +1,54 @@
...
-      customTypeMock.data[key] = groupFields;
-    });
-
-    if (sliceZone) {
-      sliceZoneHandler();
-    }
-  }
-  return customTypeMock;
+  const prismicModel = CustomTypes.fromSM(model);
+  const documentMockConfig = buildDocumentMockConfig(
```
Heyrows, line 45: I'm not very familiar with mocks tbh, does this helper also generate the mocks for the slice that are inside the custom type ?

What I mean is, if there are slices attached to the custom type, will their own custom config be ignore or would it be included ?

packages/slice-machine/lib/mock/Slice.ts
```diff
@@ -1,53 +1,80 @@
...
+    primaryFields,
+    itemFields,
+  };
+}
 
-    const repeat = variation.items;
-    if (!repeat || (repeat && repeat.length === 0)) {
+export function buildSliceMockConfig(
+  model: SharedSlice,
+  legacyMockConfig: Partial<
```
Heyrows, line 37: what a weird type ^^
Can they be extracted all the legacyMockConfig type for more clarity ?

packages/slice-machine/lib/mock/Slice.ts
```diff
@@ -1,53 +1,80 @@
...
-    for (let i = 0; i < Math.floor(Math.random() * 6) + 2; i++) {
-      items.push(handler(repeat, mockConfig ? mockConfig.items || {} : {}));
-    }
-
-    mock.items = items;
-    return mock;
+  return model.variations.map((v) => {
+    return {
+      variation: v.id,
+      variations: variationConfigs,
```
Heyrows, line 56: Should it be
```suggestion
      variations: variationConfigs[v.id],
```
So that each variation is associated to it's personal mock config and the full object is no copied multiple times ?

Also, do you know why is the key called variations ?

packages/slice-machine/lib/mock/Slice.ts
```diff
@@ -1,53 +1,80 @@
...
+    };
   });
+}
 
-  return variations;
+export default function MockSlice(
+  smModel: SliceSM,
+  legacyMockConfig: Record<string, Record<string, Record<string, unknown>>> // not sure about this one.
```
Heyrows, line 63: How can we make sure ?
Testing ?

packages/slice-machine/lib/models/common/MockConfig.ts
```diff
@@ -100,7 +100,7 @@ export const SliceMockConfig = {
 
 export interface CustomTypeMockConfig {
```
Heyrows, line 101: Do we still need this or is it a legacy type also ?

packages/slice-machine/lib/models/common/widgets/Date/Mock/Form.js
```diff
@@ -1,20 +1,19 @@
...
-import { initialValues } from ".";
+import { DefaultConfig } from "@lib/mock/LegacyMockConfig";
 
 const Form = (props) => {
   return (
     <TimeConfigForm
       {...props}
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-      initialMockValues={initialValues}
+      initialMockValues={DefaultConfig.Date}
```
Heyrows, line 9: are we sure that we need this one ?
I mean we set it a couple of lines below so it feels weird to set it twice.
+ it doesn't seem that we set it twice on the other widgets.

packages/slice-machine/lib/models/common/widgets/Group/ListItem/index.js
```diff
@@ -59,10 +59,15 @@ const CustomListItem = ({
     setSelectMode(false);
   };
 
+  /* eslint-disable */
```
Heyrows, line 61: ignoring the code block ?

We can ignore specific errors if you want but the whole file is too much, we wouldn't know if we introduce new errors

Heyrows: Especially if you introduced new errors, you wouldn't know.

packages/slice-machine/lib/models/common/widgets/Group/ListItem/index.js
```diff
@@ -59,10 +59,15 @@ const CustomListItem = ({
...
   const getFieldMockConfig = ({ apiId }) => {
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-argument
-    return CustomTypeMockConfig.getFieldMockConfig(mockConfig, apiId);
+    return CustomTypeMockConfig.getGroupFieldMockConfig(
```
Heyrows, line 63: unless i'm wrong, that function doesn't exists, this should instant fail.

packages/slice-machine/lib/models/common/widgets/StructuredText/Mock/Form.js
```diff
@@ -23,10 +25,12 @@ const HandlePatternTypes = ({
 }) => {
   const { theme } = useThemeUI();
 
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
```
Heyrows, line 28: Normally this is not something that we do.
Even though I totally understand that it's your role to fixe those issues, it's still painful to see new errors being ignored.

packages/slice-machine/lib/models/common/widgets/Timestamp/Mock/Form.js
```diff
@@ -1,11 +1,12 @@
 import TimeConfigForm from "components/TimeConfigForm";
-
-import { initialValues } from ".";
+import { DefaultConfig } from "@lib/mock/LegacyMockConfig";
 
 const Form = (props) => {
-  return <TimeConfigForm {...props} initialMockValues={initialValues} />;
+  return (
+    <TimeConfigForm {...props} initialMockValues={DefaultConfig.Timestamp} />
```
Heyrows, line 6: it seems it is the second time set the initial value twice.
Still unsure on why we need to do that.

packages/slice-machine/package.json
```diff
@@ -38,9 +38,11 @@
   },
   "dependencies": {
     "@babel/register": "^7.11.5",
+    "@prismicio/api-renderer": "^0.1.1",
+    "@prismicio/mocks": "^0.1.1",
     "@prismicio/slice-simulator-com": "^0.2.0",
-    "@prismicio/types-internal": "^0.2.7",
-    "@faker-js/faker": "^6.0.0",
```
Heyrows, line 40: hallelujah !

packages/slice-machine/package.json
```diff
@@ -38,9 +38,11 @@
...
+    "@prismicio/types-internal": "^0.2.8",
+    "@reach/menu-button": "^0.16.2",
```
Heyrows: Weird it says you added this dependency, it's suppose to be already there.

packages/slice-machine/server/src/api/custom-types/save.ts
```diff
@@ -33,6 +33,6 @@ export default async function handler(req: { body: SaveCustomTypeBody }) {
     CustomTypeMockConfig.getCustomTypeMockConfig(updatedMockConfig, model.id)
   );
   // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
-  Files.write(mockPath, mocked);
+  Files.write(mockPath, mocked as object);
```
Heyrows, line 36: weird that we would need to cast again here.

packages/slice-machine/tests/project/customtypes/about/index.json
```diff
@@ -23,6 +23,38 @@
         },
         "type": "Boolean"
       },
+      "myGroup": {
```
Heyrows: We never push on the test project, but since it's going to be remove, we don't care.

packages/slice-machine/lib/mock/CustomType.ts
```diff
@@ -1,88 +1,54 @@
...
+import {
+  CustomTypes,
+  CustomTypeSM,
+} from "@slicemachine/core/build/models/CustomType";
+
+function buildDocumentMockConfig(
+  model: CustomType,
+  legacyMockConfig: CustomTypeMockConfig
+): DocumentMockConfig {
+  const widgets = flattenWidgets(model);
```
Heyrows, line 23: what does this do ?
Does it like list the widgets from both primary and sliceZone ?

Heyrows (changes requested):
Globally good.
- For a custom type, i'm still unsure about repeatable zones as we don't seem to pass the mocks config for the Slices that could be inside, i'm unsure on how it is linked together.
- We need tests to verify mocks are generated the same way as they used to.
  - Not to test the library you made, but test that Slice Machine doesn't generate wrong mocks that will be used by other applications of the eco-system.

packages/slice-machine/lib/mock/Slice.ts
arnaudlewis: that how the SM team declared it and commented it and it seems to be okay so far so let's not change it I guess. It will disappear soon I hope.

packages/slice-machine/lib/mock/Slice.ts
arnaudlewis: I don't really care here, it's about having a proper model to generate mocks, the variation imageURL is useless here.

packages/slice-machine/package.json
arnaudlewis: idk

packages/slice-machine/lib/models/common/widgets/Timestamp/Mock/Form.js
arnaudlewis: I don't get your message, can you reexplain pls? :)

packages/slice-machine/lib/models/common/widgets/Group/ListItem/index.js
arnaudlewis: it does

arnaudlewis: I created it myself, that's what is supposed to fix the group issue.

arnaudlewis: but you're right I don't see it anymore that's weird!

packages/slice-machine/lib/models/common/widgets/Group/ListItem/index.js
arnaudlewis: yeah but since the linter in combined with the formater, I had issue to use eslint-ignore. Build issue again

packages/slice-machine/lib/models/common/widgets/Date/Mock/Form.js
arnaudlewis: this is how it's build and I didn't intend to change how the form behave so I kept the exact same setup.

packages/slice-machine/lib/models/common/MockConfig.ts
arnaudlewis: it's still used in some places to manipulate the generated legacy mock config so for now it can remain there.

packages/slice-machine/lib/mock/Slice.ts
arnaudlewis: no you can't, the lib expect a slice to provide a list of config not a single one, it must match the type of a slice not a variation.
The `variation` param before is an optional one to specify which one you want to generate mocks for eventually

packages/slice-machine/lib/mock/Slice.ts
arnaudlewis: what do you mean? create a type alias for just this line?

packages/slice-machine/lib/mock/Slice.ts
arnaudlewis: yes

packages/slice-machine/lib/mock/CustomType.ts
arnaudlewis: I called it legacy because we already planned to kill it so it's more or less localised and named as a soon-to-be-killed thing ;)

packages/slice-machine/lib/mock/CustomType.ts
arnaudlewis: I kept the same behavior, right now you can generate either mocks for shared slices or for custom type but the slices inside the custom type don't have mocks, it's an empty array that is returned.
But, The lib can generate mocks for the slice inside the custom type as long as you also provide shared slice models along with the one from the custom type.
But there is the question wether or not we want to generate or just concatenate mocks already generated for each slice.
Anyway, the solution was to keep what we have today, meaning an empty array. And ultimately we don't use custom types mocks.

packages/slice-machine/lib/mock/CustomType.ts
arnaudlewis: Not at all, CustomTypeSM a used everywhere in SM, it should be converted back to the prismic model only if writing it on FS and sending it to the custom types API that's it, the server has SM types everywhere too.
Here we convert it simply because the mocks lib expects a prismic model.

packages/slice-machine/lib/mock/CustomType.ts
arnaudlewis: it flatten the widgets meaning that it returns an array of widgets without the `tabs` anymore. You can look in types-internal

packages/slice-machine/lib/env/versions.ts
arnaudlewis: probably a type issue I had, I can check if that's gone

packages/slice-machine/tests/project/customtypes/about/index.json
arnaudlewis: ok

packages/slice-machine/server/src/api/custom-types/save.ts
arnaudlewis: not really, generated mocks are of type unknown.

packages/slice-machine/lib/mock/LegacyMockConfig.ts
arnaudlewis: That's my definition based on the code more than the UI, you had no clear contract for mock config so far.

Heyrows: cool, thanks for that.
Sorry you had to be the one doing it.

packages/slice-machine/lib/mock/Slice.ts
Heyrows: If we don't care and the base model can have an image urls it means that there is a problem with the types no ?
What i mean is, if the function using Variations and the type were the same, you wouldn't have to do that no ?

packages/slice-machine/lib/mock/CustomType.ts
Heyrows: Yeah i was thinking about concat them too also.
But i guess it should be done here but just when we need it for whatever apps we need it.

packages/slice-machine/lib/mock/Slice.ts
Heyrows: `the lib expect a slice to provide a list of config`, but each variation can have it's own config, so why would you need the informations for the other variations on a specific variation ?

packages/slice-machine/lib/models/common/MockConfig.ts
Heyrows: That's why I asked in a comment above about typing the legacy mocks

Slice Legacy Mocks,
Custom type legacy Mocks,
....

I just don't like that there is multiple old legacy code a bit scattered that we might not know about.
I feel like on that part, the work is half-done because we might have old code with old types still running here and there without us knowing about it.

packages/slice-machine/lib/models/common/widgets/Group/ListItem/index.js
Heyrows: Yeah, I checked multiple times to see if I was wrong haha

packages/slice-machine/lib/models/common/widgets/Timestamp/Mock/Form.js
Heyrows: It is about having a props `initialMockValue` ➕  the `Form.initialValue = DefaultConfig.....`

packages/slice-machine/server/src/api/custom-types/save.ts
Heyrows: That's what I don't get.
In the `mock` function above, it call the mock library, so the types should be known.

packages/slice-machine/lib/mock/CustomType.ts
Heyrows: I do understand that it was just for this case about using external libs expecting this type.
I just think those cases might happen often (pushing data, writing other files, ...).

So i'm wondering if this is not better to have the server handle data compatible with other APIs by default and only use the SM models for rendering purpose (which is why they were created).

packages/slice-machine/lib/env/versions.ts
Heyrows: thanks 😄

packages/slice-machine/lib/mock/CustomType.ts
Heyrows: I understand, soon-to-be-killed it might require a few month ^^'

packages/slice-machine/lib/mock/Slice.ts
Heyrows: Nice, that could be awesome if later we want to allow the user to add that to the mock config.

packages/slice-machine/lib/mock/Slice.ts
Heyrows: I mean typing all the different LegacyMocks so that they can be re-used (either in the UI or anywhere else).

packages/slice-machine/lib/models/common/widgets/Date/Mock/Form.js
Heyrows: But don't you think it is weird ?

packages/slice-machine/lib/models/common/widgets/Group/ListItem/index.js
Heyrows: I understand that it could have been frustrating.
But it could be done like that 
image
and that way, we don't suppress futur errors.

packages/slice-machine/lib/mock/CustomType.ts
Heyrows: I see, so it's all about primary fields, it does not really represent the model in the end but I get that it's easier to read.

One question tho, if one 2 tabs you've got a widget with the same API ID, would that still work ?

packages/slice-machine/lib/mock/CustomType.ts
arnaudlewis: I don't think that client/server is good criteria to decide where we should use a given model.
Slicemachine UI is meant to have its own model with its own constraints and everything so I'd rather keep it that way and only jump back to prismic model when contacting prismic external apps, it's more straightforward and meaningful in that case. A model should correspond to a usage, not based on where it's used

packages/slice-machine/lib/mock/CustomType.ts
arnaudlewis: I hope not but it's on SM team hands to define that and it's still legacy, it's not meant to evolve in its current state.

packages/slice-machine/lib/mock/Slice.ts
arnaudlewis: yeah they could share a type but the thing is, SM and prismic don't share the exact same type so it's the same as having the fallback, one way or another we need a value here, having a URL we don't use over an empty string is not relevant here.

packages/slice-machine/lib/mock/CustomType.ts
arnaudlewis: It was not supported before and it's an enhancement that we might consider at some point we I don't think that should be the focus right now considering that this feature is basically useless since we don't have a similar simulator for custom types.
Also it'll be a new behavior to introduce on SM so it'll be probably be up to the SM team to evolve those in the future depending on the needs.

packages/slice-machine/lib/mock/Slice.ts
arnaudlewis: the lib expect a slice definition not a variation one.

packages/slice-machine/lib/mock/Slice.ts
arnaudlewis: They are typed already partly done thanks to the new validators but I don't see the benefit of doing more here.
It wasn't typed before and now we have a contract for each field to support the library.
I don't get why we should create types for some code that is meant to disappear?
Especially in the context of this PR that has a goal of getting rid of this old thing.
If you end up deciding that we'll keep the current mock config, then definitely you'll need to improve the typing but that's not something we should focus right now again.

packages/slice-machine/lib/models/common/MockConfig.ts
arnaudlewis: see previous answers.

packages/slice-machine/lib/models/common/widgets/Date/Mock/Form.js
arnaudlewis: It might but I can't fix all SM code in my PR, there are a bunch of clunky places and I didn't intend to fix them all

packages/slice-machine/lib/models/common/widgets/Group/ListItem/index.js
arnaudlewis: I'll look if it was lost in a merge

packages/slice-machine/lib/models/common/widgets/StructuredText/Mock/Form.js
arnaudlewis: code changes and still rely on old code that is not typed, it has to happen unless we get rid of JS

packages/slice-machine/server/src/api/custom-types/save.ts
arnaudlewis: yes but then it's converted to API format to support your current behavior and the API doesn't have return types

packages/slice-machine/lib/mock/CustomType.ts
arnaudlewis: it's how custom types works, each api ID is unique in the whole custom type, tabs are here for wroom and then scrapped in the api so no problem here, always been the case.

packages/slice-machine/tests/lib/mock/Slice.test.js
```diff
@@ -1,164 +0,0 @@
...
-    __edModule: true,
-    LoremIpsum: jest.fn().mockImplementation(() => {
-      return {
-        generateParagraphs: jest.fn().mockReturnValue("Some text."),
-      };
-    }),
-  };
-});
-
-describe("MockSlice", () => {
```
MarcMcIntosh, line 17: Removed because?

packages/slice-machine/lib/mock/CustomType.ts
Heyrows: Alright, I agree then that the whole UI as a specific model.
In that optic I suggest that all the functions related to transforming a model into the other one and vice-versa should be located in the IO module.

packages/slice-machine/lib/mock/CustomType.ts
Heyrows: ok cool :)

packages/slice-machine/lib/env/versions.ts
arnaudlewis: the type `PackageJson` doesn't contain `content` which is returned as `any` hence the casting for now.

packages/slice-machine/lib/mock/CustomType.ts
arnaudlewis: I disagree on the location, it's still model related to go back and forth with prismic, IO is only about input output so writing and reading the filesystem in that case. And that conversion is specific to SM hence the `model` folder of SM close to the SM model.

MarcMcIntosh:
CI is failing as some external packages are not available, that will ned to be addressed.

I'm not approving this with out any tests.

If you can't/won't add them then write each bug fix / feature down in gerkin format (feature...given....when...then) and we'll make tickets and add them to the sprint and decided to ether implement them independently or finish the work on this PR.

packages/slice-machine/tests/lib/mock/Slice.test.js
MarcMcIntosh: Forgot to submit that a while ago

packages/slice-machine/tests/lib/mock/Slice.test.js
```diff
@@ -1,5 +1,4 @@
 import MockSlice from "../../../lib/mock/Slice";
```
arnaudlewis, line 5: Why this file `Slice.test.js` is even back here? The library generate the mocks, it makes no sense to test again the output of the lib... seems like overtesting here. it’s like testing that React actually renders components.

MarcMcIntosh: > Why this file Slice.test.js is even back here

Chances are it was originally added because something broke before, and we didn't want it to happen again.

>The library generate the mocks, it makes no sense to test again the output of the lib

Is the lib likely to change?

> seems like over-testing here

Could be, If there are practices for to prevent over testing, but we're playing catchup with what's already there.

> it’s like testing that React actually renders components

oddly enough react comes with a just utils package for that https://reactjs.org/docs/test-utils.html, it has some uses.

sintinilorenzo: there's logic done from slice-machine itself. The purpose of the test is not to check the library is generating the right data, it's to test that we are creating the right objects and passing the right arguments to the library functions. There's a bunch of object manipulation, object deconstruction (`...`), array methods (`.reduce`, `.map`) and we should test them :) 
image

chengoconnell (approved):
Nice testing!

MarcMcIntosh (approved):

packages/slice-machine/tests/lib/mock/Slice.test.js
arnaudlewis: it’s true but since everything is typed, I consider that the output is correctly built, that’s how I would consider it here. But you’re right there is still a bit of logic here. And the Slice.test.ts that we had before was about testing what was generated and I consider that part useless, it should be part of the lib. 
But testing the logic that builds the config for the lib makes sense. Thanks for that explanation @sintinilorenzo

PR #404, opened 2022-03-09
Kje/slice empty state

Miniplop:
Slice page screenshot 

image

Custom type page screenshot 

image

### DIFF
--- a/packages/slice-machine/lib/builders/SliceBuilder/Header/index.tsx
+++ b/packages/slice-machine/lib/builders/SliceBuilder/Header/index.tsx
@@ -1,7 +1,6 @@
 import React, { useState } from "react";
 import { Box, Flex, Text, Link as ThemeLinK } from "theme-ui";
 import MetaData from "./MetaData";
-import { FiLayers } from "react-icons/fi";
 import VariationModal from "./VariationModal";
 import Link from "next/link";
 import { useRouter } from "next/router";
@@ -9,6 +8,7 @@ import * as Links from "../links";
 import VariationPopover from "./VariationsPopover";
 import SaveButton from "./SaveButton";
 import type { ContextProps } from "@src/models/slice/context";
+import { MdHorizontalSplit } from "react-icons/md";
 
 const Header: React.FC<{
   Model: ContextProps["Model"];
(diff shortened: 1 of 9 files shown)


packages/slice-machine/lib/builders/SliceBuilder/Header/index.tsx
```diff
@@ -65,7 +65,7 @@ const Header: React.FC<{
                 <Link href="/slices" passHref>
                   <ThemeLinK variant="invisible">
                     <Flex sx={{ alignItems: "center" }}>
-                      <FiLayers /> <Text ml={2}>Slices</Text>
+                      <MdHorizontalSplit /> <Text ml={2}>Slices</Text>
```
Heyrows, line 68: a bit unsure why that would change when this is a PR about empty state in the lists.

packages/slice-machine/package.json
```diff
@@ -88,7 +88,6 @@
     "semver": "^7.3.2",
     "serve-static": "^1.14.1",
     "slash": "3.0.0",
-    "sm-commons": "^0.0.22",
```
Heyrows, line 91: guess that's why the package-lock has been modifed.

packages/slice-machine/pages/index.tsx
```diff
@@ -158,28 +158,35 @@ const CustomTypes: React.FunctionComponent = () => {
...
-          }
-        />
+        <Flex
+          sx={{
+            flex: 1,
+            justifyContent: "center",
+            alignItems: "center",
+          }}
+        >
+          <EmptyState
```
Heyrows, line 113: If the empty state is shared between Cts and Slices, the flex tag around should go inside the empty state.

Heyrows (changes requested):

packages/slice-machine/package.json
Miniplop: yes :)

packages/slice-machine/pages/index.tsx
Miniplop: I don't know as the flex around is here for positioning the empty state inside the page itself, if we do so that mean this component will always be in full page in the middle of his parent. It's a choice to do between having a more reusable component or a more context tied component. 

What is your preference ?

packages/slice-machine/lib/builders/SliceBuilder/Header/index.tsx
Miniplop: Because there is no ticket for doing that, and the designers ask me for doing the modification.

Heyrows (approved):

packages/slice-machine/pages/index.tsx
Miniplop: We keep like that :)

PR #405, opened 2022-03-09
Remove warnings / Change some design stuff

Miniplop:
- Remove the warnings and configErrors from the serverState (not used anymore by the frontend)
- Remove the warnings page
- Remove the log in warning
- Refactor the env store
- Add loader on the slice and ct empty state


packages/slice-machine/pages/_app.tsx
```diff
@@ -62,7 +62,7 @@ function MyApp({ Component, pageProps }: AppContext & AppInitialProps) {
     }
 
     const { store, persistor } = configureStore({
-      environment: { env: serverState.env, warnings: [], configErrors: {} },
+      environment: serverState.env,
```
Heyrows, line 65: nice

packages/slice-machine/server/src/api/state.ts
```diff
@@ -104,7 +76,11 @@ export default async function handler(
   };
 
   return {
-    ...serverState,
+    customTypes: serverState.customTypes,
+    remoteCustomTypes: serverState.remoteCustomTypes,
+    libraries: serverState.libraries,
+    remoteSlices: serverState.remoteSlices,
+    clientError: serverState.clientError,
```
Heyrows: do we still need to send this property ?
It seems only used in the backend on the push handler.

packages/slice-machine/src/modules/environment/index.ts
```diff
@@ -80,15 +69,15 @@ export const getIsTrackingAvailable = (
...
+    store.environment.manifest.tracking === undefined ||
+    store.environment.manifest.tracking
   );
 };
 
 export const getStorybookUrl = (
   store: SliceMachineStoreType
 ): string | null => {
-  return store.environment.env.manifest.storybook || null;
+  return store.environment.manifest.storybook || null;
```
Heyrows, line 80: seems way better

packages/slice-machine/src/modules/environment/index.ts
```diff
@@ -141,9 +130,7 @@ export const environmentReducer: Reducer<
     case getType(refreshStateCreator):
       return {
         ...state,
-        env: action.payload.env,
-        warnings: action.payload.warnings,
-        configErrors: action.payload.configErrors,
+        ...action.payload.env,
```
Heyrows, line 133: are you sure we still send an object with an env property and not the end directly in the payload ?

packages/slice-machine/src/modules/environment/types.ts
```diff
@@ -1,9 +1,3 @@
 import { FrontEndEnvironment } from "@lib/models/common/Environment";
-import Warning from "@models/common/Warning";
-import { ConfigErrors } from "@models/server/ServerState";
 
-export type EnvironmentStoreType = {
-  env: FrontEndEnvironment;
-  warnings: ReadonlyArray<Warning>;
-  configErrors: ConfigErrors;
-};
+export type EnvironmentStoreType = FrontEndEnvironment;
```
Heyrows, line 3: Was it for the naming coherence or to not rename everythign that you decided to keep that ?

Heyrows (changes requested):

packages/slice-machine/pages/_app.tsx
Miniplop: cool :)

packages/slice-machine/server/src/api/state.ts
Miniplop: I think we can can also remove it

packages/slice-machine/src/modules/environment/index.ts
Miniplop: coool 👍

packages/slice-machine/src/modules/environment/types.ts
Miniplop: Just for keeping the name convention for store type

packages/slice-machine/src/modules/environment/index.ts
Miniplop: Yes I'm sure look a the the refresh state action :),

Heyrows (approved):

PR #407, opened 2022-03-11
Improve CI

Miniplop:
- Add a prettier check in the CI + pass prettier on sm ui to change some ignored errors


packages/slice-machine/package.json
```diff
@@ -32,7 +32,8 @@
     "slicemachine": "start-slicemachine --port 9999",
     "lint": "eslint --ext .ts,.tsx,.js,.jsx .",
     "lint:precommit": "eslint",
-    "audit": "better-npm-audit audit -l high -p -x 1054172",
+    "prettier:check": "prettier --check .",
```
bapmrl: Does that mean the CI will only check formatting in `packages/slice-machine`, but not in all other packages? Yet, `lint-staged` formats every package so it doesn't seem to behave the same way :)

Miniplop: You right I change that :)

packages/slice-machine/lib/models/common/widgets/Image/components.js
```diff
@@ -87,6 +87,7 @@ export const ConstraintForm = ({ prefix, required, display, error = {} }) => {
   const createSetField =
     // eslint-disable-next-line @typescript-eslint/no-unsafe-return
 
+
```
Heyrows, line 90: AHHHHHHHHHHH

Heyrows (approved):

PR #408, opened 2022-03-11
Improve the dev-server command performance

Miniplop:
Use concurrently package to build while reloading the server


packages/slice-machine/package.json
```diff
@@ -19,10 +19,8 @@
     "clean-build": "rimraf ./build/client",
     "postinstall": "node postinstall",
     "prepublishOnly": "npm run bundle",
-    "serve": "node -r esm ./build/server/src/index.js",
-    "compile": "tsc --project server/tsconfig.json && node build/server/src/index.js",
-    "dev-server": "cross-env PORT=9999 CWD=./tests/project ENV=SM nodemon -e ts,js --watch server --watch lib --ignore build/ --exec \"npm run compile\"",
-    "prod-server": "cross-env PORT=9999 CWD=./tests/project nodemon -e ts,js --watch server --watch lib --ignore build/ --exec \"npm run compile\"",
+    "dev-server": "cross-env PORT=9999 CWD=./tests/project ENV=SM concurrently \"tsc -w --project server/tsconfig.json\" \"nodemon -e ts,js --watch server --watch lib --ignore build/ build/server/src/index.js\"",
```
Heyrows, line 22: as there is no ticket associated, could you explain quickly
- the need
- what are the argument concurrently takes

Miniplop: concurrently take command you want to run in parallel, here : 
- tsc to compile the code in the the build folder
- nodemon to reload the sever when the build is changing

The goal of this is to reduce the waiting time when doing modification in the server or the lib folder. Before this it took ~ 2/4s now it's ~ 1/2s. 

The best solution is to use ts-node, because this package compile and run typescript on the fly but it require extra config for working with the module_alias package.

Heyrows: hmmmm

Heyrows (approved):

PR #414, opened 2022-03-16
[WIP] Mm/reponame prompt

MarcMcIntosh:
Quick fix for the init script where the repo name is printed as undefined


Miniplop (approved):
Tested, that's good 👍

PR #416, opened 2022-03-16
Clean component model

Miniplop:
Here the modifications in the component I did :

- `sliceName` key was removed (because already in model)
- `fileName` key was removed 
- `isDirectory` key was removed
- all the informations remaining in the `infos`object were moved to the component root object
- the `meta` object was removed
- `mock` object is now in the root object
- `nameConflict` key was removed because it wasn't used anymore


packages/core/src/libraries/component.ts
```diff
@@ -114,9 +100,9 @@ function getFileInfoFromPath(
   if (match) {
     const maybeFileComponent = findComponentFile(files, componentName);
     if (maybeFileComponent) {
-      return { ...splitExtension(maybeFileComponent), isDirectory: true };
+      return { ...splitExtension(maybeFileComponent) };
```
Heyrows, line 103:
```suggestion
      return splitExtension(maybeFileComponent);
```

packages/core/src/models/Library.ts
```diff
@@ -6,20 +6,12 @@ import { VariationMock } from "./Variation";
 import { SliceAsObject } from "./Slice";
 
 export interface ComponentInfo {
```
Heyrows, line 8: look a lot like the component model, do you think we could link them ?

packages/core/src/models/Library.ts
```diff
@@ -33,19 +25,17 @@ export const ComponentInfo = {
   },
 };
 
-export interface ComponentMetadata {
-  id: string;
-  name?: string;
-  description?: string;
-}
-
 export interface Component {
```
Heyrows: i can smell the conflicts with Marc's PR already.

packages/slice-machine/lib/builders/SliceBuilder/Header/index.tsx
```diff
@@ -130,11 +128,6 @@ const Header: React.FC<{
           />
         </Flex>
       </Box>
-      <MetaData
```
Heyrows, line 133: I didn't saw anything to open that modal, i'm guessing it was unused, do you confirm ?

packages/slice-machine/lib/models/common/ComponentUI.ts
```diff
@@ -67,7 +67,7 @@ export const ComponentUI = {
     return {
       ...component,
```
Heyrows, line 68: To lighten the model even more, I think we should find a way to compute screenshot paths here so that we just build Urls straight away, just for the front.

I don't think the screenshotPaths are used anywhere else.
But i'm guessing that would be too much for now and can be done later :)

packages/core/src/models/Library.ts
```diff
@@ -33,19 +25,17 @@ export const ComponentInfo = {
...
   from: string;
   href: string;
```
Heyrows, line 30: how about deleting that one ? I'm not sure it's used anymore.

Heyrows:
Also, I think PathToSlice property could be rebuilt if it's only used in backend. (rebuilt from library name, sliceName and all)

packages/slice-machine/lib/builders/SliceBuilder/Header/index.tsx
Miniplop: Yes exactly, it's an abandoned feature I think :)

packages/core/src/libraries/component.ts
Miniplop: I think this is here for having a pure function

packages/core/src/models/Library.ts
Miniplop: Yes also I think the Arnaud PR will introduce a lot of change in the core so we will have to handle some conflicts between PRs :/

packages/slice-machine/lib/models/common/ComponentUI.ts
Miniplop: Nop there are not used elsewhere but as this two parts of code are separated for now it will be hard to do that without changing a pretty big part in the core. 

As there are already 2 PR on the core open, I think we can do that after stabilizing the core

packages/core/src/models/Library.ts
Miniplop: Done

packages/core/src/models/Library.ts
Miniplop: Linked

Heyrows (approved):

PR #419, opened 2022-03-18
Refactor core models

arnaudlewis:
## Goals ##

Today, SM has its own custom models to represent a custom type or a Shared Slice.
On the other side, we have a library `prismic-internal-types` that is the shared prismic model to represent also the custom type. This model is now shared across different apps like the editor, the prismic API and the Custom Type API.

So far, the SM representation was enough to deal with the CT api since only the structure of the data was relevant but not the types themselves.
Now that we are integrating the editor into SM, it becomes necessary to have a full compatibility between these two.

## Benefits ##
- No more custom models on slicemachine meaning that if the model is supposed to evolve elsewhere, SM will be able to retrieve these evolution and integrate them.
- No more Complex AsArray / AsObject everywhere (without considering a few other in between models)
=> We'll have the CT model coming from the lib + one and only one specific model to SM based on the widgets from the lib.
- All the widgets types are now coming from the lib so all types will have runtime validators and no more any config or any fields anywhere.

## What has been done ##
- Remove AsArray / AsObject models + in between models like `CustomTypeJsonModel`
- Spread the SM specific type everywhere and convert back to the prismic model only on IO operations (save to filesystem / read from filesystem) or at query time while retrieving data from the CT api
- Migrate all the widgets to the new types
- Tests that will verify that the structure of the new SM model hasn't change from before.

## Main Impacts ##
- Types in the `core`
- All the helpers in `library` to manipulate the model (like addWidget, removeWidget ...)
- All the related reducers
- All IO operations (save/read)
- Push operations (we only manipulate SM model until the very last moment)
- tests

Obviously all parts of the code relying on the types will also have changes but if should just be types and not behaviour or logic.


packages/slice-machine/server/src/api/custom-types/index.ts
```diff
@@ -1,32 +1,22 @@
...
-        {
-          ...CustomType.fromJsonModel(key, jsonCustomType),
-          previewUrl: Files.exists(pathToPreview)
-            ? `${env.baseUrl}/api/__preview?q=${encodeURIComponent(
-                path.join(path.dirname(p), "index.png")
-              )}&uniq=${Math.random()}`
-            : undefined,
-        },
-      ];
+      const jsonCustomType: CustomType = Files.readJson(p);
```
Miniplop: CustomType.readCustomType

packages/slice-machine/server/src/api/common/LibrariesState.ts
```diff
@@ -1,6 +1,7 @@
 import type Models from "@slicemachine/core/build/models";
 import { BackendEnvironment } from "@lib/models/common/Environment";
 import probe from "probe-image-size";
+import { Slices } from "@slicemachine/core/build/src/models/Slice";
```
Miniplop: 👀

packages/slice-machine/server/src/api/custom-types/index.ts
```diff
@@ -1,32 +1,22 @@
...
-  CustomType,
-  CustomTypeJsonModel,
-  ObjectTabs,
-} from "@lib/models/common/CustomType";
 import Files from "@lib/utils/files";
 import { CustomTypesPaths } from "@lib/models/paths";
+import {
+  CustomTypes,
+  CustomTypeSM,
+} from "@slicemachine/core/build/src/models/CustomType/index";
```
Miniplop: 👀

packages/slice-machine/server/src/api/custom-types/push.ts
```diff
@@ -10,16 +10,18 @@ import { ApiResult } from "@lib/models/server/ApiResult";
 
 import { ComponentUI } from "@lib/models/common/ComponentUI";
 import { Tab } from "@lib/models/common/CustomType/tab";
-import { CustomType } from "@lib/models/common/CustomType";
 import { RequestWithEnv } from "../http/common";
+import {
+  CustomTypes,
+  CustomTypeSM,
+} from "@slicemachine/core/build/src/models/CustomType/index";
```
Miniplop: 👀

packages/slice-machine/server/src/api/io/CustomType.ts
```diff
@@ -0,0 +1,15 @@
+import Files from "@lib/utils/files";
+import { CustomType } from "@prismicio/types-internal/lib/customtypes/CustomType";
+import {
+  CustomTypes,
+  CustomTypeSM,
+} from "@slicemachine/core/build/src/models/CustomType/index";
```
Miniplop: :eyes

packages/slice-machine/lib/models/common/widgets/withGroup.ts
```diff
@@ -1,4 +1,4 @@
-export { Boolean } from ".";
+export { Boolean } from "./";
```
Miniplop, line 1: 👀

packages/slice-machine/lib/models/server/ServerState.ts
```diff
@@ -1,8 +1,8 @@
 import ServerError from "./ServerError";
 import { FrontEndEnvironment } from "@lib/models/common/Environment";
 import { LibraryUI } from "@lib/models/common/LibraryUI";
-import { CustomType, ObjectTabs } from "@lib/models/common/CustomType";
-import { Models } from "@slicemachine/core";
+import { SliceSM } from "@slicemachine/core/build/src/models";
```
Miniplop: 👀

packages/slice-machine/server/src/api/custom-types/push.ts
```diff
@@ -72,10 +74,10 @@ export default async function handler(req: RequestWithEnv): Promise<ApiResult> {
     .customType(id as string)
     .model();
 
-  let model;
+  let model: CustomTypeSM;
   try {
     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-argument
-    model = Files.readJson(modelPath);
+    model = CustomTypes.toSM(Files.readJson(modelPath));
```
Miniplop: IO.utils.read

packages/core/src/libraries/component.ts
```diff
@@ -143,7 +140,7 @@ export function getComponentInfo(
 
   const model = fromJsonFile(path.join(slicePath, "model.json"), (payload) =>
```
Miniplop, line 141: Replace this by the IO.Slice.read no ?

packages/core/src/libraries/component.ts
```diff
@@ -158,8 +155,9 @@ export function getComponentInfo(
     return;
   }
 
-  const screenshotPaths = (model.variations || [])
-    .map((v: VariationAsObject) => {
+  const smModel = Slices.toSM(model);
```
Miniplop, line 158: weird to transform here, where the model is coming from ? Why the model as not the good format ?

Miniplop:
LGTM

Miniplop (approved):

packages/core/src/libraries/component.ts
arnaudlewis: Yeah it's because we have validation right before that fromn the FS, meaning that we have the object version.
I realise that sometimes we validate right away what comes from the FS and sometimes we cast.
If we want validation everywhere, it'll be the same problem as today. So either we throw and it won't alter the output type otherwise it can be invalid (meaning undefined or error) so we should handle this at every place where we read/write, something we don't do today because of the optimist cast

arnaudlewis: @Miniplop Either we rework how we handle error with this PR or we keep the same behavior as today and we dedicate a PR to rework the error handling alone.

PR #421, opened 2022-03-21
Update tracking plan with two separated sources

Miniplop:


package.json
```diff
@@ -14,8 +14,10 @@
     "build": "lerna run build && npm run test",
     "prettier:check": "prettier --check .",
     "pretty": "prettier --write .",
+    "bump:alpha": "lerna version prerelease --preid alpha --exact --yes",
     "publish:alpha": "lerna publish from-package --dist-tag alpha --yes --no-verify-access",
-    "publish:production": "NEXT_PUBLIC_SEGMENT_KEY=ED3O1FVsa8DteNyRmAuDyMcPoftJ6VdP lerna publish from-package",
+    "bump:production": "lerna version --conventional-graduate --exact",
+    "publish:production": "NEXT_PUBLIC_SM_UI_SEGMENT_KEY=cGjidifKefYb6EPaGaqpt8rQXkv5TD6P PUBLIC_SM_INIT_SEGMENT_KEY=ED3O1FVsa8DteNyRmAuDyMcPoftJ6VdP lerna publish from-package",
```
Heyrows, line 20: This is why I talked about having independent commands to deploy stuff.
Because the more we add env keys, the more this command is long.

Also, having a script that deploy 1 by 1 packages, would be good to have total control over the order of deployment and all.

package.json
```diff
@@ -14,8 +14,10 @@
     "build": "lerna run build && npm run test",
     "prettier:check": "prettier --check .",
     "pretty": "prettier --write .",
+    "bump:alpha": "lerna version prerelease --preid alpha --exact --yes",
     "publish:alpha": "lerna publish from-package --dist-tag alpha --yes --no-verify-access",
-    "publish:production": "NEXT_PUBLIC_SEGMENT_KEY=ED3O1FVsa8DteNyRmAuDyMcPoftJ6VdP lerna publish from-package",
+    "bump:production": "lerna version --conventional-graduate --exact",
```
Heyrows, line 19: love that, i'll modify the release process right away.

packages/init/src/index.ts
```diff
@@ -23,7 +23,8 @@ async function init() {
   const maybeRepositorySubdomain = findArgument(process.argv, "repository");
 
   Tracker.get().initialize(
-    process.env.NEXT_PUBLIC_SEGMENT_KEY || "JfTfmHaATChc4xueS7RcCBsixI71dJIJ",
+    process.env.PUBLIC_SM_INIT_SEGMENT_KEY ||
+      "JfTfmHaATChc4xueS7RcCBsixI71dJIJ",
```
Heyrows, line 27: I think the default key here is not needed anymore. As SIroc holds the default value.

.github/workflows/deploy-alpha.yml
```diff
@@ -28,9 +28,10 @@ jobs:
       - name: Version And Publish
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
-          NEXT_PUBLIC_SEGMENT_KEY: JfTfmHaATChc4xueS7RcCBsixI71dJIJ
+          NEXT_PUBLIC_SM_UI_SEGMENT_KEY: Ng5oKJHCGpSWplZ9ymB7Pu7rm0sTDeiG
```
Heyrows, line 31: are you sure the keys are not inverted ?
I'm stating this because we have much more events on SM-UI so I would think you created a new source for the init only.

Miniplop: As we sent server side events on the first source we cannot send now client event through it. So we are force to do the opposite :)

packages/init/src/index.ts
Miniplop: it's more defensive to do it that way, if Siroc configuration change. "Bretelle and ceinture" 📦

package.json
Miniplop: Cool

package.json
Miniplop: I think we can split it now

.github/workflows/deploy-alpha.yml
Heyrows: ah shit, segment is really no good on that.

Heyrows (approved):

PR #424, opened 2022-03-22
[Fix] Fix open slice simulator

Heyrows:
## Description

- Small fix to initialise the tracker on the `_app` before the page component is rendered.


packages/slice-machine/components/Simulator/index.tsx
```diff
@@ -33,7 +33,7 @@ export default function Simulator() {
 
   useEffect(() => {
     // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
Miniplop: so we can remove this right ?

packages/slice-machine/pages/_app.tsx
```diff
@@ -78,6 +81,12 @@ function MyApp({ Component, pageProps }: AppContext & AppInitialProps) {
       },
     });
 
+    Tracker.get().initialize(
```
Miniplop, line 84: nice

packages/slice-machine/components/Simulator/index.tsx
MarcMcIntosh: is this rule still needed?

packages/slice-machine/pages/_app.tsx
```diff
@@ -78,6 +81,12 @@ function MyApp({ Component, pageProps }: AppContext & AppInitialProps) {
...
+      process.env.NEXT_PUBLIC_SM_UI_SEGMENT_KEY ||
```
MarcMcIntosh, line 85: Does this env var make it into the build?

packages/sm-commons/methods/actions.js
```diff
@@ -39,39 +38,6 @@ function readConfig(p) {
   }
 }
 
-function readJsonPackage(p) {
```
MarcMcIntosh, line 42: is this function not used any-where?

MarcMcIntosh (commented):
A few questions, but it looks good

packages/sm-commons/methods/actions.js
Heyrows: no, that was an old start command apparently. It's coming from the cherry pick.

packages/slice-machine/components/Simulator/index.tsx
Heyrows: apparently not :D I removed it

Miniplop (commented):
LGTM

Miniplop (approved):

PR #428, opened 2022-03-25
Add repo into the init start script if the user use the param

Miniplop:


packages/init/__tests__/utils/tracker.test.ts
```diff
@@ -84,7 +84,7 @@ describe("InitTracker", () => {
     const smTracker = new InitTracker();
     smTracker.initialize(dumpSegmentKey);
     // Anonymous call
-    smTracker.trackInitStart();
+    smTracker.trackInitStart(undefined);
```
MarcMcIntosh, line 87: it's weird to have to pass undefined to a function

packages/init/src/index.ts
```diff
@@ -28,7 +28,7 @@ async function init() {
     isTrackingAvailable
   );
 
-  Tracker.get().trackInitStart();
+  Tracker.get().trackInitStart(maybeRepositorySubdomain);
```
MarcMcIntosh, line 31: if this needs the repo name it could be moved further down in the process?

MarcMcIntosh (commented):
Maybe consider moving the call that needs the repo name until after the prompt

MarcMcIntosh (approved):

PR #429, opened 2022-03-29
Mm/sm json validation

MarcMcIntosh:
## Types of changes

more io-ts :/

Try to prevent app from crashing when framework is missing from `sm.json`

## Checklist:


packages/core/src/node-utils/framework.ts
```diff
@@ -45,7 +45,7 @@ export function defineFramework({
 }: {
   cwd: string;
   supportedFrameworks?: Frameworks[];
-  manifest?: Manifest;
+  manifest?: Partial<Manifest>;
```
arnaudlewis: why a partial Manifest? You could just send the framework instead, it would be more accurate in that case

packages/core/src/node-utils/manifest.ts
```diff
@@ -1,10 +1,14 @@
...
 import { Manifest } from "../models/Manifest";
+import { pipe } from "fp-ts/lib/function";
+import { fold } from "fp-ts/lib/Either";
+import { formatValidationErrors } from "io-ts-reporters";
+import * as t from "io-ts";
 
 export function createManifest(cwd: string, manifest: Manifest): void {
   const manifestPath = SMConfigPath(cwd);
-  Files.write(manifestPath, JSON.stringify(manifest, null, "\t"), {
+  Files.write(manifestPath, JSON.stringify(manifest, null, 2), {
```
arnaudlewis: the Files write helper already stringify and format so it can be:
```
Files.write(manifestPath, manifest, { recursive: false })
```

packages/slice-machine/lib/env/manifest.ts
```diff
@@ -125,8 +145,12 @@ function handleManifest(cwd: string): ManifestInfo {
       Models.Manifest.decode(json),
       fold(
         // failure handler
-        () => {
-          throw new Error();
+        (errors) => {
```
arnaudlewis: You should not make some extra validation after using a codec otherwise it means that your validator is not accurate and you have two successive layers of validation. All the validation of `validate` should not be part of your codec.
should be something like:
```
getOrElseW(
  (errors) => {
    const messages = formatValidationErrors(errors);
    messages.forEach((error) => {
      console.error("[sm.json] " + error);
    });
    return undefined // or encode error
  }
)(Models.Manifest.decode(json))
```

And here you still cast + validate in the error handler, you cast a type that was rejected by the codec so are sure already that the json doesn't match the type that you cast on it.

packages/core/src/node-utils/manifest.ts
```diff
@@ -62,6 +66,29 @@ export function patchManifest(cwd: string, data: Partial<Manifest>): boolean {
   return true;
 }
 
+export function createOrUpdateManifest(
+  cwd: string,
+  data: Partial<Manifest>
+): void {
+  const maybeManifest: FileContent<Manifest> = retrieveManifest(cwd);
+
+  const updatedManifest = pipe(
```
arnaudlewis: no need for cast pipe + fold, use directly getOrElseW if you return directly the valid structure and do nothing in the Success handler.

packages/core/src/node-utils/manifest.ts
```diff
@@ -62,6 +66,29 @@ export function patchManifest(cwd: string, data: Partial<Manifest>): boolean {
   return true;
 }
 
+export function createOrUpdateManifest(
+  cwd: string,
+  data: Partial<Manifest>
+): void {
+  const maybeManifest: FileContent<Manifest> = retrieveManifest(cwd);
```
arnaudlewis: you should have a runtime validation on the retrieved manifest, not on the updated version of it since the data as parameter are typed controlled already so the validation should be as a first thing right at the moment you read from the file system.

packages/slice-machine/lib/env/manifest.ts
```diff
@@ -49,6 +54,10 @@ export function extractRepo(parsedRepo: ParseResult): string | undefined {
   }
 }
 
+function validateFramework(framework?: string): boolean {
```
arnaudlewis: It should probably part of the frameworkCodec or at least called from it so you can reuse it then in the Manifest Codec.

arnaudlewis (changes requested):
Tests are not fully reviewed, it will require an update of the code before reevaluating what we test

packages/core/src/node-utils/manifest.ts
MarcMcIntosh: k

packages/core/src/node-utils/manifest.ts
MarcMcIntosh: :/

packages/core/src/node-utils/manifest.ts
MarcMcIntosh: k

packages/core/src/node-utils/manifest.ts
MarcMcIntosh: are you asking me to re-write the "file-helpers" to use io-ts?

packages/slice-machine/lib/env/manifest.ts
MarcMcIntosh: It's not me, this is the code was already there. 

I did try to remove that validation function, but here's a few issues.

+ in some places in the code framework is needed, in others it's not.

packages/core/src/node-utils/manifest.ts
arnaudlewis: The file helper supports runtime validation already hence the readEntity, you have example in the code with io ts already

packages/slice-machine/lib/env/manifest.ts
arnaudlewis: The only concern is the model, is the framework required for sm.json? In any case that should be part of the codec. Optional or not if it's there it should be properly validated as a framework but if the code requires it in some places it means that it's required for your app to function properly

packages/core/src/node-utils/framework.ts
MarcMcIntosh: Was there for when I made the framework mandatory, I'll have to check somethings but there was a part off the init script where either create or patch where called....

Heyrows:
As discussed:

If you can remove the depedency `parse-domain` then that would be great 👍 
One less dependency is less audit problems.
And if you it's only used in validation then i think it's easily replaceable :)

packages/core/src/node-utils/manifest.ts
```diff
@@ -62,9 +64,34 @@ export function patchManifest(cwd: string, data: Partial<Manifest>): boolean {
   return true;
 }
 
+export function createOrUpdateManifest(
```
Heyrows: doesn't seem like a necessary function given what we had before.
As Arnaud said on one of his command, if we want to validate the File system when we are readying it then it seems better to keep using the same helpers as before and use io-ts on the retrieveManifest function.

packages/init/__tests__/configure-project.test.ts
```diff
@@ -87,6 +93,9 @@ describe("configure-project", () => {
   const mkdirMock = mkdir as jest.Mock;
 
   test("it should create a new manifest if it doesn't exist yet", () => {
+    // vol.fromJSON({
```
Heyrows: :)

packages/slice-machine/bin/start.js
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env node
 
-require("../build/scripts/start/index.js");
+// eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
```
Heyrows, line 3: :o why did you had issue with the linter on that, i'm curious. it's like one of those files we never touch.

packages/slice-machine/lib/env/manifest.ts
```diff
@@ -23,6 +25,7 @@ export enum ManifestState {
   MissingEndpoint = "MissingEndpoint",
   InvalidEndpoint = "InvalidEndpoint",
   InvalidJson = "InvalidJson",
+  InvalidFramework = "InvalidFramework",
 }
 
 const Messages = {
```
Heyrows, line 24: I think this can be cleaned as most of those issues will be handled in the validator

packages/slice-machine/lib/env/manifest.ts
```diff
@@ -49,63 +55,6 @@ export function extractRepo(parsedRepo: ParseResult): string | undefined {
   }
 }
 
-function validateEndpoint(endpoint: string, parsedRepo: ParseResult): boolean {
```
Heyrows, line 52: nice to see so much removal.
And on top of that since it's in the validator we won't have to worry about doing that elsewhere or on another app.

packages/slice-machine/lib/env/manifest.ts
```diff
@@ -123,13 +72,32 @@ function handleManifest(cwd: string): ManifestInfo {
...
+            state: ManifestState.InvalidJson,
+            message,
+            content: null,
+          };
         },
         // success handler
-        (manifest) => validate(manifest)
+        (manifest) => {
+          const endpoint = fromUrl(manifest.apiEndpoint);
+          const parsedRepo = parseDomain(endpoint);
```
Heyrows: can this code fails here ?
I guess not if the manifest is validated before

packages/slice-machine/lib/env/manifest.ts
```diff
@@ -123,13 +72,32 @@ function handleManifest(cwd: string): ManifestInfo {
 
     return pipe(
       Models.Manifest.decode(json),
-      fold(
+      fold<t.Errors, Models.Manifest, ManifestInfo>(
         // failure handler
-        () => {
-          throw new Error();
+        (errors) => {
+          const messages = formatValidationErrors(errors, {});
```
Heyrows: Is this really necessary ?

You're already sending custom errors through so is this needed ?

packages/slice-machine/scripts/start/index.ts
```diff
@@ -17,10 +17,7 @@ import { validateSession } from "./validateSession";
 
 async function run(): Promise<void> {
   const cwd: string = process.cwd(); // project running the script
-  const port: string =
-    findArgument(process.argv, "pepe").value ||
-    findArgument(process.argv, "port").value ||
-    "9999";
+  const port: string = findArgument(process.argv, "port").value || "9999";
```
Heyrows, line 20: thank you haha

packages/slice-machine/scripts/start/index.ts
```diff
@@ -47,7 +44,17 @@ async function run(): Promise<void> {
...
+    console.error(`[slice-machine] An unexpected error occurred. Exiting...`);
+    console.error("Full error: ", err);
+  });
+}
+
+export default start;
+
+const isCli = require.main === module;
+
+if (isCli) {
```
Heyrows: What's this ?

packages/slice-machine/scripts/start/validateManifest.ts
```diff
@@ -53,6 +53,11 @@ See below for more info 👇`,
       return { isManifestValid: false };
     }
 
+    case ManifestState.InvalidFramework: {
+      console.log("Set framework in your config to a valid value");
```
Heyrows: Why display this if we have custom errors now coming from the validator itself ?

What if we have multiple errors while validating the Manifest, don't you display them all ?

packages/slice-machine/lib/env/manifest.ts
```diff
@@ -49,63 +55,6 @@ export function extractRepo(parsedRepo: ParseResult): string | undefined {
...
-  }
-  return {
-    state: ManifestState.Valid,
-    message: Messages[ManifestState.Valid],
-    content: manifest,
-    repo,
-  };
-}
-
 function handleManifest(cwd: string): ManifestInfo {
```
Heyrows: I think this could be simplified.
We have functions in the core to retrieve the manifest.

packages/slice-machine/lib/env/manifest.ts
```diff
@@ -123,13 +72,32 @@ function handleManifest(cwd: string): ManifestInfo {
...
+          return {
+            state: ManifestState.Valid,
+            message: Messages[ManifestState.Valid],
+            content: manifest,
+            repo,
+          };
+        }
       )
     );
   } catch (e) {
```
Heyrows: do we still need that catch ?
Sound redundant

Heyrows (changes requested):

packages/slice-machine/lib/env/manifest.ts
MarcMcIntosh: it's for the call to readFileSync further up :)

packages/slice-machine/scripts/start/index.ts
MarcMcIntosh: I like being able to call the script directly and/or use it as a model.

So this checks if the script has been required or run directly by `node ../start/index`

packages/slice-machine/bin/start.js
MarcMcIntosh: I changed the start script to be a function :)

packages/init/__tests__/configure-project.test.ts
MarcMcIntosh: yeah I was going to refactor this so we get more visibility on the core by not mocking it

packages/core/src/node-utils/manifest.ts
MarcMcIntosh: If it goes in the retrieveManifest, there's going to be a lot of warnings happening before the exit message,
Screenshot 2022-04-06 at 12 02 39

Heyrows: Why ?

The core when validating returns IO-TS errors or custom errors that are catch by the code that uses it no ?

packages/slice-machine/bin/start.js
Heyrows: In which purpose :) ?

packages/slice-machine/scripts/start/index.ts
Heyrows: hmmmm instead of targeting this file you just need to target the BIN file and that's it you don't need this code.

packages/slice-machine/lib/env/manifest.ts
Heyrows: Then this can be eliminated using helpers from the Core

packages/slice-machine/lib/env/manifest.ts
MarcMcIntosh: yes, some of the types don't provide custom messages

packages/slice-machine/bin/start.js
MarcMcIntosh: `../build/scripts/start/index.js`

now exports a function, so I could test or run it :)

packages/core/src/node-utils/manifest.ts
MarcMcIntosh: `retrieveManifest` is called a lot thought slice-machine, if we do the validation in it then we get a result like that

packages/slice-machine/scripts/start/index.ts
MarcMcIntosh: fair enough

packages/core/src/node-utils/manifest.ts
MarcMcIntosh: or are you saying do it anyway?

Heyrows (approved):
Not convinced but we want to deploy it quickly. We'll have to come back on that later.

PR #435, opened 2022-03-31
Authenticate user to Intercom through segment.

Miniplop:
To authenticate a user and see actions to intercom through segment, we need to pass the intercom ID.
That will allow us to see where users struggle to use the product.


MarcMcIntosh (commented):
I have a few questions

Is the hash stored in the shared config and in redux persist?

Heyrows:
The hash will be stored in the shared config, the same as the intercomID basically.

MarcMcIntosh (approved):

PR #442, opened 2022-04-07
fix: start script running migrations before verifying sm.json

MarcMcIntosh:
If sm.json is invalid json
## Types of changes

## Checklist:

### DIFF
--- a/packages/core/src/node-utils/manifest.ts
+++ b/packages/core/src/node-utils/manifest.ts
@@ -35,6 +35,10 @@ export function retrieveManifest(
     };
   }
 
+  if (!content) {
+    throw new Error("Could not parse sm.json");
+  }
+
   return pipe(
     Manifest.decode(content),
     fold<t.Errors, Manifest, FileContent<Manifest, t.Errors>>(
--- a/packages/core/src/node-utils/manifest.ts
+++ b/packages/core/src/node-utils/manifest.ts
@@ -36,6 +36,7 @@ export function retrieveManifest(
   }
 
   if (!content) {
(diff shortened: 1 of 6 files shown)


Heyrows (approved):

packages/core/src/node-utils/manifest.ts
```diff
@@ -35,6 +35,10 @@ export function retrieveManifest(
     };
   }
 
+  if (!content) {
+    throw new Error("Could not parse sm.json");
```
Heyrows, line 40: lets call it temporary shall we !

packages/slice-machine/lib/env/manifest.ts
```diff
@@ -42,43 +42,51 @@ export function extractRepo(parsedRepo: ParseResult): string | undefined {
...
+        state: ManifestState.InvalidJson,
+        message,
+        content: null,
+      };
+    }
 
-  if (maybeManifest.errors) {
-    const messages = formatValidationErrors(maybeManifest.errors, {});
-    const message = messages.map((error) => "[sm.json] " + error).join("\n");
+    if (maybeManifest.content === null) {
```
Heyrows, line 66: This can't actually happen right ? Since we're catching it.

Heyrows (changes requested):

packages/slice-machine/lib/env/manifest.ts
MarcMcIntosh: It's unlikely but in the typescript thinks it can

Heyrows (approved):

PR #454, opened 2022-04-15
[Technical improvement] Replace and restructure internal project.

Heyrows:
A simple PR to update our file structure by
- providing an up to date next project with all features available to develop on.
- put the end to end tests up where they can be scoped outside of any packages.


MarcMcIntosh:
should probably call the folder e2e

MarcMcIntosh:
Last time I checked, lerna doesn't link from/to private packages.

There's a few work arounds for this.

MarcMcIntosh:
+ Maybe remove some of the slices and custom-types and add a .gitignore file to the folders to prevent to much noise on the project.

Defiantly ignore the videos.

Also I noticed that the paths in the `libraries-state` file have gone from relative to abosolute.
This will break when run on a different machine or by a different user

Screenshot 2022-04-19 at 18 49 45

Heyrows:
> should probably call the folder e2e

Yeah good idea, was thinking about that just didn't found yet how to specify a folder in the cypress config.

Heyrows:
> Last time I checked, lerna doesn't link from/to private packages. There's a few work arounds for this.

I'm not sure to understand what you mean.
Lerna doesn't need to handle the APP folder, these are just projects without SliceMachine being installed (so that we use them as filesystem)

Heyrows:
> * Maybe remove some of the slices and custom-types and add a .gitignore file to the folders to prevent to much noise on the project.

I don't think we should as those projects are meant to be complete so that we can like use it and develop in real time condition with as much use cases as we can.
<br>
> Defiantly ignore the videos.

I didn't understand that part sorry :x.
<br>
> Also I noticed that the paths in the `libraries-state` file have gone from relative to abosolute. This will break when run on a different machine or by a different user

Yeah, luckily this file is regenerated as soon as we launch Slice-machine but weird indeed.

e2e-projects/next/.slicemachine/libraries-state.json
```diff
@@ -0,0 +1,3848 @@
...
+            "primary": {}
+          }
+        },
+        "meta": {
+          "fileName": "index",
+          "extension": "js"
+        },
+        "screenshotPaths": {
+          "default-slice": {
+            "path": "../../e2e-projects/next/.slicemachine/assets/slices/ecommerce/ExternalVideoSlice/default-slice/preview.png",
```
MarcMcIntosh, line 588: weird that it's from the route of the project

e2e-projects/next/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "cimsirp",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "dev": "next dev -p 3500",
```
MarcMcIntosh, line 6: is this the right port for running in dev?

e2e-projects/next/slices/ecommerce/index.js
```diff
@@ -0,0 +1,44 @@
+// This file is generated by Slice Machine. Update with care!
```
MarcMcIntosh, line 1: Does this work with the slice-zone?

e2e-projects/next/slices/navigation/index.js
```diff
@@ -0,0 +1,14 @@
+// This file is generated by Slice Machine. Update with care!
```
MarcMcIntosh, line 1: slice-zone finds these right?

e2e-projects/next/sm.json
```diff
@@ -0,0 +1,10 @@
+{
+	"_latest": "0.3.3",
+	"apiEndpoint": "https://cimsirp.prismic.io/api/v2",
+	"libraries": [
+		"@/slices/ecommerce",
```
MarcMcIntosh, line 5: okay that makes sense

MarcMcIntosh (approved):

PR #461, opened 2022-05-06
Mm/tracking create custom type

MarcMcIntosh:
Notes on write the tests, i found some errors/warnings in my the console that happen when opening the custom-type creation form.

## Types of changes

## Checklist:

### DIFF
--- a/packages/slice-machine/components/Forms/CreateCustomTypeModal.tsx
+++ b/packages/slice-machine/components/Forms/CreateCustomTypeModal.tsx
@@ -15,7 +15,6 @@ import { ModalKeysEnum } from "@src/modules/modal/types";
 import { isLoading } from "@src/modules/loading";
 import { LoadingKeysEnum } from "@src/modules/loading/types";
 import { FormikErrors } from "formik";
-import { getRepoName } from "@src/modules/environment";
 
 import Tracker from "@src/tracker";
 
@@ -24,13 +23,11 @@ const CreateCustomTypeModal: React.FC = () => {
     useSliceMachineActions();
 
   const {
-    repoName,
     customTypeIds,
     isCreateCustomTypeModalOpen,
     isCreatingCustomType,
     customTypeLabels,
   } = useSelector((store: SliceMachineStoreType) => ({
-    repoName: getRepoName(store),
     customTypeIds: selectAllCustomTypeIds(store),
     customTypeLabels: selectAllCustomTypeLabels(store),
     isCreateCustomTypeModalOpen: isModalOpen(
@@ -55,7 +52,6 @@ const CreateCustomTypeModal: React.FC = () => {
       id,
       name,
       repeatable,
-      repo: repoName,
     });
     createCustomType(id, name, repeatable);
   };
--- a/packages/slice-machine/components/Forms/CreateCustomTypeModal.tsx
+++ b/packages/slice-machine/components/Forms/CreateCustomTypeModal.tsx
@@ -50,8 +50,8 @@ const CreateCustomTypeModal: React.FC = () => {
     repeatable: boolean;
   }) => {
     const name = label || id;
-    // notes repo name could be empty, and will the tracker be initialized?
-    void Tracker.get().trackCreatCustomType({
+
+    void Tracker.get().trackCreateCustomType({
       id,
       name,
       repeatable,
--- a/packages/slice-machine/components/Forms/CreateCustomTypeModal.tsx
+++ b/packages/slice-machine/components/Forms/CreateCustomTypeModal.tsx
@@ -15,17 +15,22 @@ import { ModalKeysEnum } from "@src/modules/modal/types";
 import { isLoading } from "@src/modules/loading";
 import { LoadingKeysEnum } from "@src/modules/loading/types";
 import { FormikErrors } from "formik";
+import { getRepoName } from "@src/modules/environment";
+
+import Tracker from "@src/tracker";
 
 const CreateCustomTypeModal: React.FC = () => {
   const { createCustomType, closeCreateCustomTypeModal } =
     useSliceMachineActions();
 
   const {
+    repoName,
     customTypeIds,
     isCreateCustomTypeModalOpen,
     isCreatingCustomType,
     customTypeLabels,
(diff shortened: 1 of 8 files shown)


packages/slice-machine/components/Forms/CreateCustomTypeModal.tsx
```diff
@@ -35,6 +40,26 @@ const CreateCustomTypeModal: React.FC = () => {
...
+    id,
+    label,
+    repeatable,
+  }: {
+    id: string;
+    label: string;
+    repeatable: boolean;
+  }) => {
+    const name = label || id;
+    // notes repo name could be empty, and will the tracker be initialized?
```
Heyrows: Is this a question for the reviewer or are you stating something 😄 ?

packages/slice-machine/src/tracker.ts
```diff
@@ -188,13 +188,28 @@ export class SMTracker {
...
   }
 
-  trackOnboardingSkip(screenSkipped: number): Promise<void> {
+  async trackOnboardingSkip(screenSkipped: number): Promise<void> {
     return this.#trackEvent(EventType.OnboardingSkip, {
       screenSkipped,
     });
   }
+
+  trackCreatCustomType(customTypeInfo: {
```
Heyrows:
```suggestion
  trackCreateCustomType(customTypeInfo: {
```

packages/slice-machine/src/tracker.ts
```diff
@@ -188,13 +188,28 @@ export class SMTracker {
...
+    const { id, name, repo, repeatable } = customTypeInfo;
+    const type = repeatable ? "repeatable" : "single";
+
+    const data = { id, name, repo, type };
+
+    return this.#trackEvent(EventType.CreateCustomType, data);
+  }
 }
 
+// is this supposed to be a singleton?
```
Heyrows: yes

packages/slice-machine/src/tracker.ts
```diff
@@ -188,13 +188,28 @@ export class SMTracker {
...
+  async trackOnboardingSkip(screenSkipped: number): Promise<void> {
     return this.#trackEvent(EventType.OnboardingSkip, {
       screenSkipped,
     });
   }
+
+  trackCreatCustomType(customTypeInfo: {
+    id: string;
+    name: string;
+    repo?: string;
```
Heyrows: I looked at the ticket and asked around for what this property looks like and it's used.
It seems that it would be used by Amplitude to link events to a Repository.

Knowing that, I asked François and Tony about how to use it properly making sure we don't use it the wrong way.

On an event, it should look like this.
```
  "context": {
    "groupId": {
      "Repository": ....
   }
```

After looking at how analytics is made for SliceMachine, we can add a context object on every tack call.

For that, we would need to add a new internal property to the tracker -> `repository`
And then we can use it on every track event call that we make.

2 things tho
- the repository property must be given on initialisation (backend + frontend + init ?)
- the repository property must be updated whenever we detect a changes in the SM.json (i guess that would only be for the frontend)

Heyrows (changes requested):

packages/slice-machine/components/Forms/CreateCustomTypeModal.tsx
MarcMcIntosh: Things to look into.

The tracker should be initialised in `_app.tsx` not sure about which cases would cause the repo to be null

packages/slice-machine/src/tracker.ts
MarcMcIntosh: typo... pretty sure my keyboard needs a clean.

packages/slice-machine/src/tracker.ts
MarcMcIntosh: Could also be done this way

```
class Singleton {
  count: number;
  private static instance: Singleton;

  private constructor() {
    this.count = 0
  }
  
  static getInstance(): Singleton {
    if (!Singleton.instance) {
      Singleton.instance= new Singleton();
    }
    return Singleton.instance;
  }
  
  log() {
    console.log("Hello " + this.count)
    this.count++
  }
}

 var a = Singleton.getInstance()
 var b = Singleton.getInstance()
 a.log()
 b.log()
 b.log()
 a.log()

```

packages/slice-machine/src/tracker.ts
MarcMcIntosh: Should that not have been done before estimating the ticket?

Heyrows: I guess you are right 😄.
It was added to a ticket not really knowing the subject I guess.

Here's the investigation to make it real 😉

packages/slice-machine/src/tracker.ts
Heyrows: Yes I guess so, but in the end it would give the same result no ?

packages/slice-machine/src/tracker.ts
Heyrows: haha, happens

packages/slice-machine/components/Forms/CreateCustomTypeModal.tsx
Heyrows: I guess the repo is null until the redux store is initialised 😄.

Heyrows: What I mean was more like.
If there is a comment, does it means that the work is not finished ?
Or it is a side comment about another subject ?

In that last case, is a comment the better way to retain the information and discuss about it ?

packages/slice-machine/src/tracker.ts
```diff
@@ -188,11 +188,25 @@ export class SMTracker {
...
+  async trackOnboardingSkip(screenSkipped: number): Promise<void> {
     return this.#trackEvent(EventType.OnboardingSkip, {
       screenSkipped,
     });
   }
+
+  async trackCreateCustomType(customTypeInfo: {
+    id: string;
+    name: string;
+    repo?: string;
```
Heyrows: You forgot this :)

Heyrows (changes requested):

Heyrows (approved):

PR #462, opened 2022-05-09
Mm/tracking add context

MarcMcIntosh:
## Types of changes
Pass context object to segment

## Checklist:

### DIFF
--- a/packages/init/src/index.ts
+++ b/packages/init/src/index.ts
@@ -48,7 +48,7 @@ async function init() {
     Tracker.get().identifyUser(user.profile.shortId, user.profile.intercomHash);
   }
 
-  void Tracker.get().trackInitIdentify(maybeRepositorySubdomain);
+  void Tracker.get().trackInitIdentify();
 
   // retrieve tokens for api calls
   const config = Prismic.PrismicSharedConfigManager.get();
@@ -65,6 +65,8 @@ async function init() {
     maybeRepositorySubdomain
   );
 
+  Tracker.get().setRepository(repositoryDomainName);
+
   // Install the required dependencies in the project.
   await installRequiredDependencies(cwd, frameworkResult.value);
 
@@ -80,10 +82,7 @@ async function init() {
     isTrackingAvailable
   );
 
-  await Tracker.get().trackInitDone(
-    frameworkResult.value,
-    repositoryDomainName
-  );
+  await Tracker.get().trackInitDone(frameworkResult.value);
 
   // Ask the user to run slice-machine.
   displayFinalMessage(cwd);
--- a/packages/init/src/index.ts
+++ b/packages/init/src/index.ts
@@ -28,7 +28,7 @@ async function init() {
     isTrackingAvailable
   );
 
-  Tracker.get().trackInitStart(maybeRepositorySubdomain);
+  void Tracker.get().trackInitStart(maybeRepositorySubdomain);
 
   console.log(
     logs.purple(
@@ -48,7 +48,7 @@ async function init() {
     Tracker.get().identifyUser(user.profile.shortId, user.profile.intercomHash);
   }
 
-  Tracker.get().trackInitIdentify(maybeRepositorySubdomain);
+  void Tracker.get().trackInitIdentify(maybeRepositorySubdomain);
 
   // retrieve tokens for api calls
   const config = Prismic.PrismicSharedConfigManager.get();
@@ -80,7 +80,10 @@ async function init() {
     isTrackingAvailable
   );
 
-  Tracker.get().trackInitDone(frameworkResult.value, repositoryDomainName);
+  await Tracker.get().trackInitDone(
+    frameworkResult.value,
+    repositoryDomainName
+  );
 
   // Ask the user to run slice-machine.
   displayFinalMessage(cwd);
--- a/packages/init/src/index.ts
+++ b/packages/init/src/index.ts
@@ -73,7 +73,7 @@ async function init() {
   const sliceLibPath = lib ? await installLib(cwd, lib, branch) : undefined;
 
   // configure the SM.json file and the json package file of the project..
-  configureProject(
+  await configureProject(
     cwd,
     base,
     repositoryDomainName,
@@ -81,8 +81,6 @@ async function init() {
     sliceLibPath,
     isTrackingAvailable
   );
-
-  await Tracker.get().trackInitDone(frameworkResult.value);
 
   // Ask the user to run slice-machine.
   displayFinalMessage(cwd);
--- a/packages/init/src/utils/tracker.ts
+++ b/packages/init/src/utils/tracker.ts
@@ -36,25 +36,27 @@ export class InitTracker {
   _trackEvent(
     eventType: EventType,
     attributes: Record<string, unknown> = {}
-  ): void {
-    if (!this._isTrackingPossible(this.#client)) {
-      return;
-    }
+  ): Promise<void> {
+    return new Promise((resolve) => {
+      if (!this._isTrackingPossible(this.#client)) {
+        return resolve();
+      }
 
-    const identifier = this._createSegmentIdentifier();
-
-    try {
-      this.#client.track({
+      const identifier = this._createSegmentIdentifier();
+      const payload = {
         event: eventType,
         ...identifier,
         properties: attributes,
         ...(this.#repository
           ? { context: { groupId: { Repository: this.#repository } } }
           : {}),
+      };
+
+      return this.#client.track(payload, () => {
+        // if(_) return reject(_)
+        return resolve();
       });
-    } catch {
-      // If the client is not correctly setup we are silently failing as the tracker is not a critical feature
-    }
+    });
   }
 
   _identifyEvent(userId: string, intercomHash: string): void {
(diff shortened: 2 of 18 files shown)


packages/slice-machine/lib/env/resolveAliases.ts
```diff
@@ -12,7 +12,10 @@ type PackageWithModuleAliases = JsonPackage & {
 const isAPackageHasModuleAliases = (
   jsonPackage: JsonPackage | PackageWithModuleAliases
 ): jsonPackage is PackageWithModuleAliases => {
-  return jsonPackage.hasOwnProperty("_moduleAliases");
+  return (
+    typeof jsonPackage === typeof {} &&
```
Heyrows, line 16: I don't understand why this file keeps giving eslint errors on different branches and the solution is always different 😂 .

```suggestion
    typeof jsonPackage === 'object' &&
```

packages/slice-machine/lib/env/versions.ts
```diff
@@ -13,7 +13,7 @@ export async function getPackageChangelog(
   dependencyCwd: string
 ): Promise<PackageChangelog> {
   const pkg = NodeUtils.retrieveJsonPackage(dependencyCwd);
-  if (!pkg.exists || !pkg.content?.name)
+  if (!pkg.exists || !pkg.content || !pkg.content?.name)
```
Heyrows, line 16: i'm not sure about this, it's not use in this IF
I think the fact that there is not `{}` + prettier gave an eslint error for not reason 😓

packages/slice-machine/server/src/api/services/getEnv.ts
```diff
@@ -50,7 +50,7 @@ function extractRepo(parsedRepo: ParseResult): string | undefined {
         return parsedRepo.subDomains[0];
       }
     default:
-      return;
+      return "";
```
Heyrows, line 53: Should this return an error or throw ?
Our app is kinda secured here by the fact that there is the validation on the API endpoint.
If one day for any reason that validation is broken, I think having an empty string here would silently broke the application, what do you think ?

packages/slice-machine/src/tracker.ts
```diff
@@ -42,6 +65,31 @@ export class SMTracker {
 
   /** Private methods **/
 
+  async #setupAnalyticsBrowser(
```
Heyrows: This is not what was expected.
The goal here was to pass the repository in the context of each track call.
image

Heyrows: For that, you don't need to rewrite a client or anything, the calls are already there (function #trackEvent).
You just need to store the repository in the class object and use it on every track event.

Heyrows (changes requested):
As far as I know from when we made the ticket together on Friday, this does not solve the issue.
Maybe there are some information I don't have ?
Does it work, has it been tested with Amplitude ?

packages/slice-machine/lib/env/versions.ts
MarcMcIntosh: It was a linting error, I guess there's been a small change somewhere

packages/slice-machine/server/src/api/services/getEnv.ts
MarcMcIntosh: It's on the todo list to verify the sm.json, atm in order to reach this case they will have to modify the `sm.json` at run time. 

The main issue here is that the repo-name can't be undefined.

packages/slice-machine/src/tracker.ts
MarcMcIntosh: ffs

packages/slice-machine/lib/env/resolveAliases.ts
MarcMcIntosh: Probably a minor/path update to a package

packages/slice-machine/lib/env/versions.ts
Heyrows: Not but I mean if we look after that, it returns an object that does not even use the `pkg.content.name` that's why I think we don't really need to change anything here.

And if we want to fix an error then we could just add `{}` on the return statement 😄

packages/slice-machine/src/tracker.ts
```diff
@@ -65,41 +44,19 @@ export class SMTracker {
...
   ): Promise<void> {
     if (!this.#isTrackingPossible(this.#client)) {
       return;
     }
+
     return this.#client
       .then((client): void => {
-        void client.track(eventType, attributes);
+        void client.track(eventType, attributes, {
+          context: { groupId: { Repository: this.#repository } },
```
Heyrows, line 58: neat, clean and localised, I love that ! 👍

Heyrows (approved):

Heyrows (changes requested):
Sorry I just noticed, this works ok but we still are missing acceptance criteria from the ticket
- The tracker from the Slice Machine Init have not been updated as well (only part of our events are being link to a repo then).

packages/slice-machine/server/src/api/services/getEnv.ts
Heyrows: Ahhh right I see what you mean now.
When Slice-Machine starts, we are Ok because we do validate the SM.json.
But if a change happens while Slice Machine is running, then we can't just throw an error yet because we don't handle them atm.

Thanks 👍

packages/init/src/index.ts
```diff
@@ -80,7 +80,10 @@ async function init() {
     isTrackingAvailable
   );
 
-  Tracker.get().trackInitDone(frameworkResult.value, repositoryDomainName);
+  await Tracker.get().trackInitDone(
```
Heyrows: I see how it would fix the issue of the Init event not being sent,
From a user perspective, does it run smoothly ?

What I means is, it is located between 2 steps, if we wait for it, does that means the user doesn't see any messages or anything happening until the event is sent ?

packages/init/src/utils/tracker.ts
```diff
@@ -89,23 +95,29 @@ export class InitTracker {
 
   identifyUser(userId: string, intercomHash: string): void {
     this.#userId = userId;
-    this._identifyEvent(userId, intercomHash);
+    return this._identifyEvent(userId, intercomHash);
   }
 
-  trackDownloadLibrary(library: string): void {
-    this._trackEvent(EventType.DownloadLibrary, { library });
+  trackDownloadLibrary(library: string): Promise<void> {
```
Heyrows, line 105: I like that you unified the return type of the functions sending events

packages/init/src/utils/tracker.ts
```diff
@@ -89,23 +95,29 @@ export class InitTracker {
...
+  trackInitStart(repoDomain: string | undefined): Promise<void> {
+    if (repoDomain) this.#repository = repoDomain;
+    return this._trackEvent(EventType.InitStart);
   }
 
-  trackInitDone(framework: Models.Frameworks, repoDomain: string): void {
-    this._trackEvent(EventType.InitDone, { framework, repo: repoDomain });
+  trackInitDone(
+    framework: Models.Frameworks,
+    repoDomain: string
```
Heyrows: I don't think we need to pass the repoDomain at that step, it should already be there.

packages/init/src/index.ts
```diff
@@ -48,7 +48,7 @@ async function init() {
     Tracker.get().identifyUser(user.profile.shortId, user.profile.intercomHash);
   }
 
-  Tracker.get().trackInitIdentify(maybeRepositorySubdomain);
+  void Tracker.get().trackInitIdentify(maybeRepositorySubdomain);
```
Heyrows: We shouldn't update the repository here, because it hasn't been validated yet.

Ex: If the user does not belong to the repository or is not an Admin, he will have to select another repository in the Step `chooseOrCreateARepository`.

Heyrows: I think Identifying the user is One things, updating the repo in the internal tracker is something else.

packages/init/src/utils/tracker.ts
```diff
@@ -89,23 +95,29 @@ export class InitTracker {
...
-  trackDownloadLibrary(library: string): void {
-    this._trackEvent(EventType.DownloadLibrary, { library });
+  trackDownloadLibrary(library: string): Promise<void> {
+    return this._trackEvent(EventType.DownloadLibrary, { library });
   }
 
-  trackInitIdentify(repoDomain: string | undefined): void {
-    this._trackEvent(EventType.InitIdentify, { repo: repoDomain });
+  trackInitIdentify(repoDomain: string | undefined): Promise<void> {
+    if (repoDomain) this.#repository = repoDomain;
```
Heyrows: modifying the Repo should be a function on it's own, we don't want to have to do that for every event.

Heyrows (changes requested):
Small things, the main work is done, well done.

packages/init/src/utils/tracker.ts
MarcMcIntosh: but it can be passed through these events, if it is it's own function. It'll have to be called a few times

packages/init/src/index.ts
MarcMcIntosh: means it takes a second or two longer before showing the message.

packages/init/src/utils/tracker.ts
Heyrows: As far as I know, only once.
When the repo has been selected / validated.

Do you think there is another case where the repo would be modifed ?

packages/init/src/index.ts
Heyrows: I think for that part we might want to Ask designers and UX what they think you know.

Heyrows: Like saying we have to have this but it does not belong to any step, how can we make it user friendly.

Heyrows (approved):
great work 👍

PR #468, opened 2022-05-17
track when a custom type is saved

MarcMcIntosh:
## Types of changes

## Checklist:

### DIFF
--- a/packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header.tsx
+++ b/packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header.tsx
@@ -2,18 +2,19 @@ import React from "react";
 import { Box, Button, Spinner, Text } from "theme-ui";
 
 import Header from "../../../../components/Header";
-import useSliceMachineActions from "@src/modules/useSliceMachineActions";
+import useSliceMachineActions from "../../../../src/modules/useSliceMachineActions";
 import { MdSpaceDashboard } from "react-icons/md";
 import { useSelector } from "react-redux";
-import { SliceMachineStoreType } from "@src/redux/type";
+import { SliceMachineStoreType } from "../../../../src/redux/type";
 import {
   selectCurrentCustomType,
   selectCustomTypeStatus,
   selectIsCurrentCustomTypeHasPendingModifications,
-} from "@src/modules/selectedCustomType";
+} from "../../../../src/modules/selectedCustomType";
 import { isLoading } from "@src/modules/loading";
 import { LoadingKeysEnum } from "@src/modules/loading/types";
-import { CustomTypeStatus } from "@src/modules/selectedCustomType/types";
+import { CustomTypeStatus } from "../../../../src/modules/selectedCustomType/types";
+import Tracker from "../../../../src/tracker";
 
 const CustomTypeHeader = () => {
   const {
@@ -39,6 +40,11 @@ const CustomTypeHeader = () => {
       return {
         onClick: () => {
           saveCustomType();
+          void Tracker.get().trackCustomTypeSaved({
+            id: currentCustomType.id,
+            name: currentCustomType.label,
+            type: currentCustomType.repeatable ? "repeatable" : "single",
+          });
         },
         children: (
           <span>
--- a/packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header.tsx
+++ b/packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header.tsx
@@ -14,7 +14,6 @@ import {
 import { isLoading } from "@src/modules/loading";
 import { LoadingKeysEnum } from "@src/modules/loading/types";
 import { CustomTypeStatus } from "../../../../src/modules/selectedCustomType/types";
-import Tracker from "../../../../src/tracker";
 
 const CustomTypeHeader = () => {
   const {
@@ -40,11 +39,6 @@ const CustomTypeHeader = () => {
       return {
         onClick: () => {
           saveCustomType();
-          void Tracker.get().trackCustomTypeSaved({
-            id: currentCustomType.id,
-            name: currentCustomType.label,
-            type: currentCustomType.repeatable ? "repeatable" : "single",
-          });
         },
         children: (
           <span>
--- a/packages/slice-machine/src/modules/selectedCustomType/sagas.ts
+++ b/packages/slice-machine/src/modules/selectedCustomType/sagas.ts
@@ -1,20 +1,15 @@
 import { call, fork, put, select, takeLatest } from "redux-saga/effects";
-import { openToasterCreator, ToasterType } from "@src/modules/toaster";
+import { openToasterCreator, ToasterType } from "../toaster";
 import { getType } from "typesafe-actions";
-import { withLoader } from "@src/modules/loading";
-import { LoadingKeysEnum } from "@src/modules/loading/types";
-import {
-  pushCustomTypeCreator,
-  saveCustomTypeCreator,
-} from "@src/modules/selectedCustomType/actions";
-import {
-  selectCurrentCustomType,
-  selectCurrentMockConfig,
-} from "@src/modules/selectedCustomType/index";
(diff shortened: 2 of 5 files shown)


packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header.tsx
```diff
@@ -2,18 +2,18 @@ import React from "react";
 import { Box, Button, Spinner, Text } from "theme-ui";
 
 import Header from "../../../../components/Header";
-import useSliceMachineActions from "@src/modules/useSliceMachineActions";
```
Heyrows, line 5: It's not the first time I see a change of import in your PR, is there a specific reason to do that ?

I mean it's not like the code didn't work before so is it because you prefer relative path ?

Heyrows, line 5: It's not the first time I see a change of import in your PR, is there a specific reason to do that ?

I mean it's not like the code didn't work before so is it because you prefer relative path ?

packages/slice-machine/src/modules/selectedCustomType/sagas.ts
```diff
@@ -30,6 +25,11 @@ export function* saveCustomTypeSaga() {
     }
 
     yield call(saveCustomType, currentCustomType, currentMockConfig);
+    void Tracker.get().trackCustomTypeSaved({
```
Heyrows, line 28: Interesting choice to put the tracking.

- Was it tested ?
- What happens when saving the custom type fails ?
- Why did you choose to put the tracking here instead of after a direct action ?

I'm not saying one approach is better than the other, i'd like for us to set a rule to follow from now on 😄.
Because at the moment, all the events are located with the code itself, not in the saga yet.
That's an interesting debate.

Heyrows, line 28: Interesting choice to put the tracking.

- Was it tested ?
- What happens when saving the custom type fails ?
- Why did you choose to put the tracking here instead of after a direct action ?

I'm not saying one approach is better than the other, i'd like for us to set a rule to follow from now on 😄.
Because at the moment, all the events are located with the code itself, not in the saga yet.
That's an interesting debate.

packages/slice-machine/src/modules/useSliceMachineActions.ts
```diff
@@ -1,31 +1,28 @@
...
 import {
   openSetupDrawerCreator,
   closeSetupDrawerCreator,
   toggleSetupDrawerStepCreator,
   checkSimulatorSetupCreator,
   connectToSimulatorIframeCreator,
-} from "@src/modules/simulator";
+} from "./simulator";
 import ServerState from "@models/server/ServerState";
-import { createCustomTypeCreator } from "@src/modules/availableCustomTypes";
```
Heyrows, line 25: These import, I feel changing them makes total sense, if the imported stuff are on the same level or above I feel it's okay.

But I'm not sure to see the need to change them everywhere you see one :)

Heyrows, line 25: These import, I feel changing them makes total sense, if the imported stuff are on the same level or above I feel it's okay.

But I'm not sure to see the need to change them everywhere you see one :)

packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header.tsx
MarcMcIntosh: Multiple reasons.
+ some times the alias path is longer.
+ my ide doesn't work with them.
+ the paths aren't resolved at compile time, so global module resolution is modified by the `module-alias` lib
+ `module-alias` doesn't work in projects that use hoisting. https://github.com/ilearnio/module-alias/issues/118
+ Node can do this natively https://github.com/ilearnio/module-alias/issues/113

packages/slice-machine/src/modules/selectedCustomType/sagas.ts
MarcMcIntosh: Yes it was tested.... the test was committed and you can check segment

I figured you would say put in the action when i was refactoring.

If it goes in the UI part, then it get's called regardless of success or failure, this way if ` call(saveCustomType` throws it should not be called.

MarcMcIntosh: duplicated

packages/slice-machine/src/modules/useSliceMachineActions.ts
MarcMcIntosh: see previous comment :)

packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header.tsx
Heyrows: - sometimes the alias path is long ✅ 
- my ide doesn't work with them 🟥 , pretty sure your IDE should be able to do that
- use of module alias, I do agree that is not a good practice.

Overall, if the path is long I would say ok but here replacing "@src" with "../../../../" doesn't seems right and honestly the aliases looses their purpose.

packages/slice-machine/src/modules/selectedCustomType/sagas.ts
Heyrows:
```
If it goes in the UI part, then it get's called regardless of success or failure, this way if call(saveCustomType throws it should not be called.
```

Interesting, I like that.
Which rule do you think we could come up with for the next events ?

Like, all events can't be done that way, so what would you state as good guidelines for creating events ?

packages/slice-machine/tests/pages/cts.spec.tsx
```diff
@@ -430,4 +430,99 @@ describe("Custom Type Builder", () => {
...
+    const uid = screen.getByText("UID");
+    fireEvent.click(uid);
+
+    const saveFieldButton = screen.getByText("Add");
+
+    await act(async () => {
+      fireEvent.click(saveFieldButton);
+    });
+
+    expect(fakeTracker).toHaveBeenCalledWith(
```
Heyrows, line 512: This is wrong no ?
Like the tracker shouldn't even had been called if the save to filesystem actually failed.

packages/slice-machine/tests/pages/cts.spec.tsx
```diff
@@ -430,4 +430,99 @@ describe("Custom Type Builder", () => {
...
+      "SliceMachine Custom Type Field Added",
+      { id: "uid", name: customTypeId, type: "UID", zone: "static" },
+      { context: { groupId: { Repository: "repoName" } } }
+    );
+
+    const saveCustomType = screen.getByText("Save to File System");
```
Heyrows, line 518: I'm guessing these came from the previous test

packages/slice-machine/src/modules/selectedCustomType/sagas.ts
MarcMcIntosh: I don't think there should be cases where sometimes it goes here and other times goes there, it's messy and inconsistent.

Make a redux middleware, it's a more flexible way to do it because all that's being done there is waiting of specific action types and is more uniform.

packages/slice-machine/tests/pages/cts.spec.tsx
MarcMcIntosh: It's part of the set up to make a saveable change to the custom-type

packages/slice-machine/tests/pages/cts.spec.tsx
MarcMcIntosh: Yes it's part of the set-up so there's a change that can be saved

packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header.tsx
MarcMcIntosh: > my ide doesn't work with them 🟥 , pretty sure your IDE should be able to do that
Well it doesn't.

Heyrows (approved):

PR #471, opened 2022-05-19
Mm/tracking create slice

MarcMcIntosh:
## Types of changes

## Checklist:

### DIFF
--- a/packages/slice-machine/src/modules/slices/index.ts
+++ b/packages/slice-machine/src/modules/slices/index.ts
@@ -70,7 +70,8 @@ export function* createSliceSaga({
   )) as SagaReturnType<typeof createSlice>;
   void Tracker.get().trackCreateSlice({
     id: payload.sliceName,
-    name: payload.libName,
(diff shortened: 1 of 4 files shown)


packages/slice-machine/src/modules/slices/index.ts
```diff
@@ -67,7 +68,12 @@ export function* createSliceSaga({
     payload.sliceName,
     payload.libName
   )) as SagaReturnType<typeof createSlice>;
+  void Tracker.get().trackCreateSlice({
+    id: payload.sliceName,
+    name: payload.libName,
```
Heyrows: are you sure that the properties are right ?

`LibName` doesn't sound like a `SliceName`
and
`SliceName` doesn't sound like `sliceId`.

MarcMcIntosh: Those are the only two properties given in the request when creating a slice,

Heyrows (approved):
looks good :)

PR #475, opened 2022-05-27
[Technical debt] Rework Slice Machine UI backend client.

Heyrows:
## Types of changes

- [x] Remove default/fake client for a simpler version
- [x] Change all the tests related to backend Environment
- [x] Remove extra logic to know if the user is logged in or not.

The idea behind this PR is to simplify the Client's logic and also relay the logic of the user being logged in or to outside APIs.
ex:
- The client tries to get Remote Slices
- the custom type API respond with a 403 because the user is not logged in
- the client just passed down the result to the frontend
- the frontend displays the login modal when receiving a 403

## Description

PR that brings changes to Slice Machine UI backend by simplifying the Client.

Extra:
- [ ] When the client is instantiate in Dev mode, read a json file on the OS (like manifest) that contains the url of the APIs. (useful for dev environment).


packages/slice-machine/lib/models/common/Environment.ts
```diff
@@ -1,11 +1,13 @@
 import PrismicData from "./PrismicData";
 import type { Models } from "@slicemachine/core";
-import DefaultClient from "./http/DefaultClient";
-import FakeClient from "./http/FakeClient";
 import { PackageChangelog } from "./versions";
 import { PackageManager } from "./PackageManager";
 
+import { ApplicationMode } from "../server/ApplicationMode";
```
MarcMcIntosh, line 6: what is applicaction mode?

packages/slice-machine/lib/models/server/ApplicationMode.ts
```diff
@@ -0,0 +1,16 @@
+import type { Models } from "@slicemachine/core";
+
+export enum ApplicationMode {
+  PROD = "prod",
+  STAGE = "stage",
+  DEV = "dev",
+}
+
+export function getApplicationMode(
```
MarcMcIntosh, line 9: hmm, that makes sense

packages/slice-machine/lib/models/server/Client/ApisEndpoints.ts
```diff
@@ -0,0 +1,20 @@
+export interface ApisEndpoints {
```
MarcMcIntosh, line 1: nice

packages/slice-machine/lib/models/server/Client/index.ts
```diff
@@ -0,0 +1,159 @@
...
+    body?: Body,
+    method: FetchMethods = FetchMethods.GET
+  ): Promise<Response> {
+    const headers = {
+      repository: this.repository,
+      Authorization: `Bearer ${this.authenticationToken}`,
+      "User-Agent": "slice-machine",
+    };
+
+    return fetch(url, {
```
MarcMcIntosh, line 55: Global fetch only exists in node v18 and above.

packages/slice-machine/lib/models/server/Client/upload.ts
```diff
@@ -2,19 +2,21 @@ import fs from "fs";
...
   filename,
   pathToFile,
-}: {
-  url: string;
-  fields: { [key: string]: string };
-  key: string;
-  filename: string;
-  pathToFile: string;
-}): Promise<number | undefined> {
+}: UploadParams): Promise<number | undefined> {
```
MarcMcIntosh, line 19: can this be undefined?

packages/slice-machine/package.json
```diff
@@ -29,7 +29,7 @@
     "bundle": "npm run build && npm run export-build",
     "lint": "eslint --cache --cache-location .caches/eslint --cache-strategy content --ext .ts,.tsx,.js,.jsx .",
     "lint:precommit": "eslint",
-    "audit": "better-npm-audit audit -l high -p",
+    "audit": "better-npm-audit audit -l high -p -x 1070404",
```
MarcMcIntosh, line 32: ohh, what happened?

packages/slice-machine/server/src/api/auth/status.ts
```diff
@@ -12,7 +12,7 @@ export default async function handler(
       return { status: "pending" };
     }
 
-    const profile = await getAndSetUserProfile(req.env, authToken);
+    const profile = await getAndSetUserProfile(req.env);
```
MarcMcIntosh: does it not need the auth token any more?

also, I'm not a fan of passing the whole env/context to a unit unless it uses it all. it's more readable/testable to pass the required parts only.

packages/slice-machine/server/src/api/custom-types/index.ts
```diff
@@ -26,8 +26,6 @@ const handleMatch = (matches: string[]) => {
 const fetchRemoteCustomTypes = async (
   env: BackendEnvironment
 ): Promise<{ remoteCustomTypes: CustomTypeSM[] }> => {
-  if (!env.isUserLoggedIn) return { remoteCustomTypes: [] };
```
MarcMcIntosh, line 29: what happens when the user is not logged in?

MarcMcIntosh: ...same thing

packages/slice-machine/server/src/api/services/getAndSetUserProfile.ts
```diff
@@ -1,16 +1,11 @@
-import DefaultClient from "@lib/models/common/http/DefaultClient";
 import { BackendEnvironment } from "@lib/models/common/Environment";
 import { UserProfile } from "@slicemachine/core/build/models";
-import preferWroomBase from "../../../../lib/utils/preferWroomBase";
 import { PrismicSharedConfigManager } from "@slicemachine/core/build/prismic";
 
 export async function getAndSetUserProfile(
-  env: BackendEnvironment,
-  authToken: string
+  env: BackendEnvironment
```
MarcMcIntosh: the whole env isn't needed here, 
like if you where to run this piece of code independently, a whole env is needed when only a small part of it is used :)

packages/slice-machine/server/src/api/services/getEnv.ts
```diff
@@ -105,7 +116,6 @@ export default async function getEnv(
         cwd,
         manifest: manifestInfo.content,
       }),
-      isUserLoggedIn: !!prismicData.value.auth && !!repo,
```
MarcMcIntosh, line 108: wondering to myself if this get sent to the client?

packages/slice-machine/server/src/api/services/uploadScreenshotClient.ts
```diff
@@ -37,52 +38,42 @@ export const upload = async (
   filePath: string
 ): Promise<{ s3ImageUrl?: string; err?: ApiError }> => {
   console.log("[slice/push]: uploading variation preview");
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/await-thenable
```
MarcMcIntosh, line 40: nice

MarcMcIntosh: no warnings right?

packages/slice-machine/tests/server/env.test.js
```diff
@@ -256,8 +264,120 @@ describe("getEnv", () => {
...
-    expect(env.isUserLoggedIn).toBeTruthy();
-    expect(env.client.base).toEqual("https://wroom.io");
-    expect(env.client.auth).toEqual("biscuits");
+    expect(env.client.authenticationToken).toEqual("biscuits");
+  });
+
+  test("it's client should use production apisEndpoints", async () => {
+    fs.reset();
+    fs.use(
+      Volume.fromJSON(
```
MarcMcIntosh, line 273: Absolute nightmare trying to do this in the init, I'll have to try again :)

packages/slice-machine/tests/server/env.test.js
```diff
@@ -229,20 +236,21 @@ describe("getEnv", () => {
     expect(env.framework).toEqual("vanillajs");
   });
 
-  test("it should take the auth from .prismic and base from sm.json", async () => {
+  test("it's client should use autentication token from .prismic", async () => {
```
MarcMcIntosh, line 239: user's probably won't run into the situation where the toke is not for prismic.io

packages/slice-machine/tests/lib/mock/Slice.test.js
```diff
@@ -1,6 +1,8 @@
 import MockSlice from "../../../lib/mock/Slice";
 import faker from "@faker-js/faker";
 
+global.console = { ...global.console, error: jest.fn() };
```
MarcMcIntosh, line 4: maybe not do this, as it's not reverted.
maybe 
`jest.spyOn(console, "error").mockImplimentation(()=>void))`
would work?

packages/slice-machine/tests/components/CreateCustomTypeModal.spec.tsx
```diff
@@ -14,6 +14,7 @@ import { AnalyticsBrowser } from "@segment/analytics-next";
 import Tracker from "../../src/tracker";
 
 jest.mock("next/dist/client/router", () => require("next-router-mock"));
+global.console = { ...global.console, error: jest.fn() };
```
MarcMcIntosh, line 17: yeah you might not want to over-ride the global console.error, as that's how bad tests are reported using a spy should work :)

packages/slice-machine/tests/__mocks__/fetchMock.ts
```diff
@@ -0,0 +1,17 @@
+export function FetchMock(
+  expectedUrl: string,
+  status: number = 200,
+  defaultValue: Record<string, unknown> = {}
+) {
+  global.fetch = jest.fn((url: RequestInfo | URL) => {
```
MarcMcIntosh: interesting, usually I would use `nock` or `msw` to intercept the network responses.

MarcMcIntosh (commented):
A few small comments,

But the big question is.

Should we move the "client" up to the "core" since some of the route and request logic is very simillar?

packages/slice-machine/lib/models/common/Environment.ts
Heyrows: I made it so that we know at every moment which mode we are in without having to use the repository endpoint everywhere.

packages/slice-machine/lib/models/server/ApplicationMode.ts
Heyrows: thank you ? 😂

packages/slice-machine/lib/models/server/Client/upload.ts
Heyrows: I didn't modify the code of this one actually, but I can imagine that if we return a number it's because we have a result and an undefined could mean we had a 500.

packages/slice-machine/package.json
Heyrows: An error was triggered on the audit, when I checked it there was no version available yet to upgrade a dependency to.
So I ignored it for now.

packages/slice-machine/server/src/api/auth/status.ts
Heyrows: The auth token is inside the client that this function calls, that's why we don't need it anymore.
Maybe we don't need the full environment, tbh I didn't modify much beside removing the token 😅

packages/slice-machine/server/src/api/custom-types/index.ts
Heyrows: If the user is not logged in, we'll know when we contact the API.
The API will return a 403 and we'll return that to the frontend that will display a login modal.

It's a case that is covered a bit below this code.

packages/slice-machine/server/src/api/services/uploadScreenshotClient.ts
Heyrows: no 😄 , I had to fix some stuff but I tried removing all the diisable I could find haha

packages/slice-machine/server/src/api/services/getEnv.ts
Heyrows: I'm not sure to understand 😅.

But for this, if the repository is not define, things should break earlier or in a different way (your work on errors management).
Then for the authentication token, if it is null or wrong, the value will be sent on a call and the API called will return 403.

packages/slice-machine/tests/server/env.test.js
Heyrows: nope you're right, but we might haha.

packages/slice-machine/tests/lib/mock/Slice.test.js
Heyrows: Yeah I'll try, I just wasn't sure which test was writing the error as I didn't wrote them myself.

packages/slice-machine/tests/__mocks__/fetchMock.ts
Heyrows: Actually it's because I got an error saying "fetch is not defined".
I used fetch in the code because I find it simple, non-library-dependent.

It's just when the test ran, it would just break 😅, so I made a helper to mock jest.

packages/slice-machine/lib/models/common/Environment.ts
MarcMcIntosh: neat

packages/slice-machine/lib/models/server/Client/index.ts
Heyrows: decided together -> adding `node-fetch` as hard dependency

MarcMcIntosh (approved):

PR #476, opened 2022-06-03
Client package

Heyrows:
## Description

This PR is about creating a client package that will hold common api calls between Slice Machine application.
The client is extendable and internal helpers are available to make a specific client per application (example of this in the init package).


packages/client/__tests__/__mocks__/aclCreateResult.ts
```diff
@@ -0,0 +1,10 @@
+import type { AclCreateResult } from "../../src/models/Acl";
```
MarcMcIntosh: __mocks__ directory is a special directory in jest https://jestjs.io/docs/manual-mocks
is there a module called `aclCreateResult`?

Also, technically these are stubs or fixtures

packages/client/__tests__/__mocks__/customType.ts
```diff
@@ -0,0 +1,23 @@
+import { CustomType } from "@prismicio/types-internal/lib/customtypes/CustomType";
```
MarcMcIntosh: same as above

packages/client/__tests__/__mocks__/sharedSlice.ts
```diff
@@ -0,0 +1,235 @@
+import {
```
MarcMcIntosh: same as above

packages/client/__tests__/__mocks__/userProfile.ts
```diff
@@ -0,0 +1,10 @@
+import { UserProfile } from "@slicemachine/core/build/models";
```
MarcMcIntosh: same as above

packages/client/.eslintignore
```diff
@@ -0,0 +1,3 @@
+build
+node_modules
+tests
```
MarcMcIntosh, line 3: __tests__

packages/client/src/index.ts
```diff
@@ -0,0 +1,189 @@
...
+      this.apisEndpoints = StageApisEndpoints;
+    else {
+      // Dev
+      this.apisEndpoints = ProductionApisEndpoints;
+    }
+  }
+
+  // setters to provide flexibility
+  updateAuthenticationToken(newToken: string) {
+    this.authenticationToken = newToken;
```
MarcMcIntosh, line 54: could return `this` for chain-ability?

packages/init/__tests__/choose-or-create-repo.test.ts
```diff
@@ -17,12 +17,16 @@ import { createRepository } from "../src/utils/create-repo";
 
 import nock from "nock";
 import { Models } from "@slicemachine/core";
+import { ApplicationMode } from "@slicemachine/client";
+import { Client } from "../src/utils";
 
 import * as fs from "fs";
 
 jest.mock("fs");
 jest.mock("../src/utils/create-repo");
```
MarcMcIntosh, line 26: why is the test called create-repo mocking a util called `create-repo`?

packages/init/__tests__/logInOrBypass.test.ts
```diff
@@ -37,29 +37,12 @@ describe("loginOrBypass", () => {
...
-  test("is should validate the session and get the users profile", async () => {
+  test("it should validate the session and get the users profile", async () => {
     jest.spyOn(fs, "lstatSync").mockImplementation(() => ({} as fs.Stats));
     jest.spyOn(fs, "writeFileSync");
 
-    jest
-      .spyOn(fs, "readFileSync")
-      .mockReturnValue(JSON.stringify({ cookies: fakeCookie, base: fakeBase }));
-
-    nock("https://auth.prismic.io")
```
MarcMcIntosh, line 54: is this no longer called?

packages/init/__tests__/logInOrBypass.test.ts
```diff
@@ -73,54 +56,35 @@ describe("loginOrBypass", () => {
...
-    expect(result?.profile?.intercomHash).toEqual(intercomHash);
-    expect(result?.profile?.userId).toEqual(userId);
+    expect(result.email).toEqual(email);
+    expect(result.shortId).toEqual(shortId);
+    expect(result.intercomHash).toEqual(intercomHash);
+    expect(result.userId).toEqual(userId);
 
     expect(stderr.output).toContain(email);
-
-    expect(fs.writeFileSync).toHaveBeenLastCalledWith(
```
MarcMcIntosh, line 90: if it's not writing out the user will have to login every time the init script is run?

packages/init/__tests__/logInOrBypass.test.ts
```diff
@@ -73,54 +56,35 @@ describe("loginOrBypass", () => {
...
-      ),
-      "utf8"
-    );
   });
 
-  test("user hasn't logged in for a few days", async () => {
+  test("it should consider the session invalid and ask the user to login", async () => {
     jest.spyOn(fs, "lstatSync").mockImplementation(() => ({} as fs.Stats));
 
-    jest
```
MarcMcIntosh, line 104: same

packages/init/__tests__/logInOrBypass.test.ts
```diff
@@ -134,20 +98,12 @@ describe("loginOrBypass", () => {
         lastName,
       });
 
-    const result = await loginOrBypass(fakeBase);
+    const result = await loginOrBypass();
```
MarcMcIntosh: how does it know which base to use?

packages/init/__tests__/utils/auth.test.ts
```diff
@@ -1,36 +1,33 @@
...
 
-import { Roles } from "@slicemachine/core/src/models";
 import { Endpoints } from "@slicemachine/core/build/prismic";
 import fs from "fs";
 import nock from "nock";
+import { ApplicationMode } from "@slicemachine/client";
 
-describe("auth", () => {
-  beforeEach(() => nock.cleanAll());
+Client.initialize(ApplicationMode.PROD, "biscuits");
```
MarcMcIntosh: it's is odd that it initialised before the function but not passed to it

packages/init/__tests__/utils/auth.test.ts
```diff
@@ -54,36 +52,6 @@ describe("auth", () => {
     await Auth.login(fakeBase);
   });
 
-  test("signup should always open the browser at the same url", async () => {
```
MarcMcIntosh, line 57: has this been removed?

packages/init/__tests__/utils/create-repo.test.ts
```diff
@@ -4,25 +4,27 @@ import nock from "nock";
 import { createRepository } from "../../src/utils/create-repo";
 import { stderr, stdout } from "stdout-stderr";
 import { Models } from "@slicemachine/core";
+import { Client } from "../../src/utils";
+import { ApplicationMode } from "@slicemachine/client";
+
+Client.initialize(ApplicationMode.PROD, "theBatman");
```
MarcMcIntosh: this is odd.

If it's supposed to be a singleton, it might be worth adding a destroy method so clean up can b done

packages/init/src/steps/choose-or-create-a-repository.ts
```diff
@@ -34,15 +32,9 @@ export async function promptForRepoDomain(
         type: "input",
         required: true,
         default: defaultValue,
-        transformer: (value: string) =>
-          prettyRepoName(address, value || defaultValue),
+        transformer: (value: string) => prettyRepoName(address, value),
```
MarcMcIntosh: pretty sure default value was there because in some cases the user would see "https://undefined.prismic.io"

packages/init/src/utils/index.ts
```diff
@@ -18,3 +20,18 @@ export function findArgument(args: string[], name: string): string | undefined {
 export const execCommand: (
   command: string
 ) => Promise<{ stderr: string; stdout: string }> = util.promisify(exec);
+
+export function getApplicationMode(
+  argumentValue: string | undefined
+): ApplicationMode | null {
+  switch (argumentValue) {
+    case ApplicationMode.PROD:
```
MarcMcIntosh, line 34: would a fall-through work?
```
switch(str) {
  case "a":
  case "b":
  case "c": return str
  default: return "oops"
}
```

packages/slice-machine/components/LoginModal/index.tsx
```diff
@@ -49,7 +49,7 @@ const LoginModal: React.FunctionComponent = () => {
 
   const prismicBase = preferWroomBase(env.manifest.apiEndpoint);
   const loginRedirectUrl = `${
-    buildEndpoints(prismicBase).Dashboard.cliLogin
+    buildEndpoints(prismicBase + "/").Dashboard.cliLogin
```
MarcMcIntosh, line 52: pretty sure theres a way to make sure a url ends with "/"

packages/slice-machine/server/src/api/custom-types/index.ts
```diff
@@ -26,33 +25,17 @@ const handleMatch = (matches: string[]) => {
...
-    const { remoteCustomTypes } = await (async (): Promise<{
-      remoteCustomTypes: CustomTypeSM[];
-    }> => {
-      if (res.status > 209) {
-        return { remoteCustomTypes: [] };
-      }
-      const result = await (res.json
-        ? (res.json() as Promise<Array<unknown>>)
-        : Promise.resolve([]));
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
MarcMcIntosh, line 40: nice

packages/init/src/utils/client.ts
```diff
@@ -0,0 +1,66 @@
+import * as t from "io-ts";
+import { Models } from "@slicemachine/core";
+import { PrismicSharedConfigManager } from "@slicemachine/core/build/prismic";
+import {
+  Client as ClientModel,
+  ApplicationMode,
+  getAndValidateResponse,
+} from "@slicemachine/client";
+
+export const Client = (() => {
```
MarcMcIntosh: Interesting article about implementing singeltons although it dos say some consider it an anti-pattern
https://refactoring.guru/design-patterns/singleton/typescript/example

packages/init/src/utils/create-repo.ts
```diff
@@ -1,30 +1,22 @@
 import type { Models } from "@slicemachine/core";
 import * as logs from "./logs";
-import * as Prismic from "@slicemachine/core/build/prismic";
+import { Client } from "./client";
 
 export function createRepository(
```
MarcMcIntosh, line 5: would it make more sense to pass the client as an argument ?

packages/client/src/index.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  updateRepository(repository: string | null) {
+    this.repository = repository;
+  }
+
+  // private methods
+  _fetch(
+    method: "get" | "post",
+    url: string,
+    data?: Record<string, unknown>,
```
MarcMcIntosh: axios provides argument types, for headers and bodies.

packages/client/src/utils/index.ts
```diff
@@ -0,0 +1,33 @@
...
+          `Unable to parse ${ressource}: ${JSON.stringify(response.data)}`
+        )
+      )(decoder.decode(response.data));
+    })
+    .catch((error: Error | string) => {
+      const status: number = typeof error == "string" ? 500 : getStatus(error);
+
+      const message: string =
+        axios.isAxiosError(error) && error.response
+          ? `Unable to retrieve ${ressource} with status code ${error.response.status}`
```
MarcMcIntosh, line 24: are nested ternary operators a code smell?

packages/client/src/models/ClientError.ts
```diff
@@ -0,0 +1,20 @@
...
+  message: string;
+}
+
+export function getStatus(error: Error | AxiosError): number {
+  return axios.isAxiosError(error) ? error.response?.status || 500 : 500;
+}
+
+export function getMessage(error: Error | AxiosError): string {
+  if (axios.isAxiosError(error) && error.response) {
+    return typeof error.response.data == "string"
```
MarcMcIntosh, line 14: that's the actual data returned but the request.
there's also error.response.statusText

packages/client/src/utils/upload.ts
```diff
@@ -1,22 +1,15 @@
...
 import mime from "mime";
-
-export type UploadParams = {
-  url: string;
-  fields: Record<string, string>;
-  key: string;
-  filename: string;
-  pathToFile: string;
-};
+import { UploadParameters } from "../models/UploadParameters";
```
MarcMcIntosh: are upload parameters use else where?

packages/core/src/prismic/endpoints.ts
```diff
@@ -11,9 +11,9 @@ export function buildEndpoints(base: string): Endpoints {
 
   return {
     Dashboard: {
-      index: endpoint("/dashboard"),
-      cliLogin: endpoint("/dashboard/cli/login?source=slice-machine"),
-      cliSignup: endpoint("/dashboard/cli/signup?source=slice-machine"),
+      index: endpoint("dashboard"),
```
MarcMcIntosh, line 14: so I'm seeing `/` added and remove in a few places.

should the endpoint function be able to handle both?

packages/init/src/index.ts
```diff
@@ -11,24 +11,32 @@ import {
   detectFramework,
   installLib,
 } from "./steps";
-import { findArgument, logs } from "./utils";
+import { findArgument, logs, getApplicationMode, Client } from "./utils";
 
 async function init() {
   const cwd = findArgument(process.argv, "cwd") || process.cwd();
-  const base = findArgument(process.argv, "base") || CONSTS.DEFAULT_BASE;
```
MarcMcIntosh, line 19: better let QA know not to use the `base` flag

packages/init/src/index.ts
```diff
@@ -11,24 +11,32 @@ import {
...
     process.env.PUBLIC_SM_INIT_SEGMENT_KEY ||
       "JfTfmHaATChc4xueS7RcCBsixI71dJIJ",
     isTrackingAvailable
   );
 
-  void Tracker.get().trackInitStart(maybeRepositorySubdomain);
+  void Tracker.get().trackInitStart(preSelectedRepository);
+
+  // initializing the client with what we have for now.
+  Client.initialize(mode, Prismic.PrismicSharedConfigManager.getAuth());
```
MarcMcIntosh: should pass this down to the other functions

packages/init/src/steps/choose-or-create-a-repository.ts
```diff
@@ -1,19 +1,17 @@
...
-const DEFAULT_BASE = CONSTS.DEFAULT_BASE;
 
-export function prettyRepoName(address: URL, value?: string): string {
-  const repoName = value ? logs.cyan(value) : logs.dim.cyan("repo-name");
-  return `${logs.cyan.dim(`${address.protocol}//`)}${repoName}${logs.cyan.dim(
-    `.${address.hostname}`
-  )}`;
+export function prettyRepoName(address: URL, domain: string): string {
+  return `${logs.cyan.dim(`${address.protocol}//`)}${logs.cyan(
+    domain
```
MarcMcIntosh, line 13: sometimes domain is undefined

packages/init/src/steps/choose-or-create-a-repository.ts
```diff
@@ -124,45 +116,60 @@ export function sortReposForPrompt(
...
     {
       type: "list",
-      name: "chosenRepo",
+      name: "chosenRepository",
       default: 0,
       required: true,
       message: "Connect a Prismic Repository or create a new one",
       choices,
       pageSize: numberOfDisabledRepos + 2 <= 7 ? 7 : numberOfDisabledRepos + 2,
-      // loop: false
```
MarcMcIntosh, line 158: was looping an issue when there are lots of repos.... I'll have to check

packages/init/src/utils/auth/helpers.ts
```diff
@@ -81,13 +81,17 @@ const authenticationHandler =
...
+}
+
 function buildServer(base: string, port: number, host: string): hapi.Server {
   const server = hapi.server({
     port,
     host,
     routes: {
       cors: {
-        origin: [base],
+        origin: [stripTrailingSlash(base)],
```
MarcMcIntosh, line 94: more trailing slash issues

packages/init/src/utils/auth/index.ts
```diff
@@ -1,10 +1,10 @@
...
+import { Utils } from "@slicemachine/core";
 import { startServerAndOpenBrowser } from "./helpers";
 import {
-  Communication,
   Endpoints,
   PrismicSharedConfigManager,
 } from "@slicemachine/core/build/prismic";
+import { Client } from "../client";
 
 async function startAuth({
```
MarcMcIntosh, line 9: would injecting the client make more sense?

packages/slice-machine/tests/server/services/sliceService.test.ts
```diff
@@ -6,17 +6,19 @@ import {
...
 import backendEnvironment from "../../__mocks__/backendEnvironment";
+import { Client, ApplicationMode } from "@slicemachine/client";
 import { Slices } from "@slicemachine/core/build/models/Slice";
 import { resolvePathsToScreenshot } from "@slicemachine/core/build/libraries/screenshot";
 
 const allFieldSliceModel = Slices.toSM(allFieldSliceObject);
 
-jest.mock("../../../lib/models/server/Client", () => {
+jest.mock("@slicemachine/client", () => {
+  const originalModule = jest.requireActual("@slicemachine/client");
```
MarcMcIntosh, line 18: would copying the nock objects from the client also work, at least that way you can be double sure it works as expected?

packages/slice-machine/tests/server/getAndSetUserProfile.test.ts
```diff
@@ -15,7 +15,7 @@ jest.mock(`fs`, () => {
 describe("setShortId", () => {
   afterEach(() => {
     vol.reset();
-    fetchMock.reset();
+    nock.cleanAll();
```
MarcMcIntosh, line 18: nock.restore ?
https://github.com/nock/nock#memory-issues-with-jest

packages/slice-machine/tests/server/env.test.js
```diff
@@ -236,76 +231,6 @@ describe("getEnv", () => {
     expect(env.framework).toEqual("vanillajs");
   });
 
-  test("it's client should use autentication token from .prismic", async () => {
```
MarcMcIntosh, line 239: has this changed?

packages/slice-machine/server/src/api/slices/push.ts
```diff
@@ -44,26 +45,23 @@ export async function pushSlice(
...
     };
 
-    const res = await createOrUpdate(
-      slices,
-      sliceName,
-      modelWithImageUrl,
-      env.client
-    );
-
-    if (res.status > 209) {
```
MarcMcIntosh, line 54: why 209?

MarcMcIntosh: I guess the custom-types-api can does some odd responses

packages/init/src/utils/create-repo.ts
```diff
@@ -1,30 +1,22 @@
...
-      addressUrl.hostname = `${repoDomainName}.${addressUrl.hostname}`;
+  return Client.createRepository(domain, framework)
+    .then((domain: string) => {
+      const addressUrl = new URL(Client.get().apisEndpoints.Wroom);
+      addressUrl.hostname = `${domain}.${addressUrl.hostname}`;
       const address = addressUrl.toString();
       spinner.succeed(`We created your new repository ${address}`);
 
-      return repoDomainName;
+      return domain;
```
MarcMcIntosh, line 21: domain is declared at the top of this function ;)

packages/client/README.md
```diff
@@ -0,0 +1,3 @@
+# Slice Machine Client
```
MarcMcIntosh, line 1: There's gotta be a tool to create reaedme's from js-doc style comments

packages/client/__tests__/utils/index.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    ).resolves.toBe(expectedResult);
+  });
+
+  it("should get the ressource but not validate the payload", async () => {
+    const url = "https://fakeWebsite.com";
+    const expectedResult = "Amazing";
+
+    nock(url).get("/").reply(200, expectedResult);
+
+    expect.assertions(1);
```
MarcMcIntosh, line 29: what's this do?

packages/client/__tests__/utils/index.test.ts
```diff
@@ -0,0 +1,53 @@
...
+      expect(e.message).toContain("Unable to parse");
+    });
+  });
+
+  it("should fail to the ressource the ressource", async () => {
+    const url = "https://fakeWebsite.com";
+
+    nock(url).get("/").reply(404);
+
+    expect.assertions(1);
```
MarcMcIntosh, line 44: ahh, it makes sure it throw.

would 
```
expect(() => fn(...)).rejects.toThrow(""Unable to retrieve")
```

also have worked?

packages/client/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "eslint-config-prettier": "^8.3.0",
+    "eslint-import-resolver-typescript": "^2.4.0",
+    "eslint-plugin-import": "^2.24.0",
+    "eslint-plugin-jest": "^24.4.0",
+    "jest": "^27.2.4",
+    "nock": "^13.1.3",
+    "ts-jest": "^27.0.5",
+    "types-package-json": "^2.0.39",
+    "typescript": "^4.4.3",
+    "unionfs": "^4.4.0"
```
MarcMcIntosh: tbh I would recommend using `mock-fs` over unionfs+memfs it's a lot easier to use, and less buggy :) 
https://www.npmjs.com/package/mock-fs

packages/client/src/index.ts
```diff
@@ -0,0 +1,189 @@
...
+    headers?: Record<string, unknown>
+  ): AxiosPromise {
```
MarcMcIntosh: No type paramters?

MarcMcIntosh (commented):
Few questions/observations/recommendations 

and somethings I'll have to run to check (mostly the init repo-name display and list)

packages/client/__tests__/__mocks__/aclCreateResult.ts
Heyrows: AclCreateResult is a type,
I'll rename the folder then 😄

packages/client/__tests__/__mocks__/customType.ts
Heyrows: you mean like renaming the folder ?

packages/client/.eslintignore
Heyrows: 😅 , tbh I pretty much copy pasted the configuration from the Core / Init

packages/client/src/index.ts
Heyrows: hmmmm that's a good one, didn't thought of that.

Heyrows: Not that we actually need it because we don't do things like that chained but yeah

packages/init/__tests__/choose-or-create-repo.test.ts
Heyrows: Because those tests are testing everything that revolves around creating a repo (user input, message displayed, ...) and not the function to create a repo itself.

packages/init/__tests__/logInOrBypass.test.ts
Heyrows: no, I changed the code to call the profile route which will return a 401 if the user isn't logged in.
I didn't that to avoid having to do a validate session and then call the profile.

packages/init/__tests__/logInOrBypass.test.ts
Heyrows: yes it does, it's on the "OnSuccess" of the Auth.helpers https://github.com/prismicio/slice-machine/blob/23a5c596b5abdd5d50dc53c500b35b5765c94027/packages/init/src/utils/auth/helpers.ts#L140

Heyrows: Here it was useless because in this test the session is valid.

packages/init/__tests__/logInOrBypass.test.ts
Heyrows: here the lines are just below, I didn't remove that part

packages/init/__tests__/logInOrBypass.test.ts
Heyrows: It's the application Mode, when we initialise the client.
This way we have pre-defined ApisEndpoints, and we don't pass a `base` string everywhere.

The base approach to me is too weak because it's a string and on top of that we don't really know what it looks like :x.
I find applicationMode being more intuitive (dev / stage / prod).

Even for QA, now it will be `--mode stage` instead of `--base https://.....`

packages/init/__tests__/utils/auth.test.ts
Heyrows: I'm not sure to understand `but not passed to it`.

Here as Jest run every file independently and the Client is a singleton, initialising it makes it available for the current file.

packages/init/__tests__/utils/auth.test.ts
Heyrows: Yes, the signup was not used so I got rid of it.

packages/init/__tests__/utils/create-repo.test.ts
Heyrows: We don't need it.
Each file seems to be test independently, if you try removing this line in one of the test suite, you'll see that the tests are not passing anymore.
In Slice Machine and in the Init, we do not need to destroy it.
In the tests, we also do not need to destroy it because we use the same authentication token / applicationMode

packages/init/src/steps/choose-or-create-a-repository.ts
Heyrows: does that mean the value is wrongly typed ?

packages/init/src/utils/index.ts
Heyrows: Super interesting, I didn't think the type would be inferred from that, thanks 👍

packages/slice-machine/components/LoginModal/index.tsx
Heyrows: This was though 😂.
If it was up to me I would have continue to refacto that part.
So I tried to calm myself haha.

Not adding this created CORS issue when trying to login 😓  it was a pain to figure out this.

packages/init/src/utils/client.ts
Heyrows: I checked https://stackoverflow.com/questions/1479319/simplest-cleanest-way-to-implement-a-singleton-in-javascript, it seemed to fit what Kevin did (that we know work) so I did the same 😄.

packages/init/src/utils/create-repo.ts
Heyrows: We could but since it's a singleton, it doesn't really make sense.
That would just be adding more parameters when it already is accessible.

It could make sense to make sure it is instantiated, I wanted to avoid passing it everywhere.
On top of that, the singleton approach allowed me to create more calls that are required for this application based on the Client's code.

packages/client/src/index.ts
Heyrows: For the body I think it was that, i will double check.
For the headers good idea 👍

Heyrows: Just checked, there is for header 👍 .
For body, it is inferred when you give a type to axios.

I don't think it is useful here as it doesn't affect the return type.

packages/client/src/utils/index.ts
Heyrows: I do agree, usually that's not the best.
Can be usefull in react tho.

I used it to avoid having big "ifs" with the same return statement at the end of each.
Do you want me to change it back ?

packages/client/src/models/ClientError.ts
Heyrows: Yes I know.
This specific case is for the piece of code above that validate the response.data using the validators.
As we first Reject an error, we received a string here so we have to treat it as such.

Heyrows: I wanted to differentiate both cases parsing `(message: Unable to parse ...)` and default `(message: Unable to retrieve ...)`

packages/client/src/utils/upload.ts
Heyrows: Beside being exported for external use, no.

packages/core/src/prismic/endpoints.ts
Heyrows: This function should normally be removed tbh.
I only fixed it for now, but I don't like it.

It's still based on the "base" argument and not "applicationMode".

packages/init/src/index.ts
Heyrows: Yes 😄  I planned to let them know once this PR is done and merged.
I'm sure this will make it easier for them 👍

packages/init/src/index.ts
Heyrows: Answered above, we can still discuss it but i'm resolving this one to not have multiple comment about the same thing.

packages/init/src/steps/choose-or-create-a-repository.ts
Heyrows: already talking about it in another comment.

packages/init/src/steps/choose-or-create-a-repository.ts
Heyrows: Haha I don't really know but since it was unused I just removed it :)

packages/init/src/utils/auth/helpers.ts
Heyrows: Yeah but here the function is important as it can cause CORS issues while logging.

packages/client/.eslintignore
MarcMcIntosh: :/ so there maybe more configuration that reuiqred?

packages/slice-machine/tests/server/getAndSetUserProfile.test.ts
Heyrows: I believe I tried restore once and it failed telling me on the next nock that the URL was already define.

packages/slice-machine/tests/server/services/sliceService.test.ts
Heyrows: That test was like one the first I fixed 😓.
In the end it's just testing that we call the right functions.

The fact that we make the right call is already checked in the tests of the Client.

packages/slice-machine/tests/server/env.test.js
Heyrows: In a sense yes, we do get the token and then ask the API.
I didn't wanted to refactor more for now.

packages/slice-machine/server/src/api/slices/push.ts
Heyrows: haha, more than 2xx is considered errors I guess 😅

Heyrows: We can always check the readme of the Custom type API, it is really precise.

packages/init/src/utils/create-repo.ts
Heyrows: Yes, I just wanted to keep sending back the domain as this is what the server is doing (even though we already have the value).

packages/client/README.md
Heyrows: Maybe 🤷‍♂️ , I took it from the Core and wrote a small thing.

packages/client/__tests__/utils/index.test.ts
Heyrows: It means that the test should expect at least once.

packages/client/__tests__/utils/index.test.ts
Heyrows: I think I came across this solution because It wasn't throwing.

packages/client/package.json
Heyrows: I use nock in the end so that just remnant from the Core haha.
I did a depcheck on Core / Init / Client and removed some stuff 😄

packages/client/src/index.ts
Heyrows: What do you mean ?
Like an object and typing the whole object ?

packages/client/.eslintignore
Heyrows: I was thinking, the deploy alpha might fail as it doesn't build the client first also. (same as core)
For which we might want to create a command like "build essential packages" to prevent that from happening again.

packages/init/__tests__/logInOrBypass.test.ts
MarcMcIntosh: K, but should mode be passed down?

packages/init/__tests__/utils/auth.test.ts
MarcMcIntosh: Passed down as an argument to the funciton

packages/client/__tests__/utils/index.test.ts
MarcMcIntosh: cool

packages/client/__tests__/utils/index.test.ts
MarcMcIntosh: really?

packages/client/src/index.ts
MarcMcIntosh: `_fetch<T> : AxiosPromise<T> { axios<T> }`

packages/init/src/utils/index.ts
MarcMcIntosh: it worked?.... neat :)

packages/client/package.json
MarcMcIntosh: nice

packages/init/src/steps/choose-or-create-a-repository.ts
MarcMcIntosh: yes

packages/slice-machine/components/LoginModal/index.tsx
MarcMcIntosh: the dash caused a cors issue?... weird

packages/client/src/utils/index.ts
MarcMcIntosh: not really to bothered sine it's pretty short.

packages/slice-machine/tests/server/services/sliceService.test.ts
MarcMcIntosh: Those could change over time no?

I would mock the network, saves having to assume the client works

packages/slice-machine/tests/server/getAndSetUserProfile.test.ts
MarcMcIntosh: hmmm, that's odd, not sure if changing `it` to `test` would work :/

packages/client/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "@slicemachine/core": "1.1.10-alpha.0",
+    "axios": "^0.27.2",
+    "mime": "^2.5.2",
+    "form-data": "^3.0.0",
+    "fp-ts": "^2.11.5",
+    "io-ts": "^2.2.16"
+  },
+  "devDependencies": {
+    "@babel/core": "^7.15.8",
+    "@types/axios": "^0.14.0",
```
MarcMcIntosh: axios comes with it's own types

packages/init/package.json
```diff
@@ -56,7 +57,6 @@
     "eslint-plugin-import": "^2.24.0",
     "eslint-plugin-jest": "^24.4.0",
     "jest": "^27.4.7",
-    "mock-stdin": "^1.0.0",
```
MarcMcIntosh, line 69: good catch

packages/init/src/utils/create-repo.ts
MarcMcIntosh: Which one, domain is defined on line 6 and 15 ;)
I guess they'll both be the same,

packages/init/src/utils/create-repo.ts
MarcMcIntosh: nice

MarcMcIntosh (approved):
Small comment about domain being defined on line 6 and 15, but it's no a big deal.
So approved, assuming conflicts are resolved and and the CI passed

.github/workflows/deploy-alpha.yml
```diff
@@ -28,6 +28,9 @@ jobs:
         run: |
           npm ci
           git checkout .
+      - name: Build packages used by others
+        run: |
+          npm run build-essential
```
MarcMcIntosh: cool,

But, this will need to be added to the master branch in order to work ;)

.github/workflows/test.yml
```diff
@@ -27,9 +27,8 @@ jobs:
           cache: "npm"
       - name: Install dependencies
         run: npm i
-      - name: Build slice machine core
-        working-directory: ./packages/core
-        run: npm run build
+      - name: Build essential packages used by others
+        run: npm run build-essential
```
MarcMcIntosh: This will be run with out needing to add it to master

MarcMcIntosh (dismissed):

packages/client/package.json
Heyrows: good one, done

package.json
```diff
@@ -29,19 +30,15 @@
...
   "dependencies": {
-    "@segment/analytics-next": "^1.32.1",
-    "@theme-ui/color": "^0.3.1",
     "cypress": "^9.5.4",
     "cypress-localstorage-commands": "^1.7.0",
     "cypress-wait-until": "^1.7.2",
     "jest": "^25.5.1",
     "node-fetch": "^2.6.1",
-    "start-server-and-test": "^1.14.0",
-    "validate-npm-package": "^1.0.5"
```
MarcMcIntosh, line 45: what is validate-npm-package?

packages/init/src/utils/validateRepositoryName.ts
```diff
@@ -33,5 +33,12 @@ export async function validateRepositoryName(
     return Promise.reject(new Error(msg));
   }
 
-  return client.domainExist(domain).catch(() => false);
+  return client
+    .domainExist(domain)
+    .then((isAvailable) =>
+      isAvailable
+        ? true
+        : Promise.reject(new Error(`${name} is already in use`))
```
MarcMcIntosh, line 41: should throw no?
Promise.resolve(Prosmise.reject(...))

packages/init/src/steps/starters/slices.ts
```diff
@@ -37,25 +35,35 @@ export async function sendSlicesFromStarter(
   );
   spinner.start();
 
-  const acl = await createAcl(endpoints.AclProvider, repository, authorization);
+  const acl: Acl = await client.createAcl().catch((error: ClientError) => {
+    writeError(
+      "Uploading screenshots for your slices failed, please contact us."
+    );
+    writeError(error.message, "Full error:");
+    process.exit(1);
```
MarcMcIntosh: Should it fail hard when screenshots are not uploaded?

packages/init/src/steps/sendStarterData.ts
```diff
@@ -1,32 +1,16 @@
...
+import { InitClient } from "../utils";
 
-export async function sendStarterData(
-  repository: string,
-  base: string,
-  cookies: string,
-  cwd: string
-) {
-  const smJson = retrieveManifest(cwd);
+export async function sendStarterData(client: InitClient, cwd: string) {
```
MarcMcIntosh: note to self: we will need the cookies for the documents

packages/init/src/steps/choose-or-create-a-repository.ts
```diff
@@ -1,26 +1,24 @@
...
+  )}${logs.cyan.dim(`.${address.hostname}`)}`;
 }
 
 export async function promptForRepoDomain(
-  base: string,
+  client: InitClient,
   defaultValue?: string
 ): Promise<string> {
-  const address = new URL(base);
+  const address = new URL(client.apisEndpoints.Wroom);
```
MarcMcIntosh, line 21: Wroom?

packages/client/src/utils/uploadScreenshot.ts
```diff
@@ -0,0 +1,67 @@
...
+    variationId
+  );
+  form.append("key", screenshotKey);
+
+  // adding content type
+  const fileExtension = fileName.split(".").pop();
+  form.append("Content-Type", fileExtension && mime.getType(fileExtension));
+
+  // adding the file
+  form.append("file", fs.createReadStream(filePath), {
```
MarcMcIntosh, line 49: this can fail because s3 needs to know the file length in-order to upload the screenshot, there's a solution for this in the init

packages/client/src/utils/str.ts
```diff
@@ -0,0 +1,8 @@
+const hyphenateRE = /\B([A-Z])/g;
+export function hyphenate(str: string): string {
+  return str.replace(hyphenateRE, "-$1").toLowerCase();
+}
+
+export function snakelize(str: string): string {
+  return hyphenate(str).replace(/-/g, "_");
```
MarcMcIntosh, line 7: honestly I would just use lodash for this

MarcMcIntosh (commented):
When uploading the screen shots to s3, it needs to know the size of the file or it will fail,
I've added a away to do this in the init

packages/init/src/utils/validateRepositoryName.ts
Heyrows: Promise.reject will act as a throw, we'll be able to catch it later.
I kept the same behavior as there was before.

packages/client/src/utils/str.ts
Heyrows: I took what was done previously in SM.

packages/init/src/steps/choose-or-create-a-repository.ts
Heyrows: it's the Wroom application, so prismic.io

packages/init/src/steps/starters/slices.ts
Heyrows: That's a good point, i'mm remove the exit process.
I thought that if creating the ACL fails, it's most likely something we'd want to know about immediately to fix it.

You're right tho, it's part of the screenshots.

packages/init/src/utils/validateRepositoryName.ts
Heyrows: I mean in that case, we don't catch it so inquirer knows it's an error (tested it myself 😉 )

package.json
Heyrows: it was apparently used to validate a package.json file.

packages/client/src/utils/uploadScreenshot.ts
Heyrows: ### Stage Repository
image
image

### Production Repository
image
image

Heyrows: Here's the old version of the upload: https://github.com/prismicio/slice-machine/blob/master/packages/slice-machine/lib/models/common/http/upload.ts

As you can see, I still send data the same way, i don't think I modified anything (especially because I don't know about S3 contracts haha 😂 )

packages/init/src/utils/validateRepositoryName.ts
MarcMcIntosh: I've never seen that before, so I'm a bit suspicious 
```
if(isAvvailable === false) throw new Error(`${name} is already in use`
return true
```

Heyrows: hey @MarcMcIntosh , made a small test 
image

image

I put back the error message just in case but I believe that settle this 😄

packages/init/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@slicemachine/init",
-	"version": "1.1.10-alpha.1",
+	"version": "1.1.10-alpha.4",
```
sintinilorenzo: is it normal/okay that we change the version from `alpha.1` to `alpha.4`?

packages/init/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@slicemachine/init",
-  "version": "1.1.10-alpha.1",
+  "version": "1.1.10-alpha.4",
```
sintinilorenzo, line 3: same as above

packages/init/src/steps/display-final-message.ts
```diff
@@ -2,12 +2,30 @@ import { CONSTS } from "@slicemachine/core";
...
+export function displayFinalMessage(
+  cwd: string,
+  wasStarter: boolean,
+  reponame: string,
+  base: string
+): void {
   const yarnLock = NodeUtils.Files.exists(NodeUtils.YarnLockPath(cwd));
   const command = `${yarnLock ? "yarn" : "npm"} run ${CONSTS.SCRIPT_NAME}`;
-
   console.log();
```
sintinilorenzo: is this empty console.log needed?

packages/init/package-lock.json
Heyrows: yes :) it comes from the main, which means several versions have been deployed (ticket merged), so i need the latest versions to make sure it is incremented properly for deployment when this will be merged.
I should normally have the same version as the v0.4.0.

packages/init/src/steps/display-final-message.ts
Heyrows: Haha, that's what I asked Marc when I saw it 😂

sintinilorenzo (dismissed):

sintinilorenzo (approved):

chengoconnell (approved):

PR #477, opened 2022-06-03
Mm/init send slices

MarcMcIntosh:
## Types of changes

## Checklist:


e2e-projects/next/.slicemachine/assets/slices2/MySlice/index.stories.js
```diff
@@ -0,0 +1,9 @@
+import MyComponent from '../../../../slices2/MySlice';
```
Heyrows: Please don't commit on the `e2e-projects folder`.
We try to keep it unchanged.
Remove those changes please.

packages/init/src/steps/configure-project.ts
```diff
@@ -32,6 +32,13 @@ export async function configureProject(
 
     const manifestAlreadyExistWithContent = manifest.exists && manifest.content;
 
+    const libs =
+      manifest.content &&
+      manifest.content.libraries &&
+      manifest.content.libraries.length > 0
+        ? manifest.content.libraries
```
Heyrows, line 40: nice !

packages/init/src/steps/configure-project.ts
```diff
@@ -52,7 +59,10 @@ export async function configureProject(
     const pathToSlicesFolder = NodeUtils.CustomPaths(cwd)
       .library("slices")
       .value();
-    if (!NodeUtils.Files.exists(pathToSlicesFolder)) {
+    if (
+      !NodeUtils.Files.exists(pathToSlicesFolder) &&
+      libs.includes("@/slices")
```
Heyrows, line 65: nice too !

packages/init/src/steps/install-required-dependencies.ts
```diff
@@ -52,6 +48,29 @@ export async function installRequiredDependencies(
...
+  cwd: string,
+  framework: Models.Frameworks,
+  skipDependencies: boolean
+): Promise<void> {
+  const yarnLock = NodeUtils.Files.exists(NodeUtils.YarnLockPath(cwd));
+
+  const spinner = logs.spinner("Downloading Slice Machine");
+  spinner.start();
+
+  const stderr = await (skipDependencies
```
Heyrows, line 70: in which case would we need to skip the dependencies install ?

packages/init/src/index.ts
```diff
@@ -67,11 +68,15 @@ async function init() {
 
   Tracker.get().setRepository(repositoryDomainName);
 
-  // Install the required dependencies in the project.
-  await installRequiredDependencies(cwd, frameworkResult.value);
-
   const sliceLibPath = lib ? await installLib(cwd, lib, branch) : undefined;
 
+  const wasStarter = await sendStarterData(
```
Heyrows, line 73: I think this condition will evolve later,
As the thing that tell us if it's a starter or not is the presence of documents.

Heyrows: The existence of the SM.json allows us to find slices, but later for custom types and documents, we'll know where they are located.

packages/init/src/steps/sendStarterData.ts
```diff
@@ -0,0 +1,221 @@
...
+        const pathToScreenShot = screenshotPaths[variation.id];
+        if (pathToScreenShot && pathToScreenShot.path) {
+          const imageUrl = await sendToS3(
+            model.id,
+            variation.id,
+            pathToScreenShot.path
+          ).catch(() => "");
+          return {
+            ...variation,
+            ...(imageUrl ? { imageUrl } : {}), // is image url needed by the api?
```
Heyrows: ImageUrl is not needed by the API itself, it allows Wroom to display the screenshot when selecting a Slice.

packages/init/src/steps/sendStarterData.ts
```diff
@@ -0,0 +1,221 @@
...
+          },
+        })
+        .then((res) => {
+          if (Math.floor(res.status / 100) !== 2) {
+            console.log(
+              `[ERROR: ${res.status}] SENDING SLICE ${model.id} | ${res.statusText}`
+            );
+          }
+        })
+        .catch((err) => {
```
Heyrows: I used this documentation for making the client package https://axios-http.com/docs/handling_errors

When axios receives a > 4xx then it goes to `catch`. If not, then it goes into `then`
So you don't need to check the status as well in the then :)

packages/init/src/steps/sendStarterData.ts
```diff
@@ -0,0 +1,221 @@
...
+
+    const modelPromises = components.map(async ({ screenshotPaths, model }) => {
+      const variationsReq = model.variations.map(async (variation) => {
+        const pathToScreenShot = screenshotPaths[variation.id];
+        if (pathToScreenShot && pathToScreenShot.path) {
+          const imageUrl = await sendToS3(
+            model.id,
+            variation.id,
+            pathToScreenShot.path
+          ).catch(() => "");
```
Heyrows: There is a lot of stuff that can happens in this stage.
Did we think about adding logs ?
A debug mode eventually (if we don't want to display too many things) ?

packages/init/src/steps/sendStarterData.ts
```diff
@@ -0,0 +1,221 @@
...
+      console.error(`Error code: "${res.status}"`);
+
+      throw new Error(msg);
+    }
+
+    return s3ImageUrl;
+  };
+}
+
+export async function sendStarterData(
```
Heyrows, line 8: It's a huge file, not easy to read.

packages/init/src/steps/sendStarterData.ts
```diff
@@ -0,0 +1,221 @@
...
+          } else {
+            console.log(`[ERROR] SENDING SLICE ${String(err)}`);
+          }
+        });
+    });
+
+    await Promise.all(p);
+  }
+
+  spinner.succeed();
```
Heyrows: We don't stop the spinner on error case.
We also don't stop the APP if the push fails, is it normal ?

packages/init/src/steps/sendStarterData.ts
Heyrows: Some stuff will be simplified with the Client package 😄

packages/init/src/index.ts
```diff
@@ -82,6 +87,9 @@ async function init() {
     isTrackingAvailable
   );
 
+  // Install the required dependencies in the project.
+  await installRequiredDependencies(cwd, frameworkResult.value, wasStarter);
```
Heyrows, line 91: Should we skip dependencies if it is a starter ?
I believe if we do npm install, it will just install the current versions without updating.

I do see the gain in time, i'm not sure if there is a downside to it. (like a missing dependency on the starter).

packages/init/__tests__/sendStarterData.test.ts
```diff
@@ -0,0 +1,419 @@
+import { describe, test, jest, afterEach, expect } from "@jest/globals";
```
Heyrows, line 1: I think the code of starterData should be splitted into smaller function that can be each tested independently.

Testing a huge function here is good but it's way too hard to read and if something break it will be though to find.

Heyrows (commented):
Why did you needed a fake project in the stub ?

e2e-projects/next/.slicemachine/assets/slices2/MySlice/index.stories.js
MarcMcIntosh: Initially, I couldn't get mem-fs to work so I had to improvise,

Not to sure how new files in the e2e project got added :/

packages/init/src/steps/install-required-dependencies.ts
MarcMcIntosh: If there's a starter, then the deps are already in the package.json, so don't need to be added :)

packages/init/src/index.ts
MarcMcIntosh: We'll know for now, I was thinking about this over the weekend, and if we do the plugin architecture we'll need to load the plugins to find the custom-types and slices.....

packages/init/src/steps/sendStarterData.ts
MarcMcIntosh: good to know, if there's no file or s3 doesn't work for some reason :)

packages/init/src/steps/sendStarterData.ts
MarcMcIntosh: There could errors at different stages, when dealing with multiple slices, I'll have to see what I can do

packages/init/src/index.ts
MarcMcIntosh: It shouldn't need to add deps to the starter, the starter should already have them in there. also what would happen if the starter was using an older version of slice-machine

packages/init/__tests__/sendStarterData.test.ts
MarcMcIntosh: If it was going to be a module, used by other people then yes, I would test it independently most so the user could do the set up for the consumer of the function. Here though it's by product of the test requirements.

Ideally we would be testing the entry point, that way dead code is easy to find and eliminate.

e2e-projects/next/.slicemachine/assets/slices2/MySlice/index.stories.js
MarcMcIntosh: removed

packages/init/src/steps/sendStarterData.ts
MarcMcIntosh: I'll add some additional logging in the function return by setUpS3

MarcMcIntosh: Should be better now?

packages/init/src/steps/sendStarterData.ts
MarcMcIntosh: done :)

packages/init/__tests__/install-required-dependencies.test.ts
```diff
@@ -40,7 +40,7 @@ describe("install required dependency", () => {
     expect(spy).toHaveBeenCalled();
     expect(spy).toHaveBeenCalledWith(`yarn add -D ${CONSTS.SM_PACKAGE_NAME}`);
 
-    expect(stderr.output).toContain("Downloading Slice Machine");
```
Heyrows, line 43: `toContain()` also works with part of a string :)

packages/init/src/steps/sendStarterData.ts
```diff
@@ -78,23 +78,34 @@ async function setupS3(
...
+        if (res.status !== 204) {
+          console.log(errorMessage);
+          console.log(`Error code: "${res.status}"`);
+          return null;
+        } else {
+          return `${acl.imgixEndpoint}/${key}`;
+        }
+      })
+      .catch((err) => {
+        console.log(errorMessage);
```
Heyrows: forgotten console.log ?

packages/init/src/steps/sendStarterData.ts
```diff
@@ -78,23 +78,34 @@ async function setupS3(
...
+        if (axios.isAxiosError(err) && err.response) {
+          console.log(
```
Heyrows: we have helpers to write errors

packages/init/src/steps/sendStarterData.ts
```diff
@@ -78,23 +78,34 @@ async function setupS3(
...
-
-    return s3ImageUrl;
+    const errorMessage = `[slice/push] An error occurred while uploading preview images for ${sliceName}-${variationId} - please contact support`;
+
+    return axios
+      .post(acl.values.url, form, {
+        headers: form.getHeaders(),
+      })
+      .then((res) => {
+        if (res.status !== 204) {
```
Heyrows: you don't have to check the status, it's already a success.

packages/init/src/steps/sendStarterData.ts
```diff
@@ -197,18 +213,15 @@ export async function sendStarterData(
             repository,
           },
         })
-        .then((res) => {
-          if (Math.floor(res.status / 100) !== 2) {
+        .catch((err) => {
+          if (axios.isAxiosError(err) && err.response) {
             console.log(
```
Heyrows: Same: we do have helpers to write errors.

Heyrows (commented):
There are some changes on the E2E-project folder left to reverse.

I do not agree with testing `sendStarterData` as a whole.
- you are right when you say it's not re-used, but I don't think that should stop us from making smaller function and test them altogether.
- I think that approach would make easier the testing and detection / deletion of dead code.

It's not a stopper for me as we didn't had any test before 😅, but I (personally) wouldn't want that for any other code than the init.

packages/init/src/steps/sendStarterData.ts
MarcMcIntosh: It's the same as the code in slice-machine when sending the images through the acl service in
packages/slice-machine/server/src/api/services/uploadScreenshotClient.ts

I guess there are some 200 codes that can be errors

packages/init/src/steps/sendStarterData.ts
MarcMcIntosh: yeah about half of it is the acl and image stuff :/

packages/init/src/steps/sendStarterData.ts
MarcMcIntosh: yes

packages/init/src/steps/starters/s3.ts
```diff
@@ -0,0 +1,187 @@
...
+export type ALC = {
+  values: {
+    url: string;
+    fields: Record<string, string>;
+  };
+  imgixEndpoint: string;
+  err: null | string;
+};
+
+export async function createAcl(
```
Heyrows, line 25: That's really cool, because we will be able to just delete that part and replace it

packages/init/src/steps/starters/communication.ts
```diff
@@ -0,0 +1,78 @@
...
+      return Array.isArray(res.data) ? res.data.map((model) => model.id) : [];
+    });
+}
+
+async function sendModelToPrismic(
+  repository: string,
+  authorization: string,
+  customTypesApiEndpoint: string,
+  remoteSliceIds: Array<string>,
+  model: SliceSM
```
Heyrows, line 28: That part is weird, their shouldn't be anything about the SM or non SM model here.

Heyrows: I looked at the code and I believe the Core should not know about SM/toSM
Maybe that's a deeper issue than just this PR of course but I do think there is smething to work with here.

Heyrows (approved):
Thank you for the effort 👍 , that is much clearer than before.

PR #478, opened 2022-06-08
test(e2e): start dev server and slicemachine before running the test

MarcMcIntosh:
This will enable e2e testing of the simulator and other things that require the dev server to be running
## Types of changes

## Checklist:


package.json
```diff
@@ -15,8 +15,8 @@
     "pretty": "prettier --write .",
     "prettier:check": "prettier --check .",
     "test": "lerna run test",
-    "test:e2e": "start-server-and-test 'npm run slicemachine --prefix e2e-projects/next' http://localhost:9999 'cypress run'",
-    "test:e2e:dev": "start-server-and-test 'npm run slicemachine --prefix e2e-projects/next' http://localhost:9999 'cypress open'",
+    "test:e2e": "start-server-and-test 'npm run dev --prefix e2e-projects/next' http://localhost:3500 'npm run slicemachine --prefix e2e-projects/next' http://localhost:9999 'cypress run'",
```
Heyrows, line 18: Does this internally opens two webpages ?
How does this command work ?

e2e-projects/next/package.json
```diff
@@ -23,7 +23,8 @@
     "next": "^12.1.4",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
-    "recharts": "^2.1.9"
+    "recharts": "^2.1.9",
+    "styled-jsx": "^5.0.2"
```
Heyrows, line 27: The fact that styled-jsx wasn't used before maybe it's an issue when creating new slices, like the templates might not be right

cypress/integration/slices/00-create.spec.js
```diff
@@ -1,10 +1,11 @@
 describe("Create Slices", () => {
   const name = "TestSlice";
   const lib = "slices--ecommerce"; // name of the first lib of the next project.
+  const path = "e2e-projects/next/slices/ecommerce"; // path to th library
   beforeEach(() => {
     cy.clearLocalStorageSnapshot();
     cy.cleanSliceMachineUserContext();
-    cy.task("rmrf", `${lib}/${name}`);
+    cy.task("rmrf", `${path}/${name}`);
```
Heyrows, line 8: that's nice, our test are definitely creating way too many things.

package.json
MarcMcIntosh: https://github.com/bahmutov/start-server-and-test#starting-two-servers

only cypress opens the webpages

e2e-projects/next/package.json
MarcMcIntosh: Pretty much, but it causes a errors for thee dev server

cypress/integration/slices/00-create.spec.js
MarcMcIntosh: It'll still create them, this just cleans up the target first :)

package.json
Heyrows: What are the adresses you give as options ?

cypress/integration/slices/00-create.spec.js
Heyrows: shit, we should have something to clean up after a scenario, I think that should be standard

package.json
MarcMcIntosh: it's the address `start-server-and-test` checks to make sure the server started

cypress/integration/slices/00-create.spec.js
MarcMcIntosh: The cypress documentation says to clean up first and keep the data.... usually so it can be inspected later on, like if a bad model was made, I would like to keep it on disk so I can look at it

package.json
Heyrows: ah ok, thanks 👍

cypress/integration/slices/00-create.spec.js
Heyrows: Hmmmm i see. I guess at some point we might as well commit them then.

Heyrows (approved):

PR #486, opened 2022-06-20
Mm/init no docs

MarcMcIntosh:
Allow the user to skip sending the docs from a starter by passing a `--no-docs` flag
## Types of changes

## Checklist:

### DIFF
--- a/packages/init/src/index.ts
+++ b/packages/init/src/index.ts
@@ -22,6 +22,7 @@ async function init() {
   const isTrackingAvailable =
     findArgument(process.argv, "tracking") !== "false";
   const maybeRepositorySubdomain = findArgument(process.argv, "repository");
+  const sendDocs = findArgument(process.argv, "no-docs") ? true : false;
 
   Tracker.get().initialize(
     process.env.PUBLIC_SM_INIT_SEGMENT_KEY ||
@@ -74,6 +75,7 @@ async function init() {
     repositoryDomainName,
     config.base,
     config.cookies,
+    sendDocs,
     cwd
   ); // will be false if no sm.json is found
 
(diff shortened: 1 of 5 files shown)


packages/init/src/index.ts
```diff
@@ -22,6 +22,7 @@ async function init() {
   const isTrackingAvailable =
     findArgument(process.argv, "tracking") !== "false";
   const maybeRepositorySubdomain = findArgument(process.argv, "repository");
+  const sendDocs = findArgument(process.argv, "no-docs") ? true : false;
```
Heyrows: I don't think that would work, did you tested it ?
If I look at the findArgument function it expects a value after that.

Maybe we could make another helper like "findFlag"

chengoconnell (approved):

packages/init/src/steps/sendStarterData.ts
```diff
@@ -31,5 +34,10 @@ export async function sendStarterData(
 
   await sendCustomTypesFromStarter(repository, authTokenFromCookie, base, cwd);
 
+  if (sendDocs === false) {
```
Heyrows, line 37: If the starter has documents but we don't want to send them, i'm guessing they have another use, do we want to delete them ?

From a user point of view if he/she puts the flag, I would expect my folder to be untouched i think.

MarcMcIntosh: I asked come about this yesterday and we want to delete them

packages/init/src/index.ts
MarcMcIntosh: I'll take a look, and nope I've not had the chance to run because of all the setup required. good catch though :)

MarcMcIntosh: I've run it now and it deletes the documents, just fine :)

packages/init/src/steps/sendStarterData.ts
Heyrows: Ok :)

Heyrows (approved):

PR #491, opened 2022-06-23
Ls/661 rename custom type

sintinilorenzo:
image
image

## Types of changes

## Checklist:


packages/slice-machine/components/Forms/RenameCustomTypeModal/RenameCustomTypeModal.tsx
```diff
@@ -0,0 +1,62 @@
...
+  customTypeId,
+}) => {
+  const { renameCustomType, closeRenameCustomTypeModal } =
+    useSliceMachineActions();
+
+  const handleOnSubmit = (values: { customTypeName: string }) => {
+    renameCustomType(customTypeId, values.customTypeName);
+    closeRenameCustomTypeModal();
+  };
+  const { isCreateCustomTypeModalOpen } = useSelector(
```
chengoconnell: Should this be `isRenameCustomTypeModalOpen`?

cypress/integration/customTypes/00-create.spec.js
```diff
@@ -1,24 +1,43 @@
...
+  let id = "my-test";
+
   beforeEach(() => {
+    const ID = Math.floor(10000 + Math.random() * 90000);
+    name = `My Test ${ID}`;
+    id = `my-test-${ID}`;
+
     cy.clearLocalStorageSnapshot();
     cy.cleanSliceMachineUserContext();
     cy.task("rmrf", `customtypes/${id}`);
```
MarcMcIntosh: I'd recommend checking `rmrf` works, as last time I used it it required passing the path to the project too

packages/slice-machine/components/Forms/RenameCustomTypeModal/RenameCustomTypeModal.tsx
sintinilorenzo: yes it is, but have already changed it in a later commit.

packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/index.tsx
```diff
@@ -1,3 +1,3 @@
-export { default as Header } from "./Header";
+export { default as Header } from "./Header/index";
```
MarcMcIntosh: is index needed in the path?

chengoconnell (approved):

packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header/primaryButton.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const { hasPendingModifications, customTypeStatus } = useSelector(
+    (store: SliceMachineStoreType) => ({
+      hasPendingModifications:
+        selectIsCurrentCustomTypeHasPendingModifications(store),
+      customTypeStatus: selectCustomTypeStatus(store),
+    })
+  );
+  const { saveCustomType, pushCustomType } = useSliceMachineActions();
+
+  const buttonProps = (() => {
```
MarcMcIntosh: I've not seen that being done before, a function to make the props and children for an element

packages/slice-machine/server/src/api/custom-types/rename.ts
```diff
@@ -0,0 +1,16 @@
+import getEnv from "../services/getEnv";
+import { CustomTypesPaths } from "@lib/models/paths";
+
+import { RenameCustomTypeBody } from "@lib/models/common/CustomType";
+import * as IO from "../io";
+
+export default async function handler(req: { body: RenameCustomTypeBody }) {
```
MarcMcIntosh: I'll need to check if these can throw

packages/slice-machine/server/src/api/index.ts
```diff
@@ -188,6 +189,24 @@ router.post(
   }
 );
 
+router.patch(
+  "/custom-types/rename",
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-misused-promises
```
MarcMcIntosh: does eslint have to be disabled here?

packages/slice-machine/server/src/api/io/CustomType.ts
```diff
@@ -13,3 +13,9 @@ export function readCustomType(path: string): CustomTypeSM {
 export function writeCustomType(path: string, customType: CustomTypeSM) {
   Files.write(path, CustomTypes.fromSM(customType));
 }
+
+export function renameCustomType(path: string, newCustomTypeName: string) {
```
MarcMcIntosh, line 17: trying to think if it's worth doing run time type verification here, or not :/
if so the two functions above show how it can be done :)

packages/slice-machine/src/apiClient.ts
```diff
@@ -35,6 +38,22 @@ export const saveCustomType = (
...
+export const renameCustomType = (
+  customTypeId: string,
+  newCustomTypeName: string
+): Promise<AxiosResponse> => {
+  const requestBody: RenameCustomTypeBody = {
+    customTypeId,
+    newCustomTypeName,
+  };
+
+  return axios.patch(
```
MarcMcIntosh, line 50: I think this is the first time I've seen some-one use `patch`

packages/slice-machine/src/modules/availableCustomTypes/index.ts
```diff
@@ -98,6 +114,20 @@ export const availableCustomTypesReducer: Reducer<
         ...normalizedNewCustomType,
       };
     }
+    case getType(renameCustomTypeCreator.success): {
+      const id = action.payload.customTypeId;
+      const newName = action.payload.newCustomTypeName;
+
+      const newCustomType = {
+        ...state[id],
+        local: { ...state[id].local, label: newName },
```
MarcMcIntosh, line 123: no typescript warnings here ?... wired

cypress.json
```diff
@@ -3,5 +3,6 @@
   "viewportWidth": 1920,
   "viewportHeight": 1080,
   "defaultCommandTimeout": 10000,
-  "retries": 1
+  "retries": 1,
+  "video": false
```
MarcMcIntosh, line 7: hopefully that speeds things up a bit :)

cypress/integration/customTypes/00-create.spec.js
```diff
@@ -1,24 +1,43 @@
 describe("Custom Types specs", () => {
-  const name = "My Test";
-  const id = "my-test";
+  let name = "My Test";
+  let id = "my-test";
+
   beforeEach(() => {
+    const ID = Math.floor(10000 + Math.random() * 90000);
```
sintinilorenzo: this is so that the test always creates a CT with a dynamic name. The issue with it being the same (e.g. always "My Test") is that if it creates it but then fails just after, it will always fail on any subsequent re-tries (because you will not be able to create 2 CT with the same name)

packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/index.tsx
sintinilorenzo: good point it's not. will remove :)

packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header/primaryButton.tsx
sintinilorenzo: this was already like that, all I did was copy paste it into another file since the original one was getting too big. Can refactor to change it tho

packages/slice-machine/src/apiClient.ts
sintinilorenzo: we are only changing part of the element (only the name) so I guess it's a `patch` rather than a `put` yeah?

cypress/integration/customTypes/00-create.spec.js
sintinilorenzo: ❓❓❓ what do you mean by that? What is `rmrf` and how can i check that it works?

packages/slice-machine/server/src/api/index.ts
sintinilorenzo: nope good point, will remove (other than then unsed-promise which seems to be there on every single other route)

cypress.json
sintinilorenzo: seems like it did!

packages/slice-machine/src/modules/availableCustomTypes/index.ts
sintinilorenzo: nope

packages/slice-machine/server/src/api/io/CustomType.ts
sintinilorenzo: ❓❓❓ don't think I understand what do you mean

packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header/primaryButton.tsx
sintinilorenzo: updated

packages/slice-machine/server/src/api/io/CustomType.ts
sintinilorenzo: changed

cypress/integration/customTypes/00-create.spec.js
sintinilorenzo: changed

cypress/integration/customTypes/00-create.spec.js
MarcMcIntosh: dos deleting it first not work?

PR #493, opened 2022-06-24
remove screenshot taking logic when saving a slice

sintinilorenzo:
## Types of changes

## Checklist:

### DIFF
--- a/packages/slice-machine/server/src/api/slices/save.ts
+++ b/packages/slice-machine/server/src/api/slices/save.ts
@@ -8,18 +8,17 @@ import mock from "@lib/mock/Slice";
 import { getConfig, insert as insertMockConfig } from "@lib/mock/misc/fs";
 import Files from "@lib/utils/files";
 import { SliceMockConfig } from "@lib/models/common/MockConfig";
-import { generateScreenshot } from "../screenshots/generate";
 import { BackendEnvironment } from "@lib/models/common/Environment";
 
 import onSaveSlice from "../common/hooks/onSaveSlice";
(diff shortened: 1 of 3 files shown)


packages/slice-machine/server/src/api/slices/save.ts
```diff
@@ -8,18 +8,17 @@ import mock from "@lib/mock/Slice";
...
 import onBeforeSaveSlice from "../common/hooks/onBeforeSaveSlice";
-import { SliceSaveBody, SliceSaveResponse } from "@lib/models/common/Slice";
+import { SliceSaveBody } from "@lib/models/common/Slice";
 import * as IO from "../io";
 
 export async function handler(
   env: BackendEnvironment,
   { sliceName, from, model: smModel, mockConfig }: SliceSaveBody
-): Promise<SliceSaveResponse> {
+): Promise<Record<string, never>> {
```
MarcMcIntosh, line 21: should this just return void?

MarcMcIntosh (approved):

packages/slice-machine/server/src/api/slices/save.ts
sintinilorenzo: I had void at first but then I would have had to refactor a lot of stuff because in many places fetchApi calls expect a return and have `await response = ....` so I kept if as an empty object for simplicity

MarcMcIntosh: how many places?

chengoconnell (approved):

PR #472, opened 2022-05-19
[WIP] hooks being called a different number of times between renders

MarcMcIntosh:
This happens when a user creates a slice. 

I've add a comment where this happens

## Types of changes

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. (PR: #)
- [ ] All new and existing tests are passing.

PR #505, opened 2022-07-04
Ls/663 remove the library and slice context

sintinilorenzo:
## Types of changes

## Checklist:


packages/slice-machine/src/models/slice/context.tsx
```diff
@@ -45,37 +32,16 @@ export function useModelReducer({
...
-  const [Model, store] = value;
-  return (
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-assignment
-    <SliceContext.Provider value={{ Model, store, variation }}>
-      {typeof children === "function" ? children(value) : children}
-    </SliceContext.Provider>
-  );
-};
 // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-explicit-any
 export const SliceHandler = ({ children }: { children: any }) => {
```
MarcMcIntosh: What's the correct type signature to use here?

is it `const SliceHandel: React.FC = ....` or `({ children }: { children: React.ReactNode })`

packages/slice-machine/src/models/slice/context.tsx
```diff
@@ -126,14 +92,9 @@ export const SliceHandler = ({ children }: { children: any }) => {
...
+    void router.replace(`/${lib.name}/${slice.model.name}/${variation.id}`);
   }
 
-  return (
-    <SliceProvider value={slice} variation={variation}>
-      {children}
-    </SliceProvider>
-  );
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+  return <>{typeof children === "function" ? children(slice) : children}</>;
```
MarcMcIntosh, line 98: this is odd, looks like a HOC

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -116,22 +151,22 @@ const SliceBuilder: React.FunctionComponent = () => {
...
-              store
-                .variation(variation.id)
-                .generateCustomScreenShot(
-                  Model.from,
-                  Model.model.name,
-                  setData,
-                  file
-                )
-            }
+            onHandleFile={(file: Blob) => {
```
MarcMcIntosh: nice

packages/slice-machine/src/modules/selectedSlice/push.ts
```diff
@@ -0,0 +1,18 @@
+import { fetchApi } from "@lib/builders/common/fetch";
+import SliceState from "@lib/models/ui/SliceState";
+
+export default function push(
+  slice: SliceState,
+  setData: (data: any) => void,
+  callback: () => void
+) {
+  // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
MarcMcIntosh: Would returning the promise or use await get rid of the `no-floating-promises`?

packages/slice-machine/src/modules/slices/index.ts
```diff
@@ -19,16 +19,25 @@ import { LibraryUI } from "@models/common/LibraryUI";
 import { SliceSM } from "@slicemachine/core/build/models";
 import Tracker from "../../../src/tracker";
 import { openToasterCreator, ToasterType } from "@src/modules/toaster";
+import LibraryState from "@lib/models/ui/LibraryState";
+import { useModelReducer } from "@src/models/slice/context";
+import { SliceMockConfig } from "@lib/models/common/MockConfig";
+import { push } from "connected-next-router";
```
MarcMcIntosh: Pass the Router to the connected router component then you can use next router as normal and keep the two in sync :)
https://github.com/prismicio/slice-machine/pull/472/files#diff-af0d93898897cf163cba7f962dd24f74a4a562e132bfccaf158b33ec02aa535fR114

packages/slice-machine/src/modules/selectedSlice/screenshot.ts
```diff
@@ -0,0 +1,61 @@
...
+export function generateScreenShot(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  _variationId: string,
+  libraryName: string,
+  sliceName: string,
+  setData: (data: any) => void,
+  callback: (screenshots: Screenshots) => void
+) {
+  // eslint-disable-next-line @typescript-eslint/no-floating-promises
+  fetchApi({
```
MarcMcIntosh: return, await or void if you don't want to wait for the result :)

packages/slice-machine/src/modules/selectedSlice/screenshot.ts
```diff
@@ -0,0 +1,61 @@
+import { fetchApi } from "@lib/builders/common/fetch";
+import { ScreenshotUI } from "@lib/models/common/ComponentUI";
+import { ScreenshotResponse, Screenshots } from "@models/common/Screenshots";
+
+export function generateScreenShot(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  _variationId: string,
+  libraryName: string,
+  sliceName: string,
+  setData: (data: any) => void,
```
MarcMcIntosh, line 10: what does set data do?

packages/slice-machine/src/modules/selectedSlice/screenshot.ts
```diff
@@ -0,0 +1,61 @@
...
+  callback: (variationId: string, screenshots: ScreenshotUI) => void
+) {
+  const form = new FormData();
+  form.append("file", file);
+  form.append("libraryName", libraryName);
+  form.append("sliceName", sliceName);
+  form.append("variationId", variationId);
+
+  // eslint-disable-next-line @typescript-eslint/no-floating-promises
+  fetchApi({
```
MarcMcIntosh: void, or return ?

packages/slice-machine/src/modules/selectedSlice/save.ts
```diff
@@ -0,0 +1,35 @@
+import { fetchApi } from "@lib/builders/common/fetch";
+import SliceState from "@lib/models/ui/SliceState";
+
+export default function save(
+  slice: SliceState,
+  setData: (data: any) => void,
+  callback: (SliceState: SliceState) => void
+) {
+  // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
MarcMcIntosh: void or return?

packages/slice-machine/src/modules/selectedSlice/reducer.ts
```diff
@@ -0,0 +1,273 @@
...
+    !equal(model.initialVariations, model.variations) ||
+    !equal(model.initialMockConfig, model.mockConfig);
+  let status = model.__status;
+
+  const isScreenshotModified = !equal(
+    model.screenshotUrls,
+    model.initialScreenshotUrls
+  );
+
+  // Should be the slice be pushed ?
```
MarcMcIntosh: Pushed? 

When it's touched i think that means it can be pushed, we really should think about the naming schema here as it could mean a few things

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -146,4 +181,28 @@ const SliceBuilder: React.FunctionComponent = () => {
   );
 };
 
-export default SliceBuilder;
+const SliceBuilderWithRouter = () => {
+  const router = useRouter();
```
MarcMcIntosh, line 172: defiantly pass the next router to connected router to keep them in sync :)

packages/slice-machine/components/Simulator/index.tsx
```diff
@@ -16,12 +13,26 @@ import {
...
 import { SliceMachineStoreType } from "@src/redux/type";
+import { useRouter } from "next/router";
+import { selectCurrentSlice } from "@src/modules/selectedSlice/selectors";
 
 export type SliceView = SliceViewItem[];
 export type SliceViewItem = Readonly<{ sliceID: string; variationID: string }>;
 
 export default function Simulator() {
-  const { Model, variation } = useContext(SliceContext);
+  const router = useRouter();
```
MarcMcIntosh: remember to pass Router to the connected router to keep both in sync :)
https://github.com/prismicio/slice-machine/pull/472/files#diff-af0d93898897cf163cba7f962dd24f74a4a562e132bfccaf158b33ec02aa535fR114

MarcMcIntosh (commented):
Good work, this is a prety big PR that touches a lot of old code that we rarely check.

Slice-builder, Custom-type builder, Simulator and review modal

Do we have any tests on the simulator?

packages/slice-machine/src/models/slice/context.tsx
sintinilorenzo: yah good one. I think I prefer `const SliceHandel: React.FC = .... ` out of the 2 since that way I don't have to specifically type props manually which could be wrong.

packages/slice-machine/src/models/slice/context.tsx
sintinilorenzo: you can't return `children` as it is since it's a `React.ReactNode`, while a react component expects a `ReactElement`. Placing it within a fragment fixes it since it will make it return an element (https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33006)

same thing is done elsewhere in the app I see
image

packages/slice-machine/src/modules/selectedSlice/push.ts
sintinilorenzo: yes it does :)

packages/slice-machine/src/modules/selectedSlice/screenshot.ts
sintinilorenzo: yes, changed here and for the other fetchApi calls

packages/slice-machine/src/modules/selectedSlice/screenshot.ts
sintinilorenzo: it's the function that sets the state to determine the state of the call. 
image
image

packages/slice-machine/src/modules/selectedSlice/screenshot.ts
sintinilorenzo: same as other fixes

packages/slice-machine/src/modules/selectedSlice/save.ts
sintinilorenzo: same as other fixes

packages/slice-machine/src/modules/selectedSlice/reducer.ts
sintinilorenzo: Yhink that's correct. We are checking whether our local variations match the remote ones, if not it means we have some new ones that need to be pushed. Changed the comment to be a bit more explicit and not contain info on how the variable is used, simply to specify what it means

packages/slice-machine/src/modules/selectedSlice/screenshot.ts
MarcMcIntosh: oh, so it would take a `Partial<SlicebuilderState>` ?

packages/slice-machine/components/Simulator/index.tsx
sintinilorenzo: changed!

packages/slice-machine/src/modules/slices/index.ts
sintinilorenzo: changed!

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
sintinilorenzo: changed!

packages/slice-machine/components/Simulator/index.tsx
```diff
@@ -16,12 +13,24 @@ import {
...
 
 export type SliceView = SliceViewItem[];
 export type SliceViewItem = Readonly<{ sliceID: string; variationID: string }>;
 
 export default function Simulator() {
-  const { Model, variation } = useContext(SliceContext);
+  const { Model } = useSelector((store: SliceMachineStoreType) => ({
+    Model: selectCurrentSlice(
+      store,
+      Router.router?.query.lib as string,
```
MarcMcIntosh, line 26: Isn't the routers state kept in redux now?

packages/slice-machine/pages/_app.tsx
```diff
@@ -110,7 +111,7 @@ function MyApp({ Component, pageProps }: AppContext & AppInitialProps) {
               <LoadingPage />
             ) : (
               <Provider store={smStore.store}>
-                <ConnectedRouter>
+                <ConnectedRouter Router={Router}>
```
MarcMcIntosh, line 114: nice

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -74,38 +106,42 @@ const SliceBuilder: React.FunctionComponent = () => {
...
-    () => [{ sliceID: Model.model.id, variationID: variation.id }],
-    [Model.model.id, variation.id]
+    () =>
+      Model && variation
+        ? [{ sliceID: Model.model.id, variationID: variation.id }]
+        : null,
+    [Model?.model.id, variation?.id]
   );
 
+  if (!Model || !variation || !sliceView) return null;
```
MarcMcIntosh: how come this has moved down ?

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -74,38 +106,42 @@ const SliceBuilder: React.FunctionComponent = () => {
...
         Model={Model}
-        store={store}
         variation={variation}
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-expect-error
-        onPush={() => void store.push(Model, onPush)}
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-expect-error
-        onSave={() => void store.save(Model, setData)}
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment, @typescript-eslint/no-misused-promises
```
MarcMcIntosh, line 122: no missused promises?
what's that about?

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
MarcMcIntosh: how much of the router state is in redux?

packages/slice-machine/src/models/slice/context.tsx
```diff
@@ -126,14 +90,8 @@ export const SliceHandler = ({ children }: { children: any }) => {
 
   // variation not in the URL but a default variation was found
   if (!variationParam) {
-    void router.replace(`/${lib.name}/${slice[0].model.name}/${variation.id}`);
+    void router.replace(`/${lib.name}/${slice.model.name}/${variation.id}`);
```
MarcMcIntosh: connect router provides replace :)

packages/slice-machine/src/models/slice/context.tsx
```diff
@@ -45,41 +32,18 @@ export function useModelReducer({
...
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-assignment
-    <SliceContext.Provider value={{ Model, store, variation }}>
-      {typeof children === "function" ? children(value) : children}
-    </SliceContext.Provider>
-  );
-};
-// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-explicit-any
-export const SliceHandler = ({ children }: { children: any }) => {
+export const SliceHandler: React.FC = ({ children }) => {
   const router = useRouter();
```
MarcMcIntosh: how much of router data is in in the redux state?

packages/slice-machine/src/models/slice/context.tsx
MarcMcIntosh: this part `children(slice)` looks like a HOC :)

packages/slice-machine/src/modules/selectedSlice/push.ts
```diff
@@ -0,0 +1,17 @@
+import { fetchApi } from "@lib/builders/common/fetch";
+import SliceState from "@lib/models/ui/SliceState";
+
+export default async function push(
```
MarcMcIntosh: Probably want to give this a very specific name like `pushSlice` so we (IDE's) don't get it confused with all the methods named `push`

packages/slice-machine/src/modules/selectedSlice/reducer.ts
```diff
@@ -0,0 +1,273 @@
...
+        if (variation.id === variationId) {
+          return {
+            ...acc,
+            [variationId]: screenshot,
+          };
+        }
+        if (prevModel.screenshotUrls?.[variation.id]) {
+          return {
+            ...acc,
+            [variation.id]: prevModel.screenshotUrls?.[variation.id],
```
MarcMcIntosh: does this need the the `?` since it's already checked or ?

packages/slice-machine/src/modules/selectedSlice/save.ts
```diff
@@ -0,0 +1,34 @@
+import { fetchApi } from "@lib/builders/common/fetch";
+import SliceState from "@lib/models/ui/SliceState";
+
+export default async function save(
```
MarcMcIntosh: can we give this a more descriptive name?

packages/slice-machine/src/modules/selectedSlice/reducer.ts
MarcMcIntosh: nice

packages/slice-machine/src/modules/slices/index.ts
```diff
@@ -122,11 +135,24 @@ export function* createSliceSaga({
     name: payload.sliceName,
     library: payload.libName,
   });
+  const { data: serverState } = (yield call(getState)) as SagaReturnType<
```
chengoconnell, line 139: what does this do? :D

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
sintinilorenzo: image
all the information about the location

packages/slice-machine/components/Simulator/index.tsx
sintinilorenzo: yes but only info about the `location` which would just be a string, this means it becomes hard to query for the lib or sliceName within the url.

Also the location is not always kept in redux since the redirection when clicking on a slice or custom-type is not done through `connected-next-router`. But think that's outside of this ticket scope since this was to refactor library and slice context

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
sintinilorenzo: if it's above it means that the function could return early and the hooks underneath would be conditionally called or not called. This is against the hooks rules since in React the hooks must always be called the same number of times and in the same order https://reactjs.org/docs/hooks-rules.html, https://bobbyhadz.com/blog/react-hook-usestate-called-conditionally#:~:text=The%20error%20%22React%20hook%20'useState,of%20how%20the%20error%20occurs.

packages/slice-machine/src/models/slice/context.tsx
sintinilorenzo: see other answer

packages/slice-machine/src/modules/slices/index.ts
sintinilorenzo: calls the state from the backend to re-update the library redux values

packages/slice-machine/src/models/slice/context.tsx
sintinilorenzo: ah yes, will change here and in the other places within the file!

packages/slice-machine/src/modules/selectedSlice/reducer.ts
sintinilorenzo: good catch, will remove

packages/slice-machine/src/modules/selectedSlice/save.ts
sintinilorenzo: yep, called it saveSliceApiCall to make it clear this is a backend POST request, and not an sliceMachineAction or redux action creator

packages/slice-machine/src/modules/selectedSlice/push.ts
sintinilorenzo: yep, called it pushSliceApiCall to make it clear this is a backend POST request, and not an sliceMachineAction or redux action creator

packages/slice-machine/src/models/slice/context.tsx
sintinilorenzo: think it's because a ReactNode could be a function that tale slice as param? I am not too sure tbh, it was already there and didn't wanna edit something I don't fully understand

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -35,20 +43,44 @@ const initialState: SliceBuilderState = {
...
+}
 
-  const { openLoginModal, checkSimulatorSetup, openToaster } =
-    useSliceMachineActions();
+const SliceBuilder: React.FC<SliceBuilderProps> = ({
+  initModel,
+  initVariationId,
+}) => {
+  useEffect(() => {
+    initSliceStore(initModel);
```
Heyrows: OPT: This should be done on the `page` level like it is done for the custom type.
That way the component here remains purely display, we could also change the redirection logic place.

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -35,20 +43,44 @@ const initialState: SliceBuilderState = {
...
+  }, []);
+
+  const { Model } = useSelector((state: SliceMachineStoreType) => ({
+    Model: state.selectedSlice?.Model,
+  }));
+
+  const variation = Model?.variations.find(
```
Heyrows: I'm starting to wonder if this should be a slice Store or a variation store.
The page and changes makes are all about a variation, but when we save the whole slice is saved.
Could be a good discussion to have. 😄
On the same topic as the one your brought for taking screenshot of a single variation.

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
Heyrows: this should go up right after the definition of model and variation, so that we know it's define after :)

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -74,38 +106,42 @@ const SliceBuilder: React.FunctionComponent = () => {
...
   const onTakingCustomScreenshot = () => {
-    checkSimulatorSetup(true, () =>
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
-      store
-        .variation(variation.id)
-        .generateScreenShot(Model.from, Model.model.name, setData)
-    );
+    // eslint-disable-next-line @typescript-eslint/no-misused-promises
+    checkSimulatorSetup(true, async () => {
+      await generateScreenShot(
```
Heyrows, line 107: I think this should be in a saga.
You've already done an amazing work so i'm not asking you to do it in this ticket, but i think those could be listed as another tech debt ticket to be taken care of later.

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -74,38 +106,42 @@ const SliceBuilder: React.FunctionComponent = () => {
...
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-expect-error
-        onSave={() => void store.save(Model, setData)}
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment, @typescript-eslint/no-misused-promises
+        onPush={async () => {
+          await push(Model, onPush, () => pushSlice());
+        }}
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment, @typescript-eslint/no-misused-promises
+        onSave={async () => {
+          await save(Model, setData, (state) => saveSlice(state));
```
Heyrows: same -> saga, tech debt

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -116,22 +152,23 @@ const SliceBuilder: React.FunctionComponent = () => {
...
-                .generateCustomScreenShot(
-                  Model.from,
-                  Model.model.name,
-                  setData,
-                  file
-                )
-            }
+            // eslint-disable-next-line @typescript-eslint/no-misused-promises
+            onHandleFile={async (file: Blob) => {
+              await generateCustomScreenShot(
```
Heyrows, line 142: same -> saga, tech debt

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -35,20 +43,44 @@ const initialState: SliceBuilderState = {
...
-    useSliceMachineActions();
+const SliceBuilder: React.FC<SliceBuilderProps> = ({
+  initModel,
+  initVariationId,
+}) => {
+  useEffect(() => {
+    initSliceStore(initModel);
+  }, []);
+
+  const { Model } = useSelector((state: SliceMachineStoreType) => ({
```
Heyrows: I just saw the `SliceBuilderWithRouter` I think we don't need to retrieve the model here as it's the same in the props.

And if the store changes, I believe the whole `SliceBuilderWithRouter` will re-render and so re-render this one.

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
sintinilorenzo: we return a `Promise<void>` rather than a `void`. Annoying to change however since we'd have to type everything as a `Promise<void>` instead, and as some point this type is passed to the onClick prop of a theme-ui component, whose signature we can't change

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -146,4 +183,28 @@ const SliceBuilder: React.FunctionComponent = () => {
...
+  }));
+
+  if (!Model) {
+    void router.replace("/");
+    return null;
+  }
+
+  return (
+    <SliceBuilder
+      initModel={Model}
```
Heyrows: I wonder when do we split slice data and variation data. (slice data for like displaying the title and variation data for displaying fields for instance.

If it's even doable.

packages/slice-machine/pages/_app.tsx
Heyrows, line 114: why did we needed that ?

packages/slice-machine/pages/slices.tsx
```diff
@@ -42,17 +44,22 @@ const CreateSliceButton = ({
...
-      isCreatingSlice: isLoading(store, LoadingKeysEnum.CREATE_SLICE),
-      localLibs: getLibraries(store),
-      remoteLibs: getRemoteSlices(store),
-    }));
+  const {
+    isCreateSliceModalOpen,
+    isCreatingSlice,
+    localLibs,
+    remoteLibs,
+    libraries,
```
Heyrows: - Duplicated data between `libraries` and `localLibs`
- Can you rename remoteLibs into remoteSlices.
  - I know it's not you, it's just wrong 😓

packages/slice-machine/components/Simulator/index.tsx
MarcMcIntosh: good to know :)

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
MarcMcIntosh: ah yes that old chestnut

packages/slice-machine/src/models/slice/context.tsx
MarcMcIntosh: that's understandable :)

MarcMcIntosh (approved):

packages/slice-machine/src/models/slice/context.tsx
```diff
@@ -30,7 +17,7 @@ export function useModelReducer({
   remoteSlice?: SliceSM;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   mockConfig: any;
-}): [SliceState, SliceStore] {
+}): SliceState {
```
Heyrows, line 21: I think this and those types 'SliceState' and LibraryState add an extra layer of complexity.
Hope that at some point we can get rid of it and do something a little simpler like the custom type stores.

packages/slice-machine/src/models/slice/context.tsx
Heyrows: funny, never this issue before.

packages/slice-machine/src/modules/selectedSlice/screenshot.ts
MarcMcIntosh: I though `any` was band by the linter?

packages/slice-machine/src/models/slice/context.tsx
Heyrows: Do you know why it suddenly triggered an issue ?

packages/slice-machine/src/modules/selectedSlice/push.ts
```diff
@@ -0,0 +1,17 @@
...
+  slice: SliceState,
+  setData: (data: any) => void,
+  callback: () => void
+) {
+  await fetchApi({
```
Heyrows, line 9: same -> saga, tech debt

packages/slice-machine/src/modules/selectedSlice/reducer.ts
```diff
@@ -0,0 +1,273 @@
...
+        Model: updateTouchedAndStatus(newModel),
+        variation: newVariation,
+      };
+    }
+    default:
+      return prevState;
+  }
+};
+
+const updateTouchedAndStatus = (model: SliceState) => {
```
Heyrows, line 247: We do have some tickets for the Custom type store to take remove the "initialModel" and use the custom type list.
I think we should do the same here in another ticket.

packages/slice-machine/src/modules/selectedSlice/reducer.ts
```diff
@@ -0,0 +1,273 @@
...
+  );
+
+  // True if the remote and local slice variations don't match
+  const isModelModified = !compareVariations(
+    model.remoteVariations,
+    model.initialVariations
+  );
+
+  if (isModelModified || isScreenshotModified) {
+    status = LibStatus.Modified;
```
Heyrows, line 268: If it's a new Slice and there is nothing on remote, wouldn't `isModelModfied` be true and the status displayed be `Modified` and not `New` ?

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -35,20 +43,44 @@ const initialState: SliceBuilderState = {
...
     (state: SliceMachineStoreType) => ({
       simulatorUrl: selectSimulatorUrl(state),
       isWaitingForIframeCheck: selectIsWaitingForIFrameCheck(state),
     })
   );
 
-  if (!store || !Model || !variation) return null;
-
   // We need to move this state to somewhere global to update the UI if any action from anywhere save or update to the filesystem I'd guess
   const [data, setData] = useState<SliceBuilderState>(initialState);
```
Heyrows, line 70: Shouldn't this be useless now that we have the store ?

packages/slice-machine/src/modules/selectedSlice/types.ts
```diff
@@ -0,0 +1,5 @@
+import SliceState from "@lib/models/ui/SliceState";
+
+export type SelectedSliceStoreType = {
+  Model: SliceState;
```
Heyrows, line 4: I think we should have more keys here, the SliceState type should not exist anymore and should be spread here.

Heyrows: Now if that's too much work you tell me.

packages/slice-machine/src/modules/slices/index.ts
```diff
@@ -91,11 +136,22 @@ export function* createSliceSaga({
...
+    typeof getState
+  >;
+  yield put(createSliceCreator.success({ libraries: serverState.libraries }));
   yield put(modalCloseCreator({ modalKey: ModalKeysEnum.CREATE_SLICE }));
-  // changing this to use next/router  call(Router.router.push,...) causes bugs in some contexts.
-  window.location.href = `/${payload.libName.replace(/\//g, "--")}/${
+  const addr = `/${payload.libName.replace(/\//g, "--")}/${
     payload.sliceName
   }/${variationId}`;
+  yield put(replace("/[lib]/[sliceName]/[variation]", addr));
```
Heyrows: does the replace means changing the URL and the `LOCATION_CHANGE` means triggering re-render as if the url actually changed ?

Heyrows (commented):
It's already amazing everything you did in 2 days, well done !
Left quite some questions 😅

chengoconnell (approved):

packages/slice-machine/src/models/slice/context.tsx
sintinilorenzo: don't think it triggered any issues, think it's just something a bit odd that both me and Marc don't seem to be too sure so it raised a question when I moved it and it showed up in the git diff

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
sintinilorenzo: hmm yeah I see. The issue I see on it maybe is that if you are within a variation and make changes, then go to another variation, you probably don't wanna lose your unsaved changes in the other. So we'd have to see how we can architecture it in a way that allows for that

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
sintinilorenzo: replied to a similar comment by Marc. I specifically moved it down since no hook should be conditionally rendered. 
-----
if it's above it means that the function could return early and the hooks underneath would be conditionally called or not called. This is against the hooks rules since in React the hooks must always be called the same number of times and in the same order https://reactjs.org/docs/hooks-rules.html, https://bobbyhadz.com/blog/react-hook-usestate-called-conditionally#:~:text=The%20error%20%22React%20hook%20'useState,of%20how%20the%20error%20occurs.

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
sintinilorenzo: yes agree! 
image
I mentionned it in the ticket outcome. Should I create a ticket? Where do you guys keep track of tech debt?

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
sintinilorenzo: same as before

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
sintinilorenzo: same as before

packages/slice-machine/pages/_app.tsx
sintinilorenzo: it was a suggestion from Marc he might be able to explain better. But it's basically to synch the next/router to the connected router from `connected-next-router` that we use to store the url location in redux

packages/slice-machine/src/modules/selectedSlice/push.ts
sintinilorenzo: same as before

packages/slice-machine/src/modules/slices/index.ts
sintinilorenzo: no the replace will just trigger the location change. The `yield(call(LOCATION_CHANGE))` is so that the saga waits until the redirection is finished, before triggering the next yield put. It's something me and Marc figured out when pairing cuz there was a little bug caused by the popup appearing BEFORE the redirection

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
sintinilorenzo: yeah good point! will change

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
sintinilorenzo: oh right!!! nice one for spotting

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
sintinilorenzo: ❓ don't understand what you mean by this

packages/slice-machine/pages/slices.tsx
sintinilorenzo: libraries and localLibs are not exactly the same since one is a LibraryUI[] and the other is a LibraryState[]. Kept the `libraries` since that's the one that has more information 👍🏼

packages/slice-machine/src/models/slice/context.tsx
sintinilorenzo: Yes tbh I have not fully understood why we have those, maybe we can add something related to it in the tech debt backlog?

packages/slice-machine/src/modules/selectedSlice/reducer.ts
sintinilorenzo: oh interesting, should I create a ticket?

packages/slice-machine/src/modules/selectedSlice/reducer.ts
sintinilorenzo: uff good one to spot! I fixed it here but while manually testing realise we have another issue related to it:

We update the __status in the `selectedSlice` redux state but not in the `library` state, which means when you go back to the previous page (where the status is shown in the card), it does not update unless you refresh the page :(

sintinilorenzo: not too sure what the best way of fixing this is tho...

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
sintinilorenzo: yes you're right i can remove the `!Model` check, but still need the `!variation` since that could be undefined. That's because we fetch Model from redux and then get the correct variation using a `.find` that matches the value in the url. So technically a user could manually go to a wrong URL and no variation is found (in which case the variation is undefined)

packages/slice-machine/src/modules/selectedSlice/types.ts
sintinilorenzo: Agree but would reather not change everything 😬. Ticket is quite big already, maybe something to consider in a follow up?

packages/slice-machine/src/modules/selectedSlice/reducer.ts
sintinilorenzo: guess we could listen to all redux actions of type sliceCreated, sliceSaved and slicePushed inside the library redux, and when those are triggered we can update the `__status` of the corresponding slice? What do you think

packages/slice-machine/src/modules/slices/index.ts
MarcMcIntosh: is it replace or put, or are the two similar?

sintinilorenzo: good point it's `push`. The difference between replace and push is that push adds a new route into the history route stack, while replace changes the top one. Therefore we want to use `push` so that we add on the stack and we can navigate back. 

A usecase of `replace` could be if the user tries to go to a wrong URL, in which case we want to override it so that he cannot navigate back to the wrong route.

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -44,31 +45,15 @@ const initialState: SliceBuilderState = {
...
-
-  const { Model } = useSelector((state: SliceMachineStoreType) => ({
-    Model: state.selectedSlice?.Model,
-  }));
-
-  const variation = Model?.variations.find(
-    (variation) => variation.id === initVariationId
-  );
-
+const SliceBuilder: React.FC<SliceBuilderProps> = ({ Model, variation }) => {
```
Heyrows, line 52: ❤️

packages/slice-machine/pages/slices.tsx
```diff
@@ -213,8 +203,8 @@ const SlicesIndex: React.FunctionComponent = () => {
           isCreatingSlice={isCreatingSlice}
           isOpen={isCreateSliceModalOpen}
           close={closeCreateSliceModal}
-          libraries={localLibs}
-          remoteSlices={remoteLibs}
+          libraries={libraries || []}
```
Heyrows: this should be localLibraries most likely.

packages/slice-machine/src/modules/selectedSlice/reducer.ts
Heyrows: Discussed during daily, we'll create a new ticket and tackle it right away :)

packages/slice-machine/src/modules/selectedSlice/reducer.ts
```diff
@@ -261,13 +261,16 @@ const updateTouchedAndStatus = (model: SliceState) => {
     model.initialVariations
   );
 
-  if (isModelModified || isScreenshotModified) {
+  if (
```
Heyrows: damn all those conditions seems too though to read, i'm sure we could simplify that later :)

packages/slice-machine/src/modules/slices/index.ts
```diff
@@ -144,7 +144,7 @@ export function* createSliceSaga({
   const addr = `/${payload.libName.replace(/\//g, "--")}/${
     payload.sliceName
   }/${variationId}`;
-  yield put(replace("/[lib]/[sliceName]/[variation]", addr));
+  yield put(push("/[lib]/[sliceName]/[variation]", addr));
```
Heyrows, line 147: judging from the example, https://github.com/danielr18/connected-next-router/blob/dd523db706a8605cc199814ee1a1fa38e2f5b204/examples/basic/components/navigation.js

Push seems to be enough, is `take(Location_Change)` used as some kind of "dispatch" to tell Redux an event happened ?

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
Heyrows: ah ok i see :x

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
Heyrows: I went a little bit overboard haha, we can keep that for another time.

packages/slice-machine/pages/_app.tsx
Heyrows: ok :)

packages/slice-machine/src/modules/selectedSlice/reducer.ts
Heyrows: We'll have a spike for all of those before implementing something :D

packages/slice-machine/src/modules/selectedSlice/types.ts
Heyrows: yes!  same ticket as the other one, spike on reorganizing data stores

packages/slice-machine/pages/slices.tsx
sintinilorenzo: renamed :)

packages/slice-machine/src/modules/selectedSlice/reducer.ts
sintinilorenzo: tidied up a bit with more clear if and else statements

Heyrows (approved):

PR #506, opened 2022-07-04
Aoc/696 fields redesigned

chengoconnell:
## Types of changes

## Checklist:

### DIFF
--- a/packages/slice-machine/components/FormFields/Input.js
+++ b/packages/slice-machine/components/FormFields/Input.js
@@ -1,6 +1,43 @@
 import { Field } from "formik";
 import { Box, Label, Input, Text } from "theme-ui";
 
+const getFieldStyles = (isError, isDisabled, isWarning) => {
+  const defaultFieldStyles = {
+    "&::placeholder": { color: "#C9D0D8" },
+  };
+  const fieldStyles = isDisabled
+    ? {
+        ...defaultFieldStyles,
+        border: "1px solid #E6E6EA",
+        backgroundColor: "#F3F5F7",
+      }
+    : isError
+    ? {
+        ...defaultFieldStyles,
+        border: "1px solid #E26049",
+        "&:focus": {
+          outline: "none",
+          borderColor: "#E26049",
+          boxShadow:
+            "0 0 0 3px rgba(226, 96, 73, 0.2), inset 0 1px 2px rgba(226, 96, 73, 0.2)",
+        },
+      }
+    : isWarning
+    ? {
+        ...defaultFieldStyles,
+        border: "1px solid orange",
+        "&:focus": {
+          outline: "none",
+          borderColor: "orange",
+          boxShadow:
+            "0 0 0 3px rgba(255, 165, 0, 0.2), inset 0 1px 2px rgba(255, 165, 0, 0.2)",
+        },
+      }
+    : defaultFieldStyles;
+
+  return fieldStyles;
+};
+
 const FormFieldInput = ({
   sx = {},
   field /* from Formik */,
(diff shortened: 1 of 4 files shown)


packages/slice-machine/components/FormFields/utils/getFieldStyles.tsx
```diff
@@ -0,0 +1,46 @@
...
+  isDisabled: boolean;
+  isWarning: boolean;
+}
+
+export const getFieldStyles = ({
+  isError,
+  isDisabled,
+  isWarning,
+}: getFieldStylesProps) => {
+  const defaultFieldStyles = {
```
Heyrows: That could also have been made using CSS & Classnames
```suggestion
  import cls from 'classnames'
  
  <span
    classname={cls(
      isError && 'error',
      isWarning && 'warning',
      isDisabled && 'disabled'
    )}
  ></span>
```

Heyrows: that way the CSS handles the style, we usually avoid css in the TS, it usually is a last ressort for dynamic properties for instance.

packages/slice-machine/components/FormFields/Input.js
```diff
@@ -1,5 +1,6 @@
 import { Field } from "formik";
 import { Box, Label, Input, Text } from "theme-ui";
+import { getFieldStyles } from "./utils/getFieldStyles";
 
 const FormFieldInput = ({
```
Heyrows: how much work would it required to transform this file in TS ?

packages/slice-machine/components/FormFields/Input.js
```diff
@@ -9,50 +10,53 @@ const FormFieldInput = ({
...
-  <Box sx={sx}>
-    <Label
-      variant="label.primary"
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-      htmlFor={fieldName}
-    >
-      {
+}) => {
+  return (
+    <Box sx={sx}>
```
Heyrows: what happens with your style if here we pass extra style ? (i'm guessing the property must be used if it exists 😄 )

chengoconnell: Just had a look and this property isn't actually used anywhere 😆

packages/slice-machine/components/FormFields/Input.js
chengoconnell: This file is used in Field.js which is used in a lot of other js files so transforming this will require transforming all the other js files which use Field.js

packages/slice-machine/components/FormFields/utils/getFieldStyles.tsx
chengoconnell: good idea

packages/slice-machine/components/FormFields/Input.js
chengoconnell: The sx only affects the box, not the field, so it shouldn't be affected

chengoconnell: i've just tested it and the fields remain the same

sintinilorenzo (approved):

packages/slice-machine/components/FormFields/Input.tsx
```diff
@@ -0,0 +1,120 @@
...
+interface FormFieldInputProps {
+  sx: ThemeUIStyleObject;
+  field: FieldInputProps<string>;
+  meta: FieldMetaProps<string>;
+  formField: {
+    label: string;
+    placeholder: string;
+    fieldLevelValidation?: (arg: {
+      value: string;
+      fields: any;
```
Heyrows, line 54: could `Record<string, unknown>` work ?

packages/slice-machine/components/FormFields/Input.tsx
```diff
@@ -0,0 +1,120 @@
...
+  fields,
+  initialValues,
+}: FormFieldInputProps) => {
+  console.log("fields", fields);
+
+  return (
+    <Box sx={sx}>
+      <Label
+        variant="label.primary"
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
Heyrows: can we get rid of all those eslint disable now that everything is typed :) ?

Heyrows (commented):
Please don't force push if possible, it's though to make the code review as everything needs to be re-reviewed.

packages/slice-machine/components/FormFields/Input.tsx
chengoconnell: changed! :)

packages/slice-machine/components/FormFields/Input.tsx
```diff
@@ -2,9 +2,9 @@ import { Field, FieldInputProps, FieldMetaProps } from "formik";
 import { Box, Label, Input, Text, ThemeUIStyleObject } from "theme-ui";
 
 const getFieldStyles = (
-  isError: boolean,
-  isDisabled: boolean,
-  isWarning: boolean
+  isError: boolean | undefined,
```
Heyrows: OPT: Are those distinctive conditions or could they be fused with a Typescript enum ?

Heyrows (approved):

packages/slice-machine/components/FormFields/Input.tsx
chengoconnell: Good spot, actually because i've put Boolean(meta.error), it can only be a boolean so ive just removed the undefined

PR #510, opened 2022-07-07
fix the slices status synch

sintinilorenzo:
## Types of changes

## Checklist:

### DIFF
--- a/packages/slice-machine/pages/slices.tsx
+++ b/packages/slice-machine/pages/slices.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
 import { MdHorizontalSplit } from "react-icons/md";
 import { Box, Flex, Button, Text, Spinner, Link } from "theme-ui";
 import Container from "components/Container";
@@ -20,6 +20,7 @@ import { ModalKeysEnum } from "@src/modules/modal/types";
 import { isLoading } from "@src/modules/loading";
 import { LoadingKeysEnum } from "@src/modules/loading/types";
 import { getLibrariesState, getRemoteSlices } from "@src/modules/slices";
+import { getState } from "@src/apiClient";
 
 const CreateSliceButton = ({
   onClick,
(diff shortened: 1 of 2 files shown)


chengoconnell (approved):

packages/slice-machine/pages/slices.tsx
```diff
@@ -51,6 +56,11 @@ const SlicesIndex: React.FunctionComponent = () => {
       libraries: getLibrariesState(store),
     }));
 
+  useEffect(() => {
+    //TODO: once the slice redux structure refactor is done, remove this and update the library redux state through the selectedSlice actions
+    void getState().then(({ data }) => refreshState(data));
```
Heyrows, line 61: We already refresh the state quite often using SWR.

When we push a selected Slice i'm guessing we update the list of remote Slices on the store of slices that is used to display this page, why can't we update the status with it ?

sintinilorenzo: We need to update the library state quite often not just when pushing, but whenever the status is changed (if it goes into synched, or if it goes into modified every time we save, take screenshot, rename etc...). The useSwr hook is not triggered all of these times.

So we would need to refresh the state on each of those actions which we can do with redux (but I'd do that as a follow up ticket as I explained int he spike outcome since it's quite a lot of work).

Heyrows: SWR reacts more to coming back from outside the application yeah, here i'm just wondering why don't we alter the states ourselves in the frontend as we know what happened and what changed, why do we need to refresh the whole state ?

Heyrows (approved):

packages/slice-machine/pages/slices.tsx
sintinilorenzo: seen together. Misunderstanding on what the strategy was. This is a quick fix to eliminate the bug and when we decide to tackle the tech debt on redux structure we will implement the better redux-only way of updating library state

PR #512, opened 2022-07-08
[HotFix] - v0.4.1

chengoconnell:
Hotfix for production to prevent users from creating fields with empty API ID.


packages/slice-machine/lib/forms/defaults.ts
```diff
@@ -9,6 +9,9 @@ export const validateId = ({
   fields: Array<{ key: string }>;
   initialId: string;
 }) => {
+  if (!value) {
+    return "Field is required";
```
sintinilorenzo, line 14: This might be a bit inconsistent with the error message from the label missing, which says `This field is required`. There are the errors throughout the app, may be worth using the same everywhere since it's a small change?
image
image

PR #514, opened 2022-07-11
deep merge the objects to avoid overwriting the config value

sintinilorenzo:
## Types of changes

## Checklist:

### DIFF
--- a/packages/slice-machine/lib/builders/common/EditModal/index.js
+++ b/packages/slice-machine/lib/builders/common/EditModal/index.js
@@ -1,6 +1,7 @@
 import Modal from "react-modal";
 import SliceMachineModal from "@components/SliceMachineModal";
 import deepMerge from "deepmerge";
+import { merge } from "lodash";
 
 import { Box, Close, Flex, Button, useThemeUI } from "theme-ui";
 
@@ -165,10 +166,7 @@ const EditModal = ({ close, data, fields, onSave, getFieldMockConfig }) => {
           })();
 
           // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-          const updatedValue = {
-            ...initialModelValues,
-            ...value,
-          };
+          const updatedValue = merge(initialModelValues, value);
 
           // eslint-disable-next-line @typescript-eslint/no-unsafe-call
           onSave({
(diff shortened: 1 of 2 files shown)


packages/slice-machine/lib/builders/common/EditModal/index.js
```diff
@@ -165,10 +166,7 @@ const EditModal = ({ close, data, fields, onSave, getFieldMockConfig }) => {
           })();
 
           // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-          const updatedValue = {
-            ...initialModelValues,
-            ...value,
-          };
+          const updatedValue = merge(initialModelValues, value);
```
sintinilorenzo, line 169: interesting bug: 
initialModelValues was `{type: "...", config: {...}}` (coming from the model state from the BE)
value has the same type but its config contains the data from the edit modal (therefore would not have the property `labels`).

If `lables` was put manually, since we were doing a shallow merge copy, we would override the old `config` (with labels) with the new one (without labels)

packages/slice-machine/lib/models/common/widgets/StructuredText/index.ts
```diff
@@ -31,13 +31,21 @@ const Meta = {
...
+};
+
 const schema = yup.object().shape({
   type: yup
     .string()
     .matches(/^StructuredText$/, { excludeEmptyString: true })
     .required(),
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-argument
-  config: createValidationSchema(removeProp(FormFields, "id")),
+  config: createValidationSchema(
```
Heyrows, line 43: Why do we need it in the schema ?

I mean, from my understanding, this YUP is used to validate the forms in the save modal. We don't have a fields for label and data are merged later (after validation I believe).

chengoconnell (approved):

packages/slice-machine/lib/models/common/widgets/StructuredText/index.ts
sintinilorenzo: image
if `labels` are manually added but specified in the yup schema it will fail on the `.schema.validateSync` method

Heyrows: hmmm ok so it seems to validate the initialValue too, weird haha

Heyrows (approved):

PR #515, opened 2022-07-11
avoid page reload when renaming slice

sintinilorenzo:
## Types of changes

## Checklist:

### DIFF
--- a/packages/slice-machine/src/modules/slices/index.ts
+++ b/packages/slice-machine/src/modules/slices/index.ts
@@ -169,6 +169,8 @@ export function* renameSliceSaga({
       payload.newSliceName,
       payload.libName
     );
+    //TODO: avoid refetch of state. Either manually update all the changes within the reducer,
+    //or make the renameSlice api call return the new libraries, so that it's output can be used.
     const { data: serverState } = (yield call(getState)) as SagaReturnType<
       typeof getState
     >;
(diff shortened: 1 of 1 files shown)


packages/slice-machine/src/modules/slices/index.ts
```diff
@@ -164,10 +169,15 @@ export function* renameSliceSaga({
       payload.newSliceName,
       payload.libName
     );
+    const { data: serverState } = (yield call(getState)) as SagaReturnType<
```
Heyrows, line 174: We know what changed, so why get all the state again ?
Why not simply rename the Slice and add the action to both reducers ? (selected & available)

Computing the whole state makes unnecessary calls to the API.

sintinilorenzo: I tried to change it manually in the reducer. The issue is that I had to change it lots of places because the name change will lead to a different path, therefore on top of `model.name` we also change `screenshotUrls`, `screenshotPaths` etc. Was worried of forgetting/missing some so decided it was best to refetch. 

Should i try to change it manually in the reducer?

Heyrows: Spike inc so we'll say it's ok for now but we'll need to change it later.
Maybe the server could send it directly after the call to rename, as the env is already computed this could save us some time maybe.

Heyrows (approved):

packages/slice-machine/src/modules/slices/index.ts
sintinilorenzo: added a TODO to say that

PR #517, opened 2022-07-13
init(tracking): set the repo when it's provided in process.argv

MarcMcIntosh:
Add the repository to the tracking when it's passed as a flag to the init script
## Types of changes

## Checklist:

### DIFF
--- a/packages/init/src/utils/tracker.ts
+++ b/packages/init/src/utils/tracker.ts
@@ -112,10 +112,7 @@ export class InitTracker {
 
   trackInitStart(repoDomain: string | undefined): Promise<void> {
     if (repoDomain) this.setRepository(repoDomain);
-    return this._trackEvent(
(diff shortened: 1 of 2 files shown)


packages/init/src/utils/tracker.ts
```diff
@@ -111,7 +111,11 @@ export class InitTracker {
   }
 
   trackInitStart(repoDomain: string | undefined): Promise<void> {
-    return this._trackEvent(EventType.InitStart, { repo: repoDomain });
+    if (repoDomain) this.setRepository(repoDomain);
+    return this._trackEvent(
+      EventType.InitStart,
+      repoDomain ? { repo: repoDomain } : undefined
```
Heyrows: i think we should not touch that, just in case it might break something 😅

Heyrows (approved):

PR #519, opened 2022-07-15
Ls/716 restructure slices redux config

sintinilorenzo:
## Types of changes

## Checklist:


chengoconnell (approved):

packages/slice-machine/src/modules/selectedSlice/save.ts
```diff
@@ -1,34 +1,28 @@
...
-        sliceName: slice.model.name,
-        from: slice.from,
-        model: {
-          ...slice.model,
-          variations: slice.variations,
-        },
-        mockConfig: slice.mockConfig,
+        sliceName: component.model.name,
+        from: component.from,
+        model: component.model,
```
MarcMcIntosh, line 18: are the variations still there?

packages/slice-machine/src/modules/selectedSlice/reducer.ts
```diff
@@ -157,120 +155,111 @@ export const selectedSliceReducer: Reducer<
...
-      const newModel = { ...prevModel, screenshotUrls: screenshots };
+      const screenshots: Record<string, ScreenshotUI> =
+        prevComponent.model.variations.reduce((acc, variation) => {
+          if (variation.id === variationId) {
+            return {
+              ...acc,
+              [variationId]: screenshot,
+            };
+          }
+          if (prevComponent.screenshotUrls?.[variation.id]) {
```
MarcMcIntosh, line 187: is the `.` needed before `[selector]`?

packages/slice-machine/lib/models/common/ComponentUI.ts
```diff
@@ -78,9 +78,33 @@ export const ComponentUI = {
...
+      if (variationId)
+        return component.model.variations.find((v) => v.id === variationId);
+      return component.model.variations[0];
+    }
+  },
+
+  updateVariation(component: ComponentUI, variationId: string) {
+    return (mutate: (v: VariationSM) => VariationSM): ComponentUI => {
+      const variations = component.model.variations.map((v) => {
+        if (v.id === variationId) return mutate(v);
```
MarcMcIntosh: what does `mutate` do?

MarcMcIntosh (commented):
Left a few questions,

It might be worth adding some test for redux, we have some already and it would be pretty good to have some added for this incase any future refactoring is needed :)

packages/slice-machine/src/modules/selectedSlice/reducer.ts
sintinilorenzo: image

image

Yes, without it typescrupt will interpret it as a ternarny `A ? B : C `

packages/slice-machine/lib/models/common/ComponentUI.ts
sintinilorenzo: mutate is the a function parameter of the function (passed in line 93). In this case it would be the `v => {...}` 
image

It's probably badly named, will change it to mutateCallbackFn

packages/slice-machine/src/modules/selectedSlice/save.ts
sintinilorenzo: yes, the model's own variations are updated in the redux so that we don't have duplication across the 2 places it was stored before :)

packages/slice-machine/lib/models/common/ComponentUI.ts
MarcMcIntosh: does it do any mutations?

packages/slice-machine/src/modules/selectedSlice/save.ts
MarcMcIntosh: neat

MarcMcIntosh (approved):

PR #521, opened 2022-07-18
Aoc/719 use sagas to asynchronous calls

chengoconnell:
## Types of changes

## Checklist:


packages/slice-machine/src/modules/selectedSlice/screenshot.ts
```diff
@@ -3,8 +3,6 @@ import { ScreenshotUI } from "@lib/models/common/ComponentUI";
 import { ScreenshotResponse, Screenshots } from "@models/common/Screenshots";
 
 export async function generateScreenShot(
```
chengoconnell: this function isn't used anymore now that we move the logic into sagas. Should i remove it?

sintinilorenzo: I'd say so it's dead code now, same should be for the generateCustomScreenshot.+ save + push yeah?

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -109,18 +102,25 @@ const SliceBuilder: React.FC<SliceBuilderProps> = ({
...
+    generateSliceCustomScreenshot(
+      variation.id,
+      extendedComponent.component,
+      setData,
+      file
+    );
+  };
+
+  const onPushSlice = () => {
+    pushSlice(extendedComponent, onPush);
```
sintinilorenzo: can you just write the content of `onPush` (`setData(data)...`) here rather than having a separate function for it?

For redability otherwise we end up with a bunch of `onPush` `onPushSlice` and it's hard to understand what does what

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -109,18 +102,25 @@ const SliceBuilder: React.FC<SliceBuilderProps> = ({
 
   if (!variation || !sliceView) return null;
 
-  const onTakingCustomScreenshot = () => {
-    // eslint-disable-next-line @typescript-eslint/no-misused-promises
-    checkSimulatorSetup(true, async () => {
```
sintinilorenzo, line 114: think we still want this.

packages/slice-machine/src/apiClient.ts
```diff
@@ -92,6 +94,44 @@ export const renameSlice = (
   return axios.put(`/api/slices/rename`, requestBody, defaultAxiosConfig);
 };
 
+export const generateSliceScreenshotAxios = (
```
sintinilorenzo: small detail: not great to call it generateSliceScreenshotAxios because it becomes dependent of what library we use to make the api call. If we were to use another library the function purpose would be exactly the same, but the name would become wrong

packages/slice-machine/src/modules/selectedSlice/actions.ts
```diff
@@ -71,21 +72,60 @@ export const deleteSliceWidgetMockCreator = createAction(
...
-export const saveSliceCreator = createAction("SLICE/SAVE")<{
-  extendedComponent: NonNullable<SelectedSliceStoreType>;
-}>();
+export const saveSliceCreator = createAsyncAction(
+  "SLICE/SAVE.REQUEST",
+  "SLICE/SAVE.RESPONSE",
+  "SLICE/SAVE.FAILURE"
+)<
+  {
+    extendedComponent: ExtendedComponentUI;
```
sintinilorenzo: can you change this to `NonNullable<SelectedSliceStoreType>` for consistency?
Think it's slightly better because it represents better what it is (i.e. the whole store content, except null values)

packages/slice-machine/src/modules/selectedSlice/sagas.ts
```diff
@@ -0,0 +1,328 @@
...
+    });
+    const response = (yield call(
+      saveSliceAxios,
+      extendedComponent
+    )) as SagaReturnType<typeof saveSliceAxios>;
+    if (response.status > 209) {
+      return setData({
+        loading: false,
+        done: true,
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
```
sintinilorenzo: any reason you have eslint disabled here but not in the other sagas?

packages/slice-machine/src/modules/selectedSlice/sagas.ts
```diff
@@ -0,0 +1,328 @@
...
+  payload,
+}: ReturnType<typeof renameSliceCreator.request>) {
+  try {
+    yield call(
+      renameSlice,
+      payload.sliceId,
+      payload.newSliceName,
+      payload.libName
+    );
+    //TODO: avoid refetch of state. Either manually update all the changes within the reducer,
```
sintinilorenzo: ah I have changed this in my other PR sorry! it's slightly different now.
If it's annoying to rebase and change dw about it just delete it and I will do the saga change myself :)

packages/slice-machine/src/modules/selectedSlice/actions.ts
```diff
@@ -71,21 +72,60 @@ export const deleteSliceWidgetMockCreator = createAction(
...
+    setData: (data: any) => void;
+  },
+  {
+    extendedComponent: ExtendedComponentUI;
```
Heyrows: same as what lorenzo saiid about `NonNullable<SelectedSliceStoreType>`

packages/slice-machine/src/modules/selectedSlice/sagas.ts
```diff
@@ -0,0 +1,328 @@
...
+        type: ToasterType.ERROR,
+      })
+    );
+  }
+}
+
+export function* generateSliceCustomScreenshotSaga({
+  payload,
+}: ReturnType<typeof generateSliceCustomScreenshotCreator.request>) {
+  const { variationId, component, setData, file } = payload;
```
Heyrows, line 85: i'm super surprised that you can pass functions into an action's Payload, thought actions were like raw data. Nice 👍

packages/slice-machine/src/modules/selectedSlice/sagas.ts
```diff
@@ -0,0 +1,328 @@
...
+    });
+    const response = (yield call(
+      generateSliceCustomScreenshotAxios,
+      form
+    )) as SagaReturnType<typeof generateSliceCustomScreenshotAxios>;
+    if (response.status > 209) {
+      return setData({
+        loading: false,
+        done: true,
+        status: response.status,
```
Heyrows, line 107: shouldn't we set an error here ?

Heyrows: or throw an error to be catched line 128 and open the toaster

packages/slice-machine/src/apiClient.ts
chengoconnell: Good point, would `generateSliceScreenshotApiClient` better? I wanted to distinguish this function from other `generateSliceScreenshot` function for easier exports and imports

packages/slice-machine/src/modules/selectedSlice/sagas.ts
chengoconnell: the return type of this one is `Promise<AxiosResponse>` whereas other ones have more specific types, e.g. `Promise<AxiosResponse<ScreenshotUI>>`

packages/slice-machine/src/modules/selectedSlice/sagas.ts
chengoconnell: changed np

packages/slice-machine/src/modules/selectedSlice/screenshot.ts
sintinilorenzo: can you also delete the save and push api calls that are not used anymore?

packages/slice-machine/src/modules/selectedSlice/sagas.ts
sintinilorenzo: thanks!

sintinilorenzo (approved):

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -109,41 +96,42 @@ const SliceBuilder: React.FC<SliceBuilderProps> = ({
...
-      await generateScreenShot(
+  const onTakingSliceScreenshot = () => {
+    generateSliceScreenshot(variation.id, extendedComponent.component, setData);
+  };
+
+  const onTakingSliceCustomScreenshot = (file: Blob) => {
+    checkSimulatorSetup(true, () =>
```
Heyrows, line 104: seems to be a small mistake, the simulator needs to be verified for SliceScreenshot, not CustomScreenshot.

Custom Screenshot are a custom file directly uploaded and used.
While Slice Screenshots are taken using the user's project so we need to verify it has been setup properly.

Heyrows: Looking at the old code, it seems that you are right and you modified the code right, but it seems super weird to check the Simulator Setup here and not for regular screenshots.

packages/slice-machine/src/modules/loading/types.ts
```diff
@@ -10,4 +10,8 @@ export enum LoadingKeysEnum {
   PUSH_CUSTOM_TYPE = "PUSH_CUSTOM_TYPE",
   CREATE_SLICE = "CREATE_SLICE",
   RENAME_SLICE = "RENAME_SLICE",
+  GENERATE_SLICE_SCREENSHOT = "GENERATE_SLICE_SCREENSHOT",
```
Heyrows, line 13: On the sync changes project that is coming, we would need to have a page with multple screenshot state loading at the same time.

Nothing to modify for now but we'll have to keep that in mind.

Heyrows (approved):

packages/slice-machine/src/modules/loading/types.ts
chengoconnell: Ah okay thanks for the heads up!

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
chengoconnell: True, perhaps we could investigate this in a spike too see why it was originally there and if adding it would lead to any side effects

PR #523, opened 2022-07-19
Ls/721 disable uid id field

sintinilorenzo:
image
image

## Types of changes

## Checklist:

### DIFF
--- a/packages/slice-machine/components/FormFields/Input.tsx
+++ b/packages/slice-machine/components/FormFields/Input.tsx
@@ -1,3 +1,4 @@
+import { InputType } from "@lib/forms/fields";
 import { Field, FieldInputProps, FieldMetaProps } from "formik";
 import { Box, Label, Input, Text, ThemeUIStyleObject } from "theme-ui";
 
@@ -57,15 +58,7 @@ interface FormFieldInputProps {
   sx: ThemeUIStyleObject;
   field: FieldInputProps<string>;
   meta: FieldMetaProps<string>;
-  formField: {
-    label: string;
-    placeholder: string;
-    fieldLevelValidation?: (arg: {
-      value: string;
-      fields: any;
-      initialId: string;
-    }) => boolean;
-  };
+  formField: InputType;
   fieldName: string;
   fields: Record<string, unknown>;
   initialValues?: Record<string, string>;
@@ -84,7 +77,7 @@ export const FormFieldInput = ({
 }: FormFieldInputProps) => {
   const style = meta.error
     ? InputFieldStyles.ERROR
-    : isDisabled
+    : isDisabled || formField.disabled
     ? InputFieldStyles.DISABLED
     : InputFieldStyles.DEFAULT;
 
@@ -116,7 +109,7 @@ export const FormFieldInput = ({
         {...field}
         as={Input}
         sx={getInputFieldStyles(style)}
-        disabled={isDisabled}
+        disabled={isDisabled || formField.disabled}
       />
     </Box>
   );
--- a/packages/slice-machine/components/FormFields/Input.tsx
+++ b/packages/slice-machine/components/FormFields/Input.tsx
@@ -1,22 +1,27 @@
 import { Field, FieldInputProps, FieldMetaProps } from "formik";
 import { Box, Label, Input, Text, ThemeUIStyleObject } from "theme-ui";
 
-const getFieldStyles = (
-  isError: boolean,
-  isDisabled: boolean | undefined,
-  isWarning: boolean | undefined
-) => {
+export enum InputFieldStyles {
+  ERROR,
+  DISABLED,
+  WARNING,
+  DEFAULT,
+}
+
+export const getInputFieldStyles = (type: InputFieldStyles) => {
   const defaultFieldStyles = {
     "&::placeholder": { color: "#C9D0D8" },
   };
-  const fieldStyles = isDisabled
-    ? {
+  switch (type) {
+    case InputFieldStyles.DISABLED: {
+      return {
         ...defaultFieldStyles,
         border: "1px solid #E6E6EA",
         backgroundColor: "#F3F5F7",
-      }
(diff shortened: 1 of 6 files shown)


packages/slice-machine/components/FormFields/Input.tsx
```diff
@@ -36,25 +43,22 @@ const getFieldStyles = (
           boxShadow:
             "0 0 0 3px rgba(255, 165, 0, 0.2), inset 0 1px 2px rgba(255, 165, 0, 0.2)",
         },
-      }
-    : defaultFieldStyles;
-
-  return fieldStyles;
+      };
+    }
+    case InputFieldStyles.DEFAULT: {
```
Heyrows: no need for that case, it will fall in the default if this one is not there.

packages/slice-machine/components/FormFields/Input.tsx
```diff
@@ -1,22 +1,28 @@
+import { InputType } from "@lib/forms/fields";
 import { Field, FieldInputProps, FieldMetaProps } from "formik";
 import { Box, Label, Input, Text, ThemeUIStyleObject } from "theme-ui";
 
-const getFieldStyles = (
-  isError: boolean,
-  isDisabled: boolean | undefined,
-  isWarning: boolean | undefined
-) => {
+export enum InputFieldStyles {
```
Heyrows, line 5: nice !

packages/slice-machine/lib/forms/defaults.ts
```diff
@@ -15,9 +15,7 @@ export const validateId = ({
   }
 };
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Heyrows, line 18: nice

packages/slice-machine/lib/models/common/widgets/UID/index.ts
```diff
@@ -23,24 +23,20 @@ import {
 } from "@prismicio/types-internal/lib/customtypes/widgets";
 
 const FormFields = {
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
   label: DefaultFields.label,
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
   id: {
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
     ...DefaultFields.id,
     validate: {
```
Heyrows, line 29: if it's disabled, do we still need the validate function ?

Heyrows (approved):

packages/slice-machine/lib/models/common/widgets/UID/index.ts
sintinilorenzo: no not needed since it's not really used. But if I remove it it would still be present (it would just be the default one coming from `... DefaultFields.id`). So I though it'd be best to have an unused-correct validation than an unused-incorrect validation. 

Mostly to avoid potential bugs, if the disabled field is broken/removed by accident, we want to still have the safety.

packages/slice-machine/components/FormFields/Input.tsx
sintinilorenzo: ah good point, removed

packages/slice-machine/lib/models/common/widgets/UID/index.ts
Heyrows: right 👍

Heyrows (approved):

packages/slice-machine/components/FormFields/Input.tsx
```diff
@@ -71,6 +82,12 @@ export const FormFieldInput = ({
   initialValues,
   isDisabled,
 }: FormFieldInputProps) => {
+  const style = meta.error
+    ? InputFieldStyles.ERROR
+    : isDisabled
```
chengoconnell: is it possible for there to be a meta.error AND isDisabled?

packages/slice-machine/components/FormFields/Input.tsx
```diff
@@ -84,7 +77,7 @@ export const FormFieldInput = ({
 }: FormFieldInputProps) => {
   const style = meta.error
     ? InputFieldStyles.ERROR
-    : isDisabled
+    : isDisabled || formField.disabled
```
chengoconnell, line 76: is `isDisabled` necessary if there is `formField.disabled`?

packages/slice-machine/lib/forms/fields.ts
```diff
@@ -51,7 +51,7 @@ export interface InputType {
   label: string;
   validate: any;
   defaultValue: string | undefined;
-  fieldLevelValidation: FieldLevelValidationFn | undefined;
+  fieldLevelValidation: FieldLevelValidationFn | undefined | null;
```
chengoconnell, line 54: what's the reason for this? 🤔

packages/slice-machine/components/FormFields/Input.tsx
sintinilorenzo: might be, but guess errors should take precedence so that's what we should use in case they both are present

packages/slice-machine/components/FormFields/Input.tsx
sintinilorenzo: not sure, it was already there. I assume it could be used if you need to disable the fields for other reasons, since the formField is hardcoded and fixed for the most part

packages/slice-machine/lib/forms/fields.ts
sintinilorenzo: `null` was being passed as a value for fieldLevelValidation in a form. Didn't wanna change it to `undefined ` cuz I am not too sure if it could've caused troubles

chengoconnell (approved):

PR #529, opened 2022-07-22
Ls/718 remove redux unnecesary info

sintinilorenzo:
## Types of changes

## Checklist:

### DIFF
--- a/packages/slice-machine/lib/models/common/Environment.ts
+++ b/packages/slice-machine/lib/models/common/Environment.ts
@@ -3,6 +3,7 @@ import type { Models } from "@slicemachine/core";
 import type { Client, ApplicationMode } from "@slicemachine/client";
 import { PackageChangelog } from "./versions";
 import { PackageManager } from "./PackageManager";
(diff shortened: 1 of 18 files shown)


packages/slice-machine/lib/models/common/Environment.ts
```diff
@@ -11,8 +12,7 @@ export interface BackendEnvironment {
   manifest: Models.Manifest;
   repo: string;
   changelog: PackageChangelog;
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  mockConfig: any;
+  mockConfig: CustomTypeMockConfig;
```
Heyrows, line 15: I like that it may looks like it's type but I looked at the type it's a `Record<string, any>` 😂 .
This will be improved by Arnaud's PR i believe.

Heyrows: I think that's a `GlobalMockConfig`, but it doesn't change much :)

Heyrows (approved):

PR #531, opened 2022-07-22
add ellipsis and tooltip to long slice field names

sintinilorenzo:
## Types of changes

## Description

image

## Checklist:


packages/slice-machine/components/ListItem/Header.tsx
```diff
@@ -0,0 +1,50 @@
+import { TextWithTooltip } from "@components/Tooltip/TextWithTooltip";
```
sintinilorenzo, line 1: this is the .js file converted into .tsx

packages/slice-machine/components/Tooltip/TextWithTooltip.tsx
```diff
@@ -1,38 +1,54 @@
...
         }}
-        as="h6"
+        as={as}
         ref={headingRef}
       >
         {text}
-      </Heading>
-      {showTooltip ? (
+      </Text>
+      {showTooltip && (
```
chengoconnell, line 51: https://stackoverflow.com/questions/52326260/why-does-0-true-return-0-in-javascript-instead-of-a-boolean

^ Reason why the ternary was used here instead of '&&' which could lead to bugs likes this

packages/slice-machine/components/Tooltip/TextWithTooltip.tsx
```diff
@@ -1,38 +1,54 @@
...
+  const updateTooltipVisibility = () => {
     setShowTooltip(
       headingRef.current &&
         headingRef.current.offsetWidth < headingRef.current.scrollWidth
     );
+  };
+
+  useEffect(() => {
+    updateTooltipVisibility();
+    window.addEventListener("resize", updateTooltipVisibility);
```
chengoconnell, line 28: cool! So this triggers the updateTooltipVisibility when the window gets resized?

sintinilorenzo: yes, took a bit to realise why it wasn't working lol, the showTooltip was only computed once at first render and never updated

packages/slice-machine/components/Tooltip/TextWithTooltip.tsx
sintinilorenzo: guess so but it's usually cleaner to use `condition && ComponentRenderedConditionally`. In this case the variable is only going to be a boolean or undefined.

That's what the react docs also use https://reactjs.org/docs/conditional-rendering.html

chengoconnell (approved):

packages/slice-machine/components/Tooltip/TextWithTooltip.tsx
```diff
@@ -1,38 +1,54 @@
...
     <>
-      <Heading
+      <Text
         data-for={text}
         data-tip
         sx={{
           flex: 1,
           overflow: "hidden",
           textOverflow: "ellipsis",
+          whiteSpace: "nowrap",
```
MarcMcIntosh, line 43: that's why it wasn't working

packages/slice-machine/components/ListItem/Header.tsx
MarcMcIntosh: nice

MarcMcIntosh (approved):

PR #533, opened 2022-07-25
feat: generate TypeScript types on save

lihbr:
## Types of changes

## Description

As requested by @Duaner, this PR adds an early integration of TypeScript inside Slice Machine by generating types for Slices and Custom Types on save and rename.

- A `types.ts` file is now created alongside Custom Types `index.json` model;
- A `types.ts` file is now created alongside Slices `model.json` model.

⚠ &nbsp;This code will have to be removed when switching to plugins. Plugins are already handling TypeScript integration the same way as the one implemented in this PR.

## Checklist:

- [x] All new and existing tests are passing. (existing tests were failing on Windows because of win32/posix paths inconsistencies)


packages/slice-machine/server/src/api/io/CustomType.ts
```diff
@@ -14,8 +15,21 @@ export function writeCustomType(path: string, customType: CustomTypeSM) {
...
+export function renameCustomType(
+  modelPath: string,
+  typesPath: string,
+  newCustomTypeName: string
+) {
+  const customType = readCustomType(modelPath);
   customType.label = newCustomTypeName;
-  writeCustomType(path, customType);
+  writeCustomType(modelPath, customType);
+  writeCustomTypeTypes(typesPath, customType);
```
Heyrows: nothing changes on the types, why do we need to re-write them ?

packages/slice-machine/server/src/api/io/Slice.ts
```diff
@@ -11,8 +12,19 @@ export function writeSlice(path: string, slice: SliceSM) {
...
+export function renameSlice(
+  modelPath: string,
+  typesPath: string,
+  newSliceName: string
+) {
+  const slice = readSlice(modelPath);
   slice.name = newSliceName;
-  writeSlice(path, slice);
+  writeSlice(modelPath, slice);
+  writeSliceTypes(typesPath, slice);
```
Heyrows: same question as the rename custom types, why do we need to rebuild the types if we rename them ?

packages/core/src/node-utils/paths.ts
```diff
@@ -35,6 +37,7 @@ const Paths = (cwd: string, prefix: string): Paths => ({
   customType: (id: string) => ({
     value: () => path.join(Paths(cwd, prefix).value()),
     model: () => path.join(Paths(cwd, prefix).value(), id, "index.json"),
+    types: () => path.join(Paths(cwd, prefix).value(), id, "types.ts"),
```
MarcMcIntosh: isn't the standard to call it `index.d.ts` ?

Heyrows (commented):
Should Slice template be modified to include the types ?

Thank you for contributing to Slice Machine 😁.
As you're adding a new feature, could you add the corresponding test / modify the existing one to match the expected behavior please 🙏

packages/slice-machine/server/src/api/io/CustomType.ts
lihbr: TSDocs on the generated type make use of the Slice Display name.

packages/slice-machine/server/src/api/io/Slice.ts
lihbr: TSDocs on the generated type make use of the Slice Display name here too ☺️

MarcMcIntosh (commented):
How would you test this and can it be tested?

packages/core/src/node-utils/paths.ts
lihbr: As far as I know, `*.d.ts` files are used to provide type information about their eponymous JavaScript files (e.g. `createFoo.js` & `createFoo.d.ts`)

In this case, `types.ts` doesn't represent the type of the `index.js` component (or the exports from that file). It represents the types of the Slice object (the `index.js` component eventually uses that). With that in mind we went for `types.ts` as thought more appropriate.

lihbr:
> Should Slice template be modified to include the types ?

Done, for now, it's only possible to streamline it for templates using `@prismicio/react`.

`prismic-dom`, `prismic-reactjs`, and `@prismicio/vue@2` cannot have it automated the same way because the component prop types depend on the user setup/the slice zone they use. Users, especially TypeScript ones, will still be able to take advantage of it, but manually, not from the template.

> As you're adding a new feature, could you add the corresponding test / modify the existing one to match the expected behavior please 🙏
>
> How would you test this and can it be tested?

`prismic-ts-codegen` is the library doing the bulk of the work and is already [extensively tested](https://app.codecov.io/gh/prismicio/prismic-ts-codegen). Regarding its integration with Slice Machine, we'd be mainly interested in testing that the types file is indeed created and updated when needed.

In that regard, I haven't seen tests regarding Slices/Custom Types save events (unit & e2e). If I missed them let me know where I can find them, happy to add a few. If this part of the codebase is not covered yet, I don't think it's part of the scope of this PR to start covering it (especially since this PR's additions are temporary).

packages/core/src/node-utils/paths.ts
MarcMcIntosh: ok

chengoconnell (approved):

cypress/integration/slices/00-create.spec.js
```diff
@@ -35,6 +35,7 @@ describe("Create Slices", () => {
       "eq",
       `/${lib}/${sliceName}/default`
     );
+    cy.readFile(`${path}/${sliceName}/types.ts`).should("contains", sliceName);
```
MarcMcIntosh: nice

cypress/integration/customTypes/00-create.spec.js
```diff
@@ -34,5 +36,9 @@ describe("Custom Types specs", () => {
     cy.get('[data-cy="custom-type-secondary-breadcrumb"]').contains(
       `/ ${name} - Edited`
     );
+    cy.readFile(`${path}/${id}/types.ts`).should(
```
MarcMcIntosh: nice

cypress/integration/slices/00-create.spec.js
```diff
@@ -50,5 +51,9 @@ describe("Create Slices", () => {
     cy.get('[data-cy="slice-and-variation-name-header"]').contains(
       `/ ${editedSliceName} / Default`
     );
+    cy.readFile(`${path}/${editedSliceName}/types.ts`).should(
```
MarcMcIntosh: nice

packages/slice-machine/server/src/api/slices/create/index.ts
```diff
@@ -35,6 +35,7 @@ const copyTemplate = async (
     await copy(templatePath, path.join(env.cwd, from, sliceName), {
       componentName: sliceName,
       componentId: snakelize(sliceName),
+      pascalId: pascalize(sliceName),
```
MarcMcIntosh, line 43: what's this for?

Heyrows: wait does that mean if it's rename we should change that part as well ?

lihbr: Types are generated after the pascalcased ID by convention.

Upon a Slice rename the ID wouldn't change so the type definition would remain the same and nothing would break here :)

MarcMcIntosh (approved):

MarcMcIntosh:
Update the branch :)

lihbr:
Closed in favor of #583

PR #536, opened 2022-07-27
use a safe merge function that does not overwride its arguments

sintinilorenzo:
## Types of changes

https://user-images.githubusercontent.com/69144671/181259712-d3b68c50-044b-436a-a2a7-50fb7820137a.mov

## Checklist:

### DIFF
--- a/packages/slice-machine/lib/builders/common/EditModal/index.js
+++ b/packages/slice-machine/lib/builders/common/EditModal/index.js
@@ -1,7 +1,5 @@
 import Modal from "react-modal";
 import SliceMachineModal from "@components/SliceMachineModal";
-import deepMerge from "deepmerge";
-import { merge } from "lodash";
 
 import { Box, Close, Flex, Button, useThemeUI } from "theme-ui";
 
@@ -21,6 +19,7 @@ import WidgetFormField from "./Field";
 
 import { findWidgetByConfigOrType } from "../../utils";
 import { removeProp } from "@lib/utils";
+import { deepMerge } from "@lib/utils/obj";
 
 if (process.env.NODE_ENV !== "test") {
   Modal.setAppElement("#__next");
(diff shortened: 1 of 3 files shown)


packages/slice-machine/lib/builders/common/EditModal/index.js
```diff
@@ -165,8 +164,8 @@ const EditModal = ({ close, data, fields, onSave, getFieldMockConfig }) => {
             return null;
           })();
 
-          // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-          const updatedValue = merge(initialModelValues, value);
```
sintinilorenzo, line 169: so annoying!! The lodash merge method is not a pure function and changes the actual arguments on top of returning the new merged object. This means we were updating the initialModel every time, and as a consequence there no longer was a diff between values and initialValues (which would have shown the text "Save to file system")
image

packages/slice-machine/tests/lib/utils/obj.test.ts
```diff
@@ -0,0 +1,56 @@
...
+import { deepMerge } from "../../../lib/utils/obj";
+
+describe("deepMerge", () => {
+  test("shallow value - updates value", () => {
+    const obj1 = { val1: 50, val2: 60 };
+    const obj2 = { val2: "hello" };
+    const expected = { val1: 50, val2: "hello" };
+    expect(deepMerge(obj1, obj2)).toStrictEqual(expected);
+    expect(obj1).toStrictEqual({ val1: 50, val2: 60 });
+    expect(obj2).toStrictEqual({ val2: "hello" });
```
sintinilorenzo, line 11: testing that the original parameters remain unchanged. These assertions would fail if you just use the loadash `merge` function.

packages/slice-machine/lib/builders/common/EditModal/index.js
chengoconnell: Wow super cool spot. Perhaps best to avoid lodash's merge in the future due to the side effects?

chengoconnell (approved):

Heyrows (approved):
seems good, good explanation of the issue 👍 
Thank for the doc screenshot and the extra test.

PR #538, opened 2022-07-27
Mm/ss tracking

MarcMcIntosh:
## Types of changes

Build tool for the server was changed from just `tsc` to `rollup + esbuild` this was done because tsc doesn't support compile time "define/replace". This required removing the runtime "module-alaises" package from the server.

Tracking events are passed to the server.

Couldn't test the endpoints using supertest, I think we need look into that..

It might make sense to split this PR, as there's a lot of noise.


packages/slice-machine/components/LoginModal/index.tsx
```diff
@@ -73,7 +70,7 @@ const LoginModal: React.FunctionComponent = () => {
         60
       );
 
-      void Tracker.get().identifyUser(shortId, intercomHash);
+      void Tracker.get().identifyUser();
```
Heyrows, line 73: I kinda like not having to manipulate user tokens and just having them in the back 😁

packages/slice-machine/lib/env/getEnv.ts
```diff
@@ -0,0 +1,128 @@
+import {
```
Heyrows: This is a server file only, why was it required to move it to the lib folder ?

packages/slice-machine/lib/mock/CustomType.ts
```diff
@@ -8,7 +8,7 @@ import {
   generateDocumentMock,
   renderDocumentMock,
 } from "@prismicio/mocks";
-import { CustomTypeMockConfig } from "@lib/models/common/MockConfig";
+import { CustomTypeMockConfig } from "../models/common/MockConfig";
```
Heyrows, line 11: Why was the IO folder moved from the server to the lib folder ?
It's function used only in the Server of Slice Machine so why put them on the shared folder ?
(couldn't notify on the direct file as it had been renamed)

packages/slice-machine/pages/_app.tsx
```diff
@@ -87,14 +84,8 @@ function MyApp({ Component, pageProps }: AppContext & AppInitialProps) {
...
     const tracking = getIsTrackingAvailable(state);
-    const repoName = getRepoName(state);
 
-    Tracker.get().initialize(
-      process.env.NEXT_PUBLIC_SM_UI_SEGMENT_KEY ||
-        "Ng5oKJHCGpSWplZ9ymB7Pu7rm0sTDeiG",
-      repoName,
-      tracking
-    );
+    Tracker.get().initialize(tracking);
```
Heyrows, line 88: Do we still need to initialize it ?
It tracking is disabled, the backend should be able to see it himself I guess.

packages/slice-machine/rollup-server.config.js
```diff
@@ -0,0 +1,23 @@
...
+  plugins: [
+    esbuild({
+      exclude: /node_modules/,
+      platform: "node",
+      sourceMap: false,
+      define: {
+        "process.env.NEXT_PUBLIC_SM_UI_SEGMENT_KEY": process.env
+          .NEXT_PUBLIC_SM_UI_SEGMENT_KEY
+          ? `"${process.env.NEXT_PUBLIC_SM_UI_SEGMENT_KEY}"`
+          : '""',
```
Heyrows: should this be empty or be the stage key ?

packages/slice-machine/scripts/start/startSMServer.ts
```diff
@@ -7,7 +7,7 @@ export function startSMServer(
 ) {
   const smServer: ChildProcessWithoutNullStreams = spawn(
     "node",
-    ["../../server/src/index.js"],
+    ["../../server/index.js"],
```
Heyrows, line 10: weird that this had to be changed, is it because rollup target the src folder directly ?

packages/slice-machine/server/src/api/common/error.ts
```diff
@@ -1,4 +1,4 @@
-import type { ApiError } from "@models/server/ApiResult";
+import type { ApiError } from "../../../../lib/models/server/ApiResult";
```
Heyrows, line 1: it's really weird that all those aliases have to be changed.
They simplified a lot the development, why were they modified ?

packages/slice-machine/server/src/api/index.ts
```diff
@@ -306,6 +307,15 @@ router.post(
   })
 );
 
+router.post(
+  "/s",
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises,
+  WithEnv(async (req, res): Promise<Express.Response> => {
+    await tracking(req);
```
Heyrows: should this be catch to always return 200 ?

packages/slice-machine/server/src/api/services/getEnv.ts
```diff
@@ -1,128 +1,5 @@
-import {
-  fromUrl,
-  parseDomain,
-  ParseResult,
-  ParseResultType,
-} from "parse-domain";
+import getEnv from "../../../../lib/env/getEnv";
```
Heyrows: I don't understand the need to move this file.

packages/slice-machine/server/src/api/tracking.ts
```diff
@@ -0,0 +1,63 @@
...
+const anonymousId = uuidv4();
+
+export function sendEvents(
+  analytics: Analytics,
+  event: TrackingEvents,
+  repositoryName: string,
+  userId?: string,
+  intercomHash?: string
+): void {
+  if (isGroupLibrariesEvent(event)) {
```
Heyrows, line 19: OPT: I think a switch would have ease the readability.

packages/slice-machine/server/src/api/tracking.ts
```diff
@@ -0,0 +1,63 @@
...
+      properties: event.props,
+      ...(userId ? { userId } : { anonymousId }),
+      context: { groupId: { Repository: repositoryName } },
+    });
+  }
+}
+
+export default async function handler(req: RequestWithEnv): Promise<void> {
+  const data = req.body as TrackingEvents;
+  if (isTrackingEvent(data)) {
```
Heyrows: not sure this one is needed.
If we trust the `req.body` enough to not validate it, then i guess we can trust this.

packages/slice-machine/src/hooks/useSMTracker.tsx
```diff
@@ -41,8 +41,7 @@ const useSMTracker = () => {
 
   // Handles if the user login/logout outside of the app.
   useEffect(() => {
-    if (shortId && intercomHash)
-      void Tracker.get().identifyUser(shortId, intercomHash);
+    if (shortId && intercomHash) void Tracker.get().identifyUser();
```
Heyrows, line 44: I don't think you need shortId & inside of the frontend now.
I don't think you need the identify call too as now if the user isn't logged in the backend won't find a userID and will use the anonymousID.

packages/slice-machine/server/src/api/tracking.ts
```diff
@@ -0,0 +1,63 @@
...
+
+import {
+  TrackingEvents,
+  isTrackingEvent,
+  isGroupLibrariesEvent,
+  isIdentifyUserEvent,
+} from "../../../src/tracking/types";
+import { RequestWithEnv } from "./http/common";
+
+const anonymousId = uuidv4();
```
Heyrows, line 11: This should be stored.
If not, the user will do some actions with an anonymous ID, some with another and when he will identify some events will be missing in his funnel.

packages/slice-machine/src/tracking/client.ts
```diff
@@ -0,0 +1,293 @@
...
+  CreateSlice,
+} from "./types";
+
+export class SMTracker {
+  #client = (event: TrackingEvents) =>
+    axios.post("/api/s", event).then(() => void 0);
+  #isTrackingActive = true;
+
+  initialize(isTrackingActive = true): void {
+    this.#isTrackingActive = isTrackingActive;
```
Heyrows, line 31: I would handle this on the backend because we can get it from the source of truth directly.

Heyrows: It would also save us having to manipulate this value in the frontend whenever something changes on the file system.

packages/slice-machine/src/tracking/types.ts
```diff
@@ -0,0 +1,200 @@
...
+    repoName: string;
+    manualLibsCount: number;
+    downloadedLibsCount: number;
+    npmLibsCount: number;
+    downloadedLibs: Array<string>;
+    slicemachineVersion: string;
+  };
+}
+
+export interface OpenVideoTutorials extends BaseTrackingEvent {
```
Heyrows, line 60: damn, thank you for taking the time to type all of that 👍  very clean.

packages/slice-machine/lib/env/getEnv.ts
MarcMcIntosh: it's used in the start script too :),
I noticed being in the output of the `build/server` directory after running `bundle`

packages/slice-machine/src/tracking/client.ts
MarcMcIntosh: hm, but is it worth sending extra events if they are not needed?

packages/slice-machine/server/src/api/tracking.ts
MarcMcIntosh: If he turns it off and back on again it won't be stored,

Interesting thing about node modules is that they are cached, so already work like singletons 
https://medium.com/@lazlojuly/are-node-js-modules-singletons-764ae97519af

also, in the build directory, this is all one file so we don't have to worry about the module being removed from memory ;)

packages/slice-machine/src/hooks/useSMTracker.tsx
MarcMcIntosh: In this hook, changes to short id and intercomHash trigger the event. To sure if they can go rom being defined to undefined, I'll look into it

packages/slice-machine/server/src/api/tracking.ts
MarcMcIntosh: yeah it's probably too much checking as the sendEvents also uses that function

packages/slice-machine/server/src/api/tracking.ts
MarcMcIntosh: I started with a switch, but then realised that there can be different events, I guess I can refactor i you want.

packages/slice-machine/scripts/start/startSMServer.ts
MarcMcIntosh: didn't have to be changed, but i made more sense as rollup only produces one file.

packages/slice-machine/rollup-server.config.js
MarcMcIntosh: could be either, as we have the staged key in where directly in the source, I wasn't sure if there was a reason for that so I left it in :)

packages/slice-machine/pages/_app.tsx
MarcMcIntosh: Technically we don't need initialise it, but it helps if they want to disable it.

packages/slice-machine/server/src/api/common/error.ts
MarcMcIntosh: module-alises didn't work with rollup + esbuild + typescript.
I tried a few things then figure it would be easier to remove it and re-evaluate what we're doing to support aliases as it's a bit messy having them configured in package.json, tsconfigs, babel and next.

packages/slice-machine/server/src/api/index.ts
MarcMcIntosh: tracking or withEnv?
may as-well always be a 200 so it'll stand out less for the user

packages/slice-machine/lib/mock/CustomType.ts
MarcMcIntosh: It kept showing up in the `build/server/src` directory when I ran bundle, so I guess it's required some where in the other builds and tsc places it there.

I'll recreate it and show you.

MarcMcIntosh: I think it's because it's called in `lib/utils/customtypes`

MarcMcIntosh: eeyup: right here https://github.com/prismicio/slice-machine/blob/master/packages/slice-machine/lib/utils/customTypes.ts

packages/slice-machine/src/hooks/useSMTracker.tsx
MarcMcIntosh: The hook be called on the first render, and again if shortId or intercomHash change, right?

packages/slice-machine/rollup-server.config.js
MarcMcIntosh: Done :)

packages/slice-machine/server/src/api/index.ts
MarcMcIntosh: done :)

packages/slice-machine/server/src/api/services/getEnv.ts
MarcMcIntosh: getEnv is shared by in the start/lib and here, so I swapped them around :)

packages/slice-machine/server/src/api/tracking.ts
MarcMcIntosh: done :)

packages/slice-machine/server/src/api/tracking.ts
MarcMcIntosh: done :)

packages/slice-machine/lib/mock/CustomType.ts
Heyrows: arf, feels like too many stuff are being added back to the lib when we tried to do the opposite

packages/slice-machine/server/src/api/common/error.ts
Heyrows: Ok, so taking care of the aliases in a next PR.

packages/slice-machine/server/src/api/index.ts
Heyrows: I was thinking about the tracking part since we await it, to make sure it does not return an error to the frontend

packages/slice-machine/server/src/api/common/error.ts
Heyrows: technical debt ticket* with a strategy.

packages/slice-machine/server/src/api/services/getEnv.ts
Heyrows: I checked, it's only the `getApplicationMode` that is actually being shared.
Maybe I miss some code, but if that's the case i'd prefer that we put back the getEnv and extract the `getApplicationMode` helper.

The GetEnv function is really server specific code that is kind of a central place in the server code where to find information. I'd prefer it stays there as much as possible.

packages/slice-machine/src/hooks/useSMTracker.tsx
Heyrows: Right, so this is used only for people that comes back to slice machine being logged in but our backend set the shortId and intercomHash.

Gosh this is such a edge case, i'm not even sure it would be useful anymore.

packages/slice-machine/server/src/api/tracking.ts
Heyrows: hmmmmm if you're sure i'd trust you but recommend heavy testing and that this is specified as something to test on non regression.

packages/slice-machine/src/tracking/client.ts
Heyrows: discussed on slack, we'll keep it that way just in case it slows down the server with all the extra calls.

packages/slice-machine/server/src/api/tracking.ts
MarcMcIntosh: I could add a test or that :)

packages/slice-machine/server/src/api/tracking.ts
```diff
@@ -45,11 +45,11 @@ export function sendEvents(
 
 export default async function handler(req: RequestWithEnv): Promise<void> {
   const data = req.body as TrackingEvents;
-  if (isTrackingEvent(data)) {
-    const analytics = new Analytics(
-      process.env.NEXT_PUBLIC_SM_UI_SEGMENT_KEY ||
-        "Ng5oKJHCGpSWplZ9ymB7Pu7rm0sTDeiG"
-    );
+  const trackingEnabled =
```
Heyrows, line 48: Would that work ?
```suggestion
  const trackingEnabled = typeof req.env.manifest.tracking == 'boolean' ? req.env.manifest.tracking : false
```

packages/slice-machine/rollup-server.config.js
```diff
@@ -15,7 +15,7 @@ export default {
         "process.env.NEXT_PUBLIC_SM_UI_SEGMENT_KEY": process.env
```
Heyrows: I'd remove the `NEXT` from it's name, it was used a convention for nextjs to pick it up i think.

packages/slice-machine/server/src/api/tracking.ts
MarcMcIntosh: if it's undefined (default), then it would be false, we could change that to true :)

packages/slice-machine/server/src/api/index.ts
MarcMcIntosh: it only waits because this function needed to be async, although it might throw if some data is missing :)

packages/slice-machine/server/src/api/services/getEnv.ts
MarcMcIntosh: good point, I'll extract `getApplicationMode` and put it in the lib

packages/slice-machine/lib/mock/CustomType.ts
MarcMcIntosh: yeah i think it's mostly the start script that's sharing stuff with the server + lib.

You can see the shared files by deleting the build dir and telling tsc to compile that ...module?

This one is for the start script
Screenshot 2022-07-28 at 18 37 28

packages/slice-machine/server/src/api/common/error.ts
MarcMcIntosh: Will have to think of a strategy first and see if it meets our requirements.

packages/slice-machine/server/src/api/index.ts
MarcMcIntosh: We only await it because it need to be async to work with `withEnv` 

..wait I just though of something..... 👍🏻

packages/slice-machine/src/hooks/useSMTracker.tsx
MarcMcIntosh: Well that and if the condition is removed the shortId and intercomHash will fire... even if the properties are undeffined

packages/slice-machine/server/src/api/tracking.ts
MarcMcIntosh: ^maybe

packages/slice-machine/server/src/api/services/getEnv.ts
```diff
@@ -1,5 +1,123 @@
...
+import {
+  fromUrl,
+  parseDomain,
+  ParseResult,
+  ParseResultType,
+} from "parse-domain";
 
-export default async function handler() {
-  return getEnv();
+import { Models } from "@slicemachine/core";
```
Heyrows: I'm guessing it must've been a pain to move it back here but thank you.
I really feel it does belong in the server.

packages/slice-machine/server/src/api/tracking.ts
Heyrows: ah right 😂

Heyrows: I guess it's not that important.

packages/slice-machine/tests/server/tracking.test.ts
```diff
@@ -1,22 +1,22 @@
...
+    identify: jest.fn().mockImplementation(() => Promise.resolve()),
+    group: jest.fn().mockImplementation(() => Promise.resolve()),
+  };
+});
 
 jest.mock("uuid", () => {
   return {
     __esModule: true,
-    v4: () => "uuid",
+    v4: jest.fn().mockReturnValueOnce("uuid").mockReturnValue("bad"),
```
Heyrows, line 19: funny way to make sure it's only called once, i like it 😂

Heyrows (approved):
It looks good, i'll let the QA see if everything is working fine.
Thanks for this PR 😁

packages/slice-machine/rollup-server.config.js
MarcMcIntosh: We could but we would also have to make an new env variable, some where.... I'm not sure where though :)

PR #539, opened 2022-07-28
[Tech debt] - Add dev env api endpoints

Heyrows:
⚠️  Disclaimer, the ACL provider to upload screenshots doesn't seems to be working but is not a main feature for running the E2E testing so it should be enough for now.

## Types of changes

## Checklist:


packages/slice-machine/server/src/api/services/getEnv.ts
```diff
@@ -39,17 +33,18 @@ function validate(config: Models.Manifest): ConfigErrors {
...
+    const url = new URL(apiEndpoint);
+    const host = url.host; // myrepo.prismic.io
+    const repository = host.split(".")[0]; // myrepo
+    return repository;
+  } catch (e) {
+    // Should already be covered by the start script.
+    console.error(
+      "It seems your repository API endpoint in your SM.json is wrong. Please start machine again to get instructions on how to fix it."
+    );
+    process.exit(0);
```
MarcMcIntosh, line 50: pretty sure we don't want the server to stop running if the urls not valid

packages/client/src/models/ApisEndpoints.ts
```diff
@@ -21,3 +23,33 @@ export const StageApisEndpoints: ApisEndpoints = {
...
+    console.error(`
+      It seems you're trying to use Slice Machine in DEV mode without all the environment variables
+      Here are the required environment variables
+      - Authentication API: ${authenticationApi ? "👍" : "⛔️"}
+      - Custom Type API: ${modelsApi ? "👍" : "⛔️"}
+      - Users API: ${usersApi ? "👍" : "⛔️"}
+      - Acl Provider API: ${aclApi ? "👍" : "⛔️"}
+    `);
+    // Stopping slice machine
+    process.exit(0);
```
MarcMcIntosh, line 45: I guess we're the only ones that will to deal with this exiting

MarcMcIntosh (changes requested):
are the modifications to the package-locks needed?

Edit: Yes, it's just the client's version that has been updated in the package locks.

packages/slice-machine/server/src/api/services/getEnv.ts
Heyrows: It's something only us would see, a user would have to have "wroom.test" in his API endpoint 😅.
i figured as devs, we might want to know as soon as possible if something is broken in our dev env.

What do you think should be the correct behaviour ?

MarcMcIntosh: pretty sure etEnv is used in a few places, do we want it to exit if the apiEndpointIs wrong?

MarcMcIntosh: I checked this an it looks like it should crash 
https://miro.com/app/board/uXjVOBPxAAY=/

MarcMcIntosh (approved):

PR #549, opened 2022-08-04
Mm/rename slice sync

MarcMcIntosh:
## Types of changes

## Checklist:

### DIFF
--- a/packages/slice-machine/src/modules/selectedCustomType/sagas.ts
+++ b/packages/slice-machine/src/modules/selectedCustomType/sagas.ts
@@ -5,7 +5,7 @@ import { withLoader } from "../loading";
 import { LoadingKeysEnum } from "../loading/types";
 import { pushCustomTypeCreator, saveCustomTypeCreator } from "./actions";
 import { selectCurrentCustomType, selectCurrentMockConfig } from "./index";
-import { pushCustomType, saveCustomType } from "../../../src/apiClient";
+import { pushCustomType, saveCustomType } from "../../apiClient";
 import axios from "axios";
 import { modalOpenCreator } from "../modal";
 import { ModalKeysEnum } from "../modal/types";
--- a/packages/slice-machine/src/modules/slices/index.ts
+++ b/packages/slice-machine/src/modules/slices/index.ts
@@ -296,6 +296,7 @@ export const renamedComponentUI = (
       model.name,
       newName
     ),
+    __status: LibStatus.Modified,
   };
 };
 
--- a/packages/slice-machine/src/modules/slices/index.ts
+++ b/packages/slice-machine/src/modules/slices/index.ts
@@ -148,18 +148,7 @@ export const slicesReducer: Reducer<SlicesStoreType | null, SlicesActions> = (
     }
     case getType(pushSliceCreator.success): {
       const newComponentUI = action.payload.component;
-
-      const newRemoteSlices = [...state.remoteSlices];
-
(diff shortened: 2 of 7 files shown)


packages/slice-machine/src/modules/slices/index.ts
```diff
@@ -148,18 +148,7 @@ export const slicesReducer: Reducer<SlicesStoreType | null, SlicesActions> = (
...
-      const remoteSliceIndex = state.remoteSlices.findIndex(
-        ({ id }) => id === newComponentUI.model.id
-      );
-
-      if (remoteSliceIndex !== -1) {
-        newRemoteSlices[remoteSliceIndex] = newComponentUI.model;
-      } else {
-        newRemoteSlices.push(newComponentUI.model);
-      }
+      const newRemoteSlices = action.payload.remoteSlices;
```
sintinilorenzo: why are we changing this code and adding a new payload which requires a new api call with `yield call(getState))`? what was wrong in the old logic?

packages/slice-machine/src/modules/slices/index.ts
```diff
@@ -296,6 +289,7 @@ export const renamedComponentUI = (
       model.name,
       newName
     ),
+    __status: LibStatus.Modified,
```
sintinilorenzo, line 302: nice!

packages/slice-machine/src/modules/selectedCustomType/sagas.ts
```diff
@@ -5,7 +5,7 @@ import { withLoader } from "../loading";
 import { LoadingKeysEnum } from "../loading/types";
 import { pushCustomTypeCreator, saveCustomTypeCreator } from "./actions";
 import { selectCurrentCustomType, selectCurrentMockConfig } from "./index";
-import { pushCustomType, saveCustomType } from "../../../src/apiClient";
+import { pushCustomType, saveCustomType } from "../../apiClient";
```
sintinilorenzo, line 8: :)

packages/slice-machine/src/modules/slices/index.ts
MarcMcIntosh: remote slices was not being updated, so when pushing a new slice it was not updated in the remoteSlice property

sintinilorenzo: yeah but why? didn't the code 

```
      const newRemoteSlices = [...state.remoteSlices];
      const remoteSliceIndex = state.remoteSlices.findIndex(
        ({ id }) => id === newComponentUI.model.id
      );

      if (remoteSliceIndex !== -1) {
        newRemoteSlices[remoteSliceIndex] = newComponentUI.model;
      } else {
        newRemoteSlices.push(newComponentUI.model);
      }
```

work? We already have the remoteSlices and what that code does it add another one, or change an existing one

sintinilorenzo: I think the actual bug was that we are returning `return { ...state, libraries: newLibraries };` instead of `return { ...state, libraries: newLibraries, remoteSlices: newRemoteSlices };`

we were computing the newRemoteSlices but not returning them. It's not necessary to add a new API call and get the state again

MarcMcIntosh: are those remote slices the remote slices in the state before making the request?

I'll see what. i can do

MarcMcIntosh: done?

sintinilorenzo (approved):

PR #552, opened 2022-08-05
add bash script to test the slice machine ini

sintinilorenzo:
## Types of changes

## Checklist:


packages/init/src/steps/choose-or-create-a-repository.ts
```diff
@@ -127,6 +129,12 @@ export async function chooseOrCreateARepository(
     );
   if (isPreSelectedValid) return preSelectedRepository;
 
+  if (isE2e) {
+    const domainName = `CYPRESS-REPO-${uuidv4()}`;
```
sintinilorenzo, line 133: might be nice here if the repository name is always the same e.g. `CYPRESS-REPO` but, before calling `createRepository` we call a `deleteRepository` method (but idk if it exists or which one it is)

chengoconnell (approved):

cypress/plugins/addAuth.js
```diff
@@ -0,0 +1,21 @@
+const fs = require("fs");
+const os = require("os");
+const path = require("path");
+const axios = require("axios");
+
+axios
+  .post("https://wroom.io/authentication/signin", {
+    email: "cypress@prismic.io",
+    password: "AhK9yohhie9ahyohn3w",
```
MarcMcIntosh: These should be an env var, we can ask mehdi and michel to and them to the repository settings, and a check to let the user know to set them when running locally.

cypress/integration/slices/00-create.spec.js
```diff
@@ -1,53 +1,25 @@
-import path from "path";
-
 describe("Create Slices", () => {
   const sliceName = "TestSlice";
   const editedSliceName = "TestSlice2";
-  const lib = "slices--ecommerce"; // name of the first lib of the next project.
-  const pathToLib = path.join("e2e-projects", "next", "slices", "ecommerce");
-  const generatedPath = path.join(
```
MarcMcIntosh, line 8: Probably keep these joins, as it makes it easier to add other OSs to the tests :)

packages/init/src/steps/choose-or-create-a-repository.ts
```diff
@@ -127,6 +129,12 @@ export async function chooseOrCreateARepository(
     );
   if (isPreSelectedValid) return preSelectedRepository;
 
+  if (isE2e) {
```
MarcMcIntosh, line 132: Is there another way to do this?

package.json
```diff
@@ -30,7 +31,8 @@
     "husky": "^7.0.2",
     "lerna": "^3.22.1",
     "lint-staged": "^11.1.2",
-    "prettier": "^2.4.1"
+    "prettier": "^2.4.1",
+    "slice-machine-ui": "^0.4.2"
```
MarcMcIntosh: why is slice-machine-ui now a dep?

MarcMcIntosh (changes requested):
Couple of changes requested

packages/init/src/steps/choose-or-create-a-repository.ts
MarcMcIntosh: like by passing the repo name as an argument?

package.json
sintinilorenzo: ah that's my bad it should not be. Think i accidentaly ran a command in the terminal, will remove

packages/init/src/steps/choose-or-create-a-repository.ts
sintinilorenzo: don't think so, it's to prevent having the user command line popup to ask the user to enter the name of repo

cypress/integration/slices/00-create.spec.js
sintinilorenzo: yes good point, will change

cypress/plugins/addAuth.js
sintinilorenzo: ah very true! will ask them

packages/init/src/steps/choose-or-create-a-repository.ts
sintinilorenzo: even if we pass the repo name as an argument (which we already can do), it checks if it's a valid existing repo in prismic. And if it's not it will open the terminal user input

MarcMcIntosh: really? if the repo name is available it should create the repo... :/

packages/init/src/steps/choose-or-create-a-repository.ts
MarcMcIntosh: there's a 30 character limit, I guess this should be fine, but also letter's need to be lowercase

cypress/integration/slices/00-create.spec.js
MarcMcIntosh: Also, can we keep the e2e tests in the 2e tests directory?

sintinilorenzo: what do you mean?

packages/init/src/steps/choose-or-create-a-repository.ts
sintinilorenzo: image
that's what I understand from the code yeah? You can pass a `preSelectedRepository` with the flag, and if it has the same name as one in prismic it will be used, otherwise it will show the `await promptForRepoDomain` later

packages/init/src/steps/choose-or-create-a-repository.ts
sintinilorenzo: changed to lowercase

packages/init/src/steps/choose-or-create-a-repository.ts
MarcMcIntosh: ah yes, that's for when --repository is passed down following the instructions,

I guess we could add extra logic to create the repo if it does not exist?

cypress/integration/slices/00-create.spec.js
MarcMcIntosh: we already have a directory for e2e-projects, shouldn't we use it?

cypress/plugins/addAuth.js
```diff
@@ -0,0 +1,21 @@
+const fs = require("fs");
+const os = require("os");
+const path = require("path");
+const axios = require("axios");
+
+axios
+  .post("https://wroom.io/authentication/signin", {
```
MarcMcIntosh: hmm, wroom.io isn't available to the pubic, maybe the url should be an env var, or depend on a "mode" flag?

cypress/integration/slices/00-create.spec.js
MarcMcIntosh: ie: `e2e-projects/next-app`

packages/init/src/steps/choose-or-create-a-repository.ts
sintinilorenzo: don't think we should in my opinion, that would mean we'll change the current behaviour of slice-machine which is not something we have discussed or intended. This way we are only adding logic for the e2e scenario, without touching/changing existing logic

cypress/integration/slices/00-create.spec.js
sintinilorenzo: ah okay, moved it there 👍🏼

cypress/plugins/addAuth.js
sintinilorenzo: what do you mean by that? The wroom.io is mentioned in quite a lot of places within the code (51 instances) 
image

cypress/integration/slices/00-create.spec.js
MarcMcIntosh: thanks :)

packages/init/src/steps/choose-or-create-a-repository.ts
MarcMcIntosh: hmm, good point 🧐

cypress/plugins/addAuth.js
MarcMcIntosh: yes, it's for prismic developers, our dev env is wroom.test, staged is wroom.io. and production is prismic.io.

Normal users don't have access to wroom.io (although, one managed it once but it should be blocked off to the public)

MarcMcIntosh (approved):

PR #553, opened 2022-08-08
Aoc/740 custom types status with logic

chengoconnell:
## Types of changes

## Checklist:

### DIFF
--- a/packages/core/src/models/CustomType/index.ts
+++ b/packages/core/src/models/CustomType/index.ts
@@ -16,6 +16,7 @@ export const CustomTypeSM = t.exact(
     }),
     t.partial({
       hash: t.string,
+      __status: t.string,
     }),
   ])
 );
--- a/packages/slice-machine/components/Badge/StatusBadgeWithTooltip.tsx
+++ b/packages/slice-machine/components/Badge/StatusBadgeWithTooltip.tsx
@@ -3,7 +3,7 @@ import { Badge, Flex, Text } from "theme-ui";
 import { FrontEndCustomType } from "../../src/modules/availableCustomTypes/types";
 import { CustomTypeStatus } from "../../src/modules/selectedCustomType/types";
 
-const StatusEnumToDisplayNameAndTooltip = (status: CustomTypeStatus) => {
+const StatusEnumToDisplayNameAndTooltip = (status?: string) => {
   switch (status) {
     case CustomTypeStatus.New:
       return {
@@ -29,6 +29,12 @@ const StatusEnumToDisplayNameAndTooltip = (status: CustomTypeStatus) => {
         statusTooltip:
           "Data from the remote repository could not be fetched (unknown error).",
       };
+    default:
+      return {
+        statusDisplayName: "Unknown",
+        statusTooltip:
+          "Data from the remote repository could not be fetched (unknown error).",
+      };
   }
 };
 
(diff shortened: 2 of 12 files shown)


packages/slice-machine/components/Badge/StatusBadgeWithTooltip.tsx
```diff
@@ -40,16 +46,15 @@ export const StatusBadgeWithTooltip: React.FC<StatusBadgeWithTooltipProps> = ({
   customType,
 }) => {
   const { statusDisplayName, statusTooltip } =
-    StatusEnumToDisplayNameAndTooltip(CustomTypeStatus.New);
+    StatusEnumToDisplayNameAndTooltip(customType.local.__status);
```
sintinilorenzo: any reason `StatusEnumToDisplayNameAndTooltip` starts with a capital letter?

packages/core/src/models/CustomType/index.ts
```diff
@@ -16,6 +16,7 @@ export const CustomTypeSM = t.exact(
     }),
     t.partial({
       hash: t.string,
+      __status: t.string,
```
MarcMcIntosh, line 19: is this sent to the api or extracted out?

packages/slice-machine/server/src/api/custom-types/getCustomTypeStatus.ts
```diff
@@ -1,14 +1,12 @@
...
 
 export const getCustomTypeStatus = (
-  localCustomType?: CustomTypeSM,
+  localCustomType: CustomTypeSM,
   remoteCustomType?: CustomTypeSM
 ): CustomTypeStatus => {
-  if (!localCustomType && !remoteCustomType) return CustomTypeStatus.Unknown;
-
-  if (!remoteCustomType) return CustomTypeStatus.New;
+  if (!localCustomType || !remoteCustomType) return CustomTypeStatus.New;
```
sintinilorenzo: localCustomType is defined so will never be falsy

packages/slice-machine/src/modules/availableCustomTypes/index.ts
```diff
@@ -105,6 +106,8 @@ export const availableCustomTypesReducer: Reducer<
       };
     }
     case getType(createCustomTypeCreator.success): {
+      action.payload.newCustomType.__status = CustomTypeStatus.New;
```
sintinilorenzo: better to set the `__status` in the saga before the `createCustomTypeCreator.success` action is dispatched. Ideally in the `createCustomType` method

packages/slice-machine/src/modules/selectedCustomType/reducer.ts
```diff
@@ -58,7 +59,14 @@ export const selectedCustomTypeReducer: Reducer<
 
       return {
         ...state,
-        initialModel: state.model,
+        model: {
```
sintinilorenzo, line 62: you didn't update the status for the rename redux action

packages/slice-machine/src/modules/availableCustomTypes/index.ts
```diff
@@ -105,6 +106,8 @@ export const availableCustomTypesReducer: Reducer<
...
+
       const normalizedNewCustomType = normalizeFrontendCustomType(
         action.payload.newCustomType
       );
```
sintinilorenzo: need to update the status for the `renameCustomTypeCreator.success` redux action

sintinilorenzo: also need to update the status when `saveCustomTypeCreator.success` and `pushCustomTypeCreator.success` are called.

Right now if you push a custom type you update its status on the selectedCustomType store but not on the availableCustomTypes, so I think if you push and navigate back to the main page the status will not have updated

packages/core/src/models/CustomType/index.ts
chengoconnell: extracted out i believe. This info won't be in prismic.io at all

packages/slice-machine/src/modules/availableCustomTypes/index.ts
```diff
@@ -114,13 +122,49 @@ export const availableCustomTypesReducer: Reducer<
         ...normalizedNewCustomType,
       };
     }
+
+    case getType(saveCustomTypeCreator.success): {
+      if (!state) return state;
+      const customTypeId = action.payload.customTypeId;
```
sintinilorenzo: think we should pass to the payload the actual `customType`, not just its id. Since the CT we are saving is now different than the outdated one in the availableCustomType.

so where you have the `saveCustomTypeCreator.success({ customTypeId: currentCustomType.id })` you can do `saveCustomTypeCreator.success({ customType: currentCustomType})`, and then you can use this payload to just set it (making sure to upodate the __status too)

sintinilorenzo (approved):

PR #565, opened 2022-08-15
Ls/734 add auth error page

sintinilorenzo:
## Types of changes

## Checklist:

### DIFF
--- a/packages/slice-machine/components/ChangesEmptyPage/AuthErrorPage.tsx
+++ b/packages/slice-machine/components/ChangesEmptyPage/AuthErrorPage.tsx
@@ -1,0 +1,32 @@
+import React from "react";
+import { Button, Flex, Text } from "theme-ui";
+
+export const AuthErrorPage = () => {
+  return (
+    <Flex sx={{ justifyContent: "center", alignItems: "center", flex: 1 }}>
+      <Flex
+        sx={{
+          flexDirection: "column",
+          alignItems: "center",
+          transform: "translateY(-50%)",
+        }}
+      >
+        <Text
+          sx={{
+            fontWeight: 600,
+            fontSize: "24px",
+            lineHeight: "32px",
+            mb: "8px",
+            mt: "2px",
+          }}
+        >
+          Could not access your repository
+        </Text>
+        <Text sx={{ fontSize: "13px", lineHeight: "24px" }}>
+          You need to login to Prismic in order to push your changes.
+        </Text>
+        <Button sx={{ mt: "24px" }}>Login to Prismic</Button>
+      </Flex>
+    </Flex>
+  );
+};
--- a/packages/slice-machine/components/ChangesEmptyPage/AuthErrorPage.tsx
+++ b/packages/slice-machine/components/ChangesEmptyPage/AuthErrorPage.tsx
@@ -1,7 +1,9 @@
+import useSliceMachineActions from "@src/modules/useSliceMachineActions";
 import React from "react";
 import { Button, Flex, Text } from "theme-ui";
 
 export const AuthErrorPage = () => {
+  const { openLoginModal } = useSliceMachineActions();
   return (
     <Flex sx={{ justifyContent: "center", alignItems: "center", flex: 1 }}>
       <Flex
@@ -25,7 +27,9 @@ export const AuthErrorPage = () => {
         <Text sx={{ fontSize: "13px", lineHeight: "24px" }}>
           You need to login to Prismic in order to push your changes.
         </Text>
-        <Button sx={{ mt: "24px" }}>Login to Prismic</Button>
+        <Button sx={{ mt: "24px" }} onClick={() => openLoginModal()}>
+          Login to Prismic
+        </Button>
       </Flex>
     </Flex>
   );
--- a/packages/slice-machine/lib/models/server/ServerState.ts
+++ b/packages/slice-machine/lib/models/server/ServerState.ts
@@ -3,6 +3,7 @@ import { FrontEndEnvironment } from "../common/Environment";
 import { LibraryUI } from "../common/LibraryUI";
 import { SliceSM } from "@slicemachine/core/build/models";
 import { CustomTypeSM } from "@slicemachine/core/build/models/CustomType";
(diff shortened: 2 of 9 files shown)


packages/slice-machine/src/modules/userContext/index.ts
```diff
@@ -85,7 +92,32 @@ export const userContextReducer: Reducer<
...
+): AuthStatus => {
+  switch (clientError?.status) {
+    case undefined: {
+      return AuthStatus.AUTHORIZED;
+    }
+    case 403: {
+      return AuthStatus.UNAUTHORIZED;
+    }
+    case 401: {
+      return AuthStatus.FORBIDDEN;
```
sintinilorenzo, line 117: so the usage of 401 and 403 is swapped all around the app. The backend returns a 403 for unauthorised requests, and a 401 for forbidden ones. 
These states are also incorrectly (or I guess correctly to make it work ahah) in the frontend to display the right messages: 
image
image

I think we should refactor the codebase to swap them but I think we should do so in a separate refactoring ticket (since it involves changing code in multiple other places as well as changing the wroom api responses). So for the time being I kept the consistent usage of 403 being unauthorised

packages/slice-machine/pages/changes.tsx
```diff
@@ -26,6 +33,12 @@ const changes: React.FunctionComponent = () => {
     if (!isOnline) {
       return <OfflinePage />;
     }
+    if (
+      authStatus === AuthStatus.UNAUTHORIZED ||
+      authStatus === AuthStatus.FORBIDDEN
+    ) {
+      return <AuthErrorPage />;
```
sintinilorenzo, line 40: think we can pass the authStatus to the `<AuthErrorPage />` component in order to give more info to the user:
- AuthStatus.UNAUTHORIZED => tell the user he needs to login and display a login button
-  AuthStatus.FORBIDDEN => tell the user he does not have access to the repo and do not display the login button

unclear from the design so need to check with Come and/or Louis

packages/slice-machine/components/ChangesEmptyPage/AuthErrorPage.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { Button, Flex, Text } from "theme-ui";
+
+export const AuthErrorPage = () => {
+  return (
+    <Flex sx={{ justifyContent: "center", alignItems: "center", flex: 1 }}>
+      <Flex
+        sx={{
+          flexDirection: "column",
+          alignItems: "center",
+          transform: "translateY(-50%)",
```
chengoconnell, line 13: interesting! what does this do?

chengoconnell (approved):

packages/slice-machine/components/ChangesEmptyPage/AuthErrorPage.tsx
sintinilorenzo: moves the content of the page slightly up (by half of its height so that it does not reply on sizes or pixels). Was because the figma design does not have the page content centered, and this is what me and Louis though looked best when pairing on the design :)

packages/slice-machine/src/modules/userContext/index.ts
```diff
@@ -85,7 +92,32 @@ export const userContextReducer: Reducer<
...
+        authStatus: getAuthStatus(action.payload.clientError),
+      };
+    }
     default:
       return state;
   }
 };
+
+const getAuthStatus = (
+  clientError: ErrorWithStatus | undefined
```
MarcMcIntosh, line 107: would `clientError?: ErrorWithStatus` also work, or does it need to explicitly receive undefined?

packages/slice-machine/lib/models/server/ServerState.ts
```diff
@@ -14,4 +15,5 @@ export default interface ServerState {
   customTypes: ReadonlyArray<CustomTypeSM>;
   remoteCustomTypes: ReadonlyArray<CustomTypeSM>;
   remoteSlices: ReadonlyArray<SliceSM>;
+  clientError: ErrorWithStatus | undefined;
```
MarcMcIntosh: did `clientError?: ErrorWithStatus` not work?

packages/slice-machine/pages/changes.tsx
MarcMcIntosh: defiantly check with them :)

packages/slice-machine/src/modules/userContext/index.ts
MarcMcIntosh: Cool add it to those notes to the tech debt board :)

MarcMcIntosh (changes requested):

packages/slice-machine/src/modules/userContext/index.ts
sintinilorenzo: done :)

packages/slice-machine/pages/changes.tsx
sintinilorenzo: send them a message

packages/slice-machine/src/modules/userContext/index.ts
sintinilorenzo: would work both ways. 

But the way I see it `clientError?: ErrorWithStatus` means the clientError argument is optional and does not necessarily have to be passed on. `clientError: ErrorWithStatus | undefined` means that the method takes an argument (which could be undefined)

Think the method `getAuthStatus` should always take a parameter because it doesn't make sense to call  `getAuthStatus()`

packages/slice-machine/lib/models/server/ServerState.ts
sintinilorenzo: yepp, changed to clientError?: ErrorWithStatus 👍🏼

packages/slice-machine/src/modules/userContext/index.ts
MarcMcIntosh: okay :)

MarcMcIntosh (approved):

Heyrows (approved):

PR #566, opened 2022-08-15
Aoc/735 unsynced slices take screenshot

chengoconnell:
## Types of changes

## Checklist:

### DIFF
--- a/packages/slice-machine/lib/builders/SliceBuilder/SideBar/components/ImagePreview.tsx
+++ b/packages/slice-machine/lib/builders/SliceBuilder/SideBar/components/ImagePreview.tsx
@@ -16,11 +16,11 @@ interface ImagePreviewProps {
   src?: string;
   onScreenshot: () => void;
   imageLoading: boolean;
-  onHandleFile: (file: File) => void;
+  onHandleFile?: (file: File) => void;
   preventScreenshot: boolean;
 }
 
-const ImagePreview: React.FC<ImagePreviewProps> = ({
+export const ImagePreview: React.FC<ImagePreviewProps> = ({
   src,
   onScreenshot,
   imageLoading,
@@ -35,7 +35,7 @@ const ImagePreview: React.FC<ImagePreviewProps> = ({
 
   const handleFile = (file: File | undefined) => {
     if (inputFile?.current) {
-      file && onHandleFile(file);
+      file && onHandleFile && onHandleFile(file);
       inputFile.current.value = "";
     }
   };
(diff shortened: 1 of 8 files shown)


packages/slice-machine/lib/builders/SliceBuilder/SideBar/components/ImagePreview.tsx
```diff
@@ -16,11 +16,11 @@ interface ImagePreviewProps {
   src?: string;
   onScreenshot: () => void;
   imageLoading: boolean;
-  onHandleFile: (file: File) => void;
```
Heyrows, line 19: why would this becomes optional ?
On both screen that use it I believe we do have a custom screenshot button.

packages/slice-machine/lib/models/ui/Slice/index.tsx
```diff
@@ -158,6 +163,39 @@ const SliceDescription = ({
...
+  const { isSimulatorAvailableForFramework } = useSelector(
+    (state: SliceMachineStoreType) => ({
+      isSimulatorAvailableForFramework:
+        selectIsSimulatorAvailableForFramework(state),
+    })
+  );
+  const [data, setData] = React.useState<{ imageLoading: boolean }>({
+    imageLoading: false,
+  });
+  if (
```
Heyrows: i think this can be removed 😂

packages/slice-machine/lib/models/ui/Slice/index.tsx
```diff
@@ -201,11 +239,15 @@ export const SharedSlice = {
           // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-argument
           sx={bordered ? borderedSx(sx) : defaultSx(sx)}
         >
-          <SliceThumbnail
-            withShadow={false}
-            screenshotUrl={screenshotUrl}
-            heightInPx={thumbnailHeightPx}
-          />
+          {wrapperType === WrapperType.changesPage ? (
```
Heyrows, line 236: if we have the same condition on the whole content of the card, maybe we could do something about it.

packages/slice-machine/lib/builders/SliceBuilder/SideBar/components/ImagePreview.tsx
sintinilorenzo: it's because atm there is still some uncertainty on how to handle the "custom screenshot" button. So atm in the changes page we are not rendering it. The button is conditionally rendered if the `onHandleFile` method is defined.
image

packages/slice-machine/lib/models/ui/Slice/index.tsx
sintinilorenzo: ahah oops, removed

packages/slice-machine/lib/models/ui/Slice/index.tsx
sintinilorenzo: ah yes didn't realise there were 2 consecutive `WrapperType.changesPage` checks. I have joined them together :). Think it might be fine to keep it like this tho since a lot of component has logic needed in both cases

Heyrows (approved):
Cool 👍 , wouldn't have called the wrapper "changesPages", seems too specific.

PR #570, opened 2022-08-18
feature(): add ct status when user is offline

chengoconnell:
## Types of changes

## Description
- If user is offline -> display `unknown (no internet connection)` label

https://user-images.githubusercontent.com/69536010/185614165-30a61f1b-faf6-4345-9b81-4f113046adab.mov

- If user is disconnected from Prismic -> display `unknown (not connected to Prismic)` label
   - If user attempts to save a custom type while disconnected -> keep status the same and display `unknown (not connected to Prismic)` label.
   

https://user-images.githubusercontent.com/69536010/185616703-4844b22e-a585-433b-98cf-2734ef50ad80.mov

## Checklist:

### DIFF
--- a/packages/slice-machine/components/Badge/StatusBadgeWithTooltip.tsx
+++ b/packages/slice-machine/components/Badge/StatusBadgeWithTooltip.tsx
@@ -2,6 +2,7 @@ import ReactTooltip from "react-tooltip";
 import { Badge, Flex, Text } from "theme-ui";
 import { CustomTypeSM } from "@slicemachine/core/build/models/CustomType";
 import { CustomTypeStatus } from "../../src/modules/selectedCustomType/types";
+import { useNetwork } from "@src/hooks/useNetwork";
 
 const statusEnumToDisplayNameAndTooltip = (status?: string) => {
   switch (status) {
@@ -23,7 +24,19 @@ const statusEnumToDisplayNameAndTooltip = (status?: string) => {
         statusTooltip:
           "This Custom Type is in sync with the remote repository.",
       };
-    case CustomTypeStatus.Unknown:
+    case CustomTypeStatus.UnknownOffline:
+      return {
+        statusDisplayName: "Unknown",
+        statusTooltip:
+          "Data from the remote repository could not be fetched (no internet connection).",
+      };
+    case CustomTypeStatus.UnknownDisconnected:
+      return {
+        statusDisplayName: "Unknown",
+        statusTooltip:
+          "Data from the remote repository could not be fetched (not connected to Prismic).",
+      };
+    case CustomTypeStatus.UnknownDefault:
       return {
         statusDisplayName: "Unknown",
         statusTooltip:
@@ -45,6 +58,12 @@ interface StatusBadgeWithTooltipProps {
 export const StatusBadgeWithTooltip: React.FC<StatusBadgeWithTooltipProps> = ({
   customType,
 }) => {
+  const isOnline = useNetwork();
+
+  customType.__status = !isOnline
+    ? CustomTypeStatus.UnknownOffline
+    : customType.__status;
+
   const { statusDisplayName, statusTooltip } =
     statusEnumToDisplayNameAndTooltip(customType.__status);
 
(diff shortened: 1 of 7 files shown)


packages/slice-machine/components/Badge/StatusBadgeWithTooltip.tsx
```diff
@@ -23,7 +24,19 @@ const statusEnumToDisplayNameAndTooltip = (status?: string) => {
...
+        statusTooltip:
+          "Data from the remote repository could not be fetched (no internet connection).",
+      };
+    case CustomTypeStatus.UnknownDisconnected:
+      return {
+        statusDisplayName: "Unknown",
+        statusTooltip:
+          "Data from the remote repository could not be fetched (not connected to Prismic).",
+      };
+    case CustomTypeStatus.UnknownDefault:
```
Heyrows: i'd remove that case, it seems to be the same as the default case :)

packages/slice-machine/components/Badge/StatusBadgeWithTooltip.tsx
```diff
@@ -45,6 +58,12 @@ interface StatusBadgeWithTooltipProps {
 export const StatusBadgeWithTooltip: React.FC<StatusBadgeWithTooltipProps> = ({
   customType,
 }) => {
+  const isOnline = useNetwork();
+
+  customType.__status = !isOnline
```
Heyrows: Create a new object please 🙏 , modifying the existing one could create side effects.

packages/slice-machine/lib/builders/CustomTypeBuilder/Layout/Header/primaryButton.tsx
```diff
@@ -51,7 +51,9 @@ export const PrimaryButton: React.FC<PrimaryButtonProps> = ({
     [
       CustomTypeStatus.New,
       CustomTypeStatus.Modified,
-      CustomTypeStatus.Unknown,
+      CustomTypeStatus.UnknownDefault,
```
Heyrows: judging by the previus code we don't need this one

Heyrows: I think the only condition needs is
```suggestion
     customTypeStatus != CustomTypeStatus.synced
```

packages/slice-machine/src/modules/availableCustomTypes/index.ts
```diff
@@ -177,14 +182,18 @@ export const availableCustomTypesReducer: Reducer<
...
+        newLocalCustomType.__status === CustomTypeStatus.UnknownDisconnected;
 
       const remoteCustomType: CustomTypeSM | undefined = state[id].remote;
 
       const newCustomType = {
         ...state[id],
         local: {
           ...newLocalCustomType,
-          __status: getCustomTypeStatus(newLocalCustomType, remoteCustomType),
+          __status: isCustomTypeDisconnected
```
Heyrows: this is not needed, as Lorenzo mentioned, if we are able to push and it comes here, it means the users is connected.

sintinilorenzo (approved):

packages/slice-machine/src/modules/availableCustomTypes/index.ts
```diff
@@ -182,18 +182,14 @@ export const availableCustomTypesReducer: Reducer<
       const id = action.payload.customTypeId;
       const newName = action.payload.newCustomTypeName;
       const newLocalCustomType = { ...state[id].local, label: newName };
-      const isCustomTypeDisconnected =
-        newLocalCustomType.__status === CustomTypeStatus.UnknownDisconnected;
```
Heyrows: now that I think about it, we should do that if it's any Unknown.

Heyrows: Not for the push, but for the Save.
If the user is not connected to Prismic it wuld also write "New" or maybe i'm wrong and we already handle that case.

Heyrows (approved):

PR #573, opened 2022-08-19
Mm/screenshot sync

MarcMcIntosh:
Turns out the issue was caused by not saving the slice with the updated imageUrl.

Note: Creating a file hash maybe an unnecessary step

## Types of changes

## Checklist:

### DIFF
--- a/packages/core/src/node-utils/files.ts
+++ b/packages/core/src/node-utils/files.ts
@@ -172,6 +172,11 @@ async function readFileAndCreateHash(filePath: string): Promise<string> {
     .then((file) => createHash("md5").update(file).digest("hex"));
 }
 
+function readFileAndCreateHashSync(filePath: string): string {
+  const data = readString(filePath);
+  return createHash("md5").update(data).digest("hex");
+}
+
 export default {
   write,
   readBuffer,
@@ -192,4 +197,5 @@ export default {
   removeAll,
   flushDirectories,
   readFileAndCreateHash,
+  readFileAndCreateHashSync,
 };
--- a/packages/core/src/node-utils/files.ts
+++ b/packages/core/src/node-utils/files.ts
@@ -1,5 +1,6 @@
 import fs from "fs";
 import path from "path";
+import { createHash } from "crypto";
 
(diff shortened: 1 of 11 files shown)


packages/slice-machine/server/src/api/slices/push.ts
```diff
@@ -45,6 +45,8 @@ export async function pushSlice(
       variations,
     };
 
+    IO.Slice.writeSlice(modelPath, modelWithImageUrl);
```
sintinilorenzo, line 48: what is this for? why do we write to the file system during a slice push?

packages/core/src/node-utils/files.ts
```diff
@@ -165,6 +166,17 @@ function flushDirectories(directory: string, recursive = true): void {
   } catch (e) {}
 }
 
+async function readFileAndCreateHash(filePath: string): Promise<string> {
+  return fs.promises
+    .readFile(filePath, "utf-8")
+    .then((file) => createHash("md5").update(file).digest("hex"));
+}
+
+function readFileAndCreateHashSync(filePath: string): string {
```
sintinilorenzo: why do we have a sync and an async method for computing the hash?

MarcMcIntosh: I Wrote the async one first, but then found another place where it would be needed in a blocking function here https://github.com/prismicio/slice-machine/pull/573/files#diff-7dedffd7da0ef7017990f231207cbbe1adb822b8f00dff1d78398daeaeaaf954R55

`export function resolvePathsToScreenshot({` so it was a choice of either refactoring that file or adding a sync version.

packages/slice-machine/server/src/api/slices/push.ts
MarcMcIntosh: So the imageUrl is saved, which was one of the causes of the issue :)
tbh, now that the imageUrl is saved and the compareVersion function looks at imageUrls, the hash doesn't really need to be passed to the front-end.

It might be posable to refactor that part out, I'll have to check.

sintinilorenzo (approved):

chengoconnell (approved):

PR #575, opened 2022-08-23
[WIP]: Ls/736 push changes

MarcMcIntosh:
## Types of changes

## Checklist:


packages/slice-machine/src/modules/pushSaga.ts
```diff
@@ -0,0 +1,149 @@
...
+  payload,
+}: ReturnType<typeof changesPushCreator>) {
+  const { unSyncedSlices, unSyncedCustomTypes } = payload;
+
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const slicesPushResults: Array<string | undefined | number> = yield all(
+    unSyncedSlices.map((slice) => call(pushSliceSaga, slice))
+  );
+
+  const sliceAuthFailures = slicesPushResults.filter(isAuthErrorCode).length;
```
Heyrows: Redux-Sage.all seems to act like Promise.all, seems like you wouldn't be able to have more than 1 error here.

packages/slice-machine/src/modules/pushSaga.ts
```diff
@@ -0,0 +1,149 @@
...
+  return typeof n === typeof 0 && (n === 401 || n === 403);
+}
+
+export function* changesPushSaga({
+  payload,
+}: ReturnType<typeof changesPushCreator>) {
+  const { unSyncedSlices, unSyncedCustomTypes } = payload;
+
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const slicesPushResults: Array<string | undefined | number> = yield all(
```
Heyrows: What's the type in case of an error ?

packages/slice-machine/src/modules/pushSaga.ts
```diff
@@ -0,0 +1,149 @@
...
+      })
+    );
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const customTypePushResults: Array<string | undefined | number> = yield all(
+    unSyncedCustomTypes.map((ct) => call(pushCustomTypeSaga, ct))
+  );
+
+  const customTypeAuthFailures =
```
Heyrows: same as before, we shouldn't be able to have more than 1 error, this can be optimized.

packages/slice-machine/src/modules/pushSaga.ts
```diff
@@ -0,0 +1,149 @@
...
+    yield put(
+      openToasterCreator({
+        message: errorMsg,
+        type: ToasterType.ERROR,
+      })
+    );
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const customTypePushResults: Array<string | undefined | number> = yield all(
```
Heyrows: We shouldn't be able to push Custom Types if slices failed.

packages/slice-machine/src/modules/pushSaga.ts
```diff
@@ -0,0 +1,149 @@
...
+        type: ToasterType.SUCCESS,
+      })
+    );
+  }
+
+  const { data: serverState } = (yield call(getState)) as SagaReturnType<
+    typeof getState
+  >;
+  const { customTypes, ...rest } = serverState;
+  yield put(refreshStateCreator({ ...rest, localCustomTypes: customTypes }));
```
Heyrows: Here instead of doing fetching the state manually and then refreshing the state manually, we can just use SWR to force the refresh of the state.

https://stackoverflow.com/questions/61653622/next-react-swr-refresh-from-button-click-in-child-component-can-i-use-a-call

Heyrows: I'd create like a "forceRefreshState" in the `useServerState` file and export it to be used wherever we need.

packages/slice-machine/src/modules/pushSaga.ts
```diff
@@ -0,0 +1,149 @@
+import { ComponentUI } from "../../lib/models/common/ComponentUI";
```
Heyrows: I'd rename this file "pushChangesSaga" to make sure we know what's being pushed (changes name being associated to a feature).

packages/slice-machine/src/modules/pushSaga.ts
MarcMcIntosh: only if it throws :)

packages/slice-machine/src/modules/pushSaga.ts
MarcMcIntosh: depends on the error, server error number, anyother error undefined

packages/slice-machine/src/modules/pushSaga.ts
MarcMcIntosh: This can have more than one error, the other PR can only have one

packages/slice-machine/src/modules/pushSaga.ts
MarcMcIntosh: I'll have to look into that.

packages/slice-machine/src/modules/pushSaga.ts
MarcMcIntosh: done

MarcMcIntosh:
Updated for the slice to pushed in parallel, and then for the custom types to be pushed in parallel.

https://user-images.githubusercontent.com/7499570/187644713-d1cad597-9b23-4549-9808-16cee26669ca.mov

packages/slice-machine/pages/changes.tsx
```diff
@@ -32,8 +33,16 @@ const changes: React.FunctionComponent = () => {
   const isOnline = useNetwork();
   const numberOfChanges = unSyncedSlices.length + unSyncedCustomTypes.length;
 
+  const [error, setError] = useState<PUSH_CHANGES_ERRORS | null>(null);
+
+  useEffect(() => {
+    // might not need this
+    if (error) setError(null);
```
Heyrows: yeah i don't think it's needed, if we go back on that page, the error would be re-initiated.

Heyrows: maybe the whole useEffect is pointless

packages/slice-machine/src/modules/pushChangesSaga.ts
```diff
@@ -0,0 +1,104 @@
...
+import axios from "axios";
+
+export const changesPushCreator = createAction("PUSH_CHANGES")<{
+  unSyncedSlices: ReadonlyArray<ComponentUI>;
+  unSyncedCustomTypes: ReadonlyArray<CustomTypeSM>;
+  handleError: (e: PUSH_CHANGES_ERRORS | null) => void; // bit hacky but should work
+}>();
+
+export enum PUSH_CHANGES_ERRORS {
+  SLICES = "SLICES",
```
Heyrows, line 22: I know it depends on design but i'd image we'd want something more granular and precise than just a general error on Slices or Custom types.
Will be more important in the design ticket i guess.

packages/slice-machine/src/modules/pushChangesSaga.ts
```diff
@@ -0,0 +1,104 @@
...
+export function* changesPushSaga({
+  payload,
+}: ReturnType<typeof changesPushCreator>): Generator {
+  const { unSyncedSlices, unSyncedCustomTypes, handleError } = payload;
+
+  let stop: PUSH_CHANGES_ERRORS | null = null; // bit hacky for now until I figure out task cancelation
+
+  yield all(
+    unSyncedSlices.map(function* (slice) {
+      try {
```
Heyrows, line 34: According to [this doc](https://redux-saga.js.org/docs/advanced/RunningTasksInParallel), all returns a result like Promise.all that can be use to tell us if there was an error or not.

So i think i'd not catch inside but rather check the result of the all after it's finished.

packages/slice-machine/src/modules/pushChangesSaga.ts
```diff
@@ -0,0 +1,104 @@
...
+        }
+      }
+    })
+  );
+
+  if (stop) return handleError(stop);
+
+  handleError(null);
+
+  yield put(
```
Heyrows, line 84: Didn't you wanted to refresh the state also ? what happened ?

packages/slice-machine/pages/changes.tsx
MarcMcIntosh: I think so

packages/slice-machine/src/modules/pushChangesSaga.ts
MarcMcIntosh: Eeyup, from what I've seen in the design it's just a banner that says something like "these slices couldn't be pushed..."

packages/slice-machine/src/modules/pushChangesSaga.ts
MarcMcIntosh: the all method doesn't have a then or catch method :/.... it's weird like that because it would be easier if it did

packages/slice-machine/src/modules/pushChangesSaga.ts
MarcMcIntosh: hmm, I guess it could be done now :)

MarcMcIntosh: It was in the ticket, I guess i removed but I can add it back in :)

Heyrows: I'm not asking you to add it back, just thought you wanted to do it before that's why i didn't understand why it wasn't there.

packages/slice-machine/src/modules/pushChangesSaga.ts
```diff
@@ -0,0 +1,120 @@
...
+        yield call(pushSliceApiClient, slice);
+        yield put(pushSliceCreator.success({ component: slice }));
+      } catch (e) {
+        if (
+          axios.isAxiosError(e) &&
+          e.response?.status &&
+          e.response.status === 403
+        ) {
+          yield put(modalOpenCreator({ modalKey: ModalKeysEnum.LOGIN }));
+          yield cancel();
```
Heyrows, line 44: I don't think cancel is required.
image.

packages/slice-machine/src/modules/pushChangesSaga.ts
Heyrows: I think we should do
- a try catch like you did around a single call to throw a custom error if it fails
- catch the custom error at the `yield all` level in order to stop the saga and return immediately.

packages/slice-machine/pages/changes.tsx
```diff
@@ -1,34 +1,50 @@
...
+    return () => setError(null);
+  }, []);
+
+  const handlePush = () => {
+    pushChanges(unSyncedSlices, unSyncedCustomTypes, setError);
```
Heyrows, line 40: i think we should clear errors if we start pushing again

packages/slice-machine/src/modules/pushChangesSaga.ts
MarcMcIntosh: it'll keep iterating if it's not canceled :)

packages/slice-machine/pages/changes.tsx
MarcMcIntosh: can do

MarcMcIntosh: if it throws and stops, the user will be left with good and bad slices

Heyrows (approved):
The code should work but didn't you wanted to take care of replacing that part ?
```
let stop: PUSH_CHANGES_ERRORS | null = null; // bit hacky for now until I figure out task cancelation
```

PR #577, opened 2022-08-25
[Fix] Non blocking events.

### LINKED ISSUE #550: If user tracking fails to connect, slicemachine fails and is unusable
### Versions

- slice-machine-ui: 0.4.2
- node: 16.10.0

### Reproduction

When using DNS ad-blocking, slicemachine dies when invoked in develop mode and the page is loaded

<details open>
<summary>Additional Details</summary>
Console error:

```
[slice-machine] AxiosError: connect ECONNREFUSED 0.0.0.0:443
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1146:16) {
...
_currentUrl: 'https://....D6P:@api.segment.io/v1/batch',
```

I've  obscured the API key for segment.io -> Twilio tracking

</details>

### Steps to reproduce

Enable pi-hole, a DNS ad blocker or point api.segment.io to an invalid IP or localhost

### What is expected?

Customer tracking should be optional and not mandatory for the slice machine to work

Slicemachine closes when the local page is loaded

Heyrows:
Fixes #550


packages/slice-machine/server/src/api/services/analytics.ts
```diff
@@ -1,6 +1,11 @@
 import Analytics from "analytics-node";
 
-const analytics = new Analytics(process.env.PUBLIC_SM_UI_SEGMENT_KEY || "");
+const analytics = new Analytics(process.env.PUBLIC_SM_UI_SEGMENT_KEY || "", {
+  // @ts-expect-error: @types/analytics-node not covering the property errorHandler yet.
+  errorHandler: () => {
```
MarcMcIntosh, line 5: cool

MarcMcIntosh (approved):

PR #582, opened 2022-08-31
Compute status for Slice and Custom Type list frontend side

Heyrows:
## Types of changes

## Checklist:


packages/slice-machine/lib/env/resolveAliases.ts
```diff
@@ -5,6 +5,8 @@ import {
 } from "@slicemachine/core/build/node-utils";
 import moduleAlias from "module-alias";
 
+/* eslint-disable */
+// Disabling eslint for this file, it keeps throwing errors.
```
MarcMcIntosh, line 9: how come?

packages/slice-machine/lib/models/common/ModelStatus.ts
```diff
@@ -0,0 +1,30 @@
...
+  userHasAccessToPrismic: boolean
+): ModelStatus {
+  // If the user doesn't have access to Prismic models for any reason then we can't compare models properly
+  if (!userHasAccessToPrismic) return ModelStatus.Unknown;
+
+  // If there is no remote model then it's a new model created locally waiting to be pushed
+  if (!models.remote) return ModelStatus.New;
+
+  // If the models are not equals then it was modified locally
+  if (!equal(models.local, models.remote)) return ModelStatus.Modified;
```
MarcMcIntosh, line 27: depending on what we want to do for the default image (that is added by custom-types api) 
`models.local` will ned to be mapped over so that the variations with out and imageUrl will be given the default

packages/slice-machine/src/hooks/useModelStatus.ts
```diff
@@ -0,0 +1,68 @@
...
+};
+
+export const useModelStatus = (
+  models: FrontEndModel[]
+): {
+  modelsStatuses: ModelsStatuses;
+  authStatus: AuthStatus;
+  isOnline: boolean;
+} => {
+  const isOnline = useNetwork();
```
MarcMcIntosh, line 25: should useEffect be used when isOnline changes?, maybe not :/

packages/slice-machine/lib/models/common/ModelStatus.ts
MarcMcIntosh: This may have some impact on the changes page

MarcMcIntosh (changes requested):
Some questions and comments

packages/slice-machine/lib/env/resolveAliases.ts
Heyrows: No idea, last time it threw like 10 errors, I fixed a few types and I had no errors.
This time I didn't even touch the file it threw 10 errors again the same.

packages/slice-machine/lib/models/common/ModelStatus.ts
Heyrows: I see what you mean, didn't thought of that.

packages/slice-machine/src/hooks/useModelStatus.ts
Heyrows: I don't understand what you mean, hte file doesn't use any useEffect.

packages/core/src/models/CustomType/index.ts
```diff
@@ -16,7 +16,6 @@ export const CustomTypeSM = t.exact(
     }),
     t.partial({
       hash: t.string,
```
MarcMcIntosh, line 18: what's hash for?

packages/slice-machine/src/hooks/useModelStatus.ts
```diff
@@ -0,0 +1,70 @@
...
+import { useNetwork } from "./useNetwork";
+
+export interface ModelStatusInformation {
+  modelsStatuses: ModelsStatuses;
+  authStatus: AuthStatus;
+  isOnline: boolean;
+}
+
+// Slices and Custom Types needs to be separated as Ids are not unique with each others.
+export type ModelsStatuses = {
```
MarcMcIntosh: why create a new object, when the model could be extended, seems like there will be two objects to mange ?

MarcMcIntosh (commented):
Okay so there’s something I’m not 100% clear on, 

there’s a function that compares the a local and a remote to get the status…. that makes sense.

but should a new object be created in a hook containing the all of the results which then has to passed around.

Will the hook being recalled to calculate the state on re-renders?

… would it make sense to have the status kept in redux and calculate the status in the reducer because since it’s top level enough not to need be re-run during re-renders, and easy enough to scope to run on actions that refresh one or many custom-types?

packages/core/src/models/CustomType/index.ts
Heyrows: I think it's a property added by the custom type API.

Heyrows: I don't see any real use beside that

packages/slice-machine/src/hooks/useModelStatus.ts
Heyrows: Good catch, this wasn't necessary at all to create this type.

Heyrows:
I created
- a function to compute the status that can be used both frontend & backend side
- a hook to retrieve the status based on a list of models (CTs & Slices) to have only one hook between the builders and models lists, to centralise the logic.
- a hook to unSynced Cts and Slices specifically for the changes page & related

Those hooks will be called on each render.
I thought of two solutions for computing frontend
- compute on each re-render
  - Pros: less state management and less code complexity around statuses
  -  Cons: can affect performances

- have a redux store for statuses and re-compute every time we fetch the Remote models
  - Pros: could affect performances less than the other solution
  -  Cons: not much different from what we had, would had to still manage a state around that

I decided to go for option 1 and see if we have a performance issue through testing.

PR #583, opened 2022-09-02
feat: generate TypeScript types on save

lihbr:
## Types of changes

## Description

Previously in #533

As requested by @Duaner, this PR adds an early integration of TypeScript inside Slice Machine by generating types for Slices and Custom Types on save and rename.

A `prismicTypes.generated.ts` file is now maintained when:
- a custom type is created or updated (including rename)
- a slice is created or updated (included rename)

⚠ &nbsp;This code will have to be removed when switching to plugins. Plugins are already handling TypeScript integration the same way as the one implemented in this PR.

## Checklist:

- [x] All new and existing tests are passing. (existing tests were failing on Windows because of win32/posix paths inconsistencies)


lihbr:
@angeloashmore what about the file naming (`prismicTypes.generated.ts`)?
I think plugin integration would have to be smarter 🤔

TODO:
- [x] Release `prismic-ts-codegen` and update deps
- [x] Review file naming

angeloashmore:
> What about the file naming (prismicTypes.generated.ts)?

Honestly, it sounds a bit messy. I'm not saying `types.generated.ts` is any better, but I think we could come up with something cleaner while being explicit.

Maybe `prismic.d.ts`?

- Relates to Prismic.
- Identifies itself as a type declaration file.
- Is self-contained (i.e. is for Prismic only), obviating the need to identify itself as "generated" to differentiate from a general `types.ts` file.

When we decide to support multiple repositories, I *think* we can keep all types in a single `prismic.d.ts` file. We'll need to prefix type names or consolidate types into repository-specific namespaces. This allows us to continue augmenting `@prismicio/client` under a mostly consistent API.

angeloashmore:
I made the following changes:

- Changed the generated file's name to `prismic.d.ts`. See notes in [the previous comment](https://github.com/prismicio/slice-machine/pull/583#issuecomment-1238766247).
- Added a Slice Machine-specific "DO NOT EDIT" header in the file:

   ```typescript
   // Code generated by Slice Machine. DO NOT EDIT.

   import type * as prismicT from "@prismicio/types";
   import type * as prismic from "@prismicio/client";

   // ...
   ```
- Added `@prismicio/types` installation as a dev dependency to `@slicemachine/init`

Everything else looks good to go! 👍

angeloashmore (approved):

angeloashmore:
**Note**: CI seems to be failing due to changes unrelated to this PR.

packages/slice-machine/package.json
```diff
@@ -79,6 +79,7 @@
     "open": "^8.2.1",
     "ora": "6.0.1",
     "parse-domain": "^4.0.0",
+    "prismic-ts-codegen": "0.1.5",
```
MarcMcIntosh: Do we want to pin the version?

packages/slice-machine/server/src/api/custom-types/rename.ts
```diff
@@ -14,5 +16,11 @@ export default async function handler(req: RequestWithEnv) {
 
   IO.CustomType.renameCustomType(modelPath, newCustomTypeName);
 
+  IO.Types.upsert(
```
MarcMcIntosh: So this regenerates all the types ?

packages/slice-machine/lib/utils/slices.ts
```diff
@@ -0,0 +1,39 @@
...
+import Files from "./files";
+import { CustomPaths } from "../models/paths";
+import * as IO from "../io";
+
+export function getLocalSlices(
+  cwd: string,
+  libraries: string[] = []
+): SliceSM[] {
+  const slices: SliceSM[] = [];
+  for (const library of libraries) {
```
MarcMcIntosh: Mixing loops and iterative methods, for consistency is it posable to just iterative methods?

MarcMcIntosh (changes requested):
Good work, only a tiny bit of refactoring todo :)

packages/slice-machine/package.json
angeloashmore: Nope, good catch

packages/slice-machine/server/src/api/custom-types/rename.ts
angeloashmore: Correct; it will create the file if it doesn't exist, or overwrite it if it does. The file is regenerated completely each time something is created or updated based on the current Custom Type and Slice models.

MarcMcIntosh (approved):

angeloashmore:
Per our discussion in Slack, the implementation will be modified with the following behavior:

- **New projects**: Automatically opted-in. Automatically install `@prismicio/types`.
- **Existing projects**: Do not opt-in or out. Do not automatically install `@prismicio/types`. Print a message on SM start up instructing how to enable type generation or disable the message. Once `@prismicio/types` is installed, generate types. If `generateTypes: false` is present in `sm.json`, don't generate types.

packages/init/src/steps/configure-project.ts
```diff
@@ -31,6 +31,7 @@ export async function configureProject(
         : ["@/slices"];
 
     const manifestUpdated: Models.Manifest = {
+      generateTypes: true,
```
Heyrows: It's an optional option, don't put any value here so it never shows up in sm.json files by default.

packages/init/src/steps/install-required-dependencies.ts
```diff
@@ -42,7 +43,7 @@ async function addAndInstallDeps(
   const installDependencyCommand = useYarn ? "yarn add" : "npm install --save";
 
   const { stderr } = await execCommand(
-    `${installDevDependencyCommand} ${SM_PACKAGE_NAME}`
+    `${installDevDependencyCommand} ${SM_PACKAGE_NAME} ${PRISMIC_TYPES}`
```
Heyrows, line 46: are we sure we want to install the Types by default ?
As we made it an option, shouldn't we ask people if they want types or not ?

Sorry to bring that part again, I don't remember us talking about what to do for new users.

packages/slice-machine/lib/io/Types.ts
```diff
@@ -0,0 +1,71 @@
...
+
+import {
+  CustomTypes,
+  CustomTypeSM,
+} from "@slicemachine/core/build/models/CustomType/index";
+import { Slices, SliceSM } from "@slicemachine/core/build/models/Slice";
+import Files from "../utils/files";
+
+const NON_EDITABLE_FILE_HEADER =
+  "// Code generated by Slice Machine. DO NOT EDIT.";
```
Heyrows, line 14: I believe we already have something like that for the `Index.js` files generated, I would take the same for both.

packages/slice-machine/lib/io/Types.ts
```diff
@@ -0,0 +1,71 @@
...
+    sharedSliceModels: sharedSliceModels.map((model) => Slices.fromSM(model)),
+    clientIntegration: {
+      includeCreateClientInterface: true,
+      includeContentNamespace: true,
+    },
+  });
+
+  const fileContents = `${NON_EDITABLE_FILE_HEADER}\n\n${types}`;
+
+  Files.write(path.join(cwd, ".slicemachine", "prismicio.d.ts"), fileContents);
```
Heyrows: This Path if it's only one file should be a Const, most likely in the Core with helpers to find models or such things.

packages/slice-machine/lib/io/Types.ts
```diff
@@ -0,0 +1,71 @@
...
+  sharedSliceModels: SliceSM[]
+) => {
+  const manifest = retrieveManifest(cwd);
+
+  const packageJson = retrieveJsonPackage(cwd);
+  const allDependencies = {
+    ...packageJson.content?.dependencies,
+    ...packageJson.content?.devDependencies,
+  };
+  const hasTypesPackage = "@prismicio/types" in allDependencies;
```
Heyrows: I believe you created a dependency for that.

packages/slice-machine/scripts/start/validateGenerateTypes.ts
```diff
@@ -0,0 +1,38 @@
...
+  const allDependencies = {
+    ...packageJson.content?.dependencies,
+    ...packageJson.content?.devDependencies,
+  };
+  const hasTypesPackage = "@prismicio/types" in allDependencies;
+
+  if (manifest.content && "generateTypes" in manifest.content) {
+    // `generateTypes` is in manifest
+
+    if (manifest.content.generateTypes && !hasTypesPackage) {
```
Heyrows, line 25: You don't verify the value of "generateTypes", a user could have set it to true or false.
If it's false, the message under is irrelevant.

packages/slice-machine/scripts/start/validateGenerateTypes.ts
```diff
@@ -0,0 +1,38 @@
...
+      logger.warn(
+        "Slice Machine was configured to generate TypeScript types via the `generateTypes` option in `sm.json`, but the `@prismicio/types` package is missing.\nInstall `@prismicio/types` to generate TypeScript types."
+      );
+    }
+  } else {
+    // `generateTypes` is not in manifest
+
+    if (!hasTypesPackage) {
```
Heyrows, line 33: This is the same condition as above, so adding a generateTypes value would not really shut any message about types generation.

packages/slice-machine/server/src/api/custom-types/save.ts
```diff
@@ -29,13 +29,22 @@ export default async function handler(req: { body: SaveCustomTypeBody }) {
...
+
   const mocked = await mock(
     model,
     // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-argument
     CustomTypeMockConfig.getCustomTypeMockConfig(updatedMockConfig, model.id)
   );
   // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
   Files.write(mockPath, mocked as object);
+
+  IO.Types.upsert(
```
Heyrows: Did you consider the performances of generating types ?
I'm wondering if it would slow the app down.

Heyrows: I mean, getting Slices and Custom Types each time must slow things down a bit.

packages/init/src/steps/configure-project.ts
Heyrows: That way also, the day we decide to change to plugin architecture, people won't have to remove that value from their SM.json. 😄

packages/init/src/steps/install-required-dependencies.ts
lihbr: I think it makes sense to install it by default as we want to push for it to be enabled for most people

packages/init/src/steps/configure-project.ts
lihbr: Good catch!

packages/slice-machine/lib/io/Types.ts
lihbr: Updated the other one as I think this one is better phrased and less misleading ("update with care" means nothing on the other one as you cannot update the file)

packages/slice-machine/lib/io/Types.ts
lihbr: Since this will be removed, is it ok to keep it like this? Not sure it's worth the extra abstraction.

packages/slice-machine/lib/io/Types.ts
lihbr: Not sure we have a package for that no 🤔 What are you thinking about?

packages/slice-machine/scripts/start/validateGenerateTypes.ts
lihbr: If it's falsy then the second `if` doesn't pass? Maybe I'm missing something 🤔

Basically, the tree goes like this:

- `generateTypes` in manifest? (first if)
  - Yes
    - `generateTypes` is truthy and project does no have `@prismicio/types`?
      - Yes
        - Log warning (user's trying to do something incorrectly)
      - No
        - Do nothing
  - No (else)
    - Does not have `@prismicio/types`?
      - Yes
        - Log info (informing user's about a feature we want to push for)
      - No
        - Do nothing

packages/slice-machine/scripts/start/validateGenerateTypes.ts
lihbr: Pretty sure it does, see above conf

packages/slice-machine/server/src/api/custom-types/save.ts
lihbr: We haven't benchmarked `prismic-ts-codegen` yet but it's pretty fast (appears instant to me) so far (cc @angeloashmore, added that to our todo)

I think it's negligible for now. If it becomes an issue people can still use `generateTypes: false` as an escape hatch

packages/slice-machine/lib/io/Types.ts
Heyrows: Ok

packages/slice-machine/lib/io/Types.ts
Heyrows: sorry 😓, I meant in the Core you created a const to install it in the init
This has to be re-used here.

packages/slice-machine/scripts/start/validateGenerateTypes.ts
Heyrows: yes, my bad I misread that 🤦

packages/slice-machine/server/src/api/custom-types/save.ts
Heyrows: I didn't mean the library itself, more the actions to get local custom types and local slices which could be heavy.

And I guess even with the `generateTypes: false` we would still retrieve the slices and custom types.

packages/slice-machine/lib/io/Types.ts
lihbr: No worries! Didn't know about it, done :)

packages/slice-machine/server/src/api/custom-types/save.ts
lihbr: I see, ok.

We can't do anything about it when enabled, I'll refactor to prevent that when disabled :)

packages/next-slicezone/package-lock.json
```diff
@@ -7061,6 +7061,20 @@
 				"is-typedarray": "^1.0.0"
 			}
 		},
+		"node_modules/typescript": {
```
Heyrows: there shouldn't be changes here of it might publish this package.

packages/slice-machine/lib/io/Types.ts
Heyrows: I guess so if we don't use that anywhere else

packages/init/src/steps/install-required-dependencies.ts
Heyrows: I disagree as it's an option i think we should ask the user.
Nevermind for now I guess you guys are sure of yourself and taking responsibilities on this one 😉 .

packages/next-slicezone/package-lock.json
lihbr: The real question is why is there a lock file there 🙃

lihbr: Done

packages/slice-machine/server/src/api/slices/rename.ts
```diff
@@ -77,11 +75,7 @@ export async function renameSlice(req: {
     `${GeneratedPaths(env.cwd).library(libName).value()}/${newSliceName}`
   );
 
-  IO.Types.upsert(
-    env.cwd,
-    getLocalCustomTypes(env.cwd),
-    getLocalSlices(env.cwd, env.manifest.libraries)
-  );
+  IO.Types.upsert(env);
```
Heyrows, line 78: much cleaner to read :)

Heyrows (approved):

akarray:
Is there any date for official release ?

lihbr:
Hey @akarray, it will be part of next week release :)

akarray:
Hey Lucie,
Amazing :)
Thank you.

PR #590, opened 2022-09-14
Mm/tracking videos

MarcMcIntosh:
## Types of changes

## Checklist:

Added video url to tracking payload when clicking on the tutorial video

### DIFF
--- a/packages/slice-machine/components/AppLayout/Navigation/Menu/Navigation/VideoItem.tsx
+++ b/packages/slice-machine/components/AppLayout/Navigation/Menu/Navigation/VideoItem.tsx
@@ -22,11 +22,14 @@ const VideoItem: React.FC<VideoItemProps> = ({
 }) => {
   const ref = React.createRef<HTMLParagraphElement>();
   const id = "video-tool-tip";
+  const videoUrl =
+    "https://youtube.com/playlist?list=PLUVZjQltoA3wnaQudcqQ3qdZNZ6hyfyhH";
 
   const handleClose = () => {
     void Tracker.get().trackClickOnVideoTutorials(
       framework,
-      sliceMachineVersion
+      sliceMachineVersion,
+      videoUrl
     );
     onClose();
   };
@@ -48,7 +51,7 @@ const VideoItem: React.FC<VideoItemProps> = ({
         link={{
           title: "Video tutorials",
           Icon: MdPlayCircleFilled,
-          href: "https://youtube.com/playlist?list=PLUVZjQltoA3wnaQudcqQ3qdZNZ6hyfyhH",
+          href: videoUrl,
           target: "_blank",
           match: () => false,
         }}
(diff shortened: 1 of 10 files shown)


arnaudlewis (commented):
The code seems good to me but I have two questions here:
- Is it intended to track the youtube playlist URL instead of a video URL in the case of the `video tutorial` button from the navigation menu?
- Should be keep the cloudinary video IDs and the youtube playlist URL static in the middle of the code? I would simply build an object somewhere and reference it. It's easier to centralize and manage updates I'd say.

packages/slice-machine/components/AppLayout/Navigation/Menu/Navigation/VideoItem.tsx
```diff
@@ -22,11 +23,13 @@ const VideoItem: React.FC<VideoItemProps> = ({
 }) => {
   const ref = React.createRef<HTMLParagraphElement>();
   const id = "video-tool-tip";
+  const videoUrl = VIDEO_YOUTUBE_PLAYLIST_LINK;
```
arnaudlewis, line 26: unnecessary variable here since you have the const now :)

packages/slice-machine/pages/onboarding.tsx
```diff
@@ -83,32 +102,39 @@ const WelcomeSlide = ({ onClick }: { onClick: () => void }) => {
...
+  props
+) => (
   <>
     <Image sx={imageSx} src="/send.svg" />
     <Header>Push your pages to Prismic</Header>
     <SubHeader>
       Give your content writers the freedom to build whatever they need
     </SubHeader>
-    <Video publicId="SMONBOARDING/PUSH_TO_PRISMIC" />
+    <Video {...props} publicId={VIDEO_ONBOARDING_PUSH_TO_PRISMIC} />
```
arnaudlewis, line 137: why do you need to provide **version** and **framework** to the cloudinary component?

arnaudlewis (commented):
Couldn't we get framework/version directly from the video component?
It's SM context basically so I wouldn't be chocked to let the video component gather its context instead of doing so at top level and send it through props in many places.
WDYT @MarcMcIntosh ? Maybe it doesn't make sense but that's a lead.
```
framework: getFramework(store),
version: getCurrentVersion(store)
```

MarcMcIntosh:
Could do I guess, but I'll need to check for conditional rendering, as react hooks have to be called the same number of times between renders :/

arnaudlewis:
wdym by conditional rendering in that case? Where would that happen?

MarcMcIntosh:
https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

We'v had that issues before, when using hooks with map :/

But i've cheked and it should be fine to use it a the top of a conditionally rendered component. https://www.robinwieruch.de/react-conditional-hooks/

Weird how in react it used to be lift the state up, but now it's move the hooks down 😅

arnaudlewis:
Alright, I'll let you check that up. For me conditional hooks are about hooks under if statement but if it's inside a child component without condition than it should be fine.
I'll look that up too on my side to make sure I don't provide bad advice here.

arnaudlewis:
I asked Alex for a second opinion but React will ensure that you don't have conditional hooks but conditional render of a child component that contains a hook is ok. So in that case, moving hooks to collection `framework` and `version` in the Video component shouldn't be a problem.

arnaudlewis (approved):

PR #591, opened 2022-09-14
Mm/tracking screenshots

MarcMcIntosh:
Adds some tracking evets for when users add screenshots.
Also I updated `msw` because it wasn't working as expected.... now it is :)

## Types of changes

## Checklist:


packages/slice-machine/src/modules/selectedSlice/sagas.ts
```diff
@@ -56,6 +58,9 @@ export function* generateSliceScreenshotSaga({
         imageLoading: false,
       });
     }
+
+    void Tracker.get().trackScreenshotTaken({ type: "automatic" });
```
arnaudlewis, line 62: should we care if it fails and we stick to a silent error here? and same below

arnaudlewis (approved):

packages/slice-machine/src/modules/selectedSlice/sagas.ts
MarcMcIntosh: There's a warning inside of the tracker :)

arnaudlewis: yeah I saw, I was just wondering if we wanted a feedback on our side and not on the client side but I'm fine with it

PR #596, opened 2022-09-19
sm-ui(tracking): add tracking for changes page

MarcMcIntosh:
## Types of changes

## Checklist:

### DIFF
--- a/packages/slice-machine/src/modules/pushChangesSaga/index.ts
+++ b/packages/slice-machine/src/modules/pushChangesSaga/index.ts
@@ -26,10 +26,19 @@ import {
 import { LoadingKeysEnum } from "../loading/types";
 import { ApiError } from "@src/models/ApiError";
 import { SyncError } from "@src/models/SyncError";
+import { ModelStatusInformation } from "@src/hooks/useModelStatus";
+import { ModelStatus } from "@lib/models/common/ModelStatus";
+import Tracker from "@src/tracking/client";
+
+const startTimer =
+  (startTime = Date.now()) =>
+  (endTime = Date.now()) =>
+    endTime - startTime;
 
 export const changesPushCreator = createAction("PUSH_CHANGES")<{
   unSyncedSlices: ReadonlyArray<ComponentUI>;
   unSyncedCustomTypes: ReadonlyArray<CustomTypeSM>;
+  modelStatuses: ModelStatusInformation["modelsStatuses"];
   onChangesPushed: (pushed: string | null) => void;
   handleError: (e: SyncError | null) => void;
 }>();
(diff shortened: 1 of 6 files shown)


packages/slice-machine/src/modules/pushChangesSaga/index.ts
```diff
@@ -26,24 +26,59 @@ import {
...
+  const sendTracking = () =>
+    Tracker.get().trackChangesPushed({
+      customTypesCreated,
+      customTypesModified,
+      customTypesDeleted,
+      slicesCreated,
+      slicesModified,
+      slicesDeleted,
+      errors,
+      total: totalNumberOfChanges,
```
Heyrows, line 79: I think that's only good if the entire push is a success.
Should probably be computed inside the function itself using the other parameters

packages/slice-machine/src/modules/pushChangesSaga/index.ts
```diff
@@ -26,24 +26,59 @@ import {
...
+    modelStatuses,
+    onChangesPushed,
+    handleError,
+  } = payload;
   const totalNumberOfChanges: number =
     unSyncedSlices.length + unSyncedCustomTypes.length;
 
   let alreadySyncedChanges = 0;
   let stop: Parameters<typeof handleError>[0] = null;
+  let slicesCreated = 0;
```
Heyrows, line 61: Do you think this could all go in a function outside that we just call at the beginning and return helpers to be called here ?

I feel this would remove a lot of complexity from the main saga.

packages/slice-machine/src/modules/pushChangesSaga/index.ts
```diff
@@ -26,24 +26,59 @@ import {
...
+
+  const sendTracking = () =>
+    Tracker.get().trackChangesPushed({
+      customTypesCreated,
+      customTypesModified,
+      customTypesDeleted,
+      slicesCreated,
+      slicesModified,
+      slicesDeleted,
+      errors,
```
Heyrows, line 78: OPT:
I don't know if this is the right thing to do or not but I feel that PM would want to know if it failed on Slices or Custom Types.

Let's discuss this together.

packages/slice-machine/src/tracking/types.ts
```diff
@@ -172,6 +173,21 @@ export interface ScreenshotTaken extends BaseTrackingEvent {
   };
 }
 
+export interface ChangesPushed extends BaseTrackingEvent {
```
Heyrows, line 176: 👍

packages/slice-machine/src/modules/pushChangesSaga/index.ts
MarcMcIntosh: The ticket it mentioned what to do about errors, I figured we would want to know about them.

packages/slice-machine/src/modules/pushChangesSaga/index.ts
MarcMcIntosh: The variables?

packages/slice-machine/src/modules/pushChangesSaga/index.ts
MarcMcIntosh: Ok

Heyrows (approved):
Discussed with Renaud, good 👍

PR #598, opened 2022-09-21
802/simulator screenshots

danigo99:
## Types of changes

## Checklist:


packages/slice-machine/components/Simulator/components/Header/index.tsx
```diff
@@ -1,20 +1,36 @@
...
 
 import ScreenSizes, { Size } from "../ScreenSizes";
 import { ComponentUI } from "@lib/models/common/ComponentUI";
+import { useState } from "react";
+import SaveButton from "@lib/builders/SliceBuilder/Header/SaveButton";
 
 type PropTypes = {
   Model: ComponentUI;
-  variation: Models.VariationSM | undefined;
+  variation: Models.VariationSM;
```
danigo99, line 23: Was there a reason for this to be optional before? Didn't seem to have broken anything afaik

packages/slice-machine/components/Simulator/components/Header/index.tsx
```diff
@@ -39,6 +55,12 @@ const Header: React.FunctionComponent<PropTypes> = ({
   handleScreenSizeChange,
   size,
 }) => {
+  const { generateSliceScreenshot } = useSliceMachineActions();
+  const [data, setData] = useState<ScreenShotSyncState>(initialState);
+  const onTakingSliceScreenshot = () => {
+    generateSliceScreenshot(variation?.id, Model, setData);
```
danigo99: This is wrapped in a checkSimulatorSetup elsewhere. What's the purpose for it and is it useful here at all?

packages/slice-machine/server/src/api/screenshots/generate.ts
```diff
@@ -21,53 +21,28 @@ export interface ScreenshotResults {
...
-    removeCustomScreenshot(env, libraryName, sliceName, variationId)
-  );
-
-  return {
-    screenshots: screenshots,
-    failure: failure,
-  };
-}
-
 export async function generateScreenshot(
```
danigo99: This will largely be refactored since we don't need the multiple promise logic anymore. Still WIP

packages/slice-machine/src/modules/selectedSlice/sagas.ts
```diff
@@ -33,7 +33,7 @@ import Tracker from "../../tracking/client";
 export function* generateSliceScreenshotSaga({
```
danigo99: @Heyrows I think this is what you were saying should be extracted into a separate module correct? Do you think it's own GenerateSliceScreenshot module for the saga and actions is good?

packages/slice-machine/components/Simulator/components/Header/index.tsx
Heyrows: I believe it might come from the fact that the store "SelectedSlice" might be undefined at first before being updated.

packages/slice-machine/server/src/api/screenshots/generate.ts
Heyrows: nice 👍, might not even need to be on a different file anymore.

packages/slice-machine/src/modules/selectedSlice/sagas.ts
Heyrows: Today it is entirely linked to the Slice Builder page as it was the only place to take a screenshot.
Now, we'll be able to
- take a screenshot from the simulator
- use a custom screenshot from every screens

The simulator has its own module & store so it would make sense to put the `RegularScreenshotSaga` there as well.
However, not keeping the screenshots saga together would be wrong i think.

I'm no Redux expert so I don't know which code structure people usually use, what do you think @danigo99 ?

mathieud: @Heyrows Is the custom-screenshot feature also going to move to the simulator page? I think I would create a separate module (probably only action and saga since there's no state) to handle it, and the events could be called from both anywhere where it makes sense to upload a screenshot.
What do you think?

packages/slice-machine/server/src/api/screenshots/generate.ts
```diff
@@ -21,53 +21,28 @@ export interface ScreenshotResults {
...
-  // Remove custom screenshot of success
-  Object.keys(screenshots).forEach((variationId) =>
-    removeCustomScreenshot(env, libraryName, sliceName, variationId)
-  );
-
-  return {
-    screenshots: screenshots,
-    failure: failure,
-  };
-}
```
danigo99: @Heyrows this caused a regression now where the custom screenshot isn't overwritten when a new screenshot is taken. From our call yesterday I seemed to understand that this was what you expected, but then how would a user update a screenshot after adding a custom one?

packages/slice-machine/components/Simulator/components/Header/index.tsx
Heyrows: Hey, in order to be able to take screenshots in the context of the user's project running, we need him to install a new page in his project called slice-simulator. (to render independent components with the user style).

This checkSimulatorSetup is a helper to always verify the user has every setup so that we can take a screenshot.
It is very useful 😄

packages/slice-machine/src/modules/selectedSlice/sagas.ts
Heyrows: Hey, the custom screenshots feature is going to be in a modal that will be open all across the app but not in the Simulator page.

Agree those two sagas are going to be independent from any modules 😄.
It is like the `pushChangesSaga`

packages/slice-machine/server/src/api/screenshots/generate.ts
Heyrows: There are two types of screenshots
- regular -> taken using pupeteer rendering your slice in the context of your project with mock data
- custom -> custom image uploaded by the user.

Slice Machine will look in that order `custom > regular` when searching for a screenshot.

ℹ️  So
- If a user takes a regular screenshot, the custom one must be deleted in order for the regular to show up properly
- If a user adds a custom screenshot it will be taken by default so it's ok.

packages/slice-machine/components/Simulator/components/Header/index.tsx
```diff
@@ -7,14 +7,38 @@ import * as Links from "@lib/builders/SliceBuilder/links";
...
 type PropTypes = {
   Model: ComponentUI;
-  variation: Models.VariationSM | undefined;
+  variation: Models.VariationSM;
   handleScreenSizeChange: (screen: { size: Size }) => void;
   size: Size;
 };
 
+type ScreenShotSyncState = {
+  imageLoading: boolean;
```
Heyrows: i'm not sure you need all of that.
Assuming you work with Sagas, you only need here to know if the screenshot is loading or not.
We do have global states of loading also to be used in the redux store.
I believe one of the is made for screenshots

packages/slice-machine/components/Simulator/components/Header/index.tsx
```diff
@@ -39,6 +63,42 @@ const Header: React.FunctionComponent<PropTypes> = ({
   handleScreenSizeChange,
   size,
 }) => {
+  const { generateSliceScreenshot, checkSimulatorSetup } =
```
Heyrows: I believe we already check if the simulator is setup properly when entering the page itself.

Heyrows: Assuming there is a test to verify that behavior, there would be no need to verify here that the simulator is setup properly.

packages/slice-machine/components/Simulator/components/Header/index.tsx
```diff
@@ -39,6 +63,42 @@ const Header: React.FunctionComponent<PropTypes> = ({
...
+    isCheckingSetup: isLoading(state, LoadingKeysEnum.CHECK_SIMULATOR),
+  }));
+
+  const onTakingSliceScreenshot = () => {
+    checkSimulatorSetup(true, () =>
+      generateSliceScreenshot(variation.id, Model, setData)
+    );
+  };
+
+  const { isWaitingForIframeCheck, simulatorUrl } = useSelector(
```
Heyrows: we usually prefer to call only once the "useSelector".

packages/slice-machine/components/Simulator/components/Header/index.tsx
Heyrows: If we don't, then your implementation is good but we should write it down as a default in the product.

When entering this page, the user should know if we can display stuff and if he can take screenshot or not.

packages/slice-machine/server/src/api/screenshots/generate.ts
```diff
@@ -48,28 +47,23 @@ export async function generateScreenshotAndRemoveCustom(
...
 ): Promise<ScreenshotResults> {
   // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
   const slice = IO.Slice.readSlice(
     NodeUtils.CustomPaths(env.cwd).library(libraryName).slice(sliceName).model()
   );
 
-  const variationIds: VariationSM["id"][] = slice.variations.map((v) => v.id);
-
-  const promises: Promise<ScreenshotUI>[] = variationIds.map(
+  const promises: Promise<ScreenshotUI>[] = [variationId].map(
```
Heyrows: We don't need all the logic looping on the variations anymore.

packages/slice-machine/src/modules/screenshots/sagas.ts
```diff
@@ -0,0 +1,152 @@
...
+  generateSliceCustomScreenshotApiClient,
+} from "@src/apiClient";
+import { openToasterCreator, ToasterType } from "@src/modules/toaster";
+
+export function* generateSliceScreenshotSaga({
+  payload,
+}: ReturnType<typeof generateSliceScreenshotCreator.request>) {
+  const { component, variationId, setData } = payload;
+  try {
+    setData({
```
Heyrows: i don't think this setData should be needed anymore on the new page.
We either display toasts in case of error or success and we have global loading states.

Heyrows (changes requested):

packages/slice-machine/components/Simulator/components/Header/index.tsx
danigo99: You're right, it looks like the setup is already checked when the slice simulator is opened. The simulator doesn't open if the check fails so I guess it's redundant here

packages/slice-machine/server/src/api/screenshots/generate.ts
danigo99: I think when me an Mathieu wrote it we opted to leave it like this since there was a lot of work in there and we weren't sure if soon we'd have a feature like "update all variations" that would need to use it again.

If you don't think it's needed then I'll refactor the whole thing

Heyrows: It shouldn't be needed in this iteration.
From what I know they don't want automatic screenshots for multiple variations anymore.

packages/slice-machine/server/src/api/screenshots/generate.ts
```diff
@@ -29,62 +23,30 @@ export async function generateScreenshotAndRemoveCustom(
...
+      env,
+      libraryName,
+      slice,
+      variationId
+    );
+
+    removeCustomScreenshot(env, libraryName, sliceName, variationId);
+
+    return {
+      screenshots: [result],
```
Heyrows: almost perfect, just if we know there is going to be only one screenshot here I would return a `ScreenshotUI | null`, that way in the frontend using the variationID you already have, the state can be updated easily.

packages/slice-machine/src/modules/screenshots/sagas.ts
```diff
@@ -21,40 +21,14 @@ import { openToasterCreator, ToasterType } from "@src/modules/toaster";
...
-      imageLoading: true,
-    });
     const response = (yield call(
       generateSliceScreenshotApiClient,
       component.model.name,
       component.from,
       variationId
     )) as SagaReturnType<typeof generateSliceScreenshotApiClient>;
-    if (response.status > 209) {
-      return setData({
```
Heyrows: love that, so much simpler

Heyrows (changes requested):

Heyrows (commented):
I see you changed the backend so that's good but i didn't see any change to the frontend behavior when receiving the new screenshotUI.

Heyrows (approved):

PR #605, opened 2022-09-22
fix: display error in Rich Text model when no block types are selected

angeloashmore:
## Types of changes

## Description

This PR ensures an error is displayed in the Rich Text configuration modal when no block types are selected.

Before this PR, the model would not validate, but the error was not displayed.

Fixes Linear issue SM-713: https://linear.app/prismic/issue/SM-713/aau-when-edit-the-config-of-a-rich-text-field-i-see-an-error-message

## Checklist:


packages/slice-machine/lib/models/common/widgets/StructuredText/Form.js
```diff
@@ -141,12 +141,12 @@ const WidgetForm = (props) => {
         ))}
         {
           // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-          errors[accessors[isMulti ? 1 : 0]] ? (
+          errors.config?.[isMulti ? "multi" : "single"] ? (
             <Box sx={{ position: "absolute" }}>
               <Text as="span" variant="text.labelError" pl={0}>
                 {
                   // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
MarcMcIntosh: We should probably convert this file to typescript

packages/slice-machine/lib/models/common/widgets/StructuredText/Form.tsx
```diff
@@ -0,0 +1,222 @@
...
+  const {
+    initialValues,
+    values: formValues,
+    errors,
+    fields,
+    setFieldValue,
+  } = props;
+
+  const {
+    config: { single, multi },
```
Heyrows, line 113: this field is super weird 😂 , how we register either single or the multi property.

packages/slice-machine/lib/models/common/widgets/StructuredText/Form.tsx
```diff
@@ -0,0 +1,222 @@
...
+  )
+    return ctx.parent[key];
+  return null;
+}
+
+function createValidator(key: "single" | "multi") {
+  return () =>
+    yup.string().test({
+      name: key,
+      message: `Options cannot be empty`,
```
Heyrows, line 47: I tested the PR and it seems to work, I just don't understand how.
This message appears only so it has to be it, but it's not used on the Accept field but on the fields before.

That's why i can't figure out how it's triggered for the Accept field as well.

packages/slice-machine/lib/models/common/widgets/StructuredText/Form.tsx
MarcMcIntosh: eeyup, that's what the old file did so there's a few weird parts,

MarcMcIntosh: both get registered

packages/slice-machine/lib/models/common/widgets/StructuredText/Form.tsx
MarcMcIntosh: Here: https://github.com/prismicio/slice-machine/pull/605/files#diff-6510a41b30c174d7f6e9682d3fbf894ec27c13b472ce9c8fdc7bf1f606b53e8aR169

And the options in the options.ts file

Heyrows (commented):
Good 👍 , merged it before approving but it's good

PR #611, opened 2022-09-26
Pass screen width to puppeteer on screenshot

danigo99:
## Types of changes

## Checklist:

### DIFF
--- a/packages/slice-machine/components/Simulator/components/Header/index.tsx
+++ b/packages/slice-machine/components/Simulator/components/Header/index.tsx
@@ -60,11 +60,11 @@ const redirect = (
 const deviceToWidth = (device: Size) => {
   switch (device) {
     case Size.FULL:
-      return 1200;
+      return "1200";
     case Size.TABLET:
-      return 600;
+      return "600";
     case Size.PHONE:
-      return 340;
+      return "340";
   }
 };
 
--- a/packages/slice-machine/components/Simulator/components/Header/index.tsx
+++ b/packages/slice-machine/components/Simulator/components/Header/index.tsx
@@ -57,6 +57,17 @@ const redirect = (
   void router.push(params.href, params.as, params.options);
 };
 
+const deviceToWidth = (device: Size) => {
+  switch (device) {
+    case Size.FULL:
+      return 1200;
+    case Size.TABLET:
+      return 600;
+    case Size.PHONE:
+      return 340;
+  }
+};
+
 const Header: React.FunctionComponent<PropTypes> = ({
   Model,
   variation,
@@ -73,7 +84,7 @@ const Header: React.FunctionComponent<PropTypes> = ({
 
   const onTakingSliceScreenshot = () => {
     checkSimulatorSetup(true, () =>
-      generateSliceScreenshot(variation.id, Model, setData)
+      generateSliceScreenshot(variation.id, Model, setData, deviceToWidth(size))
     );
   };
 
--- a/packages/slice-machine/components/Simulator/components/Header/index.tsx
+++ b/packages/slice-machine/components/Simulator/components/Header/index.tsx
@@ -57,14 +57,14 @@ const redirect = (
   void router.push(params.href, params.as, params.options);
 };
 
-const deviceToWidth = (device: Size) => {
+const deviceToDimensions = (device: Size) => {
   switch (device) {
     case Size.FULL:
-      return "1200";
+      return { width: "1200", height: "600" };
     case Size.TABLET:
-      return "600";
+      return { width: "600", height: "600" };
     case Size.PHONE:
-      return "340";
+      return { width: "340", height: "600" };
   }
 };
 
@@ -84,7 +84,12 @@ const Header: React.FunctionComponent<PropTypes> = ({
 
   const onTakingSliceScreenshot = () => {
     checkSimulatorSetup(true, () =>
-      generateSliceScreenshot(variation.id, Model, setData, deviceToWidth(size))
+      generateSliceScreenshot(
+        variation.id,
+        Model,
+        setData,
+        deviceToDimensions(size)
+      )
     );
   };
 
--- a/packages/slice-machine/components/Simulator/components/Header/index.tsx
+++ b/packages/slice-machine/components/Simulator/components/Header/index.tsx
@@ -17,6 +17,7 @@ import { selectIsWaitingForIFrameCheck } from "@src/modules/simulator";
 import { selectSimulatorUrl } from "@src/modules/environment";
 import { isLoading } from "@src/modules/loading";
 import { LoadingKeysEnum } from "@src/modules/loading/types";
+import { ScreenDimensions } from "@lib/models/common/Screenshots";
 
 type PropTypes = {
   Model: ComponentUI;
@@ -57,14 +58,14 @@ const redirect = (
   void router.push(params.href, params.as, params.options);
 };
 
-const deviceToDimensions = (device: Size) => {
+const deviceToDimensions = (device: Size): ScreenDimensions => {
   switch (device) {
     case Size.FULL:
-      return { width: "1200", height: "600" };
+      return { width: 1200, height: 600 };
     case Size.TABLET:
-      return { width: "600", height: "600" };
+      return { width: 600, height: 600 };
     case Size.PHONE:
-      return { width: "340", height: "600" };
+      return { width: 340, height: 600 };
   }
 };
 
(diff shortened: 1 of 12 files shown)


packages/slice-machine/components/Simulator/components/Header/index.tsx
```diff
@@ -57,6 +57,17 @@ const redirect = (
   void router.push(params.href, params.as, params.options);
 };
 
+const deviceToWidth = (device: Size) => {
+  switch (device) {
+    case Size.FULL:
+      return 1200;
```
mathieud: Where do these number com from? Could we sync them better with what we actually show the user (https://github.com/prismicio/slice-machine/blob/master/packages/slice-machine/components/Simulator/components/ScreenSizes/index.tsx#L31)?

mathieud: Also the iframe height is also fixed. Is fixing the width only enough?

packages/slice-machine/server/src/api/screenshots/puppeteer.ts
```diff
@@ -14,14 +15,14 @@ export default {
   handleScreenshot: async ({
     screenshotUrl,
     pathToFile,
+    screenWidth,
   }: PuppeteerHandleProps): Promise<void> => {
     const { warning } = await testUrl(screenshotUrl);
     if (warning) throw new Error(warning);
 
-    if (!puppeteerBrowserPromise)
```
mathieud, line 21: Not sure what this did. Do you understand it and is it safe to remove?

packages/slice-machine/server/src/api/screenshots/screenshots.ts
```diff
@@ -36,6 +36,7 @@ export default async function handler({
   libraryName,
   sliceName,
   variationId,
+  screenWidth,
```
mathieud: I wonder if we could use the same type here and in the api client

packages/slice-machine/components/Simulator/components/Header/index.tsx
danigo99: @mathieud fixing the height of the viewport doesn't make a difference since puppeteer will "scroll" to take a screenshot of the whole component. I've asked Amaury if this behaviour is expected or whether puppeteer should be configured to fix height

packages/slice-machine/server/src/api/screenshots/puppeteer.ts
```diff
@@ -39,12 +43,14 @@ export default {
...
   screenshotUrl: string,
-  pathToFile: string
+  pathToFile: string,
+  screenWidth: string
 ): Promise<void> => {
   // Create an incognito context to isolate screenshots.
   const context = await browser.createIncognitoBrowserContext();
   // Create a new page in the context.
   const page = await context.newPage();
+  await page.setViewport({ width: Number(screenWidth), height: 600 });
```
Heyrows: The height shouldn't be hardcoded.
For now this feature allows the user to take a screenshot with only 3 possible values.
In the future we could image the user being able to pass any value through inputs or something like that, to provide more flexibility.
Even though the height is 600 for now, it should be controlled by the frontend entirely.
The frontend being the one to set the intention and parameters of the screenshot.

Heyrows: I think the screenWidth property should be a dimension object.

packages/slice-machine/server/src/api/screenshots/puppeteer.ts
```diff
@@ -39,12 +43,14 @@ export default {
 const generateScreenshot = async (
```
Heyrows, line 44: Adding this here, I didn't see the `captureBeyondViewport: false` option passed to the screenshot method.

Heyrows: As discussed, this would restrict the screenshot to the viewport that was previously set with the right dimensions.

packages/slice-machine/src/modules/useSliceMachineActions.ts
```diff
@@ -379,13 +379,15 @@ const useSliceMachineActions = () => {
   const generateSliceScreenshot = (
     variationId: string,
     component: ComponentUI,
-    setData: (data: any) => void
+    setData: (data: any) => void,
+    screenWidth = "1200"
```
Heyrows: I wouldn't set a default value here to avoid weird screenshot being taken if our code is fucked up.

This would be covering a code problem and user's screenshot would be fucked up.

packages/slice-machine/src/apiClient.ts
```diff
@@ -94,12 +97,10 @@ export const renameSlice = (
 };
 
 export const generateSliceScreenshotApiClient = (
-  sliceName: string,
-  libraryName: string,
-  variationId: string
+  params: ScreenshotRequest
 ): Promise<AxiosResponse<ScreenshotResponse>> => {
   return axios.get(
```
Heyrows: With so much parameters, should we transform this to be a `POST` ?

packages/slice-machine/server/src/api/screenshots/puppeteer.ts
```diff
@@ -14,20 +15,23 @@ export default {
...
+    screenWidth,
   }: PuppeteerHandleProps): Promise<void> => {
     const { warning } = await testUrl(screenshotUrl);
     if (warning) throw new Error(warning);
 
-    if (!puppeteerBrowserPromise)
+    if (!puppeteerBrowserPromise) {
       puppeteerBrowserPromise = puppeteer.launch({
-        args: [`--window-size=1200,800`],
+        defaultViewport: null,
```
Heyrows, line 26: 👍

Heyrows (changes requested):

packages/slice-machine/server/src/api/screenshots/puppeteer.ts
danigo99: This is done further down in the puppeteer screenshot function

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -106,7 +106,10 @@ const SliceBuilder: React.FC<SliceBuilderProps> = ({
 
   const onTakingSliceScreenshot = () => {
     checkSimulatorSetup(true, () =>
-      generateSliceScreenshot(variation.id, component, setData)
+      generateSliceScreenshot(variation.id, component, setData, {
```
Heyrows: guess that's raw value because this part will disappear in the next tickets 👍

packages/slice-machine/lib/models/common/Screenshots.ts
```diff
@@ -3,11 +3,15 @@ import { VariationSM } from "@slicemachine/core/build/models";
 
 export type Screenshots = Record<VariationSM["id"], ScreenshotUI>;
 
+export interface ScreenDimensions {
+  width: string;
```
Heyrows: can't we have numbers directly ?

Heyrows: As it isn't query parameters anymore but body

Heyrows (approved):
I approve because the PR is good, the string thing is minimal.

MarcMcIntosh:
Which branch is it going on to?

Heyrows:
> Which branch is it going on to?

It's going into a first ticket #598

packages/slice-machine/server/src/api/screenshots/puppeteer.ts
```diff
@@ -72,6 +72,12 @@ const generateScreenshot = async (
       await element.screenshot({
         path: pathToFile,
         captureBeyondViewport: false,
```
Heyrows: I'd remove that option.
We don't seem to use it anymore.

packages/slice-machine/src/modules/screenshots/sagas.ts
```diff
@@ -21,40 +21,15 @@ import { openToasterCreator, ToasterType } from "@src/modules/toaster";
 export function* generateSliceScreenshotSaga({
   payload,
 }: ReturnType<typeof generateSliceScreenshotCreator.request>) {
-  const { component, variationId, setData, screenDimensions } = payload;
+  const { component, variationId, screenDimensions } = payload;
```
Heyrows, line 24: really love the simplification of the saga. Great job 👍

Heyrows (approved):

PR #612, opened 2022-09-27
Changes page screenshots

hypervillain:
1. Display count of missing screenshots in each slice card
2. Open screenshots modal with list of variations and current state


packages/slice-machine/src/css/modal.css
```diff
@@ -19,7 +19,9 @@ body {
 }
 
 /* Review modal animation   */
-
+.ReactModal__Overlay {
+  z-index: 1;
```
MarcMcIntosh, line 23: what's this for?

packages/slice-machine/lib/models/ui/Slice/index.tsx
```diff
@@ -10,25 +18,29 @@ import { StatusBadge } from "@components/StatusBadge";
...
-  ...sx,
-});
+const defaultSx = (sx: ThemeUIStyleObject = {}) =>
+  ({
+    bg: "transparent",
+    border: "none",
+    position: "relative",
+    transition: "all 100ms cubic-bezier(0.215,0.60,0.355,1)",
+    ...sx,
+  } as ThemeUICSSObject);
```
MarcMcIntosh: this needed to be cast? would it not work as a return type?

packages/slice-machine/lib/models/ui/Slice/index.tsx
```diff
@@ -10,25 +18,29 @@ import { StatusBadge } from "@components/StatusBadge";
...
+    transition: "all 200ms ease-in",
+    p: 3,
+    position: "relative",
+    ...sx,
+    "&:hover": {
+      transition: "all 200ms ease-out",
+      bg: "sidebar",
+      border: (t: Theme) => `1px solid ${t.colors?.sidebar as string}`,
+    },
+  } as ThemeUICSSObject);
```
MarcMcIntosh: Is there an error when stating the return type?

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+import { isModalOpen } from "@src/modules/modal";
+import { SliceMachineStoreType } from "@src/redux/type";
+import { ModalKeysEnum } from "@src/modules/modal/types";
+
+import { Flex, Heading, Close, Box, Text } from "theme-ui";
+
+import { ComponentUI } from "@lib/models/common/ComponentUI";
+import { useState } from "react";
+
+const VariationIcon = ({ isValid }: { isValid?: boolean }) => (
```
MarcMcIntosh: Is there no return type warning?

packages/slice-machine/src/css/modal.css
hypervillain: this was previously made for one single modal, I needed it too so generalized to all modals

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  const [selectedSliceVariation, setSelectedSliceVariation] = useState<
+    [string, string]
+  >(
+    slices.some(
+      (s) =>
+        s.model.id === defaultSelectedSliceID &&
+        s.model.variations.some((v) => v.id === defaultSelectedSliceVariationID)
+    )
+      ? (defaultSelectedSliceVariation as [string, string])
```
MarcMcIntosh: this could be undefined, and empty array or a tuple of strings right?

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+        Header={({ radius }: { radius: string | number }) => (
+          <Flex
+            sx={{
+              p: "16px",
+              pl: 4,
+              alignItems: "center",
+              justifyContent: "space-between",
+              borderTopLeftRadius: radius,
+              borderTopRightRadius: radius,
+              // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
```
MarcMcIntosh: why the linter error

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
```diff
@@ -0,0 +1,205 @@
+import Card from "@components/Card";
+import { AiOutlinePicture } from "react-icons/ai";
+import { RiErrorWarningLine } from "react-icons/ri";
+import SliceMachineModal from "@components/SliceMachineModal";
```
MarcMcIntosh, line 8: is this easer than `../SliceMachineModal` ?

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
hypervillain: if it's an empty array, the previous check returns false so you end up with an array of 2 elements in both cases I believe

packages/slice-machine/lib/models/ui/Slice/index.tsx
hypervillain: it did work with return type, my bad!

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
hypervillain: no idea!

hypervillain: do you want me to change it?

packages/slice-machine/pages/changes.tsx
```diff
@@ -36,6 +39,11 @@ const changes: React.FunctionComponent = () => {
 
   const [changesPushed, setChangesPushed] = useState<string[]>([]);
   const [error, setError] = useState<SyncError | null>(null);
+  useEffect(() => {
+    return () => {
+      closeScreenshotsModal();
```
Heyrows, line 37: Shouldn't just a click outside of the modal close it ?

packages/slice-machine/pages/changes.tsx
```diff
@@ -114,7 +129,7 @@ const changes: React.FunctionComponent = () => {
           MainBreadcrumb={<Text ml={2}>Changes</Text>}
           breadrumbHref="/changes"
         />
-        {renderPageContent()}
+        <PageContent onOpenModal={openScreenshotsModal} />
```
Heyrows: the goal of the actions is that they can be called everywhere and we avoid to pass down properties over and over.

I would not pass down the property and just call `useSliceMachineAction` when you need to take a screenshot, this makes it easier to find where the action is called afterwards.

packages/slice-machine/pages/changes.tsx
```diff
@@ -80,6 +89,12 @@ const changes: React.FunctionComponent = () => {
 
   return (
     <Container sx={{ display: "flex", flex: 1 }}>
+      {unSyncedSlices.length ? (
+        <ScreenshotChangesModal
```
Heyrows: The Modal should be global, not scope to this page so we have only one instance of it.
When the open modal option is called, slices or slicesIDs should be passed to it along with a variationID to select if need be.

Why ?
- The modal is cross page, it should have a single instance
- The modal needs on the change page to be opened with different context (list of slices or single slice with pre-selected variationID), passing this context as payload makes it way easier to use and understand.

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+import { isModalOpen } from "@src/modules/modal";
+import { SliceMachineStoreType } from "@src/redux/type";
+import { ModalKeysEnum } from "@src/modules/modal/types";
+
+import { Flex, Heading, Close, Box, Text } from "theme-ui";
+
+import { ComponentUI } from "@lib/models/common/ComponentUI";
+import { useState } from "react";
+
+const VariationIcon: React.FC<{ isValid?: boolean }> = ({ isValid }) => (
```
Heyrows, line 19: put those in different files please, 1 component 1 file.
They can all be scoped in the ScreenshotChangesModal folder.

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+            </Box>
+          );
+        })}
+      </Box>
+    </Box>
+  </>
+);
+
+const ScreenshotChangesModal = ({
+  slices,
```
Heyrows, line 120: It shouldn't take any parameters, they should come from the payload of the action.

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+  onClose,
+}: {
+  slices: ComponentUI[];
+  defaultSelectedSliceVariation?: [string, string] | [];
+  onClose: () => void;
+}) => {
+  const { isOpen } = useSelector((store: SliceMachineStoreType) => ({
+    isOpen: isModalOpen(store, ModalKeysEnum.SCREENSHOTS),
+  }));
+  const [defaultSelectedSliceID, defaultSelectedSliceVariationID] =
```
Heyrows: should be an object, `{ sliceID, variationID }` makes it clearer and more secure than an array.

Heyrows: It should also be a type created for this, i see it's being used in a useState below.

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
Heyrows: I think this might fix it.
```suggestion
              borderBottom: (t) => `1px solid ${String(t.colors?.borders)}`
```

Heyrows (changes requested):

packages/slice-machine/pages/changes.tsx
hypervillain: The modal is not cross page. This is like saying that a Button is cross page because it's used in several places. The Modal _is_ passed an array of slices which _is_ the context it needs to know about

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
hypervillain: No, "1 component 1 file" is not a rule you should comply with. See this tweet from Dan Abramov and many other articles: https://twitter.com/dan_abramov/status/1439304400464449536?s=20&t=7Xi2iln4sENglrouwI5RWA. VariationIcon is solely used in that component and is split for convenience.

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
hypervillain: I agree! No idea why I went that way lol

packages/slice-machine/pages/changes.tsx
hypervillain: agreed

packages/slice-machine/pages/changes.tsx
hypervillain: This is a different problem that I tried to explain previously: because we split state from local page, if the user switches pages while the modal is open, the state of the modal does stay open, although the component is unmounted. When they come back to the page, the modal is mounted and opened. This callback is called on unmount and prevents this to happen. Try it for yourself!

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
hypervillain: Can you show me how you would implement the modal in a gist or something?

packages/slice-machine/pages/changes.tsx
Heyrows: The user shouldn't be able to switch state without clicking outside of the modal.

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
Heyrows: I use that rule for code readability.
Big files are hard to read, if this is used only here, then i'd create a new file in the same folder, it won't be used anywhere else but the code would be scoped and easier to read.

This could prevent big chunks of imports in one file and sometimes makes it easier when a fix is to make to find the right component to make the fix on.

packages/slice-machine/pages/changes.tsx
Heyrows: Like i said, i think the modal should be global, have a single instance.

That way
- it can be initialised using the event to open it, passing Slices in the payload & all the parameters needed required.
- modals are grouped together and easier to find / interact with.
image

We want this modal to be usable on many pages with different context so to avoid complexity in the code of each pages passing down functions to open the modal, initialising the modal with different state that changes, ... I would definitely makes it global and awaiting to be fed with data to open.

Heyrows: PS: i'd use SliceIDs in the payload to always have Slices up to dates (ex: when a screenshot is taken on another tab to be updated).

packages/slice-machine/pages/changes.tsx
hypervillain: cmd + arrow_left

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
MarcMcIntosh: weird, you know type-gaurds can help get rid o the need to cast you'll can find a few examples by looking for functions using the `is` keyword

arnaudlewis: Can we maybe provide some different types here to clarify a bit?
```
slices: ComponentUI[]
defaultSelectedSliceVariation?: [string, string] | []
```
It could maybe become something like:
```
slices: NonEmptyArray<ComponentUI>
defaultSelectedSliceVariation?: [string, string] | undefined
```
It would ensure that you first check that `defaultSelectedSliceVariation` exists before destructuring and so no need for casting later on.
Also if undefined, you fallback with an indexed access on slices that could be an empty array, If you can be absolutely sure it won’t be empty, you can encode it in the type with my suggestion above, WDYT @hypervillain ?
I think the cast here is not the issue, it’s more about how far the types are from the actual truth about what the app will provide.

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
arnaudlewis: Since there are already aliases, why change to relative path? Maybe it’s a more global topic aliases VS relative path that should be tackle in a different PR if you reach a consensus.

packages/slice-machine/pages/changes.tsx
arnaudlewis: Agreed, you can’t consider only mouse actions.

packages/slice-machine/pages/changes.tsx
arnaudlewis: I agree with @hypervillain about the context but you’re also right @Heyrows that you could have one instance of a modal in your app. You can’t implement it that simply, but could rely on tools like portals for instance.
I’d suggest here to consider using tools like Radix that will handle this kind of behaviour for you as well as keyboard controls, closing the modal when needed and so on.
WDYT?

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
arnaudlewis: I don’t know what are your convention on SM but either you should make sure that either a component remain in the same file if used only specifically in one context or if it’s too big you can split but you need to make sure that it cannot be used elsewhere. By exporting it you take that risk.

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
arnaudlewis: You can refer to my comment below but you’ll need a way to provide data to your modal, this can’t happen without props

arnaudlewis: I suggest a way to simplify the implementation with tools like Radix but I’d suggest you @Heyrows that if you want to keep it homemade, you should maybe provide some insights on how you would build it. You have very different approach in mind and I think you should weight in and provide a concrete example that you can then discuss about together to reach consensus.

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
hypervillain: updated

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
hypervillain: I definitely agree. VariationIcon is only used there and is mostly a way to split huge chunks of JSX

packages/slice-machine/pages/changes.tsx
hypervillain: I'm going for an in-between where the SliceBuilder has one instance of modal and the Changes page too, and switches context based on user click. You can't really see that now because I did not code the SliceCards part. See this example (I used it to test if using functions in `useState` could give us free updates but I might do things more traditionally): https://codesandbox.io/s/dazzling-borg-e86vvg?file=/src/App.js

hypervillain: The biggest problem with passing data via payload is that you don't get "free" updates that you get by passing them as props

arnaudlewis: Seems like a good in-between for now and true about props.

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
```diff
@@ -60,11 +63,12 @@ const VariationsList = ({
         }}
       >
         {slice.model.variations.map((variation) => {
+          const { sliceID, variationID } = variationSelector;
           const isSelectedVariation =
-            selectedSliceVariation[0] === slice.model.id &&
-            selectedSliceVariation[1] === variation.id;
+            sliceID === slice.model.id && variationID === variation.id;
```
Heyrows, line 68: 👍

Heyrows (approved):

PR #618, opened 2022-09-30
Mm/image form refactor

MarcMcIntosh:
## Types of changes

## Checklist:


packages/slice-machine/lib/models/common/widgets/Image/components.tsx
```diff
@@ -75,84 +78,103 @@ export const AddThumbnailButton = ({ onClick }) => (
...
+          // @ts-ignore:enable
         }}
         formField={{ label: `Name${requiredChar}`, placeholder: "main" }}
         field={
           prefix === "config.constraint"
-            ? { value: "main", readOnly: true }
-            : // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/restrict-template-expressions, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
+            ? // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+              // @ts-ignore:disable
+              { value: "main", readOnly: true }
```
hypervillain: You can add a `fieldProps` object and pass it to `FormFieldInput` instead. This way you can pass both `readOnly` and `type:number` (see comment below)

packages/slice-machine/lib/models/common/widgets/Image/components.tsx
```diff
@@ -75,84 +78,103 @@ export const AddThumbnailButton = ({ onClick }) => (
...
       <FormFieldInput
-        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
         fieldName={`${prefix}.width`}
         meta={{
           ...meta,
-          // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/restrict-template-expressions, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore:disable
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
           error: meta.error && meta.error.width,
```
hypervillain: I tested the form and it seems `error` is always a string and it actually never displayed! If you need to know where the error comes from (name, width or height) you will need additional logic/checks

packages/slice-machine/lib/models/common/widgets/Image/components.tsx
```diff
@@ -75,84 +78,103 @@ export const AddThumbnailButton = ({ onClick }) => (
...
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore:enable
         }}
         formField={{ label: `Width (px)${requiredChar}`, placeholder: " " }}
         field={{
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore:disable
           type: "number",
```
hypervillain: Pass this to `fieldProps` to not conflict with TS field type

hypervillain (changes requested):
When testing the form, I was able to save although I did not pass a valid value to width/height. It also lets you write negative values. After filling the form, it seems that the `errors` property is always a string, so things like `error: meta.error && meta.error?.name` will never display

packages/slice-machine/lib/models/common/widgets/Image/components.tsx
MarcMcIntosh: I manged to get an object but i had to change the type of the field so the validator would fail :/

packages/slice-machine/lib/models/common/widgets/Image/Form.tsx
```diff
@@ -0,0 +1,201 @@
...
+    validate: () =>
+      yup
+        .array()
+        .defined()
+        .of(
+          yup.object().test({
+            name: "Thumbnails",
+            message: "Must set name and width or height at minimum",
+            test: function (value) {
+              const hasWidth = typeof value.width === "number" && value.width;
```
Heyrows, line 40: Is 0 considered as a valid value or does it needs to be higher than 0 ?

packages/slice-machine/lib/models/common/widgets/Image/Form.tsx
```diff
@@ -0,0 +1,201 @@
...
+} from "./components";
+import { TabFields } from "@slicemachine/core/build/models/CustomType";
+
+const FormFields = {
+  label: DefaultFields.label,
+  id: DefaultFields.id,
+  constraint: {
+    validate: () =>
+      yup.object().defined().shape({
+        width: yup.number().nullable(),
```
Heyrows, line 26: Does nullable means the value is `null` if no value has been provided ?
Is it the behavior and value on Wroom and most important, is it accepted by the model validators ?

packages/slice-machine/lib/models/common/widgets/Image/Form.js
```diff
@@ -1,196 +0,0 @@
...
-              </Fragment>
-            )}
-          />
-        </Card>
-      </Col>
-    </FlexGrid>
-  );
-};
-
-FormFields.constraint = {
```
Heyrows, line 160: damn i'm glad you found that 😓

packages/slice-machine/lib/models/common/widgets/Image/Form.tsx
MarcMcIntosh: 0 is auto, so it's not a valid number (the last commit fixed this) :)

packages/slice-machine/lib/models/common/widgets/Image/Form.tsx
MarcMcIntosh: I have no idea, I didn't touch that part.
but in the model it can be null, which is auto

Heyrows (approved):

packages/slice-machine/lib/models/common/widgets/Image/components.tsx
MarcMcIntosh: found a better way :)

packages/slice-machine/lib/models/common/widgets/Image/components.tsx
MarcMcIntosh: found a better way :)

PR #620, opened 2022-10-04
Mm/fix peer deps

MarcMcIntosh:
Fixes some warnings about missing peer-deps when using warn workspaces
## Types of changes

## Checklist:


packages/next-slicezone/package-lock.json
```diff
@@ -7061,6 +7061,20 @@
 				"is-typedarray": "^1.0.0"
 			}
 		},
+		"node_modules/typescript": {
```
Heyrows: I don't think there should be changes to this file, especially because the package.json hasn't been modified.

But on the other hand, I've seen this change been added in several PRs already so maybe let's pass it once and for all as long as the package isn't published after.

packages/slice-machine/package.json
```diff
@@ -92,6 +95,7 @@
     "yup": "^0.32.9"
   },
   "devDependencies": {
+    "@babel/core": "^7",
```
Heyrows, line 98: I don't think the ticket mentioned babel core.

Heyrows (commented):
- Should those dependencies be added as peerDependencies or dependencies ?
- I see that the core, init and client also uses @prismicio/types, should they have those dependencies as well ?
  - I could see why not as Slice Machine is the only package being "installed", but asking just in case.

packages/next-slicezone/package-lock.json
MarcMcIntosh: good point, I'm not sure why it was added :/

packages/slice-machine/package.json
MarcMcIntosh: the slack thread did :)

packages/next-slicezone/package-lock.json
MarcMcIntosh: removed :)

packages/slice-machine/package.json
MarcMcIntosh: `warning "slice-machine-ui > @babel/register@7.18.9" has unmet peer dependency "@babel/core@^7.0.0-0".`

Heyrows: ok 👍

Heyrows (approved):

PR #629, opened 2022-10-07
Unify styling of main primary buttons

Heyrows:
## Types of changes

- [X] Bug fix (a non-breaking change which fixes an issue)

## Description

In order to unify the design of buttons
- I created a first component to display proper display around the loading state -> @components/Button
- modified the existing theme to match the new design for primary buttons only
- used the new button component and theme to replace the following button
  - Button to create a custom type  image
  - Button to create a slice image
  - Button to sync changes image
  - Button to save changes to the file system for both Slice and Custom Types image
  - Button to check the Slice Simulator configuration image
  - Every confirmation Button that is in a modal like the rename button image

## Checklist:


packages/slice-machine/components/Button/index.tsx
```diff
@@ -2,39 +2,72 @@ import React from "react";
...
 };
 
-const Button: React.FunctionComponent<SliceMachineButtonProps> = ({
+// Small helper to allow us to target spinner and icon in the CY
+const cyIdBuilder = (dataCy: string | undefined, id: string) => {
+  if (dataCy) return `${dataCy}-${id}`;
+  return "";
+};
+
+// If you don't use an icon, don't forget to pass a min-width property so the button doesn't change width on loading.
```
danigo99, line 26: If we're worried about this not being followed and eventually messing up the UI, it might be worth enforcing it in the type? I.e. having a IconButtonProps and IconlessButtonProps then using the union of it in the component

packages/slice-machine/components/Button/index.tsx
```diff
@@ -2,39 +2,72 @@ import React from "react";
...
+    disabled={disabled}
+    onClick={!isLoading ? onClick : undefined}
+    variant={variant}
+    {...rest}
+  >
+    {isLoading ? (
+      <>
+        <Spinner
+          size={16}
+          color="#F7F7F7"
```
danigo99: How come we're using a colour literal here instead of one from the theme file?

packages/slice-machine/components/Button/index.tsx
Heyrows: I see what you mean, good point

packages/slice-machine/components/Button/index.tsx
Heyrows: I took it as it was, good catch 👍

packages/slice-machine/components/Button/index.tsx
Heyrows: Ok so I implemented it and then realise there are some issues

- the button is used in the modal card on which we know nothing of the height of the button as the label is passed down through a property. Those are not buttons that were required to changed, but they used the previous component so i tried to keep the same behavior.
image

- For now there is only one place where we need to use the width property it's in `packages/slice-machine/lib/builders/SliceBuilder/SetupDrawer/steps/common.tsx`. It actually needs a max-width and not even a min-width property

Heyrows: I can set a minimum for the first case but it would still glitch. (as there are different label -> button width that i cannot predict).
For the second case, I tried having types like you suggested, it add quite a lot of complexity and I didn't manage to make the type works unless having a property with a special type.

danigo99: Makes sense 👍 . Probably not worth digging into much more anyway since the bug should be pretty obvious on first use if the developer misses it 😅

danigo99 (approved):

PR #631, opened 2022-10-10
Upload screenshot from modal

hypervillain:
Added "Upload custom screenshot" button, dropzone and UI to the Screenshots Modal. Updated Redux stores to reflect screenshot updates. Added screenshots modal to all slices lists + slice builder page.

TODO: remove Sidebar buttons on hover in profit of the new button


packages/slice-machine/components/ScreenshotChangesModal/Views/Default.tsx
```diff
@@ -0,0 +1,23 @@
+import { Image, Flex, Spinner } from "theme-ui";
+
+import { ViewRendererProps } from "./";
+
+export default function DefaultView({
+  screenshot,
+  isLoadingScreenshot,
+}: ViewRendererProps) {
+  return (
```
MarcMcIntosh, line 11: would `return isLoadingScreenshots ? (<Flex>.....</Flex>) : (<Image ... />)` work?
or are fragments better?

packages/slice-machine/components/ScreenshotChangesModal/Views/Default.tsx
```diff
@@ -0,0 +1,33 @@
+import { Image, Text, Flex, Spinner } from "theme-ui";
+
+import { ViewRendererProps } from "./";
+
+export default function DefaultView({
+  screenshot,
+  isLoadingScreenshot,
+}: ViewRendererProps) {
+  return (
```
MarcMcIntosh, line 11: would `return isLoadingScreenShot ? (<Flex ...) : (<Image ...>)` also work or are fragments better?

packages/slice-machine/src/modules/useSliceMachineActions.ts
```diff
@@ -402,14 +402,12 @@ const useSliceMachineActions = () => {
...
     variationId: string,
     component: ComponentUI,
-    setData: (data: any) => void,
     file: Blob
   ) => {
     dispatch(
       generateSliceCustomScreenshotCreator.request({
         variationId,
         component,
-        setData,
```
MarcMcIntosh, line 412: nice

packages/slice-machine/src/modules/slices/index.ts
```diff
@@ -210,7 +210,12 @@ export const slicesReducer: Reducer<SlicesStoreType | null, SlicesActions> = (
                   ...component,
                   screenshots: {
                     ...component.screenshots,
-                    [variationId]: screenshot,
+                    ...(screenshot
+                      ? {
+                          [variationId]: screenshot,
+                        }
+                      : null),
```
MarcMcIntosh: can or should spread be called on null?

packages/slice-machine/src/modules/slices/index.ts
```diff
@@ -210,7 +210,12 @@ export const slicesReducer: Reducer<SlicesStoreType | null, SlicesActions> = (
...
+                  } as Record<string, ScreenshotUI>,
```
MarcMcIntosh: was it necessary to cast this?

packages/slice-machine/src/modules/screenshots/sagas.ts
```diff
@@ -62,41 +63,17 @@ export function* generateSliceScreenshotSaga({
...
-      done: false,
-      error: null,
-      status: null,
-      imageLoading: true,
-    });
     const response = (yield call(
       generateSliceCustomScreenshotApiClient,
       form
     )) as SagaReturnType<typeof generateSliceCustomScreenshotApiClient>;
-    if (response.status > 209) {
```
MarcMcIntosh, line 83: what happens if the request fails for some reason?

packages/slice-machine/lib/models/common/ComponentUI.ts
```diff
@@ -8,19 +8,20 @@ import { CustomTypeMockConfig, SliceMockConfig } from "./MockConfig";
 
 export const createScreenshotUrl = (
   baseUrl: string,
-  pathToScreenshot: string
+  pathToScreenshot: string,
+  hash: string
 ): string =>
   `${baseUrl}/api/__preview?q=${encodeURIComponent(
```
MarcMcIntosh, line 14: is this localhost?

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -114,7 +114,7 @@ const SliceBuilder: React.FC<SliceBuilderProps> = ({
   };
 
   const onTakingSliceCustomScreenshot = (file: Blob) => {
-    generateSliceCustomScreenshot(variation.id, component, setData, file);
+    generateSliceCustomScreenshot(variation.id, component, file);
```
MarcMcIntosh: how are we giving feedback to the user?

packages/slice-machine/components/ScreenshotChangesModal/useCustomScreenshot.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { acceptedImagesTypes } from "@lib/consts";
+import { ComponentUI } from "@lib/models/common/ComponentUI";
+
+type CSProps = {
+  variationID: string;
+  slice: ComponentUI;
+  onHandleFile: (variationID: string, slice: ComponentUI, file: File) => void;
+};
+
+export default function useCustomScreenshot({
```
MarcMcIntosh, line 51: I'm confused is this a hook, or a component?
Where's it used?

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
```diff
@@ -200,11 +201,17 @@ const ScreenshotChangesModal = ({
...
           >
-            SliceID: {variationSelector.sliceID}
-            <br />
-            VariationID: {variationSelector.variationID}
+            <Views
+              variationID={variationSelector.variationID}
+              slice={
+                slices.find(
+                  (s) => s.model.id === variationSelector.sliceID
+                ) as ComponentUI
```
MarcMcIntosh: is this cast to ComonentUI because it may not be found?

packages/slice-machine/components/ScreenshotChangesModal/Views/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+  variationID: string;
+  isLoadingScreenshot: boolean;
+};
+
+const viewRenderer = {
+  [ScreenshotView.Default]: DefaultView,
+  [ScreenshotView.EmptyState]: EmptyState,
+};
+
+function VariationScreenshot({
```
MarcMcIntosh: Are there not any linter errors/warnings about missing return types?

packages/slice-machine/components/ScreenshotChangesModal/Views/EmptyState.tsx
```diff
@@ -0,0 +1,118 @@
...
+import useCustomScreenshot from "../useCustomScreenshot";
+
+const sharedFlexSx = {
+  justifyContent: "center",
+  alignItems: "center",
+  height: "100%",
+  width: "100%",
+};
+
+const UploadIcon = ({ isActive }: { isActive: boolean }) => (
```
MarcMcIntosh: `const UploadIcon: React.FC<{sActive: boolean }> = ({ isActive }: => (...)`
would also work here :)
.... aslo it's weird that there's no complaints from eslint that there's no return type :/

MarcMcIntosh (commented):
Left a few comments, I'm also wondering can the upload fail and how would the user know if the screenshot failed to upload?

packages/slice-machine/components/ScreenshotChangesModal/Views/Default.tsx
hypervillain: it would work the same way but my personal preference goes to using a fragment in case the layout evolves

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
hypervillain: yesss, it's not very clean but you "know" the slice exists. Do you have a better way to express this?

packages/slice-machine/components/ScreenshotChangesModal/useCustomScreenshot.tsx
hypervillain: It's a hook that returns multiple properties, one being a component

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
hypervillain: the saga is wrapped with a `withLoading` that can be queried in the component. The success is easy to check but you are right that the failure is not handled. I'll add a toaster message on error!

packages/slice-machine/lib/models/common/ComponentUI.ts
hypervillain: yes

packages/slice-machine/src/modules/screenshots/sagas.ts
hypervillain: nothing indeed, I'll add a toaster message

packages/slice-machine/src/modules/slices/index.ts
hypervillain: spread on null and undefined works properly (equal to empty object)

packages/slice-machine/src/modules/slices/index.ts
hypervillain: I think! https://github.com/microsoft/TypeScript/issues/27273

packages/slice-machine/components/ScreenshotChangesModal/Views/index.tsx
hypervillain: hmm no, to my understanding they have never been necessary but I might be mistaken?

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
hypervillain: oh actually the error toast is there already, so you get feedback when it fails (and when it succeeds, you just see the screenshot being updated)

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
hypervillain: would it be ok to replace the cast with an exclamation mark?

packages/slice-machine/src/modules/slices/index.ts
MarcMcIntosh: could be the null, (null is a object)

packages/slice-machine/src/modules/slices/index.ts
MarcMcIntosh: I think using an empty object would save having to cast the object later on :)

packages/slice-machine/components/ScreenshotChangesModal/useCustomScreenshot.tsx
MarcMcIntosh: So I had to google this pattern, I'm not sure it's a good idea or not, I'll need to take a closer look :)
https://www.reddit.com/r/reactjs/comments/9yq1l8/how_do_you_feel_about_a_hook_returning_components/?sort=confidence

https://dev.to/droopytersen/new-react-hooks-pattern-return-a-component-31bh

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
MarcMcIntosh: This could be doe with `filter` by id,  then `map` to render the view with the slice which saves having to cast anything :)

packages/slice-machine/components/ScreenshotChangesModal/useCustomScreenshot.tsx
MarcMcIntosh: There's a bit more information out there, but the comments in those ones are pretty interesting :)

packages/slice-machine/components/ScreenshotChangesModal/Views/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+  variationID,
+  slice,
+}: {
+  variationID: string;
+  slice: ComponentUI;
+}): JSX.Element {
```
MarcMcIntosh: use `React.FC`, jsx element can be a bit weird :)

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
hypervillain: I updated the code in a third way, can you check if it's any good to you?

packages/slice-machine/components/ScreenshotChangesModal/useCustomScreenshot.tsx
hypervillain: I think it would be an issue if it returned a component instance, but here it's just the component definition!

packages/slice-machine/components/ScreenshotChangesModal/Views/index.tsx
hypervillain: ahah ok, I suck at this sorry

packages/slice-machine/components/ChangesItems/ChangesItems.tsx
```diff
@@ -159,6 +130,13 @@ export const ChangesItems: React.FC<ChangesItemsProps> = ({
                   authStatus,
                   isOnline,
                 },
+                onUpdateScreenshot: (e: React.MouseEvent) => {
+                  e.preventDefault();
+                  onOpenModal({
+                    sliceFn: (s: ComponentUI[]) =>
```
MarcMcIntosh: what's sliceFn do?

packages/slice-machine/src/modules/slices/index.ts
MarcMcIntosh: Did using an empty objet not work?

packages/slice-machine/src/modules/selectedSlice/reducer.ts
```diff
@@ -104,6 +108,18 @@ export const selectedSliceReducer: Reducer<
         return prevState;
       }
     }
+    case getType(generateSliceScreenshotCreator.success):
+    case getType(generateSliceCustomScreenshotCreator.success): {
+      if (!prevState) return prevState;
```
MarcMcIntosh, line 113: so if prevState is null it stays that way?

packages/slice-machine/src/hooks/useScreenshotChangesModal.ts
```diff
@@ -0,0 +1,34 @@
...
+  modalPayload: ModalPayload;
+  onOpenModal: (p: ModalPayload) => void;
+  openScreenshotsModal: () => void;
+};
+
+export const useScreenshotChangesModal = (): Payload => {
+  const { openScreenshotsModal } = useSliceMachineActions();
+
+  const [modalPayload, setModalPayload] = useState<ModalPayload>({
+    sliceFn: (s: ComponentUI[]) => s,
```
MarcMcIntosh: what's sliceFn do?

packages/slice-machine/pages/slices.tsx
```diff
@@ -191,6 +192,15 @@ const SlicesIndex: React.FunctionComponent = () => {
                               authStatus,
                               isOnline,
                             },
+                            onUpdateScreenshot: (e: React.MouseEvent) => {
+                              e.preventDefault();
+                              onOpenModal({
+                                sliceFn: (s: ComponentUI[]) =>
```
MarcMcIntosh: what's sliceFn do?

packages/slice-machine/lib/models/ui/Slice/index.tsx
```diff
@@ -140,24 +137,17 @@ const SliceDescription = ({
...
 );
 
-const ScreenshotMissingBanner = ({
-  visible,
-  slice,
-}: {
-  visible?: boolean;
-  slice: ComponentUI;
-}) => {
+const ScreenshotMissingBanner = ({ slice }: { slice: ComponentUI }) => {
```
MarcMcIntosh: `const ScreenshotMissingBanner: React.FC<{ { slice: ComponentUI }}> = ({ slice }) =>....`

packages/slice-machine/src/modules/useSliceMachineActions.ts
```diff
@@ -402,14 +402,12 @@ const useSliceMachineActions = () => {
   const generateSliceCustomScreenshot = (
```
MarcMcIntosh, line 402: should this file be in the hooks directory?

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -142,15 +123,7 @@ const SliceBuilder: React.FC<SliceBuilderProps> = ({
...
-        SideBar={
-          <SideBar
-            component={component}
-            variation={variation}
-            onScreenshot={onTakingSliceScreenshot}
-            onHandleFile={onTakingSliceCustomScreenshot}
-            imageLoading={data.imageLoading}
-          />
-        }
+        SideBar={<SideBar component={component} variation={variation} />}
```
MarcMcIntosh, line 126: Is this a child element?

packages/slice-machine/components/StatusBadge/index.tsx
```diff
@@ -30,7 +30,6 @@ export const StatusBadge: React.FC<StatusBadgeProps> = ({
     <>
       <Text data-for={`${modelId}-tooltip`} data-tip>
         <Badge
-          mr="2"
```
MarcMcIntosh, line 33: UI team is going to have fun with this 😅

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
```diff
@@ -200,11 +203,20 @@ const ScreenshotChangesModal = ({
...
               flexBasis: 0,
               minWidth: 320,
+              maxHeight: "90%",
             }}
           >
-            SliceID: {variationSelector.sliceID}
-            <br />
-            VariationID: {variationSelector.variationID}
+            {(() => {
+              const slice = slices.find(
```
MarcMcIntosh, line 210: you could use filter and map here, or reduce and map, that way it doesn't matter if it's undefined or not :)

packages/slice-machine/components/ScreenshotChangesModal/Views/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+  );
+  const { generateSliceCustomScreenshot } = useSliceMachineActions();
+  const maybeScreenshot = slice.screenshots[variationID];
+
+  const ViewRenderer = maybeScreenshot
+    ? viewRenderer[ScreenshotView.Default]
+    : viewRenderer[ScreenshotView.EmptyState];
+
+  const { FileInputRenderer, fileInputProps } = useCustomScreenshot({
+    onHandleFile: (file: File) => {
```
MarcMcIntosh, line 53: neet

MarcMcIntosh (commented):
Left a few more comments

packages/slice-machine/components/StatusBadge/index.tsx
hypervillain: ahah why?

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
hypervillain: sort of, it passes a component as property

packages/slice-machine/src/modules/useSliceMachineActions.ts
hypervillain: to me yes but it was created after the hooks folder and outside, maybe for a reason? @Heyrows

packages/slice-machine/src/modules/selectedSlice/reducer.ts
hypervillain: it's a bit outside the scope of this PR

packages/slice-machine/src/hooks/useScreenshotChangesModal.ts
hypervillain: it filters slices!

packages/slice-machine/src/modules/slices/index.ts
hypervillain: it did!

packages/slice-machine/components/StatusBadge/index.tsx
MarcMcIntosh: margins right?

packages/slice-machine/src/hooks/useScreenshotChangesModal.ts
MarcMcIntosh: can you call it that then?

packages/slice-machine/src/modules/selectedSlice/reducer.ts
MarcMcIntosh: true, take a note of it and put it on the tech-debt board ad maybe we'l fix it later :)

packages/slice-machine/src/modules/slices/index.ts
```diff
@@ -214,8 +214,8 @@ export const slicesReducer: Reducer<SlicesStoreType | null, SlicesActions> = (
                       ? {
                           [variationId]: screenshot,
                         }
-                      : null),
-                  } as Record<string, ScreenshotUI>,
+                      : {}),
```
MarcMcIntosh, line 217: 🎉

packages/slice-machine/src/hooks/useScreenshotChangesModal.ts
hypervillain: done!

danigo99:
Something I noticed when I was working off your branch was that the screenshot was taking up the whole ScreenshotPreview component meaning that it doesn't have the background pattern like in the designs. This is the same for when there's a mobile screenshot which I don't think should be the case

image 
image

hypervillain:
@danigo99 thanks! I updated accordingly.

Capture d’écran 2022-10-19 à 11 21 21

packages/slice-machine/src/modules/useSliceMachineActions.ts
Heyrows: I think it was created there to be close to the redux logic as we modify it only when we modify redux actions.
If you want to move it I won't vote against it but I think it would be painful to find it each time we have to modify it. 👍

So do what you want, let us know.

MarcMcIntosh (approved):

PR #632, opened 2022-10-11
Open slice simulator from screenshots modal

danigo99:
## Context

[Ticket 809](https://linear.app/prismic/issue/SM-809/aauser-on-the-screenshot-modal-i-can-see-a-button-to-take-a-screenshot)

Adding a button to the screenshot modal that opens the slice simulator.

- [x] The CI is successful.

[OPT] N.B for the reviewer:

## [OPT] Preview

The cursor looks weird in the screen recording, it's just the loading icon
ezgif com-gif-maker (2)

---

image


packages/slice-machine/components/ScreenshotChangesModal/Views/index.tsx
```diff
@@ -58,9 +71,58 @@ function VariationScreenshot({
...
+      window.open(
+        `/${slice?.href}/${slice?.model.name}/${variationID}/simulator`
+      )
+    );
+
   return (
     <>
       <Flex sx={{ p: 2, pt: 0, minHeight: "48px" }}>
+        <Button
+          variant="buttons.secondarySmall"
```
hypervillain: You don't need to specify `buttons.` here because of theme-ui variant groups!
https://theme-ui.com/components/variants

danigo99:
Waiting for https://github.com/prismicio/slice-machine/pull/631 as this will be the base

packages/slice-machine/components/Button/index.tsx
```diff
@@ -34,6 +35,7 @@ export const Button: React.FunctionComponent<ButtonProps> = ({
   onClick,
   sx = {},
   variant = "primary",
+  spinnerColor = "grey01",
```
Heyrows: not sure this should be modified for consistency in the APP design, but i'm guessing you followed the design.

packages/slice-machine/components/ScreenshotChangesModal/Views/index.tsx
```diff
@@ -58,16 +72,49 @@ function VariationScreenshot({
     onHandleFile: generateSliceCustomScreenshot,
   });
 
+  const sliceView = useMemo(
```
Heyrows, line 72: interesting, didn't think useMemo could do the equivalent of a useState + useEffect.

packages/slice-machine/components/ScreenshotChangesModal/Views/index.tsx
```diff
@@ -58,16 +72,49 @@ function VariationScreenshot({
...
+          ]
+        : null,
+    [slice.model.id, variationID]
+  );
+
+  if (!sliceView) return <></>;
+
+  const openSimulator = () =>
+    checkSimulatorSetup(true, () =>
+      window.open(
```
Heyrows, line 89: OPT: Mathieu suggested a small improvement by giving is a second parameter which would be the ID, that way we can avoid to re-open pages for the same slice / variationID.

Heyrows (changes requested):
Found a glitch while trying the PR.

https://user-images.githubusercontent.com/41187899/196380981-198bb680-9fae-4504-9bb1-46d63b4256c5.mov

packages/slice-machine/components/ScreenshotChangesModal/Views/index.tsx
danigo99: I didn't actually write this, took it from the slice builder. Your comment made me read into it though and it's quite interesting: https://stackoverflow.com/questions/56028913/usememo-vs-useeffect-usestate

packages/slice-machine/components/ScreenshotChangesModal/Views/index.tsx
danigo99: Good idea 👍

packages/slice-machine/components/Button/index.tsx
danigo99: It's more about the white buttons not being able to take a grey spinner since you can't see it. I agree that it's not great though and should ideally come from the button variant. Not sure that's possible though

Heyrows: yeah that's what I understood too, we could make a function here to match variant with spinner color, would be ideal.
Not absolutely required for right now tho, it can wait.

packages/slice-machine/components/ScreenshotChangesModal/Views/index.tsx
Heyrows: Yeah super interesting as well :), seems to do the same as useCallback so I wonder what would be the difference between these two.

Heyrows (approved):

PR #633, opened 2022-10-11
make screenshot URL optional in modal so it doesn't fail before one i…

danigo99:
## Types of changes

## Description

Previously the slice preview page failed if there was no screenshot for the given slice since the URL was required for the modal. Now we don't need this to allow the page to open for the first time.

## Checklist:

### DIFF
--- a/packages/slice-machine/components/Simulator/index.tsx
+++ b/packages/slice-machine/components/Simulator/index.tsx
@@ -78,10 +78,12 @@ export default function Simulator() {
         simulatorUrl={simulatorUrl}
         sliceView={sliceView}
       />
-      <ScreenshotPreviewModal
-        sliceName={Router.router?.query.sliceName as string}
-        screenshotUrl={component.screenshots[variation.id]?.url}
-      />
+      {!!component.screenshots[variation.id]?.url && (
+        <ScreenshotPreviewModal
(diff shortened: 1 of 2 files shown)


packages/slice-machine/components/Simulator/index.tsx
```diff
@@ -80,7 +80,7 @@ export default function Simulator() {
       />
       <ScreenshotPreviewModal
```
Heyrows: OPT: I think it would make more sense to display the modal only if there is an actual screenshot.

Heyrows (approved):

packages/slice-machine/components/Simulator/index.tsx
```diff
@@ -78,10 +78,12 @@ export default function Simulator() {
         simulatorUrl={simulatorUrl}
         sliceView={sliceView}
       />
-      <ScreenshotPreviewModal
-        sliceName={Router.router?.query.sliceName as string}
-        screenshotUrl={component.screenshots[variation.id]?.url}
-      />
+      {!!component.screenshots[variation.id]?.url && (
```
Heyrows, line 81: Funny, i never really understood why people use that !! or even !!! 😅

Heyrows (approved):

PR #634, opened 2022-10-11
Mm/image form fixes

MarcMcIntosh:
Display validation error when adding an image size and name.
don't accept `e` `+` or `-` in the number field
## Types of changes

## Checklist:

### DIFF
--- a/packages/slice-machine/lib/models/common/widgets/Image/Form.tsx
+++ b/packages/slice-machine/lib/models/common/widgets/Image/Form.tsx
@@ -37,6 +37,9 @@ const FormFields = {
             name: "Thumbnails",
             message: "Must set name and width or height at minimum",
             test: function (value) {
+              if (!value.name) {
+                return false;
+              }
               const hasWidth = typeof value.width === "number" && value.width;
               const hasHeight =
                 typeof value.height === "number" && value.height;
--- a/packages/slice-machine/lib/models/common/widgets/Image/components.tsx
+++ b/packages/slice-machine/lib/models/common/widgets/Image/components.tsx
@@ -90,14 +90,11 @@ export const ConstraintForm: React.FC<{
   const [_field, meta] = useField(prefix);
 
   const requiredChar = required ? "*" : "";
-  const nameFieldName = `${prefix}.name`;
-  const widthFieldName = `${prefix}.width`;
-  const heightFieldName = `${prefix}.height`;
 
   return (
     <Fragment>
       <TexField
-        name={nameFieldName}
+        name={`${prefix}.name`}
         label={`Name${requiredChar}`}
         placeholder="main"
         sx={{ mb: 3 }}
@@ -105,13 +102,13 @@ export const ConstraintForm: React.FC<{
         meta={meta}
       />
       <NumberField
-        name={widthFieldName}
+        name={`${prefix}.width`}
         label={`Width (px)${requiredChar}`}
         placeholder=" "
         sx={{ mb: 3 }}
       />
       <NumberField
-        name={heightFieldName}
+        name={`${prefix}.height`}
         label={`Height (px)${requiredChar}`}
         placeholder=" "
       />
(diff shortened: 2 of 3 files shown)


packages/slice-machine/lib/models/common/widgets/Image/Form.tsx
```diff
@@ -37,6 +37,9 @@ const FormFields = {
             name: "Thumbnails",
             message: "Must set name and width or height at minimum",
             test: function (value) {
+              if (!value.name) {
```
Heyrows, line 40: Wouldn't that trigger the validation right at the beginning when you open the modal ?

packages/slice-machine/lib/models/common/widgets/Image/components.tsx
```diff
@@ -86,45 +86,32 @@ export const ConstraintForm: React.FC<{
...
-          ? typeof cast === "number" && isNaN(cast)
-            ? null
-            : cast
-          : cast;
-      helpers.setValue({ ...field.value, [key]: value });
-    };
+  const [_field, meta] = useField(prefix);
+
+  const requiredChar = required ? "*" : "";
+  const nameFieldName = `${prefix}.name`;
```
Heyrows: OPT: if they're used only once I wouldn't do variables for that.

packages/slice-machine/lib/models/common/widgets/Image/components.tsx
```diff
@@ -133,54 +120,53 @@ export const ConstraintForm: React.FC<{
...
+      fieldName={name}
       formField={{ label, placeholder }}
+      meta={meta}
       field={{
+        min: 0,
         type: "number",
         value: field.value || "",
-        onChange: onChangeSetField,
+        onKeyDown: (event) => {
+          if (event.key === "e" || event.key === "-" || event.key === "+") {
```
Heyrows, line 163: Why would those value be like that ?
feels highly weird to have to prevent keys like that 😓

Heyrows: At least, I would prevent everything that is not a number if we really have no other option than using that.

Heyrows (changes requested):
I tried to test it while reviewing but the inputs are not working at all, seems like I can't click on the number fields.

packages/slice-machine/lib/models/common/widgets/Image/components.tsx
```diff
@@ -133,54 +120,53 @@ export const ConstraintForm: React.FC<{
...
   return (
     <FormFieldInput
       {...props}
-      fieldName={fieldName}
-      meta={meta}
+      fieldName={name}
       formField={{ label, placeholder }}
+      meta={meta}
       field={{
+        min: 0,
```
Heyrows, line 159: I guess nobody needs the 0 value input, but I noticed that I can't type 0 in the input.

packages/slice-machine/lib/models/common/widgets/Image/components.tsx
MarcMcIntosh: `+1`, `-1`, `1e3` are all valid numbers for inputs :)

packages/slice-machine/lib/models/common/widgets/Image/components.tsx
MarcMcIntosh: Good point that was done during the refactoring :)

packages/slice-machine/lib/models/common/widgets/Image/Form.tsx
MarcMcIntosh: nope, actually that was in the old version :)

packages/slice-machine/lib/models/common/widgets/Image/components.tsx
MarcMcIntosh: done

packages/slice-machine/lib/models/common/widgets/Image/components.tsx
Heyrows: thought the E was funny tho 😂

Heyrows (approved):

PR #639, opened 2022-10-13
update screenshot card in slice builder and fix slice cards

danigo99:
## Context

Responding to issues here: https://linear.app/prismic/issue/SM-804/aauser-i-see-slices-card-redesigned-with-a-new-screenshot-button

- Fixing general card design
- Updating the screenshot preview in the slice builder.

- [ ] The CI is successful.

## [OPT] Preview

image
image

image


danigo99:
Waiting for https://github.com/prismicio/slice-machine/pull/631

packages/slice-machine/lib/models/ui/Slice/index.tsx
```diff
@@ -97,7 +97,11 @@ const SliceDescription = ({
     }}
   >
     <Flex>
-      <TextWithTooltip text={slice.model.name} as="h6" />
+      <TextWithTooltip
+        text={slice.model.name}
+        as="h6"
+        sx={{ fontWeight: "600 !important" }}
```
Heyrows, line 109: bypassing the theme 😂, always makes me laugh

Heyrows: i'm sure you had your reasons just funny to see ;)

Heyrows (approved):
Seems good, noted some design issues I found in the ticket.

packages/slice-machine/lib/models/ui/Slice/index.tsx
danigo99: It's because the theme uses the default HTML styling of the `as` tag (in this case h6) 😅

packages/slice-machine/components/UpdateScreenshotButton/index.tsx
```diff
@@ -8,19 +9,13 @@ const UpdateScreenshotButton = ({
...
-        style={{
-          marginRight: "8px",
-          position: "relative",
-          top: "3px",
-        }}
-      />
-    </Text>
-    <Text sx={{ lineHeight: "24px", fontSize: "12px" }}>Update screenshot</Text>
-  </Button>
+  <Button
```
Heyrows: love it, you know how to talk to me :D.
OPT: remove that file to use the component directly. (unsure if it's worth it).

packages/slice-machine/components/ScreenshotPreview/index.tsx
```diff
@@ -8,45 +8,42 @@ const MemoedImage = memo<{ src: string | undefined }>(({ src }) => (
 
 interface ScreenshotPreviewProps {
   src?: string;
-  sx?: ThemeUIStyleObject;
+  sx: { height: string } & ThemeUIStyleObject;
```
Heyrows: Typescript question,

If you do `sx: { height: string } & ThemeUIStyleObject;` does this mean the height property in the final object is the one coming from ThemeUISTyleObject ?

In comparison with `sx: ThemeUIStyleObject & { height: string }` where you would extend the ThemeUIStyleObject.

It's a real question, i'm not sure of the answer 😄

Heyrows (approved):
Seems good, i'm a bit unsure about the design but that will be validated by the QA / designers 😄

packages/slice-machine/components/ScreenshotPreview/index.tsx
danigo99: I don't think the order matters in type unions (A & B === B & A). In this case all it's doing is making height required since it has the same type in both objects.

The union just forces the type to match both types.

danigo99: Actually just realised the ThemeUI type is `string | number`, will update it now so the number type can also be used

Heyrows (approved):

packages/slice-machine/components/ScreenshotPreview/index.tsx
Heyrows: Ok so it match both types, always thought it was to change properties of one given type. 👍

Heyrows: thanks :)

PR #646, opened 2022-10-18
Fix screenshot button style on isLoading

danigo99:
## Context

Fixing a small design problem during the loading state of the take screenshot button.
[Linear ticket here](https://linear.app/prismic/issue/SM-869/aauser-on-the-simulator-page-the-screenshot-button-style-should-be).

## Dependencies

NA

- [x] The CI is successful.

## [OPT] Preview

ezgif com-gif-maker (3)

---

image


packages/slice-machine/components/Button/index.tsx
```diff
@@ -47,7 +47,7 @@ export const Button: React.FunctionComponent<ButtonProps> = ({
     }}
     type={type}
     form={form}
-    disabled={disabled}
+    disabled={disabled || isLoading}
```
danigo99, line 50: Imo this makes sense since the button design is the same in both states (except for the cursor which is already overwritten above).

This then allows us to use the `&:disabled` styling from ThemeUI for each variant

Heyrows: hey, makes sense to me as well, just the design was specifying a scenario where the button was loading and not disabled.

Heyrows: But i think here since loading and disabled are independent, we loose one possible case of a button being activated multiple times.
Atm it doesn't change anything tbh because we don't use that case but, just saying :)

Heyrows (approved):

PR #647, opened 2022-10-18
pinpoint @types/react to avoid installing broken version

Heyrows:
## Context

The CD broke on [the last PR merge](https://github.com/prismicio/slice-machine/actions/runs/3272930633/jobs/5384528047#step:6:577).
We did not detect this before because our CI uses `npm install` while the CD uses `npm CI`

After investigating a bit, I found on some forum mention that the last versions of `@types/react` were broken.
I verified the NPM repo of `@types/react` and the last versions that could be installed automatically were from 1-2 months ago.
I suspected that the package lock wasn't updated since then and so I pinpointed the exact version of `@types/react` and it worked locally.

- [x] The CI is successful.

[OPT] N.B for the reviewer:

## [OPT] Preview

dan-cook-MCauAnBJeig-unsplash

### DIFF
--- a/packages/slice-machine/package-lock.json
+++ b/packages/slice-machine/package-lock.json
@@ -6,7 +6,7 @@
 	"packages": {
 		"": {
 			"name": "slice-machine-ui",
-			"version": "0.5.2-dev-screenshotv2.0",
+			"version": "0.5.2-dev-screenshotv2.1",
 			"license": "MIT",
 			"dependencies": {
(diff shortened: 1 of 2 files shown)


danigo99 (approved):
Should we add a ticket to the tech debt backlog or something to revise this in a few months in case there's a fix? Otherwise we might be locked in to an old version by mistake

Heyrows:
A ticket already exist for it 😄, i guess now I see it's true value haha

PR #651, opened 2022-10-20
851/remove unsued deps

danigo99:
## Context

[Ticket link](https://linear.app/prismic/issue/SM-851/[tech-debt]-aauser-i-dont-have-any-unused-dependencies-in-the-slice)

## Dependencies

NA

- [x] The CI is successful.

## [OPT] Preview

image

Failed test here: https://github.com/prismicio/slice-machine/actions/runs/3288980562/jobs/5419969588
image

---

image


package.json
```diff
@@ -26,9 +26,11 @@
     "bump:alpha": "lerna version prerelease --preid $npm_config_preid --no-changelog --exact --yes --no-private",
     "publish:alpha": "lerna publish from-package --dist-tag alpha --no-verify-access --yes",
     "bump:production": "lerna version --conventional-graduate --exact --no-private",
-    "publish:production": "PUBLIC_SM_UI_SEGMENT_KEY=cGjidifKefYb6EPaGaqpt8rQXkv5TD6P PUBLIC_SM_INIT_SEGMENT_KEY=ED3O1FVsa8DteNyRmAuDyMcPoftJ6VdP lerna publish from-package"
+    "publish:production": "PUBLIC_SM_UI_SEGMENT_KEY=cGjidifKefYb6EPaGaqpt8rQXkv5TD6P PUBLIC_SM_INIT_SEGMENT_KEY=ED3O1FVsa8DteNyRmAuDyMcPoftJ6VdP lerna publish from-package",
+    "depcheck": "depcheck --skip-missing; lerna run depcheck"
```
Heyrows, line 30: why the use of skip missing if this is in the config already ?

packages/client/.depcheckrc
```diff
@@ -0,0 +1,4 @@
+ignores: [
```
Heyrows, line 1: could those extends the main one ?

packages/slice-machine/.depcheckrc
```diff
@@ -0,0 +1,12 @@
+ignores: [
```
Heyrows, line 1: that's interesting !
Do you think adding all those ignores is a good sign or a bad sign ?

Could it be that dependencies are misplaced and installed by other packages ?

packages/slice-machine/package.json
```diff
@@ -55,22 +54,17 @@
...
     "cors": "^2.8.5",
     "dedent": "^0.7.0",
-    "deepmerge": "^4.2.2",
     "ejs": "^3.1.7",
-    "esm": "^3.2.25",
     "express": "^4.17.1",
     "express-form-data": "^2.0.14",
     "express-http-proxy": "^1.6.3",
     "fast-deep-equal": "^3.1.3",
-    "form-data": "^3.0.0",
```
Heyrows, line 65: damn i love seeing all those removals T_T

package.json
danigo99: I didn't add a config to the root since there's nothing to ignore. So just left the flag

packages/client/.depcheckrc
danigo99: I think since none of them share the same ignores, there's no point in having a base one. We don't want to ignore things that don't exist in a sub-package in case someone accidentally adds it later

packages/slice-machine/.depcheckrc
danigo99: Some of them (newtype-ts, and monocle-ts) were added in this PR by Marc, so I didn't want to re-create that problem: https://github.com/prismicio/slice-machine/pull/620/files

@svgr/webpack is used by the next config. Not sure why depcheck ignores this.

rollup and nodemon are used in the package.json so they aren't recognised by depcheck

I've now removed the types one, because I didn't realise that depcheck actually solves them properly. 

The last 3 (esbuild, babel-plugin-module-resolver and typescript-is) I'm honestly just worried about deleting since they sound important 😂. Do you know anything about these?

danigo99: Got help from someone here. The babel-plugin is being used by the the babel config so needs to stay. typescript-is looks like it's probably not since it isn't referenced anywhere in the ts-config.

esbuild looks like it might also not be used since it's not referenced anywhere in the `package.json` build scripts. So I'll try to remove it also

package.json
Heyrows: ah ok 👍

packages/client/.depcheckrc
Heyrows: Ok, it make sense.
I was asking the question because the DevX team suggested that for our ts-config or linter-config.

packages/slice-machine/.depcheckrc
Heyrows: hey,

I don't have a lot of knowledge about dependencies because appart from the one we uses for tests, they don't change much.
I believe esbuild might have been used before we switched to rollup.

[Typescript-is](https://www.npmjs.com/package/typescript-is) seems to be for run-time validation of objects
We already do that using [io-ts](https://www.npmjs.com/package/io-ts) so I guess we don't need it if it's not used 😄

Heyrows: If you need to test the APP with less dependencies sees if everything works well, you can
- Test in dev mode
- Test in bundled mode
  - For that, run `npm run bundle` in the Slice Machine package
  - Use `npm link` to link slice machine to a side project and just run it to test if it still works 😃

danigo99: Seems good to me I believe

Heyrows (approved):
Some conflicts to take care but it seems to be ok.

PR #653, opened 2022-10-20
Fix button padding

danigo99:
Fixing back from validation defect of this ticket: https://linear.app/prismic/issue/SM-804/aauser-i-see-slices-card-redesigned-with-a-new-screenshot-button

image

### DIFF
--- a/packages/slice-machine/src/theme.ts
+++ b/packages/slice-machine/src/theme.ts
@@ -324,8 +324,7 @@ const AppTheme = (): Theme =>
         lineHeight: "16px",
         pl: 2,
         pr: 2,
-        pb: "6px",
-        pt: "6px",
+        py: "5px",
         cursor: "pointer",
         "&:hover": {
           background: "#F4F2F4",


packages/slice-machine/src/theme.ts
```diff
@@ -324,8 +324,7 @@ const AppTheme = (): Theme =>
         lineHeight: "16px",
         pl: 2,
         pr: 2,
-        pb: "6px",
```
MarcMcIntosh, line 327: Is this the only use of the secondarySmall button class?

danigo99: Yep. 2 references and they're both update screenshot buttons

MarcMcIntosh: Okay, Is that the only change required by the ticket?

danigo99: Yeah, the button height was the only comment on the ticket from the back from validation

MarcMcIntosh (approved):

PR #657, opened 2022-10-24
Screenshot changes modal dropzone (+ alert)

hypervillain:
- Added DropZone to both views of ScreenshotChanges Modal (with and without existing image).
- Added check for file type + file size in drop handler


packages/slice-machine/components/ScreenshotChangesModal/DropZone.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const DropZone: React.FC<DropZoneProps> = ({
+  children,
+  onHandleDrop,
+  isVisible = false,
+  imageTypes = acceptedImagesTypes,
+}) => {
+  const [isDragActive, setIsDragActive] = useState(false);
+
+  const handleDrop = (e: React.DragEvent<HTMLInputElement>) => {
```
danigo99: OPT: for readability I think it's worth giving `e` a more explicit name, like `event` or `dragEvent`, especially when it's used over a few different lines

danigo99 (approved):

packages/slice-machine/src/modules/toaster/index.tsx
```diff
@@ -9,6 +9,7 @@ export enum ToasterType {
   ERROR = "error",
   INFO = "info",
   LOADING = "loading",
+  SCREENSHOT_FAILED = "screenshot_failed",
```
danigo99: I don't think we need this. We should just use the `ERROR` type instead. The only reason `SCREENSHOT_CAPTURED` was a new type is because the toaster looked different

packages/slice-machine/components/ScreenshotChangesModal/DropZone.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  const handleDrop = (event: React.DragEvent<HTMLInputElement>) => {
+    event.preventDefault();
+    setIsDragActive(false);
+    const maybeFile = event.dataTransfer.files?.[0];
+    if (maybeFile) {
+      if (imageTypes.find((t) => `image/${t}` === maybeFile.type)) {
+        if (maybeFile.size > 128000000) {
+          return openToaster(
+            "File is too big. Max file size: 128Mo.",
```
danigo99:
```suggestion
            "File is too big. Max file size: 128Mb.",
```
Is it supposed to be Mb?

packages/slice-machine/src/modules/toaster/index.tsx
hypervillain: I had no clue haha, thanks

danigo99 (approved):

PR #660, opened 2022-10-26
 870/add custom screensize options

danigo99:
## Context

Adding a new tool bar to the slice simulator to allow the user to select custom screen sizes
[Ticket #870](https://linear.app/prismic/issue/SM-870/aau-i-can-select-custom-simulator-and-screenshot-screen-sizes-in-the)

## Impact / Dependencies
N/A

- [x] The CI is successful.

## [OPT] Preview

image
image

---

image


packages/slice-machine/src/theme.ts
```diff
@@ -302,6 +302,33 @@ const AppTheme = (): Theme =>
           outline: "none",
         },
       },
+      dropDownButton: {
```
Heyrows, line 306: ❓ Should this variant be named after the design system ?
That way if later we need it again, finding it will be easy.

I also tested it quickly, and was wondering if the blue border color was from the design of maybe a side effect from the default variant.
image

packages/slice-machine/components/Simulator/components/Toolbar/ScreensizeInput/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+export enum ScreenSizeOptions {
+  DESKTOP = "Desktop",
+  TABLET = "Tablet",
+  MOBILE = "Mobile",
+  CUSTOM = "Custom",
+}
+
+export const ScreenSizes: Record<string, ScreenDimensions> = {
+  [ScreenSizeOptions.DESKTOP]: {
+    width: 1200,
```
Heyrows: Where do those values come from ?
Breakpoints always seems so random on multiple websites that i'm wondering where yours come from 😄

packages/slice-machine/components/Simulator/components/Toolbar/ScreensizeInput/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+type ScreensizeInputProps = {
+  label: "H" | "W";
+  startValue: number;
+  onChange: (sizeEvent: React.ChangeEvent<HTMLInputElement>) => void;
+  isActive: boolean;
+  sx?: ThemeUIStyleObject;
+};
+
+export const ScreensizeInput: React.FC<ScreensizeInputProps> = ({
```
Heyrows, line 33: I've asked Benjamin about the behavior of the inputs, he told me that the acceptance criteria where wrong.

The input should never be disabled and the dropdown selected option should change according to the selected values in the Input (something known, or unknown = Custom).

Heyrows (changes requested):

packages/slice-machine/src/theme.ts
Heyrows: confirmed by Benjamin, it's not the right behavior.

danigo99: This is just called a secondary button in the design system. I can leave it as that and put the rest of the styling I need directly in the component if you prefer?

packages/slice-machine/components/Simulator/components/Toolbar/ScreensizeInput/index.tsx
danigo99: I've updated this now

Heyrows:
Just tested, on the slice simulator, the variation dropdown should be displayed and disabled if there is only one variation available.
Confirmed with designers (benjamin)

image

packages/slice-machine/components/Simulator/components/Header/index.tsx
```diff
@@ -102,56 +69,35 @@ const Header: React.FunctionComponent<PropTypes> = ({
...
-          Icon={AiFillCamera}
+      <Button
+        onClick={onTakingSliceScreenshot}
+        label="Take a screenshot"
+        isLoading={isSavingScreenshot}
+        Icon={AiFillCamera}
+      />
+      {isWaitingForIframeCheck && (
+        <IframeRenderer
+          dryRun
```
Heyrows: ❓ I was wondering, do we need this ?
We already have the iframeRenderer on the Simulator/Index.tsx

packages/slice-machine/components/Simulator/components/Header/index.tsx
```diff
@@ -16,59 +10,32 @@ import { selectIsWaitingForIFrameCheck } from "@src/modules/simulator";
...
+import { ScreenDimensions } from "@lib/models/common/Screenshots";
 
 type PropTypes = {
   Model: ComponentUI;
   variation: Models.VariationSM;
-  handleScreenSizeChange: (screen: { size: Size }) => void;
-  size: Size;
-};
-
-const redirect = (
```
Heyrows, line 30: ⚠️  Unsure if we want to loose the link here on the slice name.
Feels like it could be useful.

packages/slice-machine/components/Simulator/components/Toolbar/ScreensizeInput/index.tsx
Heyrows: thanks 👍  works well.

packages/slice-machine/components/Simulator/components/Toolbar/ScreensizeInput/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+      sx={{
+        alignItems: "center",
+        ...sx,
+      }}
+    >
+      <Label
+        sx={{
+          backgroundColor: "grey07",
+          border: "transparent",
+          borderRight: "1px solid #DCDBDD",
```
Heyrows: We have to avoid hardcoded colors as much as possible to use the maximal performances of the theme.

```suggestion
          borderRight: (t) => `1px solide ${String(t.colors?.myvariable)}`,
```

packages/slice-machine/components/Simulator/components/Header/index.tsx
danigo99: You're right I think. It's probably redundant here

packages/slice-machine/components/Simulator/components/Header/index.tsx
danigo99: Not sure what you mean? This redirect was only used for changing the variation and was moved to the new Toolbar component

danigo99:
Look of the variations dropdown:
image
image

packages/slice-machine/lib/builders/SliceBuilder/Header/VariationsPopover/index.tsx
```diff
@@ -6,7 +6,7 @@ import MenuList from "./MenuList";
 
 import { Button, Box, ThemeUICSSObject, ThemeUIStyleObject } from "theme-ui";
 
-import { RiArrowDropDownLine } from "react-icons/ri";
+import { RiArrowDropDownFill, RiArrowDropUpFill } from "react-icons/ri";
 
 const VarationsPopover: React.FunctionComponent<{
```
Heyrows, line 11: Seems the same as the dropdownMenu you made.

Heyrows (approved):

packages/slice-machine/lib/builders/SliceBuilder/Header/VariationsPopover/index.tsx
danigo99: Yeah but this one has some more features specific to the variations changer like creating more variations in the slice builder. It's a bit of a shame :/

PR #663, opened 2022-10-31
Fix negative and leading zero inputs in the screen dimensions

danigo99:
## Types of changes

## Checklist:

image

### DIFF
--- a/packages/slice-machine/components/Simulator/components/Toolbar/ScreensizeInput/index.tsx
+++ b/packages/slice-machine/components/Simulator/components/Toolbar/ScreensizeInput/index.tsx
@@ -25,7 +25,7 @@ export const ScreenSizes: Record<string, ScreenDimensions> = {
 
 type ScreensizeInputProps = {
   label: "H" | "W";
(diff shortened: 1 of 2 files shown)


packages/slice-machine/components/Simulator/components/Toolbar/ScreensizeInput/index.tsx
```diff
@@ -64,7 +64,7 @@ export const ScreensizeInput: React.FC<ScreensizeInputProps> = ({
         {label}
       </Label>
       <Input
-        type="number"
+        type="tel" // Prevents leading 0 from being displayed after the input is cleared
```
Heyrows: ⚠️ I've looked at bit to understand why the change and it seems to fix the issue.

However it also removes the behavior of changing the value with up and down arrows which I think for people working with rendering their components and checking breakpoints has value.

I'd rather modify the "onChange" function to remove any starting Zeros I think to keep both world.

Heyrows (approved):

PR #664, opened 2022-10-31
841/user satisfaction

Heyrows:
## Context

[Linear ticket 841](https://linear.app/prismic/issue/SM-841/aapm-i-can-track-user-satisfaction) 

Several design bugs were found while doing this ticket, here's the list that have been fixed
- screenshot modal remove the padding left of the title
- remove tags from the list of Slices in the Slice Zone (slice builder)
- space missing between "missing screenshots" and "update all screenshot"
- modal screenshots buttons to upload custom screenshot should be bigger
- height of screenshot modal should be 456px

- [x] The CI is successful.

### DIFF
--- a/packages/slice-machine/components/ScreenshotChangesModal/useCustomScreenshot.tsx
+++ b/packages/slice-machine/components/ScreenshotChangesModal/useCustomScreenshot.tsx
@@ -30,6 +30,7 @@ const FileInputRenderer: React.FC<HandleFileProp> = ({
         justifyContent: "center",
         width: 200,
         alignItems: "center",
+        flex: 1,
       }}
     >
       {children || "Select file"}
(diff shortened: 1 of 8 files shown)


packages/slice-machine/components/ScreenshotChangesModal/useCustomScreenshot.tsx
```diff
@@ -30,6 +30,7 @@ const FileInputRenderer: React.FC<HandleFileProp> = ({
         justifyContent: "center",
         width: 200,
         alignItems: "center",
+        flex: 1,
```
danigo99, line 33: I haven't tested this locally, but doesn't this make the button take up the remaining flex space and therefore be 50/50 sharing space with the open simulator button? Because the designs aren't like that I think
image

danigo99 (approved):

packages/slice-machine/components/ScreenshotChangesModal/useCustomScreenshot.tsx
Heyrows: It does takes the remaining space but they're not 50/50.
I verified the design and that part also so the capture screenshot button wouldn't change size with or without the other button :)

PR #666, opened 2022-11-01
Parallelise CI jobs

danigo99:
## Context

Tech Debt ticket: https://linear.app/prismic/issue/SM-879/[tech-debt]-spike-1-day-aadev-the-ci-can-run-parallel-jobs

## The Solution

Separating each step of the build job into different jobs that run in parallel after the installation job.

This can still be improved by looking into parallelisation of cypress e2e test since these are taking up ~80% of the total runtime 😢 

- [x] The CI is successful.

---

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -79,30 +79,6 @@ jobs:
       - name: Audit
         run: lerna run audit
 
-  build-init:
-    needs: install-deps
-    runs-on: [ubuntu-latest]
-    steps:
-      - uses: actions/checkout@v2
-      - uses: ./.github/actions/restore-cache
-      - uses: ./.github/actions/setup-node
-      - uses: ./.github/actions/build-essential
-      - name: Build slice machine init
-        working-directory: ./packages/init
-        run: npm run build
-
-  build-ui:
-    needs: install-deps
-    runs-on: [ubuntu-latest]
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -40,14 +34,59 @@ jobs:
...
+    steps: 
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/restore-cache
+      - uses: ./.github/actions/setup-node
+      - uses: ./.github/actions/build-essential
       - name: Audit
         run: lerna run audit
+
+  e2e-tests:
+    needs: install-deps
```
Heyrows, line 83: ❓ Do parallelised steps run on the same machine ?
I'm wondering if the dependencies would be installed for this job or any other jobs.

danigo99: No they don't, each job is on a separate runner. 
So first the install step happens, the node modules are cached, then all the other jobs start at the same time. In each of them the cache is restored, and the tests happen.

Heyrows: I see, so they don't actually have to install anything themselves, good 👍

Heyrows (approved):

Heyrows:
Sounds good, can't wait to see how to make it even faster.

PR #667, opened 2022-11-01
Add CT kebab menu for delete & rename

danigo99:
## Context

Starting UI to delete custom types.

Ticket here: https://linear.app/prismic/issue/SM-856/[deleted-ct]-aauser-on-the-custom-type-list-i-see-a-new-dropdown-to

## The Solution

Added a new kebab menu dropdown component and used this in the custom types table in the the custom-types page

## Impact / Dependencies

N/A

## Checklist before requesting a review
- [ ] I hereby declare my code ready for review.
- [ ] If it is a critical feature, I have added tests.
- [ ] The CI is successful.
- [ ] If there could backward compatibility issues, it has been discussed and planned.

## [OPT] Preview

image
image

---

image

PR #669, opened 2022-11-02
Move custom type renaming to custom types table

danigo99:
## Context

Ticket: https://linear.app/prismic/issue/SM-858/[delete-ct]-aauser-i-can-rename-a-custom-type-from-the-ct-list

## The Solution

Removed the edit button in the custom type editor, added onClick functionality to the rename option of the kebab menu.

## Impact / Dependencies

Waiting on #667 

- [x] The CI is successful.

## [OPT] Preview

ezgif com-gif-maker (4)

---

image

### DIFF
--- a/packages/slice-machine/components/CustomTypeTable/ctPage.tsx
+++ b/packages/slice-machine/components/CustomTypeTable/ctPage.tsx
@@ -12,8 +12,7 @@ export const CustomTypeTable: React.FC<{
   customTypes: FrontEndCustomType[];
 }> = ({ customTypes }) => {
   const { modelsStatuses, authStatus, isOnline } = useModelStatus(customTypes);
-  const [renameCustomTypeName, setRenameCustomTypeName] = useState<string>("");
-  const [renameCustomTypeId, setRenameCustomTypeId] = useState<string>("");
+  const [customTypeIdToRename, setCustomTypeIdToRename] = useState<string>("");
 
   const firstColumnWidth = "27%";
   const secondColumnWidth = "27%";
@@ -91,8 +90,7 @@ export const CustomTypeTable: React.FC<{
                         displayName: "Rename",
                         onClick: (event) => {
                           event.stopPropagation();
-                          setRenameCustomTypeName(customType.local.label || "");
-                          setRenameCustomTypeId(customType.local.id);
+                          setCustomTypeIdToRename(customType.local.id);
                           openRenameCustomTypeModal();
                         },
                         dataCy: "ct-rename-menu-option",
@@ -112,8 +110,8 @@ export const CustomTypeTable: React.FC<{
         </tbody>
       </Box>
       <RenameCustomTypeModal
-        customTypeName={renameCustomTypeName}
-        customTypeId={renameCustomTypeId}
+        customTypes={customTypes}
+        customTypeId={customTypeIdToRename}
       />
     </>
   );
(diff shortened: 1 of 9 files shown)


packages/slice-machine/components/CustomTypeTable/ctPage.tsx
```diff
@@ -1,98 +1,120 @@
...
 import { KebabMenuDropdown } from "@components/KebabMenuDropdown";
+import useSliceMachineActions from "@src/modules/useSliceMachineActions";
+import { RenameCustomTypeModal } from "@components/Forms/RenameCustomTypeModal";
 
 export const CustomTypeTable: React.FC<{
   customTypes: FrontEndCustomType[];
 }> = ({ customTypes }) => {
   const { modelsStatuses, authStatus, isOnline } = useModelStatus(customTypes);
+  const [renameCustomTypeName, setRenameCustomTypeName] = useState<string>("");
+  const [renameCustomTypeId, setRenameCustomTypeId] = useState<string>("");
```
danigo99: I'm not super happy with this way of doing things. But other than adding more complexity to the global state (in order to pass these through) I'm not sure how else to do it 😕

Heyrows: Maybe the Rename modal should have it's own state.

What I mean is, clicking the rename button would just open the modal with the right Slice / Custom type.
The modal contains it's inputs & handle it's state by itself.
On clicking the save button it triggers a saga to rename and at the end of the saga we close the modal.

What do you think ?

packages/slice-machine/src/modules/selectedCustomType/reducer.ts
```diff
@@ -317,22 +316,6 @@ export const selectedCustomTypeReducer: Reducer<
         )((group: GroupSM) => Group.reorderWidget(group, start, end))
       );
     }
-    case getType(renameCustomTypeCreator.success): {
```
Heyrows, line 320: nice, as a change that doesn't need to be saved it makes way more sense to have it here than in the builder.

packages/slice-machine/components/Forms/RenameCustomTypeModal/RenameCustomTypeModal.tsx
```diff
@@ -39,6 +41,13 @@ export const RenameCustomTypeModal: React.FC<RenameCustomTypeModalProps> = ({
     isRenamingCustomType: isLoading(store, LoadingKeysEnum.RENAME_CUSTOM_TYPE),
   }));
 
+  const customTypeName = useMemo(() => {
```
Heyrows: we could even have passed just a custom type found previously.

Heyrows (approved):

packages/slice-machine/src/modules/availableCustomTypes/index.ts
```diff
@@ -56,7 +56,8 @@ type CustomTypesActions =
   | ActionType<typeof createCustomTypeCreator>
   | ActionType<typeof renameCustomTypeCreator>
   | ActionType<typeof saveCustomTypeCreator>
-  | ActionType<typeof pushCustomTypeCreator>;
+  | ActionType<typeof pushCustomTypeCreator>
+  | ActionType<typeof renameCustomTypeCreator>;
```
mathieud: This seems duplicated

mathieud (approved):

PR #674, opened 2022-11-04
[Editor mocks] Validation on mock generation

arnaudlewis:
## Context

So far, we use to have a different format for mocks which corresponded to the format of the API.
We decided to move to the Editor format but it means that current users of SM will have to deal with incompatible mocks.
To solve that, we simple dynamically validate the mocks and if we detect that they are not there or simply invalid (common case will be the API format) then we regenerate fully the mocks.


packages/core/src/libraries/component.ts
```diff
@@ -177,17 +166,24 @@ export function getComponentInfo(
...
   /* This illustrates the requirement for apps to pass paths to mocks */
-  const maybeMock = resolvePathsToMock({
-    paths: assetsPaths,
-    from,
-    sliceName,
-  });
+  const mocks = (() => {
+    const resolvedMocks = Files.readEntityFromFile<ComponentMocks>(
+      path.join(basePath, from, sliceName, "mocks.json"),
+      (payload: any) => {
```
MarcMcIntosh: linter warning here, maybe unknown would work better

packages/core/src/libraries/mocks.ts
```diff
@@ -1,39 +0,0 @@
...
-  sliceName,
-}: {
-  path: string;
-  from: string;
-  sliceName: string;
-}): string {
-  return path.join(filePath, from, sliceName, "mocks.json");
-}
-
-export function resolvePathsToMock({
```
MarcMcIntosh, line 17: no-longer needed?

packages/core/src/models/LibrariesState.ts
```diff
@@ -1,9 +1,10 @@
+import { SharedSliceContent } from "@prismicio/types-internal/lib/content/fields/slices/SharedSliceContent";
 import { SharedSlice } from "@prismicio/types-internal/lib/customtypes/widgets/slices";
 import * as t from "io-ts";
 // import { VariationMock } from "./Slice";
```
MarcMcIntosh: nice, now the comments can be removed, (they where there to remind me about removing valiatio mock)

packages/slice-machine/lib/mock/Slice.ts
```diff
@@ -61,19 +60,11 @@ export function buildSliceMockConfig(
...
+): ComponentMocks {
   const model = Slices.fromSM(smModel);
 
   const sliceMockConfig = buildSliceMockConfig(model, legacyMockConfig);
-  const sliceModel: SharedSlice = {
-    ...model,
-    variations: model.variations.map((v) => {
-      return {
-        ...v,
-        imageUrl: "",
```
MarcMcIntosh, line 73: why did we strip out the imageUrl before?

packages/slice-machine/lib/storybook.ts
```diff
@@ -57,24 +59,22 @@ export default {
...
-    const mocks = Files.readFirstOf<
-      SharedSliceContent[],
-      Record<string, unknown>
-    >([customMocksPath, generatedMocksPath])((value: string) => {
-      return JSON.parse(value) as unknown as SharedSliceContent[];
-    });
-    if (!mocks) {
+    const mocks = Files.readEntity<ComponentMocks>(
+      mocksPath,
+      (payload: any) => {
```
MarcMcIntosh: would unknown work?

packages/slice-machine/server/src/api/common/generate.ts
```diff
@@ -20,14 +22,21 @@ export function generate(
     );
 
     components.forEach((c) => {
-      if (
-        !Files.exists(
```
MarcMcIntosh, line 24: trying to think of what this was there for :/

packages/slice-machine/server/src/api/common/generate.ts
```diff
@@ -20,14 +22,21 @@ export function generate(
...
+      const mocks = Files.readEntityFromFile<ComponentMocks>(
+        mocksPath,
+        (payload: unknown) => {
+          return getOrElseW(() => new Error("Invalid component mocks."))(
+            ComponentMocks.decode(payload)
+          );
+        }
+      );
+
+      if (!mocks || mocks instanceof Error) {
```
MarcMcIntosh: ^ happens here now

packages/slice-machine/server/src/api/common/generate.ts
```diff
@@ -20,14 +22,21 @@ export function generate(
...
+        const mocks: ComponentMocks = mock(
```
MarcMcIntosh, line 37: is there no error for mocks being declared twice?

packages/slice-machine/tests/lib/mock/Slice.test.ts
```diff
@@ -1,7 +1,7 @@
 import "@testing-library/jest-dom";
 import { WidgetTypes } from "@prismicio/types-internal/lib/customtypes/widgets";
 import { SlicesTypes } from "@prismicio/types-internal/lib/customtypes/widgets/slices";
-import { SliceSM, Slices } from "@slicemachine/core/build/models";
+import { SliceSM } from "@slicemachine/core/build/models";
 import { isRight } from "fp-ts/lib/Either";
 import MockSlice from "../../../lib/mock/Slice";
 // import allFieldSliceModel from "../../../tests/__mocks__/sliceModel";
```
MarcMcIntosh, line 7: I'll need to add this test case back in

packages/slice-machine/tests/src/modules/screenshots/utils.ts
```diff
@@ -28,7 +27,7 @@ export const getSelectedSliceDummyData = () => {
     extension: "js",
     model: dummyModel,
     screenshots: {},
-    mock: mocks as unknown as SharedSliceContent[],
+    mock: mocks as unknown as ComponentMocks,
```
MarcMcIntosh, line 30: this is where property based testing would come in handy :)

MarcMcIntosh (commented):
Left a few comments :)

packages/core/src/libraries/mocks.ts
arnaudlewis: no, I simplified the code, you'll see when reviewing, it's no longer necessary

packages/slice-machine/lib/mock/Slice.ts
arnaudlewis: We use to require it, now we have a fallback system so if not specified, the validator will automatically set a default url that is used in the wroom already

packages/slice-machine/server/src/api/common/generate.ts
arnaudlewis: It used to check if we had existing mocks. I changed the logic to check 2 things:
- If we have existing mocks
- if the mocks are valid (using SharedSliceContent validator)

If one of these case is false, we regenerate

packages/slice-machine/server/src/api/common/generate.ts
arnaudlewis: yes that's the replacement for the condition above.

packages/slice-machine/server/src/api/common/generate.ts
arnaudlewis: oh yeah weird. I'll do some renaming around here.

packages/slice-machine/tests/lib/mock/Slice.test.ts
arnaudlewis: which case?

packages/slice-machine/tests/src/modules/screenshots/utils.ts
arnaudlewis: but actually we could runtime validate the mocks here to type them properly and at least you're sure that's the right format for your tests maybe.
But yeah I guess your suggestion makes sense. 
That's something to consider for the future

packages/slice-machine/tests/lib/mock/Slice.test.ts
MarcMcIntosh: way down at the bottom, for allFieldSliceModel

cypress/integration/slices/00-create.spec.js
```diff
@@ -57,29 +57,6 @@ describe("Create Slices", () => {
     );
     cy.readFile(type).should("contains", sliceName);
 
-    cy.readFile(pathToMock, "utf-8")
```
MarcMcIntosh, line 60: what would be a good check to make sure the mocks was generated?

packages/core/src/node-utils/paths.ts
```diff
@@ -91,9 +91,6 @@ export const CustomTypesPaths = (cwd: string): Paths =>
 export const CustomPaths = (cwd: string): Paths => Paths(cwd, "");
 export const PackagePaths = (cwd: string): Paths => Paths(cwd, "node_modules");
 export const SMConfigPath = (cwd: string): string => path.join(cwd, "sm.json");
-export const LibrariesStatePath = (cwd: string): string =>
```
MarcMcIntosh, line 94: I guess we should let user know that they can delete libraries-state

MarcMcIntosh (commented):
Left a few questions :)

packages/slice-machine/tests/server/services/getEnv.test.js
```diff
@@ -296,9 +297,13 @@ describe("getEnv", () => {
...
 
-    const { env } = await getEnv(TMP);
-    expect(env.applicationMode).toEqual(ApplicationMode.DEV);
-    expect(exitSpy).not.toHaveBeenCalled();
+    try {
+      const { env } = await getEnv(TMP);
+      expect(env.applicationMode).toEqual(ApplicationMode.DEV);
+      expect(exitSpy).not.toHaveBeenCalled();
+    } catch (e) {
+      console.error(e);
```
MarcMcIntosh: why try/catch here?

packages/core/src/node-utils/paths.ts
arnaudlewis: I guess so but also there will be remaining mocks in .slicemachine, the goal would be more to tell them that they can no longer versioned `.slicemachine` all together on GitHub and clean that up. Only the mock config is a problem there but we’ll probably no longer support it.
But for sure we could have a changelog to tell them what they could clean. If no mock config, just remove `.slicemachine` all together would be the best recommandation

packages/core/src/libraries/mocks.ts
MarcMcIntosh: good

packages/core/src/node-utils/paths.ts
MarcMcIntosh: cool, that'll need to go in the change-log when ever it's written :)

MarcMcIntosh (approved):

packages/slice-machine/tests/server/services/getEnv.test.js
arnaudlewis: debug stuff left here, I removed it

PR #675, opened 2022-11-07
Ex246/custom type mocks

MarcMcIntosh:
## Types of changes

Change the format of the mocks from the api format to editor format.
Automatically replaces the old format mocks in the start script

How to verify:
Checkout the old mocks from another branch build the project and run `npm run slicemachine` from with in the `e2e-projects/next`

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


packages/slice-machine/lib/mock/CustomType.ts
```diff
@@ -39,16 +41,23 @@ function buildDocumentMockConfig(
...
+  sharedSlices: Record<string, SharedSlice>
+): CustomTypeContent | null {
   const prismicModel = CustomTypes.fromSM(model);
   const documentMockConfig = buildDocumentMockConfig(
     prismicModel,
     legacyMockConfig
   );
-  return generateDocumentMock(
+
+  const mock = generateDocumentMock(
```
arnaudlewis: can you use `DocumentMock.generate` so we don't use this helper that is no longer necessary since we don't require any renderer.
Also you can avoid the validator below, the content returned by `DocumentMock.generate` should be properly typed.
Also can you use `1.5.2` so we get a `Document` type instead of the previously named `CustomTypeContent` which is an odd name.

packages/slice-machine/lib/storybook.ts
```diff
@@ -1,7 +1,7 @@
 import path from "path";
 import TemplateEngine from "ejs";
 import { getOrElseW } from "fp-ts/Either";
-import Files from "./utils/files";
+import Files from "@slicemachine/core/build/node-utils/files";
```
arnaudlewis, line 4: nice one! I think everytime we can, we should use the files helper from the core and no longer the one from utils, it's unnecessary duplication :)

packages/slice-machine/server/src/api/custom-types/save.ts
```diff
@@ -30,13 +33,33 @@ export default async function handler(req: { body: SaveCustomTypeBody }) {
...
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-argument
-    CustomTypeMockConfig.getCustomTypeMockConfig(updatedMockConfig, model.id)
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+    CustomTypeMockConfig.getCustomTypeMockConfig(updatedMockConfig, model.id),
+    sharedSlices
   );
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
-  Files.write(mockPath, mocked as object);
+
+  if (mocked) Files.writeJson(mockPath, mocked);
```
arnaudlewis: this condition should always be defined if using the generate function from Document so no need to check anymore.

arnaudlewis (changes requested):
I opened a thread regarding the check of the mocks happening at the start and no longer while querying the backend.

packages/slice-machine/lib/storybook.ts
MarcMcIntosh: There's a duplication ticket on the tech debt board, Files isn't event the most common one 😅

packages/slice-machine/server/src/api/custom-types/save.ts
MarcMcIntosh: done

packages/slice-machine/lib/mock/CustomType.ts
MarcMcIntosh: done

packages/slice-machine/scripts/start/udpateMocks.ts
```diff
@@ -0,0 +1,126 @@
+import mockForSlice from "../../lib/mock/Slice";
```
arnaudlewis: Considering the thread with Come, can you move this whole code back to the generate file when we load the backend state ?
Also the naming I don’t know why you talk about legacy but you should rename it.
It’s gonna runtime validate the mocks all the time and handle it when it’s inaccurate not only legacy.

MarcMcIntosh: Did come write back?

legacy / old format

MarcMcIntosh: done

arnaudlewis (approved):

PR #677, opened 2022-11-08
Fix slicezone shared type deletion logic

danigo99:
## Context

This is part of a fix for this Spike: https://linear.app/prismic/issue/SM-883/spike-1day-aauser-i-can-use-slice-machine-with-legacy-slices-in-my

## The Solution

The logic for removing a missing slice from the slice zone was reversed, and thus removing all other slices that could be displayed. This logic has now been updated and tests were added for the file. 

Also updated the dependency array for the useEffect in the slice Zone component, since this was not correctly checking that the object was equal between renders and was re-rendering infinitely.

- [ ] The CI is successful.

---

image

### DIFF
--- a/packages/slice-machine/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
+++ b/packages/slice-machine/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
@@ -110,7 +110,7 @@ const SliceZone: React.FC<SliceZoneProps> = ({
         onRemoveSharedSlice(key);
       });
     }
-  }, [notFound]);
+  }, [JSON.stringify(notFound)]);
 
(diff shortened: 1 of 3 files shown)


packages/slice-machine/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
```diff
@@ -110,7 +110,7 @@ const SliceZone: React.FC<SliceZoneProps> = ({
         onRemoveSharedSlice(key);
       });
     }
-  }, [notFound]);
+  }, [JSON.stringify(notFound)]);
```
mathieud: You should be able to use `useMemo` line 103 instead of this to prevent a rerender

mathieud (approved):

PR #679, opened 2022-11-08
Open delete CT modal from the CT list

danigo99:
## Context

Design ticket for adding delete CT modal to the CT list: https://linear.app/prismic/issue/SM-860/[delete-ct]-aauser-i-can-see-a-modal-to-delete-a-custom-type-from-the

## The Solution

Create a new modal
Add new modal type to action

- [ ] The CI is successful.

## [OPT] Preview

image

---

image

### DIFF
--- a/packages/slice-machine/components/ScreenshotPreviewModal/index.tsx
+++ b/packages/slice-machine/components/ScreenshotPreviewModal/index.tsx
@@ -35,6 +35,7 @@ const ScreenshotPreviewModal: React.FunctionComponent<ScreenshotModalProps> = ({
       shouldCloseOnOverlayClick={true}
       contentLabel={"Screenshot Preview"}
       portalClassName={"ScreenshotPreviewModal"}
+      onRequestClose={closeScreenshotPreviewModal}
     >
       <Card
         radius={"0px"}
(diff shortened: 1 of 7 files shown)


packages/slice-machine/components/ScreenshotPreviewModal/index.tsx
```diff
@@ -35,6 +35,7 @@ const ScreenshotPreviewModal: React.FunctionComponent<ScreenshotModalProps> = ({
       shouldCloseOnOverlayClick={true}
       contentLabel={"Screenshot Preview"}
       portalClassName={"ScreenshotPreviewModal"}
+      onRequestClose={closeScreenshotPreviewModal}
```
Heyrows, line 38: nice catch 👍

packages/slice-machine/src/theme.ts
```diff
@@ -317,6 +319,30 @@ const AppTheme = (): Theme =>
           outline: "none",
         },
       },
+      red: {
```
Heyrows: i'm going to be picky on this one, I like the fact that you're implementing a new button variant 👍 .
To optimise future use, could we have the same name as the one used in the design system ?

Heyrows (approved):

PR #681, opened 2022-11-09
Simulator editor preview

hypervillain:
Integrated React 17 version of the editor + new simulator layout.

Some stuff should be fixed before public release:
- ✅ Current use of lodash's throttle in simulator does not work. This is probably related to us using `useMemo` here.
- Property `apiContent` passed to `IframeRenderer` is of type `unknown`. We should be able to type it.
- Data passed to Slicezone via `client.setSlicezone` is casted to `any`. This should change!
- ✅ IframeRenderer should now be stateless and NOT rely on `setSliceZoneFromSliceIDs` anymore. To do this, the property `sliceView` should be deleted and `apiContent` should be passed (when taking screenshot and when running the Slicezone in `dryRun` mode


cypress/integration/slices/00-create.spec.js
```diff
@@ -57,21 +57,6 @@ describe("Create Slices", () => {
     );
     cy.readFile(type).should("contains", sliceName);
 
-    cy.readFile(pathToMock, "utf-8")
```
MarcMcIntosh: mock still get's made right?

packages/slice-machine/src/hooks/useEditorContent.ts
```diff
@@ -0,0 +1,26 @@
...
+function useEditorContentOnce({
+  variationID,
+  slice,
+}: {
+  variationID: string;
+  slice: ComponentUI;
+}) {
+  const editorContent =
+    slice.mock?.[0] ||
+    (defaultSharedSliceContent(variationID) as SharedSliceContent);
```
arnaudlewis, line 16: `defaultSharedSliceContent` has nothing to do in `editor-fields` actually, it only returns a `SharedSliceContent which you can access here so you should simply create the function here or return directly the right json, but it's a useless dependency to create here on the content level.
The goal of the editor is to be stateless, get some json as input and output an updated json. So it should never provide any form of default data or data whatsoever.

packages/slice-machine/components/Simulator/components/Header/index.tsx
```diff
@@ -78,15 +57,28 @@ const Header: React.FunctionComponent<PropTypes> = ({
...
+          <Flex sx={{ alignItems: "center" }}>
+            <Label htmlFor="show-mock-editor">Show the mock editor</Label>
+            <Switch
+              id="show-mock-editor"
+              checked={isDisplayEditor}
+              onChange={toggleIsDisplayEditor}
+            />
+          </Flex>
+        </Flex>
+        <Button onClick={() => console.log("todo")} label="Save mock" />
```
arnaudlewis: Can we ensure that this is not visible on the UI until we actually implement the save action

packages/slice-machine/components/Simulator/index.tsx
```diff
@@ -58,22 +67,50 @@ export default function Simulator() {
...
+    () =>
+      component.mock?.[0] ||
+      (defaultSharedSliceContent(variation.id) as SharedSliceContent),
+    [component.mock, variation.id]
+  );
+
+  const [editorContent, setContent] = useState(initialContent);
+  const initialApiContent = useMemo(
+    () =>
+      renderSliceMock(sharedSlice, editorContent) as {
```
arnaudlewis, line 84: Can you add a quick comment to explain why we cast here, it's gonna be useful for the next guy

packages/slice-machine/components/Simulator/index.tsx
```diff
@@ -58,22 +67,50 @@ export default function Simulator() {
...
+        [k: string]: unknown;
       },
-    ],
-    [component.model.id, variation.id]
+    []
+  );
+  const apiContent = useThrottle(
+    {
+      ...(renderSliceMock(sharedSlice, editorContent) as object),
+      id: initialApiContent.id,
```
arnaudlewis, line 95: And also here maybe, since it's related with the cast above.

packages/slice-machine/components/Simulator/index.tsx
```diff
@@ -58,22 +67,50 @@ export default function Simulator() {
...
+    () =>
+      renderSliceMock(sharedSlice, editorContent) as {
+        id: string;
+        [k: string]: unknown;
       },
-    ],
-    [component.model.id, variation.id]
+    []
+  );
+  const apiContent = useThrottle(
```
arnaudlewis: the throttle seems really weird, you seem to set the value at every throttle interval, meaning rendering the editor content, but that'll happen all the time here, even if editor content doesn't change. That would trigger a lot of conversion for no reason if there are no user actions. It should happen only if editor changes but with a throttle on it to prevent too many transformations in a given window of time.

packages/slice-machine/components/Simulator/index.tsx
```diff
@@ -58,22 +67,50 @@ export default function Simulator() {
...
+    {
+      ...(renderSliceMock(sharedSlice, editorContent) as object),
+      id: initialApiContent.id,
+    },
+    800
   );
 
+  const [prevVariationId, setPrevVariationId] = useState(variation.id);
```
arnaudlewis: This and the if below seem completely useless here. Am I missing something?

arnaudlewis (changes requested):

arnaudlewis (approved):

PR #683, opened 2022-11-10
create API endpoint and remove files associated to a CT on delete

danigo99:
## Context

Adding functionality to delete custom type files in slice machine local.
Ticket here: https://linear.app/prismic/issue/SM-863/[delete-ct]-aauser-i-can-delete-a-custom-type-from-the-ct-list-only

## The Solution

Added API endpoint to server, and service which does the business logic

## Impact / Dependencies

This still requires some frontend work, since the button currently has no visible feedback (need to look at the network tab).
This will come form this ticket: https://linear.app/prismic/issue/SM-865/[delete-ct]-aauser-i-can-see-the-deleted-custom-types-disappear-from

- [x] The CI is successful.

---

image

### DIFF
--- a/packages/slice-machine/lib/mock/misc/fs.ts
+++ b/packages/slice-machine/lib/mock/misc/fs.ts
@@ -1,18 +1,16 @@
 import Files from "../../utils/files";
 import { MocksConfig } from "../../models/paths";
+import { GlobalMockConfig } from "@lib/models/common/MockConfig";
 
 export const getConfig = (cwd: string) => {
   const pathToMockConfig = MocksConfig(cwd);
   if (Files.exists(pathToMockConfig)) {
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
-    return Files.readJson(pathToMockConfig);
+    return Files.readJson<GlobalMockConfig>(pathToMockConfig);
   }
   return {};
 };
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export const writeConfig = (cwd: string, config: any) => {
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+export const writeConfig = (cwd: string, config: GlobalMockConfig) => {
   Files.write(MocksConfig(cwd), config);
 };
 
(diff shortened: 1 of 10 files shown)


packages/slice-machine/server/src/api/custom-types/delete.ts
```diff
@@ -0,0 +1,79 @@
...
+  } catch (err) {
+    console.error(`[custom-type/delete] Could not update the project types.`);
+    return {
+      err: err,
+      reason:
+        "Something went wrong when deleting your Custom Type. Check your terminal.",
+      status: "500",
+      type: "warning",
+    };
+  }
```
danigo99: Should these steps all fail when one fails? Or should they all be tried and we just log the ones that didn't work? Was a bit unsure of this from the ticket

mathieud (approved):

packages/slice-machine/lib/mock/misc/fs.ts
```diff
@@ -25,16 +23,13 @@ export const insert = (
...
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
   const config = getConfig(cwd);
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
   const withInsert = {
     ...config,
     ...(prefix
       ? {
           // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-assignment
           [prefix]: {
-            // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
Heyrows, line 37: nice

packages/slice-machine/server/src/api/custom-types/delete.ts
```diff
@@ -0,0 +1,97 @@
...
+import { DeleteCustomTypeQuery } from "../../../../lib/models/common/CustomType";
+import { remove as removeCtsFromMockConfig } from "../../../../lib/mock/misc/fs";
+import path, { resolve } from "path";
+
+export default async function handler(
+  req: RequestWithEnv
+): Promise<
+  { err: unknown; reason: string; status: string } | Record<string, unknown>
+> {
+  const state = await getBackendState(req.errors, req.env);
```
Heyrows, line 18: you don't need to do that, it's already in the `req: RequestWithEnv (req.env)`

packages/slice-machine/server/src/api/custom-types/delete.ts
```diff
@@ -0,0 +1,97 @@
...
+  const { id } = req.query as DeleteCustomTypeQuery;
+  const ctFolder = CustomTypesPaths(state.env.cwd).customType(id).folder();
+  const customTypeAssetsFolder = GeneratedCustomTypesPaths(state.env.cwd)
+    .customType(id)
+    .folder();
+
+  try {
+    IO.CustomType.deleteCustomType(ctFolder);
```
Heyrows, line 26: this should be awaited, so that we don't try to do the rest if it fails.

danigo99: This function is already synchronous, so we don't need to await it right? All the other fs operations we do are the same.

PR #685, opened 2022-11-11
Delete custom types saga

danigo99:
## Context

Updating state on CT deletion.
Ticket: https://linear.app/prismic/issue/SM-865/[delete-ct]-aauser-i-can-see-the-deleted-custom-types-disappear-from

## The Solution

Added delete CT saga, actions and reducer.
Added a loading state and error toasters on response from the modal.

## Impact / Dependencies

Waiting on https://github.com/prismicio/slice-machine/pull/683

- [ ] The CI is successful.

## [OPT] Preview

ezgif com-gif-maker (6)

---

image


packages/slice-machine/src/modules/availableCustomTypes/index.ts
```diff
@@ -226,6 +256,49 @@ export function* renameCustomTypeSaga({
...
+      );
+    } else {
+      yield put(
+        openToasterCreator({
+          message:
+            "An unexpected error happened while deleting your custom type.",
+          type: ToasterType.ERROR,
+        })
+      );
+    }
```
danigo99, line 288: This wasn't specified in the ticket, but it's only for the case where the error caught isn't an Axios one. Should never happen but was just to be safe.

packages/slice-machine/tests/src/modules/availableCustomTypes/index.test.ts
```diff
@@ -297,4 +337,83 @@ describe("[Available Custom types module]", () => {
...
+        deleteCustomTypeCreator.request(actionPayload)
+      );
+
+      jest.spyOn(axios, "isAxiosError").mockImplementation(() => true);
+
+      const err = Error() as AxiosError;
+      // @ts-expect-error Ignoring the type error since we only need these properties to test
+      err.response = {
+        data: { reason: "Could not delete custom type", type: "error" },
+      };
```
danigo99, line 408: This isn't great but haven't been able to find a much better alternative

packages/slice-machine/src/modules/availableCustomTypes/index.ts
```diff
@@ -161,6 +182,15 @@ export const availableCustomTypesReducer: Reducer<
         [id]: newCustomType,
       };
     }
+
+    case getType(deleteCustomTypeCreator.success): {
+      const id = action.payload.customTypeId;
+
+      delete state[id];
```
mathieud: The redux state is immutable, don't think you can use `delete` on it. I'm actually surprised it works (unless there's a call to `/state` just after).

You can use

```
return {
  ...state,
  id: undefined
}
``

File: packages/slice-machine/src/modules/availableCustomTypes/index.ts [hunk 4]
```diff
@@ -226,6 +256,49 @@ export function* renameCustomTypeSaga({
...
 }
 
+export function* deleteCustomTypeSaga({
+  payload,
+}: ReturnType<typeof deleteCustomTypeCreator.request>) {
+  try {
+    yield call(deleteCustomType, payload.customTypeId);
+    yield put(deleteCustomTypeCreator.success(payload));
+    yield put(
+      modalCloseCreator({ modalKey: ModalKeysEnum.DELETE_CUSTOM_TYPE })
```
mathieud: This happens in both cases (try and except), you could put it just once at the end

packages/slice-machine/src/modules/availableCustomTypes/index.ts
```diff
@@ -226,6 +256,49 @@ export function* renameCustomTypeSaga({
...
+        type: ToasterType.SUCCESS,
+      })
+    );
+  } catch (e) {
+    yield put(
+      modalCloseCreator({ modalKey: ModalKeysEnum.DELETE_CUSTOM_TYPE })
+    );
+    if (axios.isAxiosError(e)) {
+      const apiResponse = e.response?.data as DeleteCustomTypeResponse;
+      console.log(apiResponse);
```
mathieud: Do we want to keep this?

packages/slice-machine/components/DeleteCTModal/index.tsx
```diff
@@ -104,11 +110,14 @@ export const DeleteCustomTypeModal: React.FunctionComponent<
...
               variant="danger"
-              // eslint-disable-next-line @typescript-eslint/no-misused-promises
-              onClick={async () => {
-                await deleteCustomType(customType?.local.id as string);
-                closeDeleteCustomTypeModal();
-              }}
+              isLoading={isDeletingCustomType}
+              onClick={() =>
+                deleteCustomType(
+                  customType?.local.id ?? "",
```
mathieud: It should never happen, but do you know what happens if `customType` is actually undefined here?

danigo99: A lot of bad things 😂 it deletes all the customTypes!

mathieud (approved):

danigo99:
Updated modal: 
image

PR #686, opened 2022-11-16
Slice mocks single location

arnaudlewis:
## Context

Since we introduce the editor on SM, we decided that the slice mocks would always be located alongside the code in the slice folder. This decision was made because once the code edited, it becomes purely a user input which cannot be regenerated so It didn’t make sense to leave it in .slicemachine.

## What has been done

So far, we had two location possible for a mock, either in `.slicemachine` or in the slice folder.
To make sure it would never end up in `.slicemachine` all the helpers allowing us to target `.slicemachine` to read or write mocks there were removed. There is only on helper for mocks that will always target the slice folder.
Also, we ensured that only the `Paths` helper from the core would retained that logic instead of having it replicated in the `Paths` module on SM.


packages/slice-machine/server/src/api/common/generate.ts
```diff
@@ -20,10 +20,8 @@ export function generate(
     );
 
     components.forEach((c) => {
-      const mocksPath = CustomPaths(env.cwd)
-        .library(c.from)
-        .slice(c.model.name)
-        .mocks();
+      sliceMockPath(env.cwd, c.from, c.model.name);
```
hypervillain: this should probably be deleted

packages/slice-machine/lib/storybook.ts
```diff
@@ -59,10 +60,7 @@ export default {
       return;
 
     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
```
hypervillain: are these comments still required?

hypervillain (changes requested):

PR #698, opened 2022-11-22
Fix bug in removing deleted content relationships

danigo99:
## Context

If we didn't replace a deleted CT in a content-relationship field, then the filter would not remove this from the model. 
Bug found in this ticket: https://linear.app/prismic/issue/SM-866#comment-416fe4a4

## The Solution

Set the field value on first-render with a useEffect after the custom types are filtered to remove deleted ones.

## Impact / Dependencies

NA

- [ ] The CI is successful.

## [OPT] Preview

ezgif com-gif-maker (2)

### DIFF
--- a/packages/slice-machine/lib/models/common/widgets/ContentRelationship/Form.tsx
+++ b/packages/slice-machine/lib/models/common/widgets/ContentRelationship/Form.tsx
@@ -11,7 +11,6 @@ import { createFieldNameFromKey } from "@lib/forms";
 import { useSelector } from "react-redux";
 import { selectAllCustomTypes } from "@src/modules/availableCustomTypes";
 import { FormikProps } from "formik";
-import { useEffect } from "react";
 
 const FormFields = {
   label: DefaultFields.label,
@@ -25,11 +24,6 @@ type FormProps = {
   config: { label: string; select: string; customtypes?: string[] };
   id: string;
   // type: string; // TODO: this exists in the yup schema but this doesn't seem to be validated by formik
-};
-
-type SelectValueType = {
-  value: string;
-  label: string | null | undefined;
 };
 
 const WidgetForm = ({
@@ -46,25 +40,13 @@ const WidgetForm = ({
   }));
 
   const selectValues = formValues.config.customtypes
-    ? formValues.config.customtypes
-        .map((id) => {
-          const ct = customTypes.find(
-            (frontendCustomType) => frontendCustomType.local.id === id
-          );
-          return ct ? { value: ct.local.id, label: ct.local.label } : ct;
-        })
-        .filter((val): val is SelectValueType => val !== undefined)
+    ? formValues.config.customtypes.map((id) => {
+        const ct = customTypes.find(
+          (frontendCustomType) => frontendCustomType.local.id === id
+        );
+        return { value: ct?.local.id, label: ct?.local.label };
+      })
     : null;
-
-  // removes deleted custom type from initial values on first render
-  useEffect(
-    () =>
-      setFieldValue(
-        "config.customtypes",
-        selectValues?.map(({ value }) => value)
-      ),
-    []
-  );
 
   return (
     <FlexGrid>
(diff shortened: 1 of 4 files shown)


packages/slice-machine/lib/models/common/widgets/ContentRelationship/Form.tsx
```diff
@@ -55,6 +56,16 @@ const WidgetForm = ({
         .filter((val): val is SelectValueType => val !== undefined)
     : null;
 
+  // removes deleted custom type from initial values on first render
+  useEffect(
```
Heyrows: ❓ As we filter the values i'm guessing before setting the field value for the first time, why isn't hitting the save button afterwards already enough to be able to save the custom type ?

danigo99: Since there's no change to the field value (no explicit interaction with it), setFieldValue in the onChange was never called before. I think because of this, the onSubmit of the form doesn't so anything

packages/slice-machine/lib/models/common/widgets/ContentRelationship/index.ts
```diff
@@ -60,4 +62,21 @@ export const ContentRelationshipWidget: Widget<Link, typeof schema> = {
   FormFields,
   CUSTOM_NAME: "ContentRelationship",
   Form,
+  prepareInitialValues: (initialValues) => {
+    const customTypes = useSelector(selectAllCustomTypes);
```
Heyrows, line 70: i don't know if using the selector is a good idea here because it's not in a component or anything, but i understand that it's morel logical to have it here.

packages/slice-machine/lib/models/common/widgets/ContentRelationship/index.ts
```diff
@@ -60,4 +62,21 @@ export const ContentRelationshipWidget: Widget<Link, typeof schema> = {
   FormFields,
   CUSTOM_NAME: "ContentRelationship",
   Form,
+  prepareInitialValues: (initialValues) => {
```
Heyrows, line 69: Should this be types ? (initial values)

Heyrows (approved):

packages/slice-machine/lib/models/common/widgets/ContentRelationship/index.ts
danigo99: I think if we want to keep generic things generic, then it seems like the other options are:

1. Do this filtering in the widget form like we had before
2. Have a selector which gets the whole state and passes it to the prepareInitialValues of all the forms (seems excessive)
3. Have the selector in here

It doesn't seem to be causing any issues I think. What do you recon?

packages/slice-machine/lib/models/common/widgets/ContentRelationship/index.ts
danigo99: The type is defined in the interface here: https://github.com/prismicio/slice-machine/blob/5b6a8efaea00332316ed8ceff940f7da3fcd2a06/packages/slice-machine/lib/models/common/widgets/Widget.ts#L26

Is that what you mean?

packages/slice-machine/lib/models/common/widgets/ContentRelationship/index.ts
Heyrows: It seems good then.
I was just worried because I usually see those at the top level of a react component.
Maybe i'm too narrow minded on this one 😄

packages/slice-machine/lib/models/common/widgets/ContentRelationship/index.ts
Heyrows: yes, but I would have expected something without an `any` if possible.

The interface can have any I guess but here it's a special use case with a special type.
Would it be hard to type it? Do you already have the type maybe?

Heyrows (approved):

PR #700, opened 2022-11-23
refactor: init command for plugins

lihbr:
WIP


package.json
```diff
@@ -63,5 +63,8 @@
     "**/packages/nuxt-sm/**/*": "prettier --write --ignore-unknown",
     "**/packages/vue-slicezone/**/*": "prettier --write --ignore-unknown",
     "**/packages/sm-commons/**/*": "prettier --write --ignore-unknown"
-  }
+  },
+  "workspaces": [
+    "packages/*"
+  ]
```
angeloashmore: Does this break other packages? @MarcMcIntosh mentioned this might introduce issues with the way modules are resolved.

packages/core2/src/auth/PrismicAuthManager.ts
```diff
@@ -127,6 +138,69 @@ export class PrismicAuthManager {
...
 	}
 
+	async browserLogin(): Promise<void> {
+		return new Promise<void>(async (resolve) => {
+			// Timeout attempt after 3 minutes
+			const timeout = setTimeout(() => {
+				throw new Error(
+					"Login timeout, server did not receive a response within a 3-minute delay",
+				);
+			}, 180_000);
```
angeloashmore: Should this timeout handler also close the server if it's running?

packages/core2/src/auth/PrismicAuthManager.ts
```diff
@@ -127,6 +138,69 @@ export class PrismicAuthManager {
 		await this._writePersistedAuthState(authState);
 	}
 
+	async browserLogin(): Promise<void> {
```
angeloashmore: I think it's worth considering making this an Express middleware rather than a full server. A middleware may not be the right approach for this method, but doing so would decouple it from a specific server implementation.

I initial wrote a function like this that spun up a new temporary server with a random port. That function was turned into `createPrismicAuthManagerMiddleware()`. That function needs to be cleaned up (it turns out `/auth/status` and `/auth/start` are Slice Machine-specific and don't belong in the generic middleware), which would turn it into something that looks similar to what you have here.

packages/core2/src/auth/PrismicAuthManager.ts
```diff
@@ -127,6 +138,69 @@ export class PrismicAuthManager {
...
+				}),
+			);
+
+			// Start server
+			const server = http.createServer(h3.toNodeListener(app));
+			await new Promise<void>((resolve) => {
+				server.once("listening", () => {
+					resolve();
+				});
+				server.listen(5555);
```
angeloashmore: Should we decide to continue creating a temporary server, I think it would be nice to us a randomly available port rather than the hardcoded value here. It would eliminate the chance that two auth login attempts would conflict. It also prevents the case where someone has something unrelated on port 5555.

packages/core2/src/auth/PrismicAuthManager.ts
```diff
@@ -127,6 +138,69 @@ export class PrismicAuthManager {
...
+			});
+
+			// Open browser
+			const url = await new URL(
+				"/dashboard/cli/login?source=slice-machine",
+				APIEndpoints.PrismicWroom,
+			);
+			await open(url.toString());
```
angeloashmore: Although I know `open()` should work even when called in the browser, I'm not sure it should happen from the server. I think it's cleaner to return the URL to open from this method and let the consumer open the new window. It's possible this `open()` method would open a different browser than the one being used.

What do you think? I'm on the fence with this one.

packages/core2/src/auth/PrismicAuthManager.ts
```diff
@@ -171,9 +245,28 @@ export class PrismicAuthManager {
 		}
 	}
 
+	async getAuthenticationCookies(): Promise<
```
angeloashmore, line 240: We could probably re-use this method in `getAuthenticationToken()`.

packages/core2/src/auth/PrismicAuthManager.ts
```diff
@@ -260,7 +353,7 @@ export class PrismicAuthManager {
 		}
 	}
 
-	private async _getPersistedAuthStateFilePath(): Promise<string> {
```
angeloashmore, line 307: Oops, thanks!

packages/core2/src/lib/captureSliceSimulatorScreenshot.ts
```diff
@@ -1,6 +1,6 @@
+import * as http from "node:http";
```
angeloashmore: This will be replaced with `node-fetch` per your past comment.

packages/core2/src/managers/_RepositoryManager.ts
```diff
@@ -1,6 +1,231 @@
...
+});
+export type PrismicRepository = t.TypeOf<typeof PrismicRepository>;
+
+type SliceMachineManagerExistsRepositoryArgs = {
+	domain: string;
+};
+
+type SliceMachineManagerCreateRepositoryArgs = {
+	domain: string;
+	framework: string; // TODO: Type(?)
```
angeloashmore, line 51: Maybe adapters or plugins need to return this value.

- Adapter: `@slicemachine/adapter-s3`
- Plugins: `[@slicemachine/adapter-next]`

In the above case, the adapter can't return a framework since it is framework agnostic. The `adapter-next` plugin, however, could return a framework. This implies that multiple plugins could return a framework, which may still be accurate. If we want to track how a repository is used, we should track every framework in which it is used.

Alternatively, framework detection logic could be internal to Slice Machine and only limited to a known set of frameworks. For example, we could detect Next.js/Nuxt/Astro by manually checking package.json without the involvement of plugins. If we are only using `framework` for tracking purposes, this would be the simpler, albeit unsustainable, approach.

packages/core2/src/managers/_RepositoryManager.ts
```diff
@@ -1,6 +1,231 @@
...
+			}
+
+			return repositories;
+		} else {
+			throw new Error(`Failed to read repositories`, { cause: json });
+		}
+	}
+
+	// Should this be in manager? It's one of the few sync method
+	hasWriteAccess(repository: PrismicRepository): boolean {
```
angeloashmore, line 93: I wonder if `RepositoryManager` should be connected to the project's repository. If we did that, `hasWriteAccess()` would no longer require passing a repository.

Do you see a need for this method to be repository-agnostic?

If it becomes stateful, then yes, the method should remain in the manager.

packages/core2/src/managers/_RepositoryManager.ts
```diff
@@ -1,6 +1,231 @@
...
+			case PrismicRepositoryRoles.Owner:
+			case PrismicRepositoryRoles.Administrator:
+				return true;
+
+			default:
+				return false;
+		}
+	}
+
+	async exists(
```
angeloashmore: To keep a consistent naming scheme, could this be called `checkExists()`? Alternatively, we could call it `checkIsAvailable()` and flip the response, which I think is more inline with the way the function is used (but I'm aware it is named after the API endpoint).

The `check-` prefix comes from needing to store the result in a variable in some cases. The prefix makes it easier to read, IMO:

```typescript
const isAvailable = this.checkIsAvailable(/* ... */)
```

packages/core2/src/managers/_RepositoryManager.ts
```diff
@@ -1,6 +1,231 @@
...
+			APIEndpoints.PrismicWroom,
+		);
+
+		const body = {
+			domain: args.domain,
+			framework: args.framework, // This property appears to be optional for the API
+			// TODO: Maybe those data shouldn't be hardcoded
+			plan: "personal", // Using any other value than "personal" requires payment
+			isAnnual: "false", // This property appears to be optional for the API
+			role: "developer", // This property appears to be optional for the API
```
angeloashmore: Is these values remain as constants, could we move them into the `constants.ts` file ([see here](https://github.com/prismicio/slice-machine/blob/ca52148eb1dd6d03aeaa3fa7e80e39798cbd394d/packages/core2/src/constants.ts))? Or it could live at the top of this file as well, whichever you think makes more sense.

packages/core2/src/managers/_RepositoryManager.ts
```diff
@@ -1,6 +1,231 @@
...
+		};
+
+		const res = await this._fetch({
+			url,
+			method: "POST",
+			body,
+			userAgent: "prismic-cli/sm", // Custom User Agent is required
```
angeloashmore: Same as above comment re: moving constants to `constants.ts`.

packages/core2/src/managers/_RepositoryManager.ts
```diff
@@ -1,6 +1,231 @@
...
+		});
+		const text = await res.text();
+
+		// Endpoint returns repository name on success, which must be more than 4 characters and less than 30
+		// if (!res.ok) {
+		if (!res.ok || text.length < 4 || text.length > 30) {
```
angeloashmore, line 147: Out of curiosity, what happens if we don't check for the length? Does the API ever return something other than a repository name while the response is OK?

packages/core2/src/managers/_RepositoryManager.ts
```diff
@@ -1,6 +1,231 @@
...
+				cause: res,
+			});
+		}
+	}
+
+	async delete(args: SliceMachineManagerDeleteRepositoryArgs): Promise<void> {
+		const cookies = await this.user.getAuthenticationCookies();
+
+		const url = new URL(
+			`/app/settings/delete?_=${cookies["X_XSRF"]}`, // TODO: Maybe we want to throw early if the token is no available, or get the token another way
```
angeloashmore, line 158: I think a specialized `UserManager` method like `user.getXSRFToken()` would be useful, similar to `user.getAuthenticationToken()`.

packages/core2/src/managers/_RepositoryManager.ts
```diff
@@ -1,6 +1,231 @@
...
+		if (args.body) {
+			extraHeaders["Content-Type"] = "application/json";
+		}
+
+		return fetch(args.url.toString(), {
+			method: args.method,
+			body: args.body ? JSON.stringify(args.body) : undefined,
+			headers: {
+				// Some endpoints rely on the authorization header...
+				Authorization: `Bearer ${cookies["prismic-auth"]}`,
```
angeloashmore, line 248: We should use `user.getAuthenticationToken()` or `cookies[AUTH_COOKIE_KEY]` rather than the hardcoded `"prismic-auth"` string. `AUTH_COOKIE_KEY` would need to be moved to the shared `constants.ts` file.

packages/core2/src/root.ts
```diff
@@ -0,0 +1,47 @@
+abstract class Manager {
```
angeloashmore: This is a temporary file, right?

angeloashmore: This is a temporary file, right?

packages/core2/src/rpc/lib/readProcedureArgs.ts
```diff
@@ -1,5 +1,5 @@
-import { H3Event, getHeaders } from "h3";
 import { Buffer } from "node:buffer";
+import { H3Event, getHeaders } from "h3";
```
angeloashmore: We could consider an imports Prettier plugin: https://github.com/trivago/prettier-plugin-sort-imports

packages/init2/src/cli.ts
```diff
@@ -0,0 +1,63 @@
...
+	}
+);
+
+if (cli.flags.help) {
+	cli.showHelp();
+} else if (cli.flags.version) {
+	// eslint-disable-next-line no-console
+	console.log(`${pkgName}@${pkgVersion}`);
+	process.exit(0);
+} else {
```
angeloashmore, line 77: Do we need to handle `--help` and `--version` manually?

Fun fact: I added the `autoHelp` and `autoVersion` options back in 2017 😛 - https://github.com/sindresorhus/meow/pull/68

packages/init2/src/lib/assertExists.ts
```diff
@@ -0,0 +1,9 @@
+// TypeScript cannot use arrow functions for assertions
+export function assertExists<T>(
+	value: T,
+	name: string
+): asserts value is NonNullable<T> {
+	if (value == undefined) {
+		throw new Error(`Failed to resolve ${name}`);
```
angeloashmore: Small nit: "resolve" could be confused with Node.js module resolution.

If that's what the message means, then the function name could be clearer using something like "assertsModuleExists".

If not, then a message like "`${name} does not exist`" would be clearer.

packages/init2/src/lib/framework.ts
```diff
@@ -0,0 +1,110 @@
...
+		name: "Next.js 11-13",
+		prismicName: "next-11-13",
+		compatibility: {
+			next: "^11.0.0 || ^12.0.0 || ^13.0.0",
+		},
+		devDependencies: {
+			...DEFAULT_DEV_DEPENDENCIES,
+			"@slicemachine/adapter-next": "latest",
+		},
+	},
```
angeloashmore: This is a really clever system. 🙂 It looks like it will scale well as new incompatible versions are released.

packages/init2/src/lib/framework.ts
```diff
@@ -0,0 +1,110 @@
...
+		compatibility: {
+			nuxt: "^2.0.0",
+		},
+		devDependencies: {
+			...DEFAULT_DEV_DEPENDENCIES,
+			"@slicemachine/adapter-nuxt": "latest",
+		},
+	},
+	"next-11-13": {
+		name: "Next.js 11-13",
```
angeloashmore, line 54: I think we can leave off the version numbers in the name *if it's the latest version of the framework*. Most developers will be running `init` in new projects, which should be using the latest version by default. There's no reason for them to think "Next.js" doesn't mean the latest version.

For older versions (e.g. Nuxt 2), then there is value in explicitly stating it's installing support for an older version.

WDYT?

packages/init2/src/lib/framework.ts
```diff
@@ -0,0 +1,110 @@
...
+		},
+		devDependencies: {
+			...DEFAULT_DEV_DEPENDENCIES,
+			"@slicemachine/adapter-next": "latest",
+		},
+	},
+} as const;
+
+/** Vanilla package used when framework is not supported. */
+export const VANILLA: Framework = {
```
angeloashmore: Bikeshedding: "Universal" is more accurate IMO. "Vanilla" implies "VanillaJS," i.e. no framework, but this "fallback" adapter would be used by any framework not officially supported, such as Astro.

packages/init2/src/lib/listr.ts
```diff
@@ -0,0 +1,15 @@
+// @ts-expect-error - @types package not available
+import UpdateRenderer from "@lihbr/listr-update-renderer";
```
angeloashmore, line 2: We may want to consider forking your fork into a `prismicio-community` package, just so code depended upon in Slice Machine is tied to Prismic rather than you. Certainly not a requirement, I'll leave that decision up to you.

packages/init2/src/lib/framework.ts
```diff
@@ -0,0 +1,110 @@
...
+	name: "vanilla (no framework)",
+	prismicName: "vanilla",
+	compatibility: {},
+	devDependencies: {
+		...DEFAULT_DEV_DEPENDENCIES,
+		"@slicemachine/adapter-vanilla": "latest",
+	},
+};
+
+export const detect = async (): Promise<Framework> => {
```
angeloashmore: Generic names like `detect` can be hard to follow elsewhere in the code, even if the intention is to import the module as a namespace. What do you think of renaming this function and file to `detectFramework`?

packages/init2/src/lib/packageManager.ts
```diff
@@ -0,0 +1,64 @@
...
+
+	const execaProcess = execaCommand(command, {
+		encoding: "utf-8",
+		...args.execa,
+	});
+
+	return { execaProcess };
+};
+
+export const detect = async (): Promise<Agent> => {
```
angeloashmore: Similar to the previous comment about `framework.detect()`, what do you think of renaming this function and file to `detectPackageManager`?

packages/init2/src/lib/repositoryDomain.ts
```diff
@@ -0,0 +1,121 @@
...
+	"kolompeh",
+	"macaron",
+	"paste",
+	"pretzel",
+	"rustico",
+	"semla",
+	"strudel",
+	"taiyaki",
+	"toast",
+].filter((bakery) => bakery.length <= 11);
```
angeloashmore, line 41: This is my second favorite part of this PR.

packages/init2/src/lib/repositoryDomain.ts
```diff
@@ -0,0 +1,121 @@
...
+	"inspiring",
+	"lovable",
+	"melodious",
+	"pleasant",
+	"pretty",
+	"shiny",
+	"sugary",
+	"tall",
+	"valliant",
+].filter((adjective) => adjective.length <= 11);
```
angeloashmore, line 21: This is my favorite part of this PR.

angeloashmore (commented):
This looks really good. 🙂 I added a lot of comments, but most of them are small edits or need some discussion before committing to changes.

packages/init2/src/lib/repositoryDomain.ts
angeloashmore, line 41: I wonder if the `filter()` could be accomplished using a TypeScript type, one that only accepts words less than 12 characters.

package.json
lihbr: Well, that was tenporary as installing everything took ages.

I'd prefer to fix those resolving issues than not relying on hoisting tbf ^^'

packages/core2/src/auth/PrismicAuthManager.ts
lihbr: Definitely yes!

packages/core2/src/managers/_RepositoryManager.ts
lihbr: Yes it needs to be agnostic (c.f. usage in the init command)

This is typically the kind of method/function that I don't know where to store right now and that we need a clear guideline to define where.

packages/core2/src/managers/_RepositoryManager.ts
lihbr: Hmm, I'm not a fan of starting to tuck everything in the constant file/wondering if we should not already start offloading `core2` from the auth/repository clients into their own packages and maybe considering a constants package (all in the monorepo for now)

packages/core2/src/root.ts
lihbr: Lol yes, it was the proof of concept for the discussion on Notion, didn't know I commit it :see_no_evil:

packages/init2/src/cli.ts
lihbr: I didn't like the default formatting 😭 + wanted to force show help/version if flags are present.

packages/init2/src/lib/framework.ts
lihbr: I don't know, maybe?

Today if you say you're building a Nuxt website, people wouldn't necessarily assume it's a Nuxt 3 one.

Also explicitely stating versions gives us some room to maneover if let's say Next 14 is released and we need to perform some work to support it.

packages/init2/src/lib/framework.ts
lihbr: Definitely, I started to dance between the two styles and it was akward, will rework!

packages/init2/src/lib/listr.ts
lihbr: You don't trust me as a maintainer? 😥

jk, I think we might just have it in `lib` here since it's a small package (just one file), or fork it if we plan on using it a multiple places indeed.

packages/init2/src/lib/repositoryDomain.ts
lihbr: I feel like this overall file could be a package of its own, Sindre's style

lihbr: image

I tried this way, but TypeScript says no ^^'

packages/core2/src/auth/PrismicAuthManager.ts
lihbr: Good catch! I'm assuming it does automatically on SIGTERM but let's make it explicit

packages/core2/src/auth/PrismicAuthManager.ts
lihbr: Implemented using [`get-port`](https://www.npmjs.com/package/get-port), kept the preference for `5555`

packages/core2/src/auth/PrismicAuthManager.ts
lihbr: ~~So I refactored it into two functions:~~
~~- `getBrowerLoginInfo()`: picks an available port and return port & login URL~~
~~- `browserLogin()`: receives a port, spins the server, and await a valid auth payload for up to 3 minutes~~

~~At first, I just wanted `browserLogin()` to return a payload like `{ url: string; promise: Promise<void> }`, but it would have made it not compatible with browser usage (because the promise is not serializable).~~

Actually, I reconsidered. Just returning the URL to the browser would not work as such a pattern could result in a race condition where the login page is opened before the server is:
```typescript
const { port, url } = await auth.getBrowserLoginInfo();
window.open("https://google.com");
await auth.browserLogin({ port });
```

I think we'll ultimately need 2 methods, one for Node, one for the browser, so for now I have:
- `getLoginSessionInfo()`: returns port and URL
- `nodeLoginSession()`: receives a port and a callback for when the server started, then starts a login session for Node
- `browserLoginSesstion()`: ??? to be discussed

I'm not a fan of any of those names, so happy to discuss/rethink that process!

packages/core2/src/auth/PrismicAuthManager.ts
lihbr: I refactored it into `createPrismicAuthManagerMiddleware()`.

`browserLogin()` still uses `h3` to spin up the temporary server.

Let me know if that works :)

packages/core2/src/auth/PrismicAuthManager.ts
lihbr: Nice catch, had issues with type narrowing at first but cleaned it up since ☺️

packages/core2/src/lib/captureSliceSimulatorScreenshot.ts
lihbr: Refactored 👌

packages/core2/src/managers/_RepositoryManager.ts
lihbr: I think we might want to keep it under-engineered like this for now.

- This is used for ...? I guess (legacy?) tracking purposes only. Plugins used will be tracked another way (I'm assuming we might want to tell Segment the full list of plugins used at another point in the process)
- Relying on adapters/plugins to return a value would force the init process to await dependency installation before being able to create a repository

packages/core2/src/managers/_RepositoryManager.ts
lihbr: `checkExists` it is o/

packages/core2/src/managers/_RepositoryManager.ts
lihbr: Refactored 👌

packages/core2/src/managers/_RepositoryManager.ts
lihbr: Refactored 👌

packages/core2/src/managers/_RepositoryManager.ts
lihbr: I checked:

- for `length < 4`: HTTP 400 with plain text error
- for already existing repositories: HTTP 500 with HTML error
- for `length > 30`: HTTP 200 with plain text repository name
- for `length > 63` (the maximum length a subdomain can be): HTTP 500 with HTML error

lihbr: To be followed here: https://prismic-team.slack.com/archives/C048EJU096Z/p1670240206140689

packages/core2/src/managers/_RepositoryManager.ts
lihbr: Definitely, I don't know the endpoint though 🤔

packages/core2/src/rpc/lib/readProcedureArgs.ts
lihbr: Was mainly nitpicking, I think we might want to keep it manually, it's not that important(?)

packages/init2/src/lib/assertExists.ts
lihbr: Refactored to just hold a message

packages/init2/src/lib/framework.ts
lihbr: Renamed

packages/init2/src/lib/framework.ts
lihbr: Refactored

packages/init2/src/lib/packageManager.ts
lihbr: Refactored

lihbr:
OK, so I handled most of the review today, the remaining is to be discussed in the opened discussion @angeloashmore, thanks again!

packages/core2/src/auth/PrismicAuthManager.ts
angeloashmore: Looks good to me!

packages/core2/src/auth/PrismicAuthManager.ts
angeloashmore: Background info: SM currently uses polling along with an `/api/auth/status` endpoint that returns true/false depending on if the user is logged in. That endpoint can be reproduced using PrismicAuthManager's existing methods fairly easily.

Besides using a WebSocket or long polling, both of which we aren't prepared to handle at this time, I'm not sure of solution besides polling that could work in the browser.

I think for now we should implement the login flows individually, once for `@sm/init` and again for SM UI (via `start-slicemachine` and `slice-machine-ui`). We can share `PrismicAuthManager` logic, such as the `login()` method, and `createPrismicAuthManagerMiddleware()` as it is now with just the `/` endpoint. Everything else, such as starting the server and reacting to `onLoginCallback`, should be handled by the individual consumer.

Once we have the two implementations, we can see what kind of overlap they have and, if it makes sense, create a general abstraction.

We can chat about this on a call if you want to discuss our approach further. This is a tricky problem to solve. :)

packages/init2/src/lib/listr.ts
angeloashmore: lol, I trust you. I'm just looking out for you and the company.

`lib` sounds good to me if it's small enough to fit in a single file.

packages/init2/src/lib/repositoryDomain.ts
angeloashmore: Haha, oh well!

packages/core2/src/managers/_RepositoryManager.ts
angeloashmore: Okay, yes, lets go with the simpler `package.json` detection method. Init already does some form of this. Maybe we could move that logic into `core2` if it isn't already?

The legacy `getState()` method requires `framework` as well (I just filled it in with `""` for now).

packages/core2/src/managers/_RepositoryManager.ts
angeloashmore: `user.getXSRFToken()` would just read the user's `.prismic` cookies and return the `"X_XSRF"` value. Basically a clone of `user.getAuthenticationToken()` but returning a different cookie.

packages/core2/src/managers/_RepositoryManager.ts
angeloashmore: A constants package seems overly complicated right now. We can use methods to achieve similar abstractions, even before moving `PrismicAuthManager` and `_RepositoryManager` to their own packages (if we decide to do that).

In this case, I'd go with the following code, even if it is less performant:

```ts
const cookies = await this.user.getAuthenticationCookies();
const authenticationToken = await this.user.getAuthenticationToken();

// ...

				Authorization: `Bearer ${authenticationToken}`,
```

Just a suggestion, totally up to you which solution you think is best. 🙂

angeloashmore:
I'm trying this pattern out for `start-slicemachine`, inspired by the way you structured `@sm/init`:

- [`src/cli.ts`](https://github.com/prismicio/slice-machine/blob/9d5643637a332c4bd0a478858a597ba2b807e83d/packages/start-slicemachine/src/cli.ts): Runs the CLI process, parses arguments, handles help/version text.
- [`src/StartSliceMachineProcess.ts`](https://github.com/prismicio/slice-machine/blob/9d5643637a332c4bd0a478858a597ba2b807e83d/packages/start-slicemachine/src/StartSliceMachineProcess.ts): Contains a `run()` method with all script logic.

It uses [`mri`](https://github.com/lukeed/mri) and lazy loads `StartSliceMachineProcess`, which means the CLI is super quick to start. If you only run `--help`, for example, only `mri` is loaded, not all of `core2` and `StartSliceMachineProcess`.

It might be worth trying for `@sm/init` since it seems like we aren't using many of meow's features.

lihbr:
Moved to #736

PR #706, opened 2022-11-24
Feature/sentry express

mathieud:


packages/slice-machine/server/src/sentry-setup.ts
```diff
@@ -0,0 +1,72 @@
...
+      "https://b673ba8b041d4449a0fb0a38691882dd@o146123.ingest.sentry.io/4504179268845568",
+    release: RELEASE_NUMBER,
+    environment: isStableVersion ? process.env.NODE_ENV : "alpha",
+  });
+
+  // TODO should we refresh on getState like frontend?
+  getEnv()
+    .then((state) => {
+      Sentry.setUser({ id: state.env.prismicData.shortId });
+      // TODO is state.env.repo the same as state.env.client.repository?
```
MarcMcIntosh, line 38: good question

packages/slice-machine/server/src/sentry-setup.ts
```diff
@@ -0,0 +1,72 @@
...
+    throw new Error("Hello From Express 1");
+  });
+};
+
+export const addSentryPostHandler = (app: express.Express) => {
+  if (!useSentry()) {
+    return;
+  }
+
+  app.use(Sentry.Handlers.errorHandler());
```
MarcMcIntosh, line 69: > By default, errorHandler will capture only errors with a status code of 500 or higher. If you want to change it, provide it with the shouldHandleError callback, which accepts middleware errors as its argument and decides, whether an error should be sent or not, by returning an appropriate boolean value.

Do use 500 response codes anywhere in the app?

mathieud: We only log uncaught exceptions at the moment, which will result in a 500. 

We could add a custom handler for 400s, but it might create a lot of noise for us to deal with.

packages/slice-machine/server/src/sentry-setup.ts
mathieud: Seems like yes: https://github.com/prismicio/slice-machine/blob/master/packages/slice-machine/server/src/api/services/getEnv.ts#L103

MarcMcIntosh: Yes. 
this will just be the subdomain of the repo, which should be enough to identify users

packages/slice-machine/server/src/sentry-setup.ts
MarcMcIntosh: if it throws does express automatically create a 500?

MarcMcIntosh: I guess we'll find out when teesting

packages/slice-machine/rollup-server.config.js
```diff
@@ -19,5 +30,25 @@ export default {
...
+
+        // Auth tokens can be obtained from https://sentry.io/settings/account/api/auth-tokens/
+        // and need `project:releases` and `org:read` scopes
+        authToken: SENTRY_AUTH_TOKEN,
+
+        // Optionally uncomment the line below to override automatic release name detection
+        release: RELEASE_NUMBER,
+
+        // Specify the directory containing build artifacts
+        include: "./build/server",
```
MarcMcIntosh, line 46: note: in the server folder there should only be an `index.js` file..... if there's not it means some other part of the build system is importing from server :)

MarcMcIntosh (approved):

PR #712, opened 2022-11-29
lint: onboarding fix linter error

MarcMcIntosh:
## Context
Fix the linter error on onboarding page

- [ ] The CI is successful.

### DIFF
--- a/packages/slice-machine/components/Simulator/index.tsx
+++ b/packages/slice-machine/components/Simulator/index.tsx
@@ -54,19 +54,19 @@ export default function Simulator() {
     ScreenSizes[ScreenSizeOptions.DESKTOP]
   );
 
-  if (!component || !variation) {
-    return <div />;
-  }
-
   const sliceView = useMemo(
     () => [
       {
-        sliceID: component.model.id,
-        variationID: variation.id,
+        sliceID: component?.model.id,
+        variationID: variation?.id,
       },
     ],
-    [component.model.id, variation.id]
+    [component?.model.id, variation?.id]
   );
+
+  if (!component || !variation) {
+    return <div />;
+  }
 
   return (
     <Flex sx={{ height: "100vh", flexDirection: "column" }}>
--- a/packages/slice-machine/components/Simulator/index.tsx
+++ b/packages/slice-machine/components/Simulator/index.tsx
@@ -54,15 +54,17 @@ export default function Simulator() {
     ScreenSizes[ScreenSizeOptions.DESKTOP]
   );
 
-  const sliceView = useMemo(
-    () => [
-      {
-        sliceID: component?.model.id,
-        variationID: variation?.id,
-      },
-    ],
-    [component?.model.id, variation?.id]
-  );
+  const sliceView = useMemo(() => {
+    if (component?.model.id && variation?.id) {
+      return [
+        {
+          sliceID: component?.model.id,
+          variationID: variation?.id,
+        },
+      ];
+    }
+    return [];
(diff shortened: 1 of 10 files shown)


packages/slice-machine/components/Simulator/index.tsx
```diff
@@ -54,20 +54,22 @@ export default function Simulator() {
     ScreenSizes[ScreenSizeOptions.DESKTOP]
   );
 
+  const sliceView = useMemo(() => {
+    if (component?.model.id && variation?.id) {
+      return [
+        {
+          sliceID: component?.model.id,
```
bapmrl: 💡 I don't think the `?.` are required in the returned object.

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
```diff
@@ -155,15 +155,17 @@ const SliceBuilderWithRouter = () => {
     ),
   }));
 
+  useEffect(() => {
```
bapmrl: ⚠️ It conflicts with `chore/react-18`. Here's what I've written:
```
const SliceBuilderWithRouter = () => {
  const router = useRouter();
  const { initSliceStore } = useSliceMachineActions();

  const { component } = useSelector((store: SliceMachineStoreType) => ({
    component: selectCurrentSlice(
      store,
      router.query.lib as string,
      router.query.sliceName as string
    ),
  }));

  useEffect(() => {
    if (component) initSliceStore(component);
    else void router.replace("/");
  }, [component]);

  if (!component) {
    return null;
  }

  const variation = component.model.variations.find(
    (variation) => variation.id === router.query.variation
  );

  return <SliceBuilder component={component} variation={variation} />;
};

export default SliceBuilderWithRouter;
```

packages/slice-machine/src/hooks/useServerState.ts
```diff
@@ -10,15 +10,19 @@ const fetcher = (url: string): Promise<ServerState> =>
 
 const useServerState = () => {
   const { refreshState } = useSliceMachineActions();
+  const handleRefreshState = useCallback(refreshState, []);
```
bapmrl, line 13: 💡 Maybe a better approach would be to ensure that all actions returned by `useSliceMachineActions` have stable references (by calling `useCallback` on each of them — inside the `useSliceMachineActions` module). Yet, I think we can keep this as-is because it's simpler and introduces less changes in the `chore/react-18` PR.

bapmrl (commented):
Thank you!

packages/slice-machine/src/hooks/useServerState.ts
MarcMcIntosh: it was done purely for the linter warning 😅

packages/slice-machine/components/Simulator/index.tsx
MarcMcIntosh: good point :)

MarcMcIntosh: done

packages/slice-machine/src/hooks/useServerState.ts
MarcMcIntosh: I think they should have stable references, unless the dispatch function changes

packages/slice-machine/lib/builders/SliceBuilder/index.tsx
MarcMcIntosh: done

PR #715, opened 2022-11-30
editor-mocks: add server route to save the mocks

MarcMcIntosh:
## Context
Allow user to save edits to mocks from the simulator.

note: there is some flakyness in the e2e tests:
https://github.com/prismicio/slice-machine/actions/runs/3621078880/jobs/6104088189#step:8:385

maybe fixed here https://github.com/prismicio/slice-machine/actions/runs/3621527298

- [x] The CI is successful.


packages/slice-machine/server/src/api/slices/save-mock.ts
```diff
@@ -0,0 +1,55 @@
...
+  cwd: string,
+  libraryName: string,
+  sliceName: string,
+  mock: ComponentMocks
+) {
+  const path = sliceMockPath(cwd, libraryName, sliceName);
+  return Files.writeJson(path, mock);
+}
+
+const SaveMockBody = t.type({
```
arnaudlewis: you can replace `t.type` by `t.strict` which combines `t.type` and `t.exact` so even if you send too many data, they are filtered on the output.
Same for `SaveMockRequest` below, even if you send the whole `env`, it would only remain the `cwd` in the output so it’s only what we need that’s even better

arnaudlewis (approved):

PR #719, opened 2022-12-01
Delete index.js when slice was the last in lib

danigo99:
## Context

Issue: https://linear.app/prismic/issue/SM-861#comment-4f5d666b

## The Solution

Check if the lib is empty after updating the last slice. If so, delete index.js

- [x] The CI is successful.

---

image


danigo99:
Might be worth reading the commits separately since they each have a different strat

Heyrows (commented):
I watched both commits,
- on the first one I like the fact that we check if it is the last Slice explicitly
- on the second approach I'm not sure I understand why we delete the file first, I'm not sure to understand how that could solve the issue.

How about a third option, updating the existing functions:
- onSaveSlice is a function that regenerate the index for all libraries.
  I think we could re-use that to have functions that generate index for a specific library and 
  for all of them. We could make it very explicit that those functions are meant for this 
  purpose. `onSaveSlice` seems blurry to me, i wouldn't what happens here at first glance.
 
What do you think ?

packages/slice-machine/tests/server/slices/delete.test.ts
```diff
@@ -152,6 +152,16 @@ const CustomTypeModelWithoutSlice = {
   },
 };
 
+const emptyIndexFile = `// Code generated by Slice Machine. DO NOT EDIT.
```
Heyrows: OPT:

Could this be exported to reinforce the tests ?

packages/slice-machine/server/src/api/common/hooks/updateLibraries.ts
```diff
@@ -59,7 +59,7 @@ const createIndexFileForFrameWork = (
   return createIndexFile(lib);
 };
 
-export default async function onSaveSlice(
+export default async function reGenerateLibrariesState(
```
Heyrows: libraryStates.json is a very specific file in the file system (.slicemachine), i think it should be `generateLibrariesIndex)`

packages/slice-machine/server/src/api/slices/delete.ts
```diff
@@ -149,8 +146,7 @@ export async function deleteSlice(req: {
 
   const updateLibraries = async () => {
     try {
-      resetLibraryIndex(desiredLibrary?.path);
-      await onSaveSlice(env);
+      await reGenerateLibrariesState(env);
```
Heyrows: should we update all libraries or the one that is associated to the deleted slice ?

danigo99: I'm not sure. In the rename we update all of them, so I wasn't sure if this was because of some links between them? Otherwise we should probably just update one

packages/slice-machine/tests/server/slices/delete.test.ts
danigo99: Not really sure what you mean by this?  Is it so others can use it?

packages/slice-machine/server/src/api/slices/delete.ts
Heyrows: A library is basically a folder in your file system.
Updating a slice means updating a particular library.
For that, we just need to regenerate one index normally.
I guess it's ancient code that doesn't make sense anymore.

Is it too much work to change ?
What do you think ?

danigo99: Not much to change, have updated it 👍

packages/slice-machine/tests/server/slices/delete.test.ts
Heyrows: I meant if the original text change, this could make the test fail maybe ?
So by importing the text itself, maybe we could avoid this kind of situation.

If that's not easily do-able forget about it 😄 , i just thought of it as a way to improve the tests

Heyrows (approved):
looking good 👍 , thanks for the changes

PR #723, opened 2022-12-05
Cleanup package.json scripts

danigo99:
## Context

Want to make the package.json scripts easier to read and use. 
Ticket: https://linear.app/prismic/issue/SM-906/[tech-debt]-aadev-i-have-clear-and-small-list-of-usable-scripts-that

## The Solution

Condense unused sub-scripts into 1
Have a single start script which can be run from the root
Have a build script in the root which builds all packages

- [ ] The CI is successful.

---

image


packages/slice-machine/package.json
```diff
@@ -12,19 +12,16 @@
...
     "postinstall": "node postinstall",
     "prepublishOnly": "npm run bundle",
-    "dev-server": "cross-env PORT=9999 CWD=../../e2e-projects/next ENV=SM concurrently \"rollup -w -c ./rollup-server.config.js\" \"nodemon -e ts,js --watch server --watch lib --ignore build/ build/server/index.js\"",
-    "build": "npm run build-scripts && npm run build-server && npm run build-next-app",
-    "build-scripts": "tsc --project scripts/tsconfig.json",
-    "build-server": "rollup -c ./rollup-server.config.js",
-    "build-next-app": "cross-env NODE_OPTIONS=--max_old_space_size=8192 next build",
-    "export-build": "next export && npm run clean-build",
-    "bundle": "npm run build && npm run export-build",
+    "bundle": "npm run build && next export && rimraf ./build/client",
```
Heyrows: I think we don't need the bundle anymore.
There should just be a build command that makes the project available to be used elsewhere

packages/slice-machine/package.json
```diff
@@ -12,19 +12,16 @@
...
+    "export-build": "next export && rimraf ./build/client",
```
Heyrows: i don't think this is used anymore.

.github/workflows/deploy-alpha.yml
```diff
@@ -30,7 +30,7 @@ jobs:
           git checkout .
       - name: Build packages used by others
         run: |
-          npm run build-essential
+          npm run build:essential
```
Heyrows, line 33: If this is applied to master, all the other branches that are not synced will not be ablet odeploy anymore

danigo99: It will be applied to dev-delete for now. And I don't thin it will block deployments, because the package.json and deploy-alpha.yml will be on the same version in one branch (branches use the workflow yamls from their own branch)

packages/slice-machine/package.json
danigo99: It's used before the e2e tests in CI. Not sure if it's needed though after the bundle command is added to build

bapmrl (approved):
It seems simpler and I understand the flow 🙂

PR #724, opened 2022-12-06
setupModal variable tabs len

hypervillain:
Updates SetupModal to accept variable tabs length + design values for Tooltip and simulator page

### DIFF
--- a/packages/slice-machine/components/Simulator/components/FailedConnect/index.tsx
+++ b/packages/slice-machine/components/Simulator/components/FailedConnect/index.tsx
@@ -1,21 +1,41 @@
-import { Heading, Image, Link, Text } from "theme-ui";
+import { Image, Link, Text } from "theme-ui";
 import FullPage from "../FullPage";
 
 const FailedConnect = () => (
   <FullPage>
     <Image src="/iframe-not-running.png" sx={{ width: "320px" }} />
-    <Heading as="h3" sx={{ color: "textClear", mb: 2 }}>
-      Slice Machine can't render your Slice
-    </Heading>
     <Text
-      sx={{ color: "greyIcon", mt: 1, maxWidth: "500px", textAlign: "center" }}
+      sx={{
+        color: "textClear",
+        mb: 2,
+        fontSize: "14px",
+        lineHeight: "24px",
+        fontWeight: "600",
+      }}
     >
-      Ensure your website's development server is running by typing{" "}
-      <Text as="code" variant="styles.inlineCode">
+      Slice Machine can't render your Slice
+    </Text>
+    <Text
+      sx={{
+        color: "#86848D",
+        fontWeight: "400",
+        maxWidth: "400px",
+        textAlign: "center",
+        fontSize: "12px",
+        lineHeight: "22px",
+      }}
+    >
+      Ensure your website's development server is running by typing
+      <br />
+      <Text
+        as="code"
+        variant="styles.inlineCode"
+        sx={{ padding: "4px", borderRadius: "6px" }}
+      >
         npm run dev
       </Text>
-      &nbsp; in your terminal at the root of your website directory. If that
-      doesn't work, see the&nbsp;
+      &nbsp; in your terminal at the root of your website directory.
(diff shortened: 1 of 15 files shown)


packages/slice-machine/components/Simulator/components/FailedConnect/index.tsx
```diff
@@ -1,21 +1,41 @@
...
+        fontWeight: "600",
+      }}
+    >
       Slice Machine can't render your Slice
-    </Heading>
+    </Text>
     <Text
-      sx={{ color: "greyIcon", mt: 1, maxWidth: "500px", textAlign: "center" }}
+      sx={{
+        color: "#86848D",
```
Heyrows: should we take this from theme ui ?

packages/slice-machine/components/Simulator/components/FailedConnect/index.tsx
```diff
@@ -1,21 +1,41 @@
...
+        fontWeight: "400",
```
Heyrows: i think it's the default you need need to pass it

packages/slice-machine/components/Simulator/components/SetupModal/CodeBlock.tsx
```diff
@@ -32,10 +40,15 @@ const CodeBlockWithCopy: React.FC<{
         {...(lang ? { lang, codeClass: "is-json" } : {})}
         sx={{
           width: "100%",
-          height: "100%",
+          minHeight: "100%",
           bg: "codeBlockBackground",
           borderRadius: "6px",
           flexDirection: "column",
+          ...(fullHeightCode
```
Heyrows, line 47: unsure of what that would change, should have been better to pass an `sx` object ?

packages/slice-machine/components/Simulator/components/SetupModal/steps/excerpts.ts
```diff
@@ -0,0 +1,31 @@
+export type Excerpt = {
+  title: string;
+  excerpt: string;
+};
+
+export const InstallExcerpt = {
+  title: "Install the simulator",
+  excerpt: "Run this command to install the simulator package via npm.",
```
Heyrows: should we stick with NPM or offer more like do we have the info if the userr is using NPM or YARN in the frontend ?

packages/slice-machine/components/Simulator/components/FailedConnect/index.tsx
hypervillain: I had a conversation on this topic yesterday and adding the value to the theme would not have a strong effect: we'll drop it or rewrite it, there are already too many values in the theme and the color is not used anywhere else

packages/slice-machine/components/Simulator/components/SetupModal/CodeBlock.tsx
hypervillain: We can't use sx here because it's not a theme-ui component

packages/slice-machine/components/Simulator/components/SetupModal/steps/excerpts.ts
hypervillain: Good question, I'll ask the PM!

hypervillain: Done, code updated

packages/slice-machine/components/Simulator/components/FailedConnect/index.tsx
```diff
@@ -17,8 +17,7 @@ const FailedConnect = () => (
     </Text>
     <Text
       sx={{
-        color: "#86848D",
-        fontWeight: "400",
+        color: "failedConnectText",
```
Heyrows, line 20: I don't think that's the name used in the design system.

Heyrows (approved):

packages/slice-machine/components/Simulator/components/FailedConnect/index.tsx
hypervillain: hmm yeah I called it like that because it's only used there

PR #727, opened 2022-12-08
add new dependency when starting a next project

mathieud:
## Context

https://linear.app/prismic/issue/SM-912/aauser-i-can-install-prismicionext-in-next-projects-when-i-run-the

- [ ] The CI is successful.

### DIFF
--- a/packages/core/src/node-utils/framework.ts
+++ b/packages/core/src/node-utils/framework.ts
@@ -28,15 +28,24 @@ export function fancyName(str: Frameworks): string {
 export function detectFramework(
   pkg: JsonPackage,
   supportedFrameworks: Frameworks[] = SupportedFrameworks
-): Frameworks {
+): FrameworkWithVersion {
   const { dependencies, devDependencies, peerDependencies } = pkg;
   const deps = { ...peerDependencies, ...devDependencies, ...dependencies };
 
   const frameworkEntry: Frameworks | undefined = Object.values(
     supportedFrameworks
   ).find((f) => deps[f] && deps[f].length);
-  return frameworkEntry || Frameworks.vanillajs;
+
+  return {
+    framework: frameworkEntry || Frameworks.vanillajs,
+    version: frameworkEntry ? deps[frameworkEntry] : undefined,
+  };
 }
+
+export type FrameworkWithVersion = {
+  framework: Frameworks;
+  version: string | undefined;
+};
 
 export function defineFramework({
   cwd,
@@ -46,9 +55,9 @@ export function defineFramework({
   cwd: string;
   supportedFrameworks?: Frameworks[];
   manifest?: Manifest;
-}): Frameworks {
+}): FrameworkWithVersion {
   if (manifest?.framework && isFrameworkSupported(manifest.framework))
-    return manifest.framework;
+    return { framework: manifest.framework, version: undefined };
 
   const pkg = retrieveJsonPackage(cwd);
   if (!pkg.exists || !pkg.content) {
--- a/packages/core/src/node-utils/framework.ts
+++ b/packages/core/src/node-utils/framework.ts
@@ -1,5 +1,5 @@
 import { retrieveJsonPackage, JsonPackage } from "./pkg";
-import { Frameworks, SupportedFrameworks } from "../models/Framework";
+import { Frameworks, SupportedFrameworks, Version } from "../models/Framework";
 import { Manifest } from "../models/Manifest";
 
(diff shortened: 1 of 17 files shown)


MarcMcIntosh:
Are you sure you want to target master with this one?

packages/core/src/node-utils/framework.ts
```diff
@@ -1,5 +1,5 @@
 import { retrieveJsonPackage, JsonPackage } from "./pkg";
-import { Frameworks, SupportedFrameworks } from "../models/Framework";
+import { Frameworks, SupportedFrameworks, Version } from "../models/Framework";
```
MarcMcIntosh: do we need this ? 

later on it's 
```
version: Version; 
```
which could be `version?: string`

```
```

packages/core/src/node-utils/framework.ts
```diff
@@ -60,3 +66,33 @@ export function defineFramework({
...
+    const message =
+      "[api/env]: Unrecoverable error. Could not find package.json. Exiting..";
+    console.error(message);
+    throw new Error(message);
+  }
+
+  const { dependencies, devDependencies, peerDependencies } = pkg.content;
+  const deps = { ...peerDependencies, ...devDependencies, ...dependencies };
+
+  const version = deps[framework];
```
MarcMcIntosh, line 92: do we always all the frame work after it's name in the dep?

packages/init/__tests__/utils/versionCheck.test.ts
```diff
@@ -0,0 +1,19 @@
+import { describe, expect, test } from "@jest/globals";
+
+import { checkVersion } from "../../src/utils/versionCheck";
+
+describe("versionCheck", () => {
+  describe("checkVersion", () => {
+    test("checks the version satisfies the threshold", () => {
+      expect(checkVersion(undefined, ">=13.0.0")).toBe(false);
```
MarcMcIntosh: nice

packages/init/src/steps/install-required-dependencies.ts
```diff
@@ -16,12 +17,25 @@ const {
...
+  if (checkVersion(version, ">=13.0.0")) {
+    return PRISMIC_NEXT;
+  }
+
+  return "";
+}
+
+function depsForFramework(
+  framework: Models.Frameworks,
+  version: Models.Version
```
MarcMcIntosh: Probably better to have this as optional :)

packages/init/src/steps/install-required-dependencies.ts
```diff
@@ -35,6 +49,7 @@ function depsForFramework(framework: Models.Frameworks): string {
 
 async function addAndInstallDeps(
   framework: Models.Frameworks,
+  version: Models.Version,
```
MarcMcIntosh: could be optional?

packages/core/src/node-utils/framework.ts
mathieud: That's what I had initially, but it was repeated quite often. I thought this made it clearer. Can change if you prefer

packages/core/src/node-utils/framework.ts
mathieud: Looks like it yes https://github.com/prismicio/slice-machine/blob/master/packages/core/src/node-utils/framework.ts#L38

packages/init/src/steps/install-required-dependencies.ts
mathieud: It's always called with it currently. Are we also worried another package might use it?

packages/core/src/node-utils/framework.ts
MarcMcIntosh: https://github.com/prismicio/slice-machine/blob/master/packages/core/src/models/Framework.ts#L3
....sort of 😅

packages/init/src/steps/install-required-dependencies.ts
MarcMcIntosh: it would be weird to have to call it with `undefined`... I've seen that somewhere in a different code base

packages/core/src/node-utils/framework.ts
MarcMcIntosh: should be fine though :)

packages/core/src/node-utils/framework.ts
MarcMcIntosh: yes, 👍

MarcMcIntosh (approved):

PR #729, opened 2022-12-09
Mock retention

arnaudlewis:
## Context

When saving a slice model, we regenerate a mock from scratch for the all slice.
Now that we have a new experience with the editor, we don’t want people to loose their last changes if they simply add a new field so we change the mocks only where the model has changed and we keep everything else as is.


packages/slice-machine/lib/mock/LegacyMockConfig.ts
```diff
@@ -404,42 +413,44 @@ export function buildNestableMockConfig(
 export function buildWidgetMockConfig(
```
MarcMcIntosh, line 413: what is LegacyMock config?

packages/slice-machine/lib/mock/Slice.ts
```diff
@@ -51,20 +51,38 @@ export function buildSliceMockConfig(
...
     };
   });
 }
 
 export default function MockSlice(
-  smModel: SliceSM,
-  legacyMockConfig: Record<string, Record<string, Record<string, unknown>>> // not sure about this one.
+  sliceModel: SharedSlice,
+  legacyMockConfig: Record<string, Record<string, Record<string, unknown>>>, // not sure about this one.
+  previousMocks?: ComponentMocks | null | undefined,
```
MarcMcIntosh, line 64: does it need to be undefined if it's optional`?`

packages/slice-machine/server/src/api/slices/save.ts
```diff
@@ -36,21 +43,24 @@ export async function handler(
...
-    );
-    Files.writeJson(sliceMockPath(env.cwd, from, sliceName), mocks);
-  }
+  const mocks = mock(
+    sliceModel,
+    SliceMockConfig.getSliceMockConfig(updatedMockConfig, from, sliceName),
+    customMocks,
+    sliceDiff
+  );
+  Files.write(sliceMockPath(env.cwd, from, sliceName), mocks);
```
MarcMcIntosh: Files.writeJson has a nicer (human readable) output.

packages/slice-machine/server/src/api/slices/save.ts
```diff
@@ -36,21 +43,24 @@ export async function handler(
   console.log("\n\n[slice/save]: Updating slice model");
 
   const modelPath = CustomPaths(env.cwd).library(from).slice(sliceName).model();
+  Files.write(modelPath, sliceModel);
```
MarcMcIntosh: Files.writeJson, it has a more human friendly output

packages/slice-machine/lib/mock/Slice.ts
```diff
@@ -51,20 +51,38 @@ export function buildSliceMockConfig(
...
+        sliceModel,
+        variationMock,
+        sc
+      );
+      if (!patched.ok || !patched.result)
+        return SharedSliceMock.generate(sliceModel, sc);
+      return patched.result;
+    }
+
+    return SharedSliceMock.generate(sliceModel, sc);
```
MarcMcIntosh: would it make more sense (less nesting) to have a fefw of these conditions as an early exit?
```
if(!sliceDiff) return SharedSliceMock.generate(sliceModel, sc);
const variationMock = ...
if(!variationMock) return  SharedSliceMock.generate(sliceModel, sc);
const patched = ...
if(patchedNotGood) return SharedSliceMock.generate(sliceModel, sc);
return patched.result
```

MarcMcIntosh: Surprised this didn't break the tests :)

packages/slice-machine/lib/mock/LegacyMockConfig.ts
arnaudlewis: It’s the mock config is it is in SM. the new mock config contract is the one from the mocks library so the code for legacy mock config just transforms the SM format to the lib format.
The code was there already but now we need a type discriminant to makes things easier to match.

packages/slice-machine/lib/mock/Slice.ts
arnaudlewis: Semantically it doesn’t have the same meaning so it can be here or not and if it’s here it can be undefined or not.

PR #732, opened 2022-12-12
Fix/delete local ct only

mathieud:
## Context

Fixes a bug when deleting a CustomType. There was a mismatch between attribute missing and being undefined.

## The Solution

Update the types and code around to match the state given by the server

- [x] The CI is successful.

### DIFF
--- a/packages/slice-machine/components/ReviewModal/index.tsx
+++ b/packages/slice-machine/components/ReviewModal/index.tsx
@@ -27,6 +27,7 @@ import { getEnvironment } from "@src/modules/environment";
 import Tracker from "@src/tracking/client";
 import { selectAllCustomTypes } from "@src/modules/availableCustomTypes";
 import { getLibraries } from "@src/modules/slices";
+import { isLocalCustomType } from "@src/modules/availableCustomTypes/types";
 
 Modal.setAppElement("#__next");
 
@@ -91,10 +92,12 @@ const ReviewModal: React.FunctionComponent = () => {
         }, 0)
       : 0;
 
-  const hasSliceWithinCustomType: boolean = customTypes.some((customType) =>
-    customType.local?.tabs.some(
-      (tab) => tab.sliceZone && tab.sliceZone?.value.length > 0
-    )
+  const hasSliceWithinCustomType = customTypes.some(
+    (customType) =>
+      isLocalCustomType(customType) &&
+      customType.local?.tabs.some(
+        (tab) => tab.sliceZone && tab.sliceZone?.value.length > 0
+      )
   );
 
   const hasPushedAnHourAgo = Boolean(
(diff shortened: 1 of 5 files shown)


packages/slice-machine/src/modules/availableCustomTypes/index.ts
```diff
@@ -208,23 +211,48 @@ export const availableCustomTypesReducer: Reducer<
...
+        return {
+          ...state,
+          [ct.local.id]: deletedCT,
+        };
+      }
+      if (isNewCustomType(ct)) {
+        return omit(state, action.payload.customTypeId);
+      }
+
+      return state;
```
danigo99, line 226: [OPT] Do you think this case deserves a comment?

packages/slice-machine/src/modules/availableCustomTypes/types.ts
```diff
@@ -11,25 +11,29 @@ export type SyncedCustomType = {
...
   ct: FrontEndCustomType
 ): ct is DeletedFrontEndCustomType => !isLocalCustomType(ct);
 
 export const isNewCustomType = (ct: FrontEndCustomType): ct is NewCustomType =>
-  isLocalCustomType(ct) && !("remote" in ct);
+  !("remote" in ct);
+
+export const isSyncedCustomType = (
+  ct: FrontEndCustomType
+): ct is SyncedCustomType => "remote" in ct && "local" in ct;
```
danigo99, line 36: ❓ Is this enough? We don't need to check that local & remote !== undefined?

packages/slice-machine/components/ReviewModal/index.tsx
```diff
@@ -91,10 +92,12 @@ const ReviewModal: React.FunctionComponent = () => {
         }, 0)
       : 0;
 
-  const hasSliceWithinCustomType: boolean = customTypes.some((customType) =>
-    customType.local?.tabs.some(
-      (tab) => tab.sliceZone && tab.sliceZone?.value.length > 0
-    )
+  const hasSliceWithinCustomType: boolean = customTypes.some(
```
danigo99: ❓ Why do we need to explicitly define the type?

danigo99 (approved):

packages/slice-machine/src/modules/availableCustomTypes/index.ts
mathieud: I'd say no, the types are explicit enough

packages/slice-machine/src/modules/availableCustomTypes/types.ts
mathieud: remote being undefined is not a valid use case. It's either present and a custom type, or not defined at all

mathieud: See [here](https://github.com/prismicio/slice-machine/blob/964279b1e08823d215c38c0680569c9cd47e4afd/packages/slice-machine/src/modules/availableCustomTypes/types.ts#L4).
An attribute being undefined is what caused the crash we had before.

packages/slice-machine/components/ReviewModal/index.tsx
mathieud: Didn't notice it. Updated.

danigo99 (approved):

PR #736, opened 2022-12-12
lh/dev plugins init

lihbr:
WIP


packages/core2/src/auth/PrismicAuthManager.ts
```diff
@@ -128,6 +144,64 @@ export class PrismicAuthManager {
...
 	}
 
+	async getLoginSessionInfo(): Promise<PrismicAuthManagerGetLoginSessionInfoReturnType> {
+		// Pick a random port, with a preference for historic `5555`
+		const port = await getPort({ port: 5555 });
+
+		const url = new URL(
+			`/dashboard/cli/login?source=slice-machine&port=${port}`,
+			APIEndpoints.PrismicWroom,
+		).toString();
```
angeloashmore, line 154: What do you think of using `searchParams` instead of string concatenation? It's more type safe and (IMO) easier to read.

```typescript
const url = new URL("/dashboard/cli/login", APIEndpoints.PrismicWroom);
url.searchParams.set("source", "slice-machine")
url.searchParams.set("port", port.toString())

return {
	port,
	url: url.toString(),
}
```

Good either way. Up to you 🙂

packages/core2/src/lib/prismicrc.ts
```diff
@@ -0,0 +1,54 @@
+import * as path from "node:path";
+import * as t from "io-ts";
+import * as rc9 from "rc9";
+
+import { decode } from "./decode";
+
+const PRISMICRC = ".prismicrc";
```
angeloashmore, line 7: Is this a new rc file? I think SM currently uses the `tracking` property in `sm.json` to determine if tracking is enabled.

https://github.com/prismicio/slice-machine/blob/f38c42c7f53deb87d5b57ed8e77bd9fa36f3ac4d/packages/slice-machine/server/src/api/tracking.ts#L48-L49

packages/core2/src/managers/_AnalyticsManager.ts
```diff
@@ -0,0 +1,171 @@
...
+
+		this._enabled = readPrismicrc().telemetry !== false;
+		this._segmentClient = new SegmentClient(Analytics.SegmentKey, {
+			// Since it's a local app, we do not benefit from event batching the way a server would normally do, all tracking event will be awaited.
+			flushAt: 1,
+			// TODO: Verify that this actually does not send data to Segment when false.
+			enable: this._enabled,
+		});
+		this._anonymousID = randomUUID();
+	}
```
angeloashmore, line 95: Since this method is synchronous, we could initialize it in the constructor rather than a dedicated `init*` method.

The other clients need an `init*` method because they are async. However, if you think this method may become async in the future, or may need parameters that don't make sense to accept in the constructor, we can keep it as an init method.

packages/core2/src/managers/_PluginsManager.ts
```diff
@@ -16,4 +20,10 @@ export class PluginsManager extends BaseManager {
 
 		await this.sliceMachinePluginRunner.init();
 	}
+
+	dangerouslyCallHook: SliceMachinePluginRunner["callHook"] = (...args) => {
+		assertPluginsInitialized(this.sliceMachinePluginRunner);
+
+		return this.sliceMachinePluginRunner.callHook(...args);
+	};
```
angeloashmore, line 28: Is this being used anywhere? I feel like this is a dangerous method to include, even considering the `dangerous` warning.

Could we exclude this method or leave it commented out until it's needed? Dedicated method calls for hook calls (e.g. `readSlice()`) is preferable since it encapsulates an intent.

packages/core2/src/managers/_ProjectManager.ts
```diff
@@ -1,4 +1,5 @@
 import * as fs from "node:fs/promises";
+import { existsSync } from "node:fs";
```
angeloashmore, line 2: Any reason we use the sync version instead of the version from `fs/promises`?

packages/core2/src/managers/_ProjectManager.ts
```diff
@@ -52,23 +85,36 @@ export class ProjectManager extends BaseManager {
...
-
-			try {
-				await fs.access(configFilePath);
-				configModule = loadModuleWithJiti(path.resolve(configFileName));
-			} catch {
-				// noop
-			}
+		try {
+			await fs.access(configFilePath);
+			configModule = loadModuleWithJiti(path.resolve(configFilePath));
```
angeloashmore, line 115: Per Côme, we'll be sticking with `sm.json` for now. We'll need to refactor this file to read the JSON file instead of `slicemachine.config.ts`.

This means we can update the repository name with property replacement rather than search/replace.

packages/core2/src/managers/_RepositoryManager.ts
```diff
@@ -1,6 +1,255 @@
...
+				);
+			}
+
+			return repositories;
+		} else {
+			throw new Error(`Failed to read repositories`, { cause: json });
+		}
+	}
+
+	// Should this be in manager? It's one of the few sync method
```
angeloashmore, line 92: Could we add a "TODO:" at the start of the comment if we don't have a decision? It'll be helpful later when we can do a project-wide search for "TODO".

packages/core2/src/managers/_RepositoryManager.ts
```diff
@@ -1,6 +1,255 @@
...
+				cause: res,
+			});
+		}
+	}
+
+	async delete(args: SliceMachineManagerDeleteRepositoryArgs): Promise<void> {
+		const cookies = await this.user.getAuthenticationCookies();
+
+		const url = new URL(
+			`/app/settings/delete?_=${cookies["X_XSRF"]}`, // TODO: Maybe we want to throw early if the token is no available, or get the token another way
```
angeloashmore, line 158: If we update the previously commented URL to use `url.searchParams`, we should make that change again here.

packages/plugin-kit/src/createSliceMachinePluginRunner.ts
```diff
@@ -127,14 +130,11 @@ export class SliceMachinePluginRunner {
...
-				const raw = await import(resolve);
-				plugin = raw.default || raw;
-			} catch {
-				// If the plugin could not be imported, we will
-				// throw a custom error later.
-			}
+			// Import plugin as a file from the project
+			const raw = await createRequire(
+				path.resolve(this._project.root, "noop.js"),
+			)(resolve);
```
angeloashmore, line 136: Did you come across an issue with using `import()`? Can `import()` load CJS?

angeloashmore (approved):
Looks good! I had a couple of stylistic comments and some implementation questions.

Feel free to change whatever you feel works best and then merge. 🙂

packages/core2/src/lib/prismicrc.ts
lihbr: Yes, I think people should be able to set a global preference (as an individual, I don't want to be tracked) in the home folder, while also being able to set a local preference (as an organization, I don't want this project to be tracked) I'll update it in that fashion.

- I know we have a `~/.prismic` file there already and those should eventually be merged into one.
- I don't know about the syntax, on one hand, we should likely stick with the JSON syntax for compatibility, on another hand I don't like that the file name does not git any hint about its syntax.
- I feel like the `rc` suffix/syntax is a bit more common than non-`rc` suffixed/syntax files

To be continued~

packages/core2/src/managers/_AnalyticsManager.ts
lihbr: Will revisit yes :)

packages/core2/src/managers/_PluginsManager.ts
lihbr: I needed it in the init command to call the `command:init` hook, I didn't see it fit anywhere else 🤔

packages/core2/src/managers/_ProjectManager.ts
lihbr: There's no `fp/promises.exists` method for some reason (I guess a concurrency issue).

Here I used it over `exists` over `access` as I just care if the file exists (I'm not tampering it). Not sure if a read `access` check is equivalent, maybe 🤔

There's a comment reminding that here: https://github.com/prismicio/slice-machine/pull/736/files/3ea2b70c0bc4f1ea5174c1261eb98db2c4905265#diff-0f22f2f93929168f23ba549d2270c80ee15c84b2f456ba8c77fa3ea726a2d0f2R76-R77

packages/core2/src/managers/_ProjectManager.ts
lihbr: Could we at least support a `slicemachine.config.json` for new projects? `sm.json` feels suboptimal for a name :/

packages/core2/src/managers/_RepositoryManager.ts
lihbr: Sure!

packages/plugin-kit/src/createSliceMachinePluginRunner.ts
lihbr: The issue I had was that it was trying to load the adapter from this package rather than from the test project root. We might just want to use that as a fallback 🤔

lihbr:
OK, definitely some valid comments and ideas. Merging it anyway for now, we'll be able to get back to this PR and #700 when reviewing the dev branch, onwards~

packages/core2/src/lib/prismicrc.ts
angeloashmore: `~/.prismic` is a state persistence file, while `~/.prismicrc` is a user configuration file. They probably should be separate files since `~/.prismic` will be updated programmatically.

Unfortunately, their locations and names are confusing since they are so similar.

Alternatives:

- Use the `~/.config/prismic` directory, which follows the standard `XDG_CONFIG_HOME` location on Unix-like systems. I'm not sure if Windows has a similar convention.
- Name `~/.prismic` something different, like `.prismic-session`

packages/core2/src/managers/_PluginsManager.ts
angeloashmore: Hmm, a dedicated `triggerCommandInitHook()` or something like it would be weird.

We could get the plugin runner itself and make the call outside the managers:

```ts
const pluginRunner = manager.getSliceMachinePluginRunner();

await pluginRunner.callHook("command:init");
```

This signals explicitly working outside the manager's API and dropping down to the lower-level plugin runner.

---

Do types work properly when using `dangerouslyCallHook()`? I thought it required a generic for it to infer the hook type and its arguments.

packages/core2/src/managers/_ProjectManager.ts
angeloashmore: Ooh, right, makes sense

packages/core2/src/managers/_ProjectManager.ts
angeloashmore: I like that. It's clearer than `sm.json` and sets us up for supporting the `.js`/`.ts` version.

packages/plugin-kit/src/createSliceMachinePluginRunner.ts
angeloashmore: Oh right, `require()` lets you specify its resolution paths while `import()` doesn't (yet). I've been linking `@slicemachine/adapter-next` into `@slicemachine/plugin-kit` during development, but that's not a good solution.

Okay cool, let's leave this here. We may run into CJS vs ESM issues in the future since I assume this always loads CJS instead, but we can tackle that issue when the time comes.

PR #738, opened 2022-12-13
Add drawer for delete docs over limit

danigo99:
## Context

Ticket: https://linear.app/prismic/issue/SM-887/[delete-documents-attached]-aauser-if-i-have-documents-database-limit

## The Solution

Have implemented the design for the modal and have both open when we push changes that include a CT who's name starts with "Drawer".

## Impact / Dependencies

Since this is still waiting for the transactional push work to be completed, I just have the modal open by default when we push changes that include a CT who's name starts with "Drawer".
This will need to be reverted when the remaining work is done.

- [x] The CI is successful.

## [OPT] Preview

image

---

image


packages/slice-machine/components/DeleteDocumentsDrawer/AssociatedDocumentsCard.tsx
```diff
@@ -0,0 +1,54 @@
+import React from "react";
+import { Flex, Text, Card, Link } from "theme-ui";
+
+type CustomtypeDocuments = {
```
mathieud, line 4: We'll need to share these types with the backend when we implement it

packages/slice-machine/components/DeleteDocumentsDrawer/AssociatedDocumentsCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+        {ctDocuments.numberOfDocuments} documents
+      </Text>
+    </Flex>
+    <Link
+      href={ctDocuments.link}
+      target="_blank"
+      variant="cardSmall"
+      sx={{ color: isOverLimit ? "danger" : "purple08" }}
+    >
+      View documents
```
mathieud, line 51: Wondering if we should make the entire card clickable

packages/slice-machine/components/DeleteDocumentsDrawer/DeleteDocumentsDrawerOverLimit.tsx
```diff
@@ -0,0 +1,127 @@
...
+          </Flex>
+        )}
+      >
+        <Text sx={{ fontWeight: "bold", mb: 1, lineHeight: "24px" }}>
+          Your Custom Type{AssociatedDocuments.length > 1 && "s"} cannot be
+          deleted.
+        </Text>
+        <Text sx={{ mb: 24 }}>
+          {AssociatedDocuments.length > 1
+            ? "These Custom Types have"
```
mathieud, line 113: Don't also need a space here?

packages/slice-machine/components/DeleteDocumentsDrawer/AssociatedDocumentsCard.tsx
danigo99: Yep, the type will also likely change since we don't get the CT name from the API

packages/slice-machine/components/DeleteDocumentsDrawer/DeleteDocumentsDrawerOverLimit.tsx
danigo99: No, it's added after with both

packages/slice-machine/components/DeleteDocumentsDrawer/AssociatedDocumentsCard.tsx
danigo99: This doesn't seem to be the design. I'm not sure from a usability perspective tho? 🤔

packages/slice-machine/components/DeleteDocumentsDrawer/DeleteDocumentsDrawerOverLimit.tsx
mathieud: Read too quickly

mathieud (approved):

PR #752, opened 2022-12-14
Fix: Switch variation on simulator

arnaudlewis:
## Context

Since the merge with Simulator v2, there is a regression where we get the mock for the default-slice without any consideration of the selected variation in the UI.
We fixed that bug by reintroducing the proper code done on `dev-mocks` the original code of the integration of the editor.


packages/slice-machine/components/Simulator/index.tsx
```diff
@@ -114,17 +114,17 @@ const Simulator: ComponentWithSliceProps = ({ slice, variation }) => {
   const sharedSlice = useMemo(() => Slices.fromSM(slice.model), [slice.model]);
 
   const initialContent = useMemo<SharedSliceContent>(
-    () => slice.mock?.[0] || defaultSharedSliceContent(variation.id),
+    () =>
+      slice.mock?.find((m) => m.variation === variation.id) ||
```
MarcMcIntosh, line 118: Is this the fix?

packages/slice-machine/components/Simulator/index.tsx
```diff
@@ -114,17 +114,17 @@ const Simulator: ComponentWithSliceProps = ({ slice, variation }) => {
...
 
-  const previousInitialContent = useRef(initialContent);
   const [editorContent, setContent] = useState(initialContent);
 
-  if (previousInitialContent.current !== initialContent) {
-    previousInitialContent.current = initialContent;
+  useEffect(() => {
     setContent(initialContent);
-  }
+  }, [variation.id, initialContent]);
```
MarcMcIntosh, line 127: hmmm,
so if variaion.id or initial content change, set change the ...editor content, got it :) (was worried that it might setting the initial content)

MarcMcIntosh (approved):

packages/slice-machine/components/Simulator/index.tsx
arnaudlewis: Yea it’s only the editor. We keep track of initial content and api for the purpose of keeping slice I’d, hope we can simplify in the future

PR #753, opened 2022-12-14
refactor(core2): organize managers

lihbr:
WIP


packages/core2/src/constants/index.ts
```diff
@@ -0,0 +1,12 @@
...
+export { SLICE_MACHINE_USER_AGENT, APIEndpoints, APITokens } from "./api";
+
+export {
+	SLICE_MACHINE_NPM_PACKAGE_NAME,
+	SLICE_MACHINE_CONFIG_JS,
+	SLICE_MACHINE_CONFIG_TS,
+	SLICE_MACHINE_CONFIG_FILENAMES,
+	TS_CONFIG_FILENAME,
+	DEFAULT_SLICE_SCREENSHOT_URL,
+} from "./project";
```
angeloashmore: Rather than group constants by category (it's too subjective and could lead to strange organization), we can:

- Create one file per constant, removing categories altogether. If constants are very similar (e.g. `SLICE_MACHINE_CONFIG_JS` and `_TS`), that could be a single obj-as-const enum. Or…
- Retain a single `constants.ts` file. It's simple. It may get unwieldy eventually.

Any other ideas? I feel the categorization here (`ApplicationMode.ts`, `api.ts`, `project.ts`) is unnecessary and adds overhead.

angeloashmore: Barrel files (i.e. `index.ts`) can cause messiness in other files. Autocomplete might import constants from the individual source files (e.g. `constants/api.ts`), while importing from the barrel file in other times.

In my opinion, it's cleaner to remove `index.ts` and import from the source files directly. That ensures there's only one way to import code.

packages/core2/src/constants/project.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Slice Machine's published NPM package name.
+ */
+export const SLICE_MACHINE_NPM_PACKAGE_NAME = "slice-machine-ui";
+
+/**
+ * Slice Machine configuration filename.
+ */
```
angeloashmore: This TSDoc would only apply to the first constant. If we want to comment in the file to group code, we should use `//` comments instead.

packages/core2/src/errors.ts
```diff
@@ -47,7 +47,7 @@ import { HookError } from "@slicemachine/plugin-kit";
 // 	cause?: Error;
 // };
 
-export class SliceMachineError extends Error {
+export abstract class SliceMachineError extends Error {
```
angeloashmore, line 50: Nice catch!

packages/core2/src/managers/BaseManager.ts
```diff
@@ -46,18 +51,23 @@ export abstract class BaseManager {
 	protected get user(): UserManager {
 		return this._sliceMachineManager.user;
 	}
-	protected get project(): ProjectManager {
-		return this._sliceMachineManager.project;
+	protected get repositoryAPI(): RepositoryAPIManager {
```
angeloashmore: "RepositoryAPI" doesn't sound right. It couples the manager to the HTTP API's endpoints rather than the more general idea of managing the repository. Anything to do with repositories should go in there, not just actions supported by the API.

`repository` and `RepositoryManager` sounds better to me.

packages/core2/src/managers/repositoryAPI/RepositoryAPIManager.ts
```diff
@@ -92,9 +71,9 @@ export class RepositoryManager extends BaseManager {
 	// Should this be in manager? It's one of the few sync method
 	hasWriteAccess(repository: PrismicRepository): boolean {
 		switch (repository.role) {
-			case PrismicRepositoryRoles.SuperUser:
-			case PrismicRepositoryRoles.Owner:
-			case PrismicRepositoryRoles.Administrator:
+			case PrismicRepositoryRole.SuperUser:
```
angeloashmore: 👍 Consistency 🌈

angeloashmore (commented):
Left some comments re: organization. Here are some high-level comments:

- I think we can move `auth` into the `managers` directory. It was written outside `managers` to be reusable in other contexts (specifically `start-slicemachine` and `init`), but we can probably refactor that integration to be simpler.
- `rpc` can be moved out into its own top-level package (maybe outside the monorepo).
- One directory per manager makes more sense than grouping them IMO (as seen in `project`). I think that will scale better as it becomes clear how it add new managers and where to write co-located files (like `types.ts`).

This is looking great. I think we're headed in the right direction here. 😁

packages/core2/src/constants/index.ts
lihbr: Let's go back to individual file

packages/core2/src/managers/BaseManager.ts
lihbr: Will rename `prismicRepository`

PR #754, opened 2022-12-15
Make E2E tests independent, add E2E helpers for repetitive actions.

Heyrows:
## Context

[Linear Ticket](https://linear.app/prismic/issue/SM-908/[spike]-[e2e-test]-screenshotv2-features)

## The Solution

I've been extracting commands for simple actions such as reseting the project custom types and slices.
These are executed most of the time on a `BeforeEach` hook.

- [ ] The CI is successful.

## [OPT] Preview

mikhail-vasilyev-NodtnCsLdTE-unsplash

### DIFF
--- a/cypress/consts.js
+++ b/cypress/consts.js
@@ -1,0 +1,13 @@
+export const root = "e2e-projects/cypress-next-app";
+
+export const packageJsonFile = `${root}/package.json`;
+export const manifestFile = `${root}/sm.json`;
+
+export const assetsFolder = `${root}/.slicemachine`;
+export const typesFile = `${assetsFolder}/prismicio.d.ts`;
+export const sliceMockFile = (sliceName) => `${assetsFolder}/assets/slices/${sliceName}/mocks.json`;
+export const libraryStateFile = `${assetsFolder}/libraries-state.json`;
+
+export const customTypesFolder = `${root}/customtypes`;
+export const slicesFolder = `${root}/slices`;
+
--- a/cypress/consts.js
+++ b/cypress/consts.js
@@ -1,13 +1,13 @@
-export const root = "e2e-projects/cypress-next-app";
+export const ROOT = "e2e-projects/cypress-next-app";
 
-export const packageJsonFile = `${root}/package.json`;
-export const manifestFile = `${root}/sm.json`;
+export const PACKAGE_JSON_FILE = `${ROOT}/package.json`;
+export const MANIFEST_FILE = `${ROOT}/sm.json`;
 
-export const assetsFolder = `${root}/.slicemachine`;
-export const typesFile = `${assetsFolder}/prismicio.d.ts`;
-export const sliceMockFile = (sliceName) =>
-  `${assetsFolder}/assets/slices/${sliceName}/mocks.json`;
-export const libraryStateFile = `${assetsFolder}/libraries-state.json`;
+export const ASSETS_FOLDER = `${ROOT}/.slicemachine`;
+export const TYPES_FILE = `${ASSETS_FOLDER}/prismicio.d.ts`;
+export const SLICE_MOCK_FILE = (sliceName) =>
+  `${ASSETS_FOLDER}/assets/slices/${sliceName}/mocks.json`;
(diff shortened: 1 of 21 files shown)


cypress/consts.js
```diff
@@ -0,0 +1,13 @@
+export const root = "e2e-projects/cypress-next-app";
+
+export const packageJsonFile = `${root}/package.json`;
+export const manifestFile = `${root}/sm.json`;
+
+export const assetsFolder = `${root}/.slicemachine`;
+export const typesFile = `${assetsFolder}/prismicio.d.ts`;
+export const sliceMockFile = (sliceName) =>
```
MarcMcIntosh: this will need to be updated after the mocks branch is merged in

cypress/consts.js
```diff
@@ -0,0 +1,13 @@
+export const root = "e2e-projects/cypress-next-app";
+
+export const packageJsonFile = `${root}/package.json`;
```
MarcMcIntosh: tbh: I prefer sting constants to be name in UPPER_SNAKE_CASE..... I'm not sure if it'll catch on

cypress/e2e/customTypes/00-create.cy.js
```diff
@@ -1,64 +1,33 @@
...
+    cy.renameCustomType(customTypeId, customTypeName, "New Custom Type Name");
   });
 
   it("generates types if `generateTypes` is `true` and `@prismicio/types` is installed", () => {
     // Stub manifest.
-    cy.readFile(manifest).then((manifestContents) => {
-      cy.writeFile(manifest, { ...manifestContents, generateTypes: true });
+    cy.readFile(manifestFile).then((manifestContents) => {
+      cy.writeFile(manifestFile, { ...manifestContents, generateTypes: true });
     });
```
danigo99: OPT: This seems like a common pattern, wondering if we should create a helper for it? I.e. adding info to json files. Would just be weird to type I guess

MarcMcIntosh (commented):
Left a few comments, I think either the mock branch or this one will needed to updated when the either are merged

cypress/helpers/localStorage.js
```diff
@@ -0,0 +1,38 @@
+import "cypress-localstorage-commands";
+
+export function setSliceMachineUserContext({
+  hasSendAReview, // boolean
+  isOnboarded, // boolean
+  updatesViewed, // object
+  hasSeenTutorialsTooTip, // boolean
+}) {
```
danigo99: Can we make these all optional with defaults so we only overload the field we want to change? Then spread it below

cypress/helpers/localStorage.js
```diff
@@ -0,0 +1,38 @@
...
+        },
+        hasSeenTutorialsTooTip,
+      }),
+    })
+  );
+}
+
+export function cleanSliceMachineUserContext(
+  hasSendAReview = true,
+  isOnboarded = true
```
mdeclercq: Maybe these 2 params can be removed since they don't seem to be used.

cypress/support/commands.js
```diff
@@ -1,42 +1,42 @@
...
   "cleanSliceMachineUserContext",
-  (hasSendAReview = true, isOnboarded = true) => {
-    return cy.removeLocalStorage("persist:root");
-  }
+  cleanSliceMachineUserContext
 );
+Cypress.Commands.add("getSliceMachineUserContext", getSliceMachineUserContext);
+
+/* -- PROJECT RESET -- */
+Cypress.Commands.add("clearProject", clearProject);
```
mdeclercq, line 30: Nice abstractions, it will really help the test readability 👍
Too bad Cypress doesn't offer a way to scope commands to organise them in a better way.

cypress/helpers/localStorage.js
Heyrows: I personally prefer to state all the values as I think it makes it clearer to understand from the test itself what happens.
But you and previously Hugo seems to agree that it would be better with default values 👍.

I'll modify it 😄

MarcMcIntosh (approved):

danigo99 (approved):

PR #756, opened 2022-12-15
Fix/iframe retry

hypervillain:
- remove retry logic for iframe renderer
- useCallback for useEffect dependency in order to not call it on each render


cypress/e2e/slices/00-create.cy.js
```diff
@@ -72,6 +72,32 @@ describe("Create Slices", () => {
 
     cy.get("[data-cy=rename-slice-modal]").should("not.exist");
 
+    // add a variation
+
+    const getInputByLabel = (label) =>
```
bapmrl, line 77: 💡 We should probably move this to `support` (or the extraneous `helpers`) directory to avoid duplicating it in the future.

cypress/e2e/slices/00-create.cy.js
```diff
@@ -72,6 +72,32 @@ describe("Create Slices", () => {
...
+      cy
+        .contains("label", label)
+        .invoke("attr", "for")
+        .then((id) => cy.get("#" + id));
+
+    cy.get("button").contains("Default").click();
+    cy.contains("+ Add new variation").click();
+    getInputByLabel("Variation name*").type("foo");
+    getInputByLabel("Variation ID*").clear().type("bar");
+    cy.get("#variation-add").submit();
```
bapmrl, line 87: ❓ Is there a way to submit the `form` without having to resort on a technical identifier (and instead, using an interactive label like real users would do)? This would help us follow [best practices in E2E tests](https://testing-library.com/docs/queries/about#priority).

cypress/e2e/slices/00-create.cy.js
```diff
@@ -98,17 +124,39 @@ describe("Create Slices", () => {
 
     cy.get("[data-testid=simulator-open-button]").click();
 
-    cy.get("[contenteditable]").first().clear().type("👋");
+    cy.get("[contenteditable]").first().clear();
```
bapmrl, line 127: ⚠️ `[contenteditable]` is an implementation detail. Would the `getInputByLabel` function work in this case?

cypress/e2e/slices/00-create.cy.js
```diff
@@ -98,17 +124,39 @@ describe("Create Slices", () => {
...
+    cy.contains("foo").click();
+
+    cy.wait(1000);
+
+    cy.get("[contenteditable]").first().clear();
+    cy.get("[contenteditable]").first().type("🎉");
+
+    cy.get("[data-cy=save-mock]").click();
+
+    cy.wait(1000);
```
bapmrl, line 144: ❓ Is there a way to avoid explicit calls to `cy.wait` (because they are listed as an ["anti-pattern"](https://docs.cypress.io/api/commands/wait#Time) in the Cypress documentation)? For instance, `cy.get("button")` has a default timeout so is this specific `cy.wait` really necessary?

packages/slice-machine/components/Simulator/index.tsx
```diff
@@ -292,6 +268,7 @@ const Simulator: ComponentWithSliceProps = ({ slice, variation }) => {
               sx={{
                 height: "100%",
                 overflowY: "scroll",
+                width: "800px",
```
bapmrl: ❓ Do we want such a wide editor? Does this value come from Figma?

packages/slice-machine/tests/pages/simulator.spec.tsx
```diff
@@ -47,13 +57,41 @@ describe("simulator", () => {
     // });
   });
 
+  document.createRange = () => {
+    // https://github.com/jsdom/jsdom/issues/3002
```
bapmrl, line 61: ❓ Would it be better to specify this `createRange` function in a `setupTests` file?

packages/slice-machine/components/Simulator/index.tsx
danigo99: Yeah I agree this seems strange actually. Mathieu recently fixed an issue with the width of the sidebar and Benjamin indicated it should be 496 (like other sidebars in the app)

cypress/e2e/slices/00-create.cy.js
MarcMcIntosh: Done in another PR

cypress/e2e/slices/00-create.cy.js
MarcMcIntosh: Proably, but I'm not sure what al forms are on screen at once

cypress/e2e/slices/00-create.cy.js
MarcMcIntosh: maybe

cypress/e2e/slices/00-create.cy.js
MarcMcIntosh: waiting for it to save

packages/slice-machine/tests/pages/simulator.spec.tsx
MarcMcIntosh: depends if it's in a dom env or server env?

bapmrl: We could test if `document` has the `createRange` property and then, if it doesn't, add this function 🙂

MarcMcIntosh: jsdom doesn't provide one :)

cypress/e2e/slices/00-create.cy.js
bapmrl: Resolved in https://github.com/prismicio/slice-machine/pull/772.

PR #757, opened 2022-12-15
Improve editor state management

arnaudlewis:


packages/slice-machine/src/modules/selectedSlice/reducer.ts
```diff
@@ -189,6 +190,16 @@ export const selectedSliceReducer: Reducer<
...
       return renamedComponentUI(prevState, newSliceName);
     }
+
+    case getType(updateSelectedSliceMocks): {
+      if (!prevState) return prevState;
+      const { mocks } = action.payload;
+      return {
+        ...prevState,
+        mock: mocks,
+        screenshots: prevState?.screenshots,
```
danigo99: ❓ Do we need this? Looks like it can be omitted

danigo99 (approved):

packages/slice-machine/src/modules/selectedSlice/reducer.ts
MarcMcIntosh: I think so :)

PR #762, opened 2022-12-16
refactor: sort imports with `@trivago/prettier-plugin-sort-imports`

angeloashmore:
This PR adds `@trivago/prettier-plugin-sort-imports` to the Prettier config only in the `@slicemachine/manager` package.

The plugin is configured to sort imports in the following order:

- `node:*`
- Third party packages
- Local dependencies, grouped and sorted by relative distance (i.e. `../../../`, `../../`, `../`)

Local dependencies per relative directory distance is grouped by the followng directories:

- `lib`
- `constants`
- `types`
- `errors`
- Everything else

See one of the formatted files for an example: https://github.com/prismicio/slice-machine/blob/f6df7ca05cf6393843a44d8e045a4bf1e2c93727/packages/manager/src/managers/slices/SlicesManager.ts

## TODO

`.prettierrc` is messy because it needs to explicitly list each import group. We could probably be clever with the regexs and clean up some of the groups. We could also use `.prettierrc.cjs` and build the list programmatically.

Before spending time cleaning up the config, we have to decide if this plugin is worth using.


lihbr (commented):
I like it, but maybe we can use an ESLint alternative instead: https://www.npmjs.com/package/eslint-plugin-simple-import-sort

I'd prefer to lean more on the end linter (ESLint) and keep Prettier as vanilla as possible.

bapmrl:
I agree with @lihbr. We're using `eslint-plugin-simple-import-sort` in `prismicio/editor` and it works well.

angeloashmore:
Perfect, let's try that one out here then! I'll close this PR and open a new one when we're ready.

PR #764, opened 2022-12-16
test(cypress): onboarding

bapmrl:
## Context

Linear issue: [SM-920 [e2e test] Onboarding user experience](https://linear.app/prismic/issue/SM-920/[e2e-test]-onboarding-user-experience).

## The Solution

1. Add `cypress/e2e/user-flows/onboarding.cy.js` and remove deprecated `cypress/e2e/onboarding` tests.
2. Add custom `is.playing` assertion (to ensure videos are automatically played).
3. Use `display: none` instead of the `hidden` attribute in `packages/slice-machine/pages/onboarding.tsx` so that Cypress doesn't find multiple visible `video` elements when running the `video:visible` query.
4. Add `role="tooltip"` in `packages/slice-machine/components/AppLayout/Navigation/Menu/Navigation/VideoItem.tsx` to make it easily queryable.
5. Remove unused `data-cy` attributes.
6. Remove unused `cleanSliceMachineUserContext` helper.

- [x] The CI is successful.


cypress/e2e/user-flows/onboarding.cy.js
```diff
@@ -0,0 +1,54 @@
+describe("onboarding", () => {
+  beforeEach(() => {
+    cy.setSliceMachineUserContext({
+      hasSeenTutorialsTooTip: false,
+      isOnboarded: false,
+    });
```
mdeclercq: I wonder if we should set the localStorage for these tests. Having an empty localStorage instead of `false` values could be more representative of a user who never logged in to SM. WDYT ?

bapmrl: I've removed that, thanks! 😊

mdeclercq (approved):
LGTM, the `isPlaying` assertion is a nice addition too

PR #765, opened 2022-12-16
call changelog only when the app starts

danigo99:
## Context

Ticket: https://linear.app/prismic/issue/SM-916/[spike]-aauser-i-should-always-be-able-to-access-the-sm-changelog

## The Solution

Remove the change log call from the general getEnv function.
Create an API route which returns the change log from github
Call this to set the CL in the redux store on start or refresh

- [ ] The CI is successful.

## [OPT] Preview

---

image

### DIFF
--- a/packages/slice-machine/src/modules/environment/index.ts
+++ b/packages/slice-machine/src/modules/environment/index.ts
@@ -1,6 +1,11 @@
 import { Reducer } from "redux";
 import { EnvironmentStoreType } from "./types";
-import { ActionType, createAction, getType } from "typesafe-actions";
+import {
+  ActionType,
+  createAction,
+  createAsyncAction,
+  getType,
+} from "typesafe-actions";
 import { SliceMachineStoreType } from "@src/redux/type";
 import { FrontEndEnvironment } from "@models/common/Environment";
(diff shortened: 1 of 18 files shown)


packages/slice-machine/src/modules/environment/index.ts
```diff
@@ -55,10 +79,15 @@ export const selectIsSimulatorAvailableForFramework = (
   return simulatorIsSupported(store.environment.framework);
 };
 
-export const getChangelog = (
-  store: SliceMachineStoreType
-): PackageChangelog => {
-  return store.environment.changelog;
+export const getChangelog = (store: SliceMachineStoreType) => {
+  return (
+    store.environment?.changelog ?? {
```
MarcMcIntosh: more conditionals have been added?

MarcMcIntosh (commented):
not sure if we want to merge this one master or `dev-alpha`

packages/slice-machine/src/modules/environment/index.ts
danigo99: This was because the changelog will now be undefined on first load, so we just return a default _empty_ one to not throw errors. 

I think the optional chaining on environment might not be needed though 😕 will test it now

PR #766, opened 2022-12-16
Rename event

AlexGalays:
- Renaming the Editor Widget Used event to match the existing naming convention.
- Add the sliceId as event props so we can find out which slices were simulated but never edited.


Heyrows (changes requested):
PR is good, but the event on segment is not ok.
- The event doesn't seems to be in the library event
- The event doesn't seems to be up to date on the tracking plan

Heyrows (approved):

PR #771, opened 2022-12-19
Fetch changelog only on app restart

danigo99:
## Context

Ticket: https://linear.app/prismic/issue/SM-916/[spike]-aauser-i-should-always-be-able-to-access-the-sm-changelog

## The Solution

Remove the change log call from the general getEnv function.
Create an API route which returns the change log from github
Call this to set the CL in the redux store on start or refresh

- [ ] The CI is successful.

## [OPT] Preview

---

image


packages/slice-machine/lib/models/common/Environment.ts
```diff
@@ -23,7 +22,7 @@ export interface FrontEndEnvironment {
   intercomHash?: string;
   manifest: Models.Manifest;
   repo: string;
-  changelog: PackageChangelog;
+  changelog?: PackageChangelog;
```
MarcMcIntosh, line 25: how come it's optional now?

MarcMcIntosh (approved):

packages/slice-machine/lib/models/common/Environment.ts
danigo99: Because the change log will be undefined on first render. We need to wait for the API call to the changelog/ endpoint for it to be set.

PR #776, opened 2022-12-20
fix: iframe scaling

bapmrl:
## Context

Corresponding Linear issue: [SM-946 Slice Simulator v2 - App preview not scrollable](https://linear.app/prismic/issue/SM-946/slice-simulator-v2-app-preview-not-scrollable).

## The Solution

1. Correctly scale the iframe based on the requested screen dimensions.
2. Introduce the `useElementSize` hook and call it in `packages/slice-machine/components/Header/index.tsx`.
3. Mock `ResizeObserver` in `packages/slice-machine/tests/jest-setup.js`.
4. Remove unused `@hooks/*` alias from `packages/slice-machine/tsconfig.json`.

- [x] The CI is successful.

### DIFF
--- a/packages/slice-machine/components/Simulator/components/IframeRenderer/index.tsx
+++ b/packages/slice-machine/components/Simulator/components/IframeRenderer/index.tsx
@@ -115,7 +115,7 @@ const IframeRenderer: React.FunctionComponent<IframeRendererProps> = ({
         display: "flex",
         height: "100%",
         justifyContent: "center",
-        overflowY: "hidden",
+        overflow: "hidden",
         ...(dryRun ? { display: "none" } : {}),
       }}
     >
--- a/packages/slice-machine/components/Simulator/components/IframeRenderer/index.tsx
+++ b/packages/slice-machine/components/Simulator/components/IframeRenderer/index.tsx
@@ -93,10 +93,7 @@ const IframeRenderer: React.FunctionComponent<IframeRendererProps> = ({
       });
   }, [client, apiContent, simulatorUrl]);
 
-  const [viewportSize, setViewportSize] = useState<ScreenDimensions>({
-    height: 0,
-    width: 0,
-  });
+  const [viewportSize, setViewportSize] = useState<ScreenDimensions>();
   const viewportRef = useElementSize(({ blockSize, inlineSize }) => {
     setViewportSize({ height: blockSize, width: inlineSize });
   }, []);
@@ -131,7 +128,9 @@ const IframeRenderer: React.FunctionComponent<IframeRendererProps> = ({
             minHeight: `${iframeSize.height}px`,
             minWidth: `${iframeSize.width}px`,
             overflowY: "auto",
-            transform: `scale(${getScaling(iframeSize, viewportSize)})`,
+            ...(viewportSize
+              ? { transform: `scale(${getScaling(iframeSize, viewportSize)})` }
+              : { display: "none" }),
           }}
         />
       ) : null}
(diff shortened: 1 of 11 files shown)


benjammartin (commented):
✅

packages/slice-machine/components/Simulator/components/IframeRenderer/index.tsx
```diff
@@ -95,63 +93,66 @@ const IframeRenderer: React.FunctionComponent<IframeRendererProps> = ({
...
-      </Flex>
-    </Box>
+        />
+      ) : null}
+      {client?.connected ? <div id="__iframe-ready" /> : null}
+    </Flex>
   );
 };
 
+function getScaling(
```
Heyrows: I think this deserve some unit test
The logic isn't that simple.

packages/slice-machine/components/Simulator/components/IframeRenderer/index.tsx
```diff
@@ -95,63 +93,66 @@ const IframeRenderer: React.FunctionComponent<IframeRendererProps> = ({
...
+  { height: iframeHeight, width: iframeWidth }: ScreenDimensions,
+  { height: viewportHeight, width: viewportWidth }: ScreenDimensions
+): number {
+  const widthDiff = iframeWidth - viewportWidth;
+  const heightDiff = iframeHeight - viewportHeight;
+  if (iframeWidth > viewportWidth && widthDiff > heightDiff) {
```
Heyrows: What if both conditions are met (the wanted height AND width is bigger than the box)
Why do we scale only by one of them ?

Heyrows (approved):

packages/slice-machine/components/Simulator/components/IframeRenderer/index.tsx
bapmrl: I'll soon push a unit test 🙂

packages/slice-machine/components/Simulator/components/IframeRenderer/index.tsx
bapmrl: 1. The same scaling factor is applied to both dimensions as we want to preserve the initial aspect ratio.
2. The scaling factor to apply depends on which dimension overflows the most (either the height or the width).

Tell me if my explanation isn't clear 🙂

Heyrows: It is crystal clear with explanations. 👍 
The fact that is very clear with few explanations is an excellent argument for tests / comments

packages/slice-machine/components/Simulator/components/IframeRenderer/index.tsx
```diff
@@ -131,7 +128,9 @@ const IframeRenderer: React.FunctionComponent<IframeRendererProps> = ({
             minHeight: `${iframeSize.height}px`,
             minWidth: `${iframeSize.width}px`,
             overflowY: "auto",
-            transform: `scale(${getScaling(iframeSize, viewportSize)})`,
+            ...(viewportSize
+              ? { transform: `scale(${getScaling(iframeSize, viewportSize)})` }
+              : { display: "none" }),
```
Heyrows, line 133: Wouldn't display none make the Iframe takes no space ?
Which would make the screen jump a little bit ?

bapmrl: I don't think so because the topmost `Flex` has a `height: "100%"` and the Editor panel has a fixed width. I'll test that after lunch :)

Heyrows: hmmm 🤔
Ok, just wanted to make sure because unless the Flex says the elements should align at the end, I thought it would jump as there would only be the editor centered in the middle of the Div.

Heyrows (approved):

danigo99 (approved):

PR #779, opened 2022-12-21
WIP fix

danigo99:
- [ ] The CI is successful.

### DIFF
--- a/packages/slice-machine/server/src/api/screenshots/puppeteer.ts
+++ b/packages/slice-machine/server/src/api/screenshots/puppeteer.ts
@@ -58,12 +58,9 @@ const generateScreenshot = async (
   try {
     Files.mkdir(path.dirname(pathToFile), { recursive: true });
 
-    await page.goto(
-      `${screenshotUrl}?screenWidth=${screenDimensions.width}&screenHeight=${screenDimensions.height}`,
-      {
-        waitUntil: "load",
-      }
-    );
+    await page.goto(screenshotUrl, {
+      waitUntil: "load",
+    });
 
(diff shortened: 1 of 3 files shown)


packages/slice-machine/server/src/api/screenshots/puppeteer.ts
```diff
@@ -58,9 +58,12 @@ const generateScreenshot = async (
   try {
     Files.mkdir(path.dirname(pathToFile), { recursive: true });
 
-    await page.goto(screenshotUrl, {
-      waitUntil: "load",
-    });
+    await page.goto(
+      `${screenshotUrl}?screenWidth=${screenDimensions.width}&screenHeight=${screenDimensions.height}`,
```
Heyrows: puppeteer shouldn't have to know  how to compute this url.
I think you should look where `screenshotUrl` was computed first and assign the properties at that point.

Heyrows (changes requested):

packages/slice-machine/server/src/api/screenshots/puppeteer.ts
Heyrows: To me, i think there is a difference between
- Puppeteer needs to set his viewport so he can take the good screenshot
- The data to access is already with good dimensions (making the url with the good properties).

One could be used without the other.

mathieud: We already set the viewport (https://github.com/prismicio/slice-machine/blob/dev-alpha/packages/slice-machine/server/src/api/screenshots/puppeteer.ts#L53), not sure why it does not work anymore.

Heyrows (approved):

PR #781, opened 2022-12-21
E2E Scenario 2: Add a slice to a custom type and fill the review modal

Heyrows:
## Context

[Linear ticket](https://linear.app/prismic/issue/SM-921/[e2e-test]-custom-type-slice-association)

- [ ] The CI is successful.


cypress/e2e/user-flows/scenario_slice_association.cy.js
```diff
@@ -0,0 +1,121 @@
...
+
+  it("Add the Slice to the Custom Type", () => {
+    cy.visit(`/cts/${customTypeId}`);
+
+    cy.get("[data-cy=update-slices]").click();
+    // forcing this because the input itself is invisible and an svg is displayed
+    cy.get(`[data-cy=check-${sliceId}]`).click({ force: true });
+    cy.get("[data-cy=update-slices-modal]").submit();
+
+    cy.saveSliceModifications();
```
danigo99: Why is this saveSlice if it's in the custom type builder?

cypress/e2e/user-flows/scenario_slice_association.cy.js
```diff
@@ -0,0 +1,121 @@
+const random = Date.now();
+
+const customTypeName = `My Test ${random}`;
+const customTypeId = `my_test_${random}`;
+
+const sliceName = `TestSlice${random}`;
+const sliceId = `test_slice${random}`; // generated automatically from the slice name
+const sliceLib = "slices";
+
+describe("I am an existing SM user (Next) and I want to associate a Slice to a CT and review my experience.", () => {
```
danigo99, line 10: Are we sure that all the tests in the suite always run in order? I thought one test scenario would go in one `it()`

cypress/helpers/slices.js
```diff
@@ -46,9 +50,63 @@ export function renameSlice(lib, actualName, newName) {
...
   cy.get('[data-cy="slice-and-variation-name-header"]').contains(
     `/ ${newName} / Default`
   );
   cy.get("[data-cy=rename-slice-modal]").should("not.exist");
+  cy.readFile(SLICE_MODEL(newName)).then((model) => {
+    expect(JSON.stringify(model)).to.contain(newName);
+  });
+}
+
+export function addStaticFieldToSlice(fieldType, fieldName, fieldId) {
```
danigo99, line 63: I really like these helpers 😄

cypress/e2e/user-flows/scenario_slice_association.cy.js
Heyrows: good catch haha.
In the end saving a slice of a custom type does the same thing as they use the same button in the builder.

Funny that I used the wrong one but it still worked

cypress/e2e/user-flows/scenario_slice_association.cy.js
Heyrows: https://jestjs.io/docs/setup-teardown#order-of-execution

I believe inside the describe itself, the `it` will run one by one waiting for the the previous one to finish.

Heyrows: One scenario could go in one `it`, but then it might be tougher to trace back issues as we have bigger steps in the scenario itself.

I'd be ok to change it but I guess we should do it for all the existing scenarios.

cypress/e2e/customTypes/01-duplicate.cy.js
```diff
@@ -10,7 +10,6 @@ describe("Duplicate custom types", () => {
 
   it("when using a label that is already in use it should warn the user", () => {
     cy.visit("/");
-    cy.waitUntil(() => cy.get("[data-cy=create-ct]")).then(() => true);
```
mathieud, line 13: Do you know why these were there originally? Any chance it will make the tests flaky?

cypress/helpers/customTypes.js
```diff
@@ -33,30 +32,49 @@ export function renameCustomType(id, actualName, newName) {
...
+    cy.get(`[data-cy=check-${sliceId}]`).click({ force: true });
+  });
+
+  cy.get("[data-cy=update-slices-modal]").submit();
+}
+
+export function saveCustomTypeModifications() {
+  cy.get("[data-cy=builder-save-button]").should("not.be.disabled");
+  cy.get("[data-cy=builder-save-button]").click();
+  cy.get("[data-cy=builder-save-button-spinner]").should("be.visible");
```
mathieud, line 106: Should we split actions and checks? Looking at the function name it's not easy to know what it does

cypress/e2e/user-flows/scenario_slice_association.cy.js
```diff
@@ -0,0 +1,121 @@
...
+
+  it("Create a Custom type with multiple fields", () => {
+    cy.createCustomType(customTypeId, customTypeName);
+
+    cy.addFieldToCustomType("UID", "ID Field", "uid");
+    cy.addFieldToCustomType("Key Text", "Key Text Field", "key_text_id");
+    cy.addFieldToCustomType("Rich Text", "Rich Text Field", "rich_text_id");
+    cy.saveCustomTypeModifications();
+
+    cy.reload();
```
mathieud, line 27: Do we reload to check the save worked as expected? 
We've had a. bug recently where the server state was updated correctly but the browser state wasn't. I wonder which one we'd want to test more here (also users usually don't reload)

cypress/helpers/customTypes.js
```diff
@@ -33,30 +32,49 @@ export function renameCustomType(id, actualName, newName) {
...
-  fieldName,
-  fieldId,
-  isFirstField /* boolean | undefined */
-) {
-  const addFieldButtonSelector = isFirstField
-    ? "empty-zone-add-new-field"
-    : "add-new-field";
-
-  cy.get(`[data-cy="${addFieldButtonSelector}"]`).first().click();
+export function addFieldToCustomType(fieldType, fieldName, fieldId) {
```
mathieud, line 60: Are there restriction to when this can be used? ie needs to be on a specific page, or some seed data? If so would be nice to have then explained in a comment

cypress/support/commands.js
```diff
@@ -14,8 +14,17 @@ import {
   createCustomType,
   renameCustomType,
   addFieldToCustomType,
+  addSlicesToCustomType,
+  saveCustomTypeModifications,
```
mathieud: Can we (should we?) import and load all the exported functions automatically with a loop?

packages/slice-machine/lib/builders/CustomTypeBuilder/SliceZone/UpdateSliceZoneModal.tsx
```diff
@@ -44,6 +44,7 @@ const UpdateSliceZoneModal: React.FC<UpdateSliceModalProps> = ({
         title: "Update Slice Zone",
       }}
       omitFooter={!projectHasAvailableSlices}
+      dataCy="update-slices-modal"
```
mathieud, line 47: Is there a reason we use `data-cy` everywhere but here?

Heyrows: `dataCy` is a property name so that it use `data-cy` on the actual element.

cypress/support/commands.js
Heyrows: I've never done that before but it looks like it would be nice.

cypress/e2e/user-flows/scenario_slice_association.cy.js
Heyrows: Indeed we reload to see if the state persisted. This is what is used in the scenarios.
I guess it makes more sense because if changes are persisted, the app still has a way of running.

I don't know if we'd be able to test the Redux state or if we'd want to do that.

cypress/e2e/customTypes/01-duplicate.cy.js
Heyrows: They were here originally for the spinner to finish loading.
I removed them because I didn't see any issues or flakyness while developing my scenario.

cypress/helpers/customTypes.js
Heyrows: We could definitely split actions and checks but I'd expect the button not to be clicked if it isn't clickable here for example.

cypress/helpers/customTypes.js
Heyrows: It needs to be done from the builder directly.

I didn't added the navigation and save actions because we usually adds more than 1 field before saving like a user would do.

cypress/e2e/user-flows/scenario_slice_association.cy.js
mdeclercq: I agree that reloading is a good way to make sure the persisted data can be read/understood by the app. 
I don't think the e2e tests should check the Redux state since it's an implementation detail.

mathieud (approved):

PR #783, opened 2022-12-22
Update latest version in changelog page when the versions change

danigo99:
## Context

Issue in ticket: https://linear.app/prismic/issue/SM-916#comment-e223c409
When navigating directly to /changelog, the list in the column updates but the latest version displayed on the right does not as this was wrapped in a usestate. Not we update the state when the changelog changes.

- [ ] The CI is successful.

## [OPT] Preview

ezgif com-gif-maker (8)

### DIFF
--- a/packages/slice-machine/components/Changelog/index.tsx
+++ b/packages/slice-machine/components/Changelog/index.tsx
@@ -15,13 +15,15 @@ export default function Changelog() {
     })
   );
 
+  const latestVersion = changelog.versions[0] || null;
+
   const [selectedVersion, setSelectedVersion] = useState<PackageVersion | null>(
-    changelog.versions[0] || null
+    latestVersion
   );
 
   useEffect(() => {
-    setSelectedVersion(changelog.versions[0]);
-  }, [JSON.stringify(changelog)]);
+    setSelectedVersion(latestVersion);
+  }, [latestVersion]);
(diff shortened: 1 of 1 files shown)


packages/slice-machine/components/Changelog/index.tsx
```diff
@@ -15,11 +15,14 @@ export default function Changelog() {
...
   );
 
-  // Null is when no version are found (edge case)
   const [selectedVersion, setSelectedVersion] = useState<PackageVersion | null>(
     changelog.versions[0] || null
   );
 
+  useEffect(() => {
+    setSelectedVersion(changelog.versions[0]);
+  }, [JSON.stringify(changelog)]);
```
Heyrows: why the json stringify ?

danigo99: We can't just use changelog because the shallow comparison doesn't see them as equal. So causes a infinite renders. But it can compare 2 strings 😕

danigo99:
The last commit is optional, just a small improvement to show the loading spinner when we're waiting for the changelog to fetch


https://user-images.githubusercontent.com/47107427/209132771-9619baab-eea0-45be-9420-49b809e8748b.mov

Heyrows (approved):

PR #792, opened 2023-01-03
fix(slice-machine-ui): add source maps to .npmignore

bapmrl:
## Context

Corresponding Linear issue: [SM-958 (SPIKE 1Day)slice-machine-ui package size doubled between 0.5.2 and 0.6.0](https://linear.app/prismic/issue/SM-958/spike-1dayslice-machine-ui-package-size-doubled-between-052-and-060).

## The Solution

Add `.map` files located in `packages/slice-machine/build` and `packages/slice-machine/out` to `packages/slice-machine/.npmignore`.

## Impact / Dependencies

N/A

- [x] The CI is successful.

### DIFF
--- a/packages/slice-machine/.npmignore
+++ b/packages/slice-machine/.npmignore
@@ -26,3 +26,7 @@ tests/project/node_modules
 
 # prov
 /api-dump
+
+# source maps
+/build/**/*.map
+/out/**/*.map


packages/slice-machine/.npmignore
```diff
@@ -26,3 +26,7 @@ tests/project/node_modules
 
 # prov
 /api-dump
+
+# source maps
+/build/**/*.map
```
Heyrows, line 31: hey, i'm not sure but maybe source map have been added for Sentry ?
Do you know if Sentry still works if we don't have the source maps ? Or is it maybe that source maps are uploaded when creating a Sentry release ?

bapmrl: Hello,

Yes, source maps are automatically generated and uploaded by the `@sentry/nextjs` and `@sentry/rollup-plugin` packages when we build `slice-machine-ui`. However, once uploaded, the generated source maps are not deleted (and there's no easy way to do that) and get published to npm. That's why I've decided to add them to the `.npmignore` file.

Heyrows: Ok, 👌 

I don't know much about how source map works so one last question.
If an error is triggered in Sentry, will it still be able to link the error with the source map ?

bapmrl: Yes, as long as we've uploaded the source maps to Sentry 😊 Consumers don't need them (and they're [hidden](https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/#use-hidden-source-map) anyway).

Heyrows: Thanks

Heyrows (approved):

PR #794, opened 2023-01-04
Add multi-tab for code in SetupModal

hypervillain:
[Linear ticket](https://linear.app/prismic/issue/SM-917/aauser-i-can-see-multitab-to-code-block-in-setupmodal-in-slice)

- Add tabs for npm and yarn in Setup modal
- Optimize tabs to prevent unnecessary reload

To test it:
- Remove prop `localSimulatorURL` from `sm.json`
- Access the simulator
- Check the first Setup tab
- alternate between yarn and npm: both code and copy should match
- the other code blocks should still match previous design

How it looks
Capture d’écran 2023-01-03 à 17 29 00


packages/slice-machine/components/Simulator/components/SetupModal/index.tsx
```diff
@@ -88,7 +89,30 @@ const SetupModal: React.FC<{ isOpen: boolean }> = ({ isOpen }) => {
...
+          >
+            <Step
+              stepNumber={i + 1}
+              linkToStorybookDocs={linkToStorybookDocs}
+              linkToTroubleshootingDocs={linkToTroubleshootingDocs}
+            />
+          </TabPanel>
+        );
+      }),
+    [setupData]
```
mathieud: Could you add the missing dependencies?

packages/slice-machine/components/Simulator/components/SetupModal/VersionBadgeList.tsx
```diff
@@ -0,0 +1,54 @@
...
+const VersionBadgeList: FC<{
+  versions: string[];
+  selectedVersion: string;
+  setSelectedVersion: (s: string) => void;
+}> = ({ versions, selectedVersion, setSelectedVersion }) => (
+  <Flex sx={{ maxWidth: "fit-content", p: "16px 16px 0 16px" }}>
+    {versions.map((version, i) => (
+      <VersionBadge
+        key={version}
+        sx={versions[i + 1] ? { mr: "12px" } : undefined}
```
mathieud: We could have this on the `Flex` parent, avoiding the condition

```
'& > *': {
  mr: "12px",
}
```

packages/slice-machine/components/Simulator/components/SetupModal/CodeBlock.tsx
```diff
@@ -5,9 +5,16 @@ import { MdCheck, MdContentCopy } from "react-icons/md";
...
+import VersionBadgeList from "./VersionBadgeList";
+
+type Code = {
+  text: string;
+  version: string;
+};
 
 const CodeBlockWithCopy: React.FC<{
-  children: string;
+  children?: string;
```
mathieud: I'm not a fan of having two possible source. I'd rather have `children: string | Code[]` and deal with it in the component. It feels cleaner, makes the interface simpler and easier to understand and user errors less likely (here could be called with neither children nor codelist)

mathieud (commented):
Do we need the changes to the `package.lock`?

packages/slice-machine/components/Simulator/components/SetupModal/CodeBlock.tsx
hypervillain: Yeah me either. But I'm not a fan of children being anything but a node. I'll find something else

packages/slice-machine/components/Simulator/components/SetupModal/index.tsx
hypervillain: ✅

packages/slice-machine/components/Simulator/components/SetupModal/VersionBadgeList.tsx
hypervillain: ✅

packages/slice-machine/components/Simulator/components/SetupModal/CodeBlock.tsx
hypervillain: Would it be ok for you if I removed `children` and only kept `code`, of type `Code | Code[]` ?

hypervillain: https://github.com/prismicio/slice-machine/pull/794/commits/88613f6af5ae95fc842bdbe565e46716e77ee802#diff-9d2f70d013cc7c10e6d41982d7fdd59904c6c12627482ffd9a9ae8546146024bR16

I updated, LMK!

mathieud: Yep really nice, exactly what I had in mind

packages/slice-machine/components/Simulator/components/SetupModal/CodeBlock.tsx
```diff
@@ -62,43 +65,58 @@ const CodeBlockWithCopy: React.FC<{
...
+              >
+                Copy&nbsp;
+                {isCopied ? (
+                  <MdCheck size={16} color={String(theme?.colors?.success)} />
+                ) : (
+                  <MdContentCopy size={16} style={{ marginLeft: "8px" }} />
+                )}
+              </Button>
+            </Flex>
+            {codeList?.length && codeList.length > 1 ? (
```
mathieud: nit: `codeList` is always defined

packages/slice-machine/components/Simulator/components/SetupModal/VersionBadgeList.tsx
```diff
@@ -0,0 +1,61 @@
...
+  <Flex
+    sx={{
+      maxWidth: "fit-content",
+      p: "16px 16px 0 16px",
+      "& > button:not(:last-child)": {
+        mr: "12px",
+      },
+    }}
+  >
+    {versions.map((version, i) => (
```
mathieud: `i` is not used anymore

PR #805, opened 2023-01-13
ci: use matrix strategy and track `win32` platform

lihbr:
## Matrix Strategy

This PR speeds up CI feedback by using a per-package matrix strategy, this allows most packages to get feedback on their tests in under 4 minutes (best case 3m, worst case 6m). Today, to get this CI feedback, the best case is 4 minutes, the worst one is 10 minutes.

The matrix strategy also makes the CI simpler to read, with each step being compartmentalized rather than all packages being grouped together (which sometimes messed up the logs in a confusing way).

## Windows Suite

This PR also adds a Windows suite to run unit tests on `win32` (non-POSIX) platform. This helps to ensure the software we're building will work seamlessly for even more users, and as implementing this PR is already helped us catch and fix a few unknown bugs 🎉

The Windows suite is, without surprise, slower than the Linux suite, so we can eventually make it unrequired to complete the release process. However, I think keeping it is important so we can always be sure we're doing our best efforts to support every user.

---

While the matrix strategy makes the CI faster, the Windows suite slows it down a bit. Overall the CI is as fast as the current one, but feedback is much faster to appear and organization makes it simpler to navigate.

Other changes:
- Uses `npm ci` to install dependencies more quickly (npm will just follow the lock file instead of resolving it)
- Type check code with `tsc --noEmit` (there are some gotchas on a few packages so for now it just continues on errors)
- Actions were maintained to move away from deprecated and soon-to-be-removed APIs (cache & node v2 to v3)
- Build essentials build artifacts are cached on a per-run basis so that it's only built once for all packages~ (this also allows the process to fail early because most of the tests rely on essential packages being built)
- Node version has been downgraded from 18 to 16 (the soon-to-be latest LTS) to ensure we're testing among the less capable API our users could have

image


.github/workflows/test-and-deploy-alpha.yml
```diff
@@ -3,27 +3,96 @@ name: test-and-deploy-alpha
...
+
+      - name: Set up Node
+        uses: ./.github/actions/setup-node
+
       - name: Install dependencies
-        run: npm i
-      - name: List all dependencies
-        shell: bash
-        run: npm list
+        run: npm ci --legacy-peer-deps
```
mdeclercq: Since the package `better-npm-audit` is used for auditing, would it make sense to use the `--no-audit` option here? 
I wonder if it makes it faster

lihbr: Probably yes, we could add that :)

.github/workflows/test-and-deploy-alpha.yml
```diff
@@ -3,27 +3,96 @@ name: test-and-deploy-alpha
...
+        uses: actions/checkout@v2
+
+      - name: Cache node_modules and artifacts
+        uses: ./.github/actions/restore-cache
+
+      - name: Set up Node
+        uses: ./.github/actions/setup-node
+
+      - name: Install dependencies
+        run: npm ci --legacy-peer-deps
```
MarcMcIntosh: does it need the `--legacy-peer-deps` flag?

.github/workflows/test-and-deploy-alpha.yml
```diff
@@ -33,67 +102,60 @@ jobs:
...
+
+      - name: Restore cache
+        uses: ./.github/actions/restore-cache
+
+      - name: Set up Node
+        uses: ./.github/actions/setup-node
+
+      - name: Unit
+        run: npm run unit --workspace ${{ matrix.package }}
+        env:
```
MarcMcIntosh: do the unit tests nee env vars?

packages/slice-machine/test/core/libraries/libraries.test.ts
```diff
@@ -133,7 +134,9 @@ test("it finds non local libs", () => {
 
   const result = libraries(TMP, [libName]);
   expect(result[0].isLocal).toEqual(false);
-  expect(result[0].components[0].pathToSlice).toEqual(`${libName}/slices`);
+  expect(result[0].components[0].pathToSlice).toEqual(
+    path.join(`${libName}/slices`)
```
MarcMcIntosh: shouldn't this fail on windows because of the `/`?

.github/workflows/test-and-deploy-alpha.yml
lihbr: We had to add it as it appeared some dependencies of `slice-machine-ui` don't (officially?) support React 19, making the install fail otherwise.

.github/workflows/test-and-deploy-alpha.yml
lihbr: I have no clue to be honest, this is just taken from the current workflow: https://github.com/prismicio/slice-machine/blob/master/.github/workflows/test-and-deploy-alpha.yml#L27-L36

packages/slice-machine/test/core/libraries/libraries.test.ts
lihbr: POSIX path delimiters (`/`) are interpreted correctly by Windows (`win32` platforms) in the case of path resolution here and Node imports, so I assume it's alright.
In any case, this test is not relevant anymore as it's now the responsibility of plugins to perform such resolutions, and likely to be removed in #812

.github/actions/build-essential/action.yml
```diff
@@ -1,7 +1,7 @@
-name: 'Build essential packages'
-description: 'Build essential packages'
+name: Build essential packages
+description: Build essential packages
 runs:
-  using: 'composite'
+  using: composite
   steps:
     - name: Build essential packages used by others
       shell: bash
```
MarcMcIntosh, line 7: is bash an option on windows?

lihbr: I guess so since it passed, however I don't think it has much utility here so we could remove it, let me know

.github/workflows/test-and-deploy-alpha.yml
angeloashmore: `@reach/menu-button` was giving me trouble without `--legacy-peer-deps`. It only officially supports React 17, so you'll get a peer deps error from npm without the flag.

.github/workflows/test-and-deploy-alpha.yml
lihbr: Done

mdeclercq (approved):
Great work, it's nice to have added the test execution on Windows 👏 
LGTM

MarcMcIntosh (approved):

PR #808, opened 2023-01-17
Sm 963/mocks with variations

MarcMcIntosh:
## Context
ticket: https://linear.app/prismic/issue/SM-963/%5Bspike%5D-adding-new-slice-variation-resets-editor-mocks-of-the-existing

Adding new Slice variation resets editor mocks of the existing variations

Scenario

Create a Slice 

Open the Simulator, set mocks on the Default variation and save them

Got to the slice page, create a new slice variation, save

Refresh the simulator

Actual behaviour

Mocks are re-generated on the Default variation

Expected behaviour

Mocks of the Default variation are preserved

## The Solution

return the variation mock :)

- [x] The CI is successful.

## [OPT] Preview

82060872_598821160679421_6173575182179455260_n

### DIFF
--- a/packages/slice-machine/lib/mock/Slice.ts
+++ b/packages/slice-machine/lib/mock/Slice.ts
@@ -65,13 +65,16 @@ export default function MockSlice(
   sliceDiff?: SliceDiff | undefined
 ): ComponentMocks {
   const sliceMockConfig = buildSliceMockConfig(sliceModel, legacyMockConfig);
+
   return sliceMockConfig.map((sc) => {
     if (!sliceDiff) return SharedSliceMock.generate(sliceModel, sc);
 
     const variationMock = previousMocks?.find(
       (m) => m.variation === sc.variation
     );
-    if (!variationMock) return SharedSliceMock.generate(sliceModel, sc);
+    if (!variationMock) {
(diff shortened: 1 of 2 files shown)


packages/slice-machine/lib/mock/Slice.ts
```diff
@@ -65,22 +65,29 @@ export default function MockSlice(
...
       variationMock,
       sc
     );
-    if (!patched.ok || !patched.result)
-      return SharedSliceMock.generate(sliceModel, sc);
+
+    if (!patched.ok || !patched.result) {
+      // odd case here where patched is false with the error "Error: The model of the content with variation default has not changed.", but here it would be regenerated
+      // return SharedSliceMock.generate(sliceModel, sc);
+      return variationMock;
```
mathieud, line 87: Are we mixing 2 cases here?
- something actually bad happened when generating the new mock data, and we should probably regenerate from scratch
- there is no diff and we don't want to change anything (and could we detect this just after line 70 and return early?)

MarcMcIntosh: I could split them up :)

mathieud (approved):

PR #810, opened 2023-01-17
Add e2e tests covering screenshots from the slice simulator

danigo99:
## Context

Ticket: https://linear.app/prismic/issue/SM-990/[e2e-test]-slice-simulator-screenshots

- [ ] The CI is successful.

## [OPT] Preview

image

### DIFF
--- a/cypress/e2e/slices/02-simulator-screenshots.cy.js
+++ b/cypress/e2e/slices/02-simulator-screenshots.cy.js
@@ -1,0 +1,83 @@
+import { SlicePage } from "../../pages/slices/slicePage";
+import { SimulatorPage } from "../../pages/simulator/simulatorPage";
+import { ScreenshotModal } from "../../pages/simulator/screenshotModal";
+
+describe("I am an existing SM user and I want to take a screenshot from the slice simulator", () => {
+  const random = Date.now();
+
+  const slice = {
+    id: `test_screenshots${random}`,
+    name: `TestScreenshots${random}`,
+    library: "slices",
+    newVariationName: "foo",
+  };
+
+  const slicePage = new SlicePage();
+  const simulatorPage = new SimulatorPage();
+  const screenshotModal = new ScreenshotModal();
+
+  before("Cleanup local data and create a new slice", () => {
+    cy.clearProject();
+    cy.setSliceMachineUserContext({});
+    cy.createSlice(slice.library, slice.id, slice.name);
+
+    // add a variation
+    cy.get("button").contains("Default").click();
+    cy.contains("+ Add new variation").click();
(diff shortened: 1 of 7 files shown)


cypress/e2e/slices/02-simulator-screenshots.cy.js
```diff
@@ -0,0 +1,83 @@
...
+  const simulatorPage = new SimulatorPage();
+  const screenshotModal = new ScreenshotModal();
+
+  before("Cleanup local data and create a new slice", () => {
+    cy.clearProject();
+    cy.setSliceMachineUserContext({});
+    cy.createSlice(slice.library, slice.id, slice.name);
+
+    // add a variation
+    cy.get("button").contains("Default").click();
```
mdeclercq: In case you didn't notice there is a `addVariationToSlice()` helper in https://github.com/prismicio/slice-machine/blob/master/cypress/helpers/slices.js#L120 
But I see that in your version you input a different variation id so it might suit your use case

cypress/pages/simulator/simulatorPage.js
```diff
@@ -0,0 +1,129 @@
...
+    return this;
+  }
+
+  /**
+   * Change the slice variations using the variations dropdown in the simulator.
+   *
+   * @param {string} targetVariation the new variation value.
+   */
+  changeVariations(targetVariation) {
+    cy.get("button").contains("Default").click();
```
mdeclercq: it could be nice to have a way to identify the variation selector without using the variation name otherwise this would only work when the selected variation is `Default`. Maybe an aria-label or something. 
But it's something we could introduce the day we need it.

cypress/pages/simulator/simulatorPage.js
```diff
@@ -0,0 +1,129 @@
...
+      .should("have.css", "minHeight")
+      .and("eq", `${expectedHeight}px`);
+
+    return this;
+  }
+
+  takeAScreenshotAndOpenModal() {
+    this.takeScreenshotButton.click();
+
+    cy.contains("Screenshot taken", { timeout: 30000 }).should("be.visible");
```
mdeclercq, line 111: out of curiosity, was the screenshot feature slow when running the tests locally ? I'm asking because i'm surprised to see that the test is relatively fast when running in the ci.

danigo99: This actually only happened when I had left the test server running for a long time between tests. Any time I ran it from fresh it would take a normal amount of time (around 5s). But it does seem fast on CI!

mdeclercq (approved):
Nicely done, LGTM

PR #819, opened 2023-01-24
implement e2e scenario 008

mathieud:
## Context

https://linear.app/prismic/issue/SM-989/[e2e-test]-slice-simulator-editor-mocks

- [ ] The CI is successful.


cypress/e2e/user-flows/scenario_008.cy.js
```diff
@@ -0,0 +1,110 @@
...
+      .addNewWidgetField("SimpleTextField", "Key Text")
+      .addNewWidgetField("RichTextField", "Rich Text")
+      .addNewWidgetField("NumberField", "Number")
+      .addNewWidgetField("BooleanField", "Boolean")
+      .addNewWidgetField("SelectField", "Select")
+      .addNewWidgetField("ImageField", "Image")
+      .save();
+
+    // force mock value for the boolean and select fields
+    cy.modifyFile(SLICE_MOCK_FILE(SLICE.name), (mock) =>
```
MarcMcIntosh: oh I understand this issue now, you could check it against the codec ?

cypress/e2e/user-flows/scenario_008.cy.js
```diff
@@ -0,0 +1,108 @@
...
+    cy.contains("Scenario008 • SecondVariation").should("be.visible");
+
+    // editor # change field values
+    editor.type("SimpleTextField", "SimpleTextContent");
+    editor.type("RichTextField", "RichTextContent");
+    editor.type("NumberField", "42", "have.value");
+    editor.toggleBooleanField("BooleanField");
+    editor.select("SelectField", "2");
+
+    editor.changeImage("ImageField").then((newImageSrc) => {
```
mdeclercq: If you want to avoid having to nest the rest of the test just to access `newImageSrc`, you could use an alias instead:
```js
    editor.changeImage("ImageField").invoke('attr', 'src').as('newImageSrc');
```
Then write the rest of the test without nesting it, and using the alias only when you need it:
```js
    cy.get("@newImageSrc").then(newImageSrc => {
      cy.get("[alt='An ananas maybe']")
        .invoke("attr", "src")
        .should("contain", newImageSrc.substr(0, newImageSrc.indexOf("?") + 1));
    })
```
For that to work, you would also have to update the `changeImage()` method to remove the last line `.then(([img]) => img.src)`. It would actually be less specific in case other tests want to access another attribute.

It's just an idea, I'm not sure which approach is more readable.

mathieud: Nice, it worked! And it does look nicer

mdeclercq: 👌

cypress/e2e/user-flows/scenario_008.cy.js
```diff
@@ -0,0 +1,109 @@
+import { SimulatorPage } from "../../pages/simulator/simulatorPage";
```
Heyrows: I think we're trying not to name our files that way.
Maybe `scenario_08_edit_mock_from_simulator` or something like that owuld be cool.

Have some kind of convention. (this could also be a topic for E2E tech debt 😄 )

cypress/e2e/user-flows/scenario_008.cy.js
```diff
@@ -0,0 +1,109 @@
...
+  let slice = new SlicePage();
+  let simulator = new SimulatorPage();
+  let editor = new EditorPage();
+
+  beforeEach(() => {
+    cy.clearProject();
+    cy.setSliceMachineUserContext({});
+  });
+
+  it("creates slice and tests the editor mocks", () => {
```
Heyrows: are we sure we want one IT only ? or one per action in the scenario ?

Heyrows: The name of the action should be clearer.
What exactly are you testing ? the display ? a specific screen or functionality ?

cypress/e2e/user-flows/scenario_008.cy.js
Heyrows: That is a really good idea to verify our fixtures are always up to date, make a unit test or special test to verify the fixture are always validated.

cypress/e2e/user-flows/scenario_008.cy.js
```diff
@@ -0,0 +1,109 @@
...
+              },
+            }
+      )
+    );
+
+    simulator.setup();
+    slice.openSimulator();
+
+    // Wait for the editor to be fully loaded
+    cy.contains("label > span", "Title").should("be.visible");
```
Heyrows: is there some page object for the editor as well ?

Heyrows: Like "editor.contains(title)"

cypress/pages/simulator/editorPage.js
```diff
@@ -0,0 +1,56 @@
+export class EditorPage {
```
Heyrows, line 1: is there a root element to the editor we could use ?
To limit and fasten the search of elements maybe.

cypress/pages/slices/slicePage.js
```diff
@@ -49,4 +49,41 @@ export class SlicePage {
...
+    return this;
+  }
+
+  /**
+   * add a new widget through the modal form.
+   *
+   * @param {string} name Name of the widget field.
+   * @param {string} type Type of the widget.
+   */
+  addNewWidgetField(name, type) {
```
Heyrows, line 77: is this different from `cypress/helpers/slice` ?

cypress/e2e/user-flows/scenario_008.cy.js
mathieud: The problem separating in multiple `it` sections is that the page has to be reloaded between each block, which is quite slow currently, and not what the user would do. Open for discussion though

cypress/pages/slices/slicePage.js
mathieud: It replaces it

cypress/e2e/user-flows/scenario_008.cy.js
Heyrows: I opened the discussion because i'm not clear on what we want to do for this actually 😅 
I think @mdeclercq  would be able to answer what's the best practice for us

cypress/e2e/user-flows/scenario_008_edit_mock_from_simulator.cy.js
```diff
@@ -58,7 +58,7 @@ describe("Scenario 008", () => {
     slice.openSimulator();
 
     // Wait for the editor to be fully loaded
-    cy.contains("label > span", "Title").should("be.visible");
+    editor.contains("Title").should("be.visible");
```
Heyrows, line 61: editor.root.contains would have worked also, to avoid creating contains for each page objects

Heyrows (approved):

cypress/e2e/user-flows/scenario_008.cy.js
mdeclercq: I like the idea of splitting the scenario into multiple steps to make it easier to understand but I don't think it's the right approach for e2e testing with Cypress.
They actually discourage having multiple `it()` tests relying on each other, see https://docs.cypress.io/guides/references/best-practices#Having-tests-rely-on-the-state-of-previous-tests
@mathieud, good point on page reload slowness, I experienced that too :)

MarcMcIntosh (approved):

PR #822, opened 2023-01-26
prevent rerender loop in screenshot

hypervillain:
Fixes https://linear.app/prismic/issue/SM-983/[performance]-screenshot-page-renders-continuously-causing-high-cpu

### DIFF
--- a/packages/slice-machine/pages/[lib]/[sliceName]/[variation]/screenshot.tsx
+++ b/packages/slice-machine/pages/[lib]/[sliceName]/[variation]/screenshot.tsx
@@ -29,7 +29,7 @@ const Screenshot: ComponentWithSliceProps = ({ slice, variation }) => {
     variationID: variation.id,
   });
 
-  if (!simulatorUrl) {
+  if (simulatorUrl === undefined) {
     return null;
   }
 
--- a/packages/slice-machine/src/hooks/useEditorContent.ts
+++ b/packages/slice-machine/src/hooks/useEditorContent.ts
@@ -1,3 +1,4 @@
+import { useMemo } from "react";
 import { ComponentUI } from "@lib/models/common/ComponentUI";
 import { defaultSharedSliceContent } from "@src/utils/editor";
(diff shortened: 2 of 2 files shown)


packages/slice-machine/src/hooks/useEditorContent.ts
```diff
@@ -15,12 +16,12 @@ function useEditorContentOnce({
     slice.mock?.find((m) => m.variation === variationID) ||
     defaultSharedSliceContent(variationID);
 
-  const apiContent = {
+  const apiContent = useMemo(() => ({
     ...(renderSliceMock(Slices.fromSM(slice.model), editorContent) as object),
     id: slice.model.id,
-  };
+  }), []);
```
mathieud: Does adding the dependencies here trigger the bug?

hypervillain: you're right! Updated

mathieud (approved):

packages/slice-machine/pages/[lib]/[sliceName]/[variation]/screenshot.tsx
```diff
@@ -29,7 +29,7 @@ const Screenshot: ComponentWithSliceProps = ({ slice, variation }) => {
     variationID: variation.id,
   });
 
-  if (!simulatorUrl) {
+  if (simulatorUrl === undefined) {
```
Comment (MarcMcIntosh) [inline packages/slice-machine/pages/[lib]/[sliceName]/[variation]/screenshot.tsx:32]: can simulator url be an other falsely value?

MarcMcIntosh (approved):
Just one question :)

MarcMcIntosh (changes requested):
Which branch should this one target?

packages/slice-machine/pages/[lib]/[sliceName]/[variation]/screenshot.tsx
Comment (hypervillain) [reply packages/slice-machine/pages/[lib]/[sliceName]/[variation]/screenshot.tsx:32]: no but Typescript complains if you use a non-boolean value in conditions

PR #824, opened 2023-01-26
Rename stories in renamed slice folder

hypervillain:
Fixes https://linear.app/prismic/issue/SM-1000/aauser-i-want-to-be-able-to-use-storybook-after-i-rename-a-slice

### DIFF
--- a/packages/slice-machine/server/src/api/slices/rename.ts
+++ b/packages/slice-machine/server/src/api/slices/rename.ts
@@ -1,11 +1,13 @@
-import { BackendEnvironment } from "../../../../lib/models/common/Environment";
+import fs from "fs";
+import path from "path";
+
 import * as Libraries from "@slicemachine/core/build/libraries";
 import {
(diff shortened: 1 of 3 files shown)


packages/slice-machine/server/src/api/slices/rename.ts
```diff
@@ -70,14 +70,29 @@ export async function renameSlice(req: {
     `${CustomPaths(env.cwd).library(libName).value()}/${newSliceName}`
```
MarcMcIntosh: do `/` paths work on windows?
maybe `path.join` would be better?

hypervillain: I believe they're equivalent in Windows and code editors do accept /

hypervillain: but happy to make the change

MarcMcIntosh (approved):

PR #825, opened 2023-01-27
[Reunification] Call transactional push endpoint

Heyrows:
## Context

[Linear ticket](https://linear.app/prismic/issue/SM-994/[reunification]-aauser-i-can-push-changes-in-a-transactional-way)

- [ ] The CI is successful.

## [OPT] Preview

fauzan-saari-AmhdN68wjPc-unsplash


packages/slice-machine/components/CustomTypeTable/ctPage.tsx
```diff
@@ -2,19 +2,23 @@ import { StatusBadge } from "../StatusBadge";
...
 import { RenameCustomTypeModal } from "@components/Forms/RenameCustomTypeModal";
 import { DeleteCustomTypeModal } from "@components/DeleteCTModal";
+import {
+  LocalAndRemoteCustomType,
+  LocalOnlyCustomType,
+} from "@lib/models/common/ModelData";
 
 export const CustomTypeTable: React.FC<{
-  customTypes: LocalFrontEndCustomType[];
+  customTypes: (LocalOnlyCustomType | LocalOnlyCustomType)[];
```
danigo99: Is this a mistake?

```suggestion
  customTypes: (LocalAndRemoteCustomType | LocalOnlyCustomType)[];
```

packages/slice-machine/components/DeleteDocumentsDrawer/DeleteDocumentsDrawerOverLimit.tsx
```diff
@@ -91,7 +99,8 @@ const DeleteDocumentsDrawerOverLimit: React.FunctionComponent = () => {
               label="Try again"
               variant="primary"
               onClick={() => {
-                console.log("Pushing");
+                closeModals();
+                pushChanges(); // TODO: should this confirm deletion? Otherwise we get the other modal
```
danigo99: Check with @comeprismic

packages/slice-machine/components/DeleteDocumentsDrawer/index.tsx
```diff
@@ -17,35 +17,22 @@ import { useSelector } from "react-redux";
...
+} from "@lib/models/common/ModelData";
+import { ToasterType } from "@src/modules/toaster";
 
-export const sortDocumentCards = (
-  documents: Readonly<Array<CustomtypeDocuments>>
-) =>
-  [...documents].sort(
-    (doc1, doc2) => doc2.numberOfDocuments - doc1.numberOfDocuments
-  );
+export function getCTName(ct: LocalOrRemoteCustomType | undefined): string {
```
danigo99: Pretty sure this type needs to be `LocalAndRemote | LocalOnly`.

But not sure how much difference this makes

packages/slice-machine/components/DeleteDocumentsDrawer/index.tsx
```diff
@@ -17,35 +17,22 @@ import { useSelector } from "react-redux";
...
-  { ctName: "Features6", numberOfDocuments: 6, link: "https://prismic.io/" },
-  { ctName: "Legals2", numberOfDocuments: 23, link: "https://prismic.io/" },
-  { ctName: "Features7", numberOfDocuments: 6, link: "https://prismic.io/" },
-  { ctName: "Features8", numberOfDocuments: 6, link: "https://prismic.io/" },
-  { ctName: "Features9", numberOfDocuments: 6, link: "https://prismic.io/" },
-  { ctName: "Features10", numberOfDocuments: 6, link: "https://prismic.io/" },
-  { ctName: "Features11", numberOfDocuments: 6, link: "https://prismic.io/" },
-  { ctName: "Features12", numberOfDocuments: 6, link: "https://prismic.io/" },
-];
+  return (hasLocal(ct) ? ct.local.label : ct.remote.label) ?? 'John "CT" Doe';
```
danigo99: Maybe a better default 😅

packages/slice-machine/src/modules/availableCustomTypes/types.ts
```diff
@@ -1,46 +1,6 @@
...
 
-export type NewCustomType = {
-  local: CustomTypeSM;
-};
-
-export type SyncedCustomType = {
-  local: CustomTypeSM;
-  remote: CustomTypeSM;
-};
-export type LocalFrontEndCustomType = NewCustomType | SyncedCustomType;
```
danigo99, line 11: I think this type was quite useful. We use it a lot everywhere, might be worth moving into `ModelData.ts`?

packages/slice-machine/lib/models/common/ModelData.ts
```diff
@@ -0,0 +1,63 @@
...
+export function hasLocal<T extends LocalOrRemote>(
+  obj: T
+): obj is T extends LocalAndRemote | LocalOnly ? T : never {
+  return "local" in obj;
+}
+
+export function hasRemote<T extends LocalOrRemote>(
+  obj: T
+): obj is T extends LocalAndRemote | RemoteOnly ? T : never {
+  return "remote" in obj && obj.remote !== undefined;
```
danigo99: ❓ Just realised I left the !== undefined here from when I was fixing something.
Not sure if this is something we want to keep?

packages/slice-machine/lib/models/common/ModelStatus/compareCustomTypeModels.ts
```diff
@@ -1,17 +1,14 @@
...
+  | { status: ModelStatus.Synced; model: LocalAndRemoteCustomType } {
   // If Custom Types are not equals then it was modified locally
-  if (!equal(models.local, models.remote)) {
-    return ModelStatus.Modified;
-  }
-
-  return ModelStatus.Synced;
+  if (!equal(model.local, model.remote))
+    return { status: ModelStatus.Modified, model };
+  return { status: ModelStatus.Synced, model };
```
danigo99: ❓ Why are we returning the model? I can't see where this is providing value

packages/slice-machine/lib/models/common/ModelStatus/compareCustomTypeModels.ts
```diff
@@ -1,17 +1,14 @@
...
-  SyncedCustomType,
-} from "@src/modules/availableCustomTypes/types";
 import equal from "fast-deep-equal";
 import { ModelStatus } from ".";
+import { LocalAndRemoteCustomType } from "../ModelData";
 
-export type FrontEndCtModel = FrontEndCustomType;
-
-export function compareCustomTypeModels(models: SyncedCustomType) {
+export function compareCustomTypeModels(
```
danigo99: 💡 I think this name is a bit deceiving? Since we're specifically comparing a customType's local model to its remote, I would suggest something like `compareCustomTypeLocalToRemote`

packages/slice-machine/lib/models/common/ModelStatus/compareSliceModels.ts
```diff
@@ -1,68 +1,26 @@
...
-  s: FrontEndSliceModel
-): s is DeletedFrontEndSliceModel => !isLocalSlice(s);
-
-export const getSliceProp = <key extends keyof SliceSM>(
-  s: FrontEndSliceModel,
-  property: key
-): SliceSM[key] => (isLocalSlice(s) ? s.local[property] : s.remote[property]);
+import { LocalAndRemoteSlice } from "../ModelData";
 
 export function compareSliceModels(
```
danigo99: Same here about the the function name

packages/slice-machine/lib/models/common/ModelStatus/compareSliceModels.ts
```diff
@@ -1,68 +1,26 @@
...
   const areScreenshotsEqual = compareScreenshots(
-    models.remote,
-    models.localScreenshots
+    model.remote,
+    model.localScreenshots
   );
-  if (!areScreenshotsEqual) {
-    return ModelStatus.Modified;
-  }
+  if (!areScreenshotsEqual) return { status: ModelStatus.Modified, model };
```
danigo99: Also same question about returning the model

packages/slice-machine/lib/models/common/ModelStatus/index.ts
```diff
@@ -12,22 +22,77 @@ export enum ModelStatus {
...
-  if (!("local" in models) || !models.local) return ModelStatus.Deleted;
+  if (!hasLocal(model)) return { status: ModelStatus.Deleted, model };
 
   // If there is no remote model then it's a new model created locally waiting to be pushed
-  if (!("remote" in models) || !models.remote) return ModelStatus.New;
+  if (!hasRemote(model)) return { status: ModelStatus.New, model };
 
-  if ("localScreenshots" in models) return compareSliceModels(models);
-  return compareCustomTypeModels(models);
+  if ("localScreenshots" in model) return compareSliceModels(model);
```
danigo99: ❓ is `localScreenshots` the best distinguishing feature of a slice and CT? I thought `type` might be an alternative
Is it even safe? What happens if you haven't added a screenshot yet?

packages/slice-machine/lib/models/common/normalizers/slices.ts
```diff
@@ -0,0 +1,27 @@
+import { LocalOrRemoteSlice } from "../ModelData";
+import { Component, Library, SliceSM } from "@slicemachine/core/build/models";
+
+export const normalizeFrontendSlices = (
+  localLibraries: ReadonlyArray<Library<Component>>,
+  remoteSlices: ReadonlyArray<SliceSM>
```
danigo99, line 6: Why aren't we using the new models?

packages/slice-machine/lib/models/common/widgets/ContentRelationship/index.ts
```diff
@@ -9,7 +9,7 @@ import { Link } from "@prismicio/types-internal/lib/customtypes/widgets/nestable
 import { WidgetTypes } from "@prismicio/types-internal/lib/customtypes/widgets";
 import { useSelector } from "react-redux";
 import { selectAllCustomTypes } from "@src/modules/availableCustomTypes";
-import { isLocalCustomType } from "@src/modules/availableCustomTypes/types";
+import { hasLocal } from "../../ModelData";
```
danigo99: [nit] Can we use the alias to avoid this big relative import?

danigo99:
```suggestion
import { hasLocal } from "@lib/models/common/ModelData";
```

packages/slice-machine/server/src/api/push-changes.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  /* -- Assemble the models together and compute their statuses -- */
+  const slicesModels: ReadonlyArray<LocalOrRemoteSlice> =
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    normalizeFrontendSlices(localSlices, remoteSlices);
+  const customTypeModels: ReadonlyArray<LocalOrRemoteCustomType> =
+    Object.values(
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+      normalizeFrontendCustomTypes(localCustomTypes, remoteCustomTypes)
+    );
```
danigo99: @mathieud can we remove these two eslint disables now?

packages/slice-machine/src/modules/availableCustomTypes/types.ts
```diff
@@ -1,46 +1,6 @@
...
-): ct is DeletedFrontEndCustomType => !isLocalCustomType(ct);
-
-export const isNewCustomType = (ct: FrontEndCustomType): ct is NewCustomType =>
-  !("remote" in ct);
-
-export const isSyncedCustomType = (
-  ct: FrontEndCustomType
-): ct is SyncedCustomType => "remote" in ct && "local" in ct;
-
-export const getCustomTypeProp = <key extends keyof CustomTypeSM>(
```
danigo99, line 38: ❓ I think this was cool. Is there a reason why it was removed in favour of just an `id` one?

packages/slice-machine/src/modules/pushChangesSaga/index.ts
```diff
@@ -1,294 +1,105 @@
...
-          case 401:
-          case 403: {
-            // Opening the login modal
-            yield put(modalOpenCreator({ modalKey: ModalKeysEnum.LOGIN }));
-
-            // Canceling the saga
-            yield cancel();
-
-            break;
-          }
```
danigo99, line 172: ❗❗❗ I think we're missing this case in the current implementation

cypress/e2e/user-flows/transactional-push.cy.js
```diff
@@ -0,0 +1,151 @@
...
+  function mockPushError() {
+    cy.intercept("POST", "/api/push-changes", {
+      statusCode: 500,
+      body: null,
+    });
+  }
+
+  before("Cleanup local data and create a new slice", () => {
+    cy.clearProject();
+    cy.setSliceMachineUserContext({});
```
mathieud: We can remove from here, it's already in the beforeEach

cypress/e2e/user-flows/transactional-push.cy.js
```diff
@@ -0,0 +1,151 @@
...
+  it("Creates, updates and deletes slices and custom types", () => {
+    cy.createSlice(slice.library, slice.id, slice.name);
+    cy.createCustomType(customType.id, customType.name);
+
+    const menu = new Menu();
+    const slicePage = new SlicePage();
+
+    menu.navigateTo("Changes");
+
+    cy.pushLocalChanges();
```
mathieud, line 57: It would be nice to have a better mechanism than this to make sure the state is clean

packages/client/src/index.ts
```diff
@@ -224,4 +233,50 @@ export class Client {
...
+          throw error;
+        },
+        (rawLimit: RawLimit) => {
+          const limit = { ...rawLimit, type: limitType };
+          return limit;
+        }
+      )(RawLimit.decode(potentialLimit));
+    }
+
+    return this._post(`${this.apisEndpoints.Models}bulk`, body)
```
mathieud, line 262: I find this really hard to read and follow

packages/client/src/models/BulkChanges.ts
```diff
@@ -0,0 +1,85 @@
...
+    customTypes: t.array(
+      t.type({
+        id: t.string,
+        numberOfDocuments: t.number,
+        url: t.string,
+      })
+    ),
+  }),
+});
+export type RawLimit = t.TypeOf<typeof RawLimit>;
```
mathieud, line 78: Should this be defined in the `client`? It forces the UI to import from it which can cause issues (`import Limit from "client"` fails because another export uses `fs`)

packages/slice-machine/components/CustomTypeTable/ctPage.tsx
mathieud: Also we had a type representing this before

packages/slice-machine/components/DeleteDocumentsDrawer/DeleteDocumentsDrawerOverLimit.tsx
mathieud: I think we should send the diff in here instead of recomputing it in the backend
- we have it already, no need to do the work a second time
- guarantees what the user sees is what we push
- will make it easy to allow the user to select what to push

mathieud: Frontend would send something like `{ id: string, operation: string }[]` and the server would still be responsible for creating the full payload from the data in the fs

packages/slice-machine/components/DeleteDocumentsDrawer/DeleteDocumentsDrawerOverLimit.tsx
```diff
@@ -104,22 +113,28 @@ const DeleteDocumentsDrawerOverLimit: React.FunctionComponent = () => {
...
 
-        {sortDocumentCards(AssociatedDocuments).map((ctDocuments) => (
+        {modalData.details.customTypes.map((ct) => (
           <AssociatedDocumentsCard
-            key={ctDocuments.ctName}
-            ctDocuments={ctDocuments}
+            key={ct.id}
             isOverLimit
+            ctName={getCTName(
+              availableCustomTypes.find(
```
mathieud: Should we move this to the selector, or even store it in the reducer to avoid recomputing it?

packages/slice-machine/components/DeleteDocumentsDrawer/index.tsx
mathieud: Don't think it can be `LocalOnly`. If you never pushed it you can delete it

packages/slice-machine/components/DeleteDocumentsDrawer/index.tsx
mathieud: I find this one quite inspired

packages/slice-machine/components/DeleteDocumentsDrawer/index.tsx
```diff
@@ -76,19 +63,27 @@ const ConfirmationDialogue: React.FC<{
...
+  const { pushChanges, closeModals, openToaster } = useSliceMachineActions();
 
   const [hasConfirmed, setHasConfirmed] = useState(false);
 
+  if (!isDeleteDocumentsDrawerOpen) return null;
+
+  if (!modalData?.details.customTypes) {
+    openToaster("No change data", ToasterType.ERROR);
+    return null;
+  }
```
mathieud: There's a lot of similar code across the modals. Maybe something worth sharing

packages/slice-machine/lib/models/common/ModelData.ts
```diff
@@ -0,0 +1,63 @@
...
+  | LocalOnlySlice
+  | RemoteOnlySlice;
+export type LocalAndRemoteSlice = LocalAndRemote<SliceSM> & {
+  localScreenshots: Partial<Record<string, Screenshot>>;
+};
+export type LocalOnlySlice = LocalOnly<SliceSM> & {
+  localScreenshots: Partial<Record<string, Screenshot>>;
+};
+export type RemoteOnlySlice = RemoteOnly<SliceSM> & {
+  localScreenshots: Partial<Record<string, Screenshot>>;
```
mathieud, line 64: This should be removed again, but causes a bug in the guard function (thinks a remote slice is a remote CT)

packages/slice-machine/lib/models/common/ModelData.ts
```diff
@@ -0,0 +1,63 @@
...
+export type LocalOnlyCustomType = LocalOnly<CustomTypeSM>;
+export type RemoteOnlyCustomType = RemoteOnly<CustomTypeSM>;
+
+// Slices
+
+export type LocalOrRemoteSlice =
+  | LocalAndRemoteSlice
+  | LocalOnlySlice
+  | RemoteOnlySlice;
+export type LocalAndRemoteSlice = LocalAndRemote<SliceSM> & {
```
mathieud, line 57: I've been thinking we should store the result of `computeStatus` in this structure. I think it would make the code simpler and definitely help typescript identify the objects 
```
type LocalSlice = {
  type: "slice",
  status: "new",
  local: Slice,
  localScreenshots: ...
}
```

packages/slice-machine/lib/models/common/ModelData.ts
```diff
@@ -0,0 +1,63 @@
...
+};
+export type RemoteOnlySlice = RemoteOnly<SliceSM> & {
+  localScreenshots: Partial<Record<string, Screenshot>>;
+};
+
+// Models
+
+export type LocalOrRemoteModel = LocalOrRemoteCustomType | LocalOrRemoteSlice;
+
+export function getModelId<M extends CustomTypeSM | SliceSM>(
```
mathieud, line 71: The method allowing to fetch any field seemed useful. We could have used it for the CT name. Any reason it's not here?

packages/slice-machine/lib/models/common/ModelStatus/compareSliceModels.ts
```diff
@@ -72,16 +30,16 @@ const stripImageUrl = (slice: SliceSM) => ({
 
 function compareScreenshots(
   remoteModel: SliceSM,
-  localScreenshots: Record<string, Screenshot>
+  localScreenshots: Partial<Record<string, Screenshot>>
```
mathieud, line 31: Does the `Partial` do something here?

packages/slice-machine/lib/models/common/normalizers/slices.ts
```diff
@@ -0,0 +1,27 @@
...
+): LocalOrRemoteSlice[] => {
+  const slices: Record<string, LocalOrRemoteSlice> = {};
+
+  const localComponents = localLibraries.reduce<Component[]>(
```
mathieud: Looks like a `flatMap`

packages/slice-machine/server/src/api/push-changes.ts
mathieud: Yes we should be able to

packages/slice-machine/server/src/api/push-changes.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  /* -- Compute the request body -- */
+
+  const sliceChanges = slicesModels.reduce(
```
mathieud: All this could be sent by the frontend. 
If not I'd change this for a `.filter().map()`, would be easier to read

packages/slice-machine/server/src/api/push-changes.ts
```diff
@@ -0,0 +1,197 @@
...
+    (
+      acc: (SliceInsertChange | SliceUpdateChange | SliceDeleteChange)[],
```
mathieud: Typescript should be able to infer this

packages/slice-machine/src/apiClient.ts
```diff
@@ -158,6 +160,12 @@ export const pushSliceApiClient = (
     .then((response) => response.data);
 };
 
+export const pushChanges = (
+  payload: PushChangesPayload
+): Promise<AxiosResponse<Limit | null>> => {
+  return axios.post("/api/push-changes", payload, defaultAxiosConfig);
```
mathieud, line 166: @danigo99 You said we needed to handle auth issues here?

packages/slice-machine/src/hooks/useModelStatus.ts
```diff
@@ -24,11 +25,12 @@ export interface ModelStatusInformation {
   isOnline: boolean;
 }
 
-const isSliceModel = (m: FrontEndModel): m is FrontEndSliceModel =>
+// THIS DOES NOT WORK
```
mathieud: There is a problem as `localScreenshots` don't exist in the delete slices

packages/slice-machine/src/hooks/useModelStatus.ts
```diff
@@ -41,17 +43,17 @@ export const useModelStatus = (
 
   const modelsStatuses: ModelStatusInformation["modelsStatuses"] =
     models.reduce(
-      (acc: ModelStatusInformation["modelsStatuses"], model: FrontEndModel) => {
-        const status: ModelStatus = computeModelStatus(
-          model,
-          userHasAccessToModels
-        );
+      (
+        acc: ModelStatusInformation["modelsStatuses"],
```
mathieud: Type should be infered

packages/slice-machine/src/hooks/useModelStatus.ts
```diff
@@ -74,3 +76,53 @@ export const useModelStatus = (
...
+    const { status } = computeModelStatus(model, userHasAccessToModels);
+
+    return {
+      ...acc,
+      [hasLocal(model) ? model.local.id : model.remote.id]: status,
+    };
+  }, {} as { [sliceId: string]: ModelStatus });
+}
+
+export const useModelStatus2 = ({
```
mathieud: This version looks much simpler to me. We never have the slices and customTypes combined in the application

packages/slice-machine/src/hooks/useUnSyncChanges.ts
```diff
@@ -49,17 +48,19 @@ export const useUnSyncChanges = (): UnSyncChanges => {
...
+  });
 
   const localComponents: ComponentUI[] = libraries.flatMap(
     (lib) => lib.components
   );
 
   const deletedComponents: ComponentUI[] = slices
-    .filter(isDeletedSlice)
+    .filter(
+      (slice): slice is RemoteOnlySlice => hasRemote(slice) && !hasLocal(slice)
```
mathieud: We should have this guard defined somewhere

packages/slice-machine/src/modules/pushChangesSaga/index.ts
```diff
@@ -1,294 +1,109 @@
...
+          modalKey:
+            response.data?.type === LimitType.SOFT
+              ? ModalKeysEnum.DELETE_DOCUMENTS_DRAWER
+              : ModalKeysEnum.DELETE_DOCUMENTS_DRAWER_OVER_LIMIT,
+        })
+      );
+      return;
+    }
+
+    // TODO: find a better way of doing this
```
mathieud: @danigo99 Do you remember what this was for?

packages/slice-machine/components/CustomTypeTable/ctPage.tsx
Heyrows: Seems like a mistake haha, we chose to differentiate the types of data and less what they represent (local or remote).

This is a bit more precise than the previous type which stated "you have a local value" without knowing the state of the remote one.

packages/slice-machine/components/DeleteDocumentsDrawer/index.tsx
Heyrows: right, the type would have to be `RemoteOnly` or `LocalAndRemote` to be deleted after wards.

packages/slice-machine/components/DeleteDocumentsDrawer/index.tsx
Heyrows: 😂 damn, shouldn't this function return undefined actually and where it's used we would use the default value ?
Does this default values comes from you or the design ? Because it's quite funny

packages/slice-machine/lib/models/common/ModelData.ts
Heyrows: Actually while developing we found that some part of the code would add `remote: undefined` but this is something the type doesn't not expect.

That's why we didn't put it in the remote check. To make sure the data are not wrong but covered by tricks here and there in the app.

We actually quite to struggle the piece of code that introduced the data deformation. 😓

packages/slice-machine/src/modules/availableCustomTypes/types.ts
Heyrows: I also think this is useful but i think it creates too much layer of types.

packages/slice-machine/lib/models/common/ModelStatus/compareCustomTypeModels.ts
Heyrows: Hey, the model returned is actually type depending on the status.
If you look at the type definition of the functions, you'll see it return tuples of status and data typed.
This whole file refactor was so we could re-use the logic and have a model typed.

This is useful when testing the status of a model then using `switch case` / `ifs` to process each case.

packages/slice-machine/lib/models/common/ModelStatus/index.ts
Heyrows: good point, to test out.
Ideally the localScreenshots would be empty if there is none but the property itself would still be here so this check wouldn't fail.

packages/slice-machine/lib/models/common/normalizers/slices.ts
Heyrows: For now the slice store is still shaped with those types, so we wan't to make a second ticket refactor the store and data interactions around it 😄

packages/slice-machine/src/modules/pushChangesSaga/index.ts
Heyrows: It is true if we just look at this function in itself, but at the same time when the user arrive on the pages, if he's not connected he won't see the changes and won't be able to push.

Heyrows: But you're right, for safety maybe we could put it back.

packages/slice-machine/src/modules/availableCustomTypes/types.ts
Heyrows: As much as we saw, getCustomTypeProp was only use to get the ID.
+ we needed 2 helpers (one for slices, one for CT)

Now we have one that returns the ID specifically.
Unless there's need to access X properties, i'd still stick with specific helpers.

In most cases and ideally (meaning propably not right now) the app would know on some screens which special type to use.

ex:
- slice list -> slices with local value
- changes page -> all type of data -> call for those helpers or `hasLocal / hasRemote`

cypress/e2e/user-flows/transactional-push.cy.js
Heyrows: What do you mean ?

packages/client/src/index.ts
Heyrows: What improvement would do you think could be made ?

packages/client/src/models/BulkChanges.ts
Heyrows: I'm not sure to understand, the Raw Limit is defined in the Client package already.

Heyrows: The UI normally import the Limit as a Type so i'm not sure if it would import `fs` doing so.

packages/slice-machine/components/DeleteDocumentsDrawer/DeleteDocumentsDrawerOverLimit.tsx
Heyrows: I'm glad you asked the question because it's an important one that I struggled myself with too 😄.

Should the frontend or the backend control what's being done.
- the frontend can be out of date so it would push wrong things (ex, when someone else push changes, you wouldn't see it straight away so the changes might not be the same and things will fail)
- the goal of pushing things is to push your file system as is in the database. By doing it in the backend, we ensure that this actually happens as much as possible.

There is more to say, it's an interesting question, i'll keep some for the meeting this afternoon.

packages/slice-machine/components/DeleteDocumentsDrawer/index.tsx
Heyrows: I'm not sure to understand which similar code you're talking about.

packages/slice-machine/lib/models/common/ModelData.ts
Heyrows: hmmmm odd, would be funny to see.

packages/slice-machine/lib/models/common/ModelStatus/compareSliceModels.ts
Heyrows: It defines that no any string has a value, it would be the same as `Record<string, Screenshot | undefined>`

packages/slice-machine/lib/models/common/normalizers/slices.ts
Heyrows: Interesting, i didn't knew flatMap existed in JS.

packages/slice-machine/server/src/api/push-changes.ts
Heyrows: I'm not sure to understand what you'd actually filter here.

packages/slice-machine/server/src/api/push-changes.ts
Heyrows: I would have thought so too 😓  but i think it's because we change the original type that typescript doesn't like if we don't declare the type of the accumulator.

packages/slice-machine/src/hooks/useModelStatus.ts
Heyrows: ahhh nice 👍

Heyrows: We could use the property of a slice which is sharedSlice something
Or we could use io-ts decoder even if it seems a bit much.

packages/slice-machine/src/hooks/useUnSyncChanges.ts
Heyrows: good point

packages/slice-machine/src/modules/pushChangesSaga/index.ts
danigo99: I've just added it back now along with the test

packages/slice-machine/components/DeleteDocumentsDrawer/index.tsx
danigo99: I think returning the ID as the default would make sense :)

packages/slice-machine/components/DeleteDocumentsDrawer/index.tsx
Heyrows: Review: The type of custom type to display in teh drawer should be RemoteOnly for deleted custom types.
The name would be easier to get through the remote model.

packages/slice-machine/components/DeleteDocumentsDrawer/index.tsx
Heyrows: Review:
- Ask in engineering about the types why is the label potentially undefined
- Ask design for default value.

packages/slice-machine/src/modules/availableCustomTypes/types.ts
Heyrows: Review: If we find a good name for it, we create it.

packages/slice-machine/lib/models/common/ModelData.ts
Heyrows: Review: Amaury's gonna try to remote the == undefined part and see if something breaks.

packages/slice-machine/lib/models/common/ModelStatus/compareCustomTypeModels.ts
Heyrows: Review: this helper can send the status only.

packages/slice-machine/lib/models/common/ModelStatus/index.ts
Heyrows: Review: replace "localScreenshots" by `variations` in model.local

Heyrows: Review: change function "isSliceModel" in `useModelStatus.ts`

packages/slice-machine/lib/models/common/widgets/ContentRelationship/index.ts
Heyrows: Review: do it

packages/slice-machine/server/src/api/push-changes.ts
Heyrows: Review: @mathieud has a fix for this.

packages/slice-machine/components/DeleteDocumentsDrawer/DeleteDocumentsDrawerOverLimit.tsx
danigo99: ^^ confirmed that we show the confirmation modal. Just need to remove the TODO here.

cypress/e2e/user-flows/transactional-push.cy.js
Heyrows: Review: Create a helper that totally cleans the project and the remote Repository that would be usable everywhere.

packages/client/src/index.ts
Heyrows: Review: Amaury and Mathieu go over the code to clarify it.

packages/client/src/models/BulkChanges.ts
Heyrows: Review: Baptiste go crazy

packages/slice-machine/components/DeleteDocumentsDrawer/DeleteDocumentsDrawerOverLimit.tsx
Heyrows: Review: ask Côme about allowing users to select what to push.

packages/slice-machine/components/DeleteDocumentsDrawer/DeleteDocumentsDrawerOverLimit.tsx
Heyrows: Review:
- we'll wait for optimisation to see if there's an issue with this.
- getCTName shouldn't be used when the data is undefined.

packages/slice-machine/components/DeleteDocumentsDrawer/index.tsx
Heyrows: @mathieud

packages/slice-machine/server/src/api/push-changes.ts
Heyrows: @mathieud

packages/slice-machine/src/hooks/useModelStatus.ts
Heyrows: Reviewer: much clearer, let's use that !

packages/slice-machine/src/hooks/useUnSyncChanges.ts
Heyrows: Review: create type guards for those

packages/client/src/models/BulkChanges.ts
bapmrl: I thought we could use TypeScript's `infer` to build a type to extract the success type of the `Promise`, but `Limit` is the success type and not `RawLimit` so...

In this instance @mathieud, I would suggest either:
1. Duplicating the `RawLimit` type (it's fine as long as we only have 2 definitions and `RawLimit` is kind of an implementation detail of the client).
2. Moving the `RawLimit` type to the core package.

packages/slice-machine/components/DeleteDocumentsDrawer/DeleteDocumentsDrawerOverLimit.tsx
Heyrows: resolving this, Côme answered that it's not a priority for now.

packages/slice-machine/lib/models/common/ModelData.ts
danigo99: Nice!

PR #827, opened 2023-02-03
fix flaky e2e custom screenshots test

mdeclercq:
## Context

https://linear.app/prismic/issue/SM-1001/[spike]-upload-screenshots-e2e-test-is-flaky

The Slice screenshot was sometimes disappearing right after being successfully uploaded and displayed, making the test fail. I could not reproduce when interacting with the browser as a human. 

## The Solution

- I noticed that not reloading the Slice page after having created the slice solved the issue. i.e. not calling `slicePage.goTo(library, name)`
- Moved the test to the `slices` folder with a comprehensive name
- Used `md5` instead of `sha256` to compute the image hash as `sha256` seemed unnecessary (not linked to the flakiness)

- [ ] The CI is successful.

image

### DIFF
--- a/cypress/pages/Menu.js
+++ b/cypress/pages/Menu.js
@@ -1,5 +1,10 @@
 export class Menu {
+  /**
+   * @param {('Custom Types'|'Slices'|'Changes')} label - The menu item to click
+   */
   navigateTo(label) {
-    return cy.get("aside").contains(label).click();
+    cy.get("aside").contains(label).click();
+    cy.get("main a").contains(label);
+    return this;
   }
 }
--- a/cypress/e2e/user-flows/scenario_custom_screenshots.cy.js
+++ b/cypress/e2e/user-flows/scenario_custom_screenshots.cy.js
@@ -77,6 +77,8 @@ describe("I am an existing SM user and I want to upload screenshots on variation
   it("Error displayed when non-image files are uploaded", () => {
     slicePage.goTo(slice.library, slice.name);
     slicePage.addVariation("Error handling");
+    cy.saveSliceModifications();
+
     slicePage.openScreenshotModal();
     cy.contains("Select file").selectFile(
       {
--- a/cypress/e2e/user-flows/scenario_custom_screenshots.cy.js
+++ b/cypress/e2e/user-flows/scenario_custom_screenshots.cy.js
@@ -3,12 +3,12 @@ import { SliceCard } from "../../pages/slices/sliceCard";
 import { SlicePage } from "../../pages/slices/slicePage";
 import { Menu } from "../../pages/Menu";
 
-describe.skip("I am an existing SM user and I want to upload screenshots on variations of an existing Slice", () => {
+describe("I am an existing SM user and I want to upload screenshots on variations of an existing Slice", () => {
   const random = Date.now();
 
   const slice = {
-    id: `test_screenshots${random}`,
-    name: `TestScreenshots${random}`,
+    id: `test_custom_screenshots${random}`,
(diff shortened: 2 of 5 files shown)


cypress/helpers/images.js
```diff
@@ -3,10 +3,10 @@ export function isSameImageAs(subject, referenceImage) {
   cy.wrap(subject)
     .should(([img]) => expect(img.complete).to.be.true)
     .then(([img]) => cy.request({ url: img.src, encoding: "base64" }))
-    .then((response) => cy.task("sha256", response.body))
+    .then((response) => cy.task("md5", response.body))
```
Heyrows: i'd be careful with this and either use functions from the code itself to compute the hash or try to get as close as possible.

cypress/pages/Menu.js
```diff
@@ -1,5 +1,10 @@
 export class Menu {
+  /**
+   * @param {('Custom Types'|'Slices'|'Changes')} label - The menu item to click
+   */
   navigateTo(label) {
-    return cy.get("aside").contains(label).click();
+    cy.get("aside").contains(label).click();
+    cy.get("main a").contains(label);
```
Heyrows: what would that one check ?

cypress/pages/Menu.js
```diff
@@ -1,5 +1,10 @@
 export class Menu {
+  /**
+   * @param {('Custom Types'|'Slices'|'Changes')} label - The menu item to click
```
Heyrows, line 3: super nice !!

Heyrows (approved):
Maybe the Before Each was starting before the first test actually finishes.
I don't have much means of verifying it unless running the tests X times on my machine so i think it's okay to let the CI tell us if it's okay or not in the long run.

I'd share with the team a way to detect the test easily so that we can know if this is the same issue again next time.

cypress/pages/Menu.js
mdeclercq: To identify the page title in the main panel: 
image
It would have been nice to have a `h1` or `h2` to locate this but it didn't exist. Maybe we could just add some aria labels like `role="heading" aria-level="1"` as described [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role), wdyt?

cypress/helpers/images.js
mdeclercq: Do you mean we should directly call something like https://github.com/prismicio/slice-machine/blob/165337d4da2f80238c156480f73a64016b07f7a7/packages/core/src/utils/str.ts#L15-L17 ?
tbh I wish Cypress offered a better way to compare an image to a local one, to avoid having to define these.

cypress/pages/Menu.js
Heyrows: ah i see.
I thought it would an element of the menu itself that's why I found it confusing.

cypress/helpers/images.js
Heyrows: Yeah that's what I meant, using what the code actually uses to test the same thing.
Or getting closer as possible (i thought the code was using sha256

mdeclercq: 👍 I'll revert that commit

cypress/pages/Menu.js
mdeclercq: you're right, it shouldn't be part of the Menu page object, I'll move it away.

mdeclercq: done

cypress/helpers/images.js
mdeclercq: done

MarcMcIntosh (approved):

PR #835, opened 2023-02-10
Fix express route with env type

hypervillain:
Following a comment from @MarcMcIntosh regarding misused promises in Express route handlers, this PR narrows the type of `withEnv` to comply to Express RouteHandler type, and removes all unnecessary comments in API endpoint


danigo99 (approved):
Looks good!!

❓ There are a few endpoints which don't use `withEnv` and instead get their env inside the handler, do we want to refactor these now so we can get rid of the misused-promises there?

MarcMcIntosh (approved):

PR #799, opened 2023-01-09
feat!: plugins support

angeloashmore:
> **Note**
>
> **Please open issues on Linear in the SM x Plugin team**
>
> - 🐛 [**Submit a Bug Report**](https://linear.app/prismic/team/SMX/new?template=9ad3c0ef-7a5d-4abc-bb60-673808614898)
> - 💬 [**Submit a Code Review**](https://linear.app/prismic/team/SMX/new?template=0c2318e8-3173-4d86-bd20-81b28fc6e0fc)
> - 🙋‍♀️ [**Submit a Feature Request**](https://linear.app/prismic/team/SMX/new?template=c4c8b05e-fb78-4ae5-9c76-160411704bf3)
> 
> Using Linear will help us thread discussions and track changes more easily than through the comment system on GitHub.

## Overview

The current Slice Machine architecture is purpose-built for two frameworks: Next.js and Nuxt. Over time, we have learned that adding new framework integrations and maintaining existing ones is currently not scalable nor flexible. We now know these limitations primarily stem from architectural decisions that were beneficial at the time of Slice Machine’s creation, but are now detrimental to its growth.

An ideal Slice Machine architecture exhibits the following attributes:

- Allows user customization to match a project’s framework and conventions.
- Empowers its community to improve and extend its functionality.
- Requires minimal development effort to maintain and improve.

This PR implements a plugin-based architecture that enables these attributes.

With this new architecture, we can better serve developers using and building Slice Machine.

### Implementation status and next steps

In order to open this PR, we have conducted experiments, proposed architectural changes, and implemented a new Slice Machine architecture.

Opening this PR represents the start of the internal review process. The process will require significant effort to test, review, discuss, and approve before merging and publishing.

The following checklist shows overall where we are in the implementation process:

- [x]  Internal Request for Comments
- [x]  Implementation experiments
- [x]  Implementation proposals
- [x]  Technical implementation
- [x]  Internal code review
- [x]  Internal quality assurance
- [x]  Public release

### Glossary

<details>
<summary>View glossary terms</summary>

- **plugin**: An npm package that provides core Prismic functionality in Slice Machine and/or extends Slice Machine’s functionality.
- **adapter**: A specific type of plugin that Slice Machine uses as a source of truth for data, such as Custom Type JSON models and Slice Library metadata.
- **project**: An environment where Slice Machine is configured and used. A Next.js app with Slice Machine installed is a “project.”
[...]

PR #838, opened 2023-02-10
Fix generate index before rename

hypervillain:
Refactor of library index file on save, rename, push + generate before rename to prevent errors in user's project console


lihbr:
Handled in #799

PR #840, opened 2023-02-14
track changes pushed and limit reached

Heyrows:
## Context

[Linear ticket](https://linear.app/prismic/issue/SM-993/[reunification]-aapm-i-can-track-users-pushing-changes-in-slice)

- [ ] The CI is successful.

## [OPT] Preview

Screenshots in the linear tickets

nikolett-emmert-Blq2GgF49yM-unsplash


packages/slice-machine/pages/changes.tsx
```diff
@@ -40,10 +41,22 @@ const Changes: React.FunctionComponent = () => {
     return () => {
       closeModals();
     };
+    // Do not remote the eslint disable, fixing the eslint warning creates a bug that prevent from opening all modals
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
danigo99, line 46: ❤️

packages/slice-machine/server/src/api/tracking.ts
```diff
@@ -18,12 +18,12 @@ export function sendEvents(
 ): void {
   if (isGroupLibrariesEvent(event)) {
     analytics.group({
-      ...(userId ? { userId } : { anonymousId }),
+      ...(userId !== undefined ? { userId } : { anonymousId }),
```
danigo99, line 21: ❓ What prompted this?

packages/slice-machine/components/DeleteDocumentsDrawer/HardDeleteDocumentsDrawer.tsx
```diff
@@ -45,7 +47,7 @@ export const HardDeleteDocumentsDrawer: React.FunctionComponent = () => {
           variant="primary"
           onClick={() => {
             closeModals();
-            pushChanges();
+            pushChanges(true);
```
danigo99: I think this is wrong. We don't want it to push the changes when you retry without a confirmation first. This is what we clarified with Côme the other day 🙂

packages/slice-machine/src/modules/pushChangesSaga/trackPushChangesSuccess.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  const total = unSyncedSlices.length + unSyncedCustomTypes.length;
+  const missingScreenshots: number = unSyncedSlices.reduce(
+    (sum: number, slice: ComponentUI) => sum + countMissingScreenshots(slice),
+    0
+  );
+  const duration = Date.now() - startTime;
+
+  return Tracker.get().trackChangesPushed({
+    customTypesCreated: customTypesStats.customTypesCreated,
```
mathieud: Could this be
```
 {
   ...customTypesStats,
   ...slicesStats,
   // rest
}
```

packages/slice-machine/server/src/api/tracking.ts
Heyrows: eslint errors so I fixed them :')

packages/slice-machine/components/DeleteDocumentsDrawer/HardDeleteDocumentsDrawer.tsx
Heyrows: hmmm ok, I wasn't sure.
I just assumed that the action of pushing changes in case of a hard delete (that require manual deletion of documents)
Should be treated as a true by default because in facts, documents have been deleted for this.

Heyrows: If you clarified with Come then i'll change it to false, thank for telling me 😄

packages/slice-machine/src/modules/pushChangesSaga/trackPushChangesSuccess.ts
Heyrows: definitely 😄 , i was hesitant.

packages/slice-machine/components/DeleteDocumentsDrawer/HardDeleteDocumentsDrawer.tsx
Heyrows: ahhhh how dumb of me yeah you're totally right we don't want to assume they actually removed all the documents.

danigo99 (commented):
Happy with the changes! 👍

danigo99 (approved):
LGTM 👍 

Just to confirm then, the references error event will be the same as the limit reached one in the dashboard right? 
And then from the type you can separate the two

Heyrows:
yes it's exactly that, just an extra property value to track this as well.

PR #842, opened 2023-02-14
Don't show references error for legacy slice

danigo99:
## Context

Issue: https://linear.app/prismic/issue/SM-994#comment-6a542572

## The Solution

We check that the slice is not a legacy slice when checking if the slices is in the local libraries.

- [ ] The CI is successful.

## [OPT] Preview

image

### DIFF
--- a/packages/slice-machine/server/src/api/pushChanges/checkCustomTypes.ts
+++ b/packages/slice-machine/server/src/api/pushChanges/checkCustomTypes.ts
@@ -15,7 +15,8 @@ export function getCustomTypesWithInvalidReferences(
       ct.tabs.some((tab) =>
         tab.sliceZone?.value.some(
           (z) =>
-            !localSliceIds.includes(z.key) && z.value.type !== SlicesTypes.Slice
+            !localSliceIds.includes(z.key) &&
+            z.value.type === SlicesTypes.SharedSlice
         )
       )
     )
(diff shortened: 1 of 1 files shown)


packages/slice-machine/server/src/api/pushChanges/checkCustomTypes.ts
```diff
@@ -12,7 +13,10 @@ export function getCustomTypesWithInvalidReferences(
   return localCustomTypes
     .filter((ct) =>
       ct.tabs.some((tab) =>
-        tab.sliceZone?.value.some((z) => !localSliceIds.includes(z.key))
+        tab.sliceZone?.value.some(
+          (z) =>
+            !localSliceIds.includes(z.key) && z.value.type !== SlicesTypes.Slice
```
Heyrows: do you think it would be worth making a helper out of it ?
To recognise legacy slices.

Heyrows (approved):

packages/slice-machine/server/src/api/pushChanges/checkCustomTypes.ts
danigo99: Fair point, we do use it in a few places.

The helper would just return `slice.type === SlicesTypes.Slice` though

arnaudlewis: I'm out of context but legacy slices are more that just `Slice` just in case. I reviewed a PR on the editor where I talk about that a bit: https://github.com/prismicio/editor/pull/129#discussion_r1106884567
Hope it helps

Heyrows (approved):

PR #844, opened 2023-02-15
Cleanup selected CT store when exiting the CT page

danigo99:
## Context

Linear ticket with bug: https://linear.app/prismic/issue/SM-1033/aauser-i-want-to-see-the-legacy-slice-migration-link-only-on-ct-using

## The Solution

The problem is that after you have just visited a custom type with legacy slices, this CT is still loaded in the `selectedCustomType` store. So when you open another one, the page is initially rendered with the info of the one in the store (since this is what the selector returns). 

This isn't a problem for most of the content on the page (this behaviour is not new) since the subsequent re-render happens to fast that you don't notice the change. But since the warning toast remains on the page for a few seconds, it becomes a problem.

image

- [ ] The CI is successful.

## [OPT] Preview

https://user-images.githubusercontent.com/47107427/219027282-9c8556f7-b3b0-4a4b-8828-ef437bdc3f36.mov

### DIFF
--- a/packages/slice-machine/pages/cts/[ct].tsx
+++ b/packages/slice-machine/pages/cts/[ct].tsx
@@ -49,10 +49,21 @@ const CustomTypeBuilderWithRouter = () => {
     })
   );
 
+  const { cleanupCustomTypeStore } = useSliceMachineActions();
+
   useEffect(() => {
     if (!selectedCustomType || !hasLocal(selectedCustomType))
       void router.replace("/");
   }, [selectedCustomType, router]);
+
+  useEffect(() => {
+    router.events.on("routeChangeStart", cleanupCustomTypeStore);
+
+    return () => {
+      cleanupCustomTypeStore();
+      router.events.off("routeChangeStart", cleanupCustomTypeStore);
+    };
+  }, []);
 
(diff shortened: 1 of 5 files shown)


packages/slice-machine/pages/cts/[ct].tsx
```diff
@@ -49,11 +49,22 @@ const CustomTypeBuilderWithRouter = () => {
...
   );
 
+  const { cleanupCustomTypeStore } = useSliceMachineActions();
+
   useEffect(() => {
     if (!selectedCustomType || !hasLocal(selectedCustomType))
       void router.replace("/");
   }, [selectedCustomType, router]);
 
+  useEffect(() => {
```
Comment (Heyrows) [inline packages/slice-machine/pages/cts/[ct].tsx:59]: I don't think you need the router events, just returning a function that executes cleanupCustomTypeStore is enough.

Heyrows (changes requested):

packages/slice-machine/pages/cts/[ct].tsx
Comment (danigo99) [reply packages/slice-machine/pages/cts/[ct].tsx:59]: You're right! Will update :)

Heyrows (approved):

PR #848, opened 2023-02-21
fix(plugins): validate `SliceMachineConfig` in `getProject()` plugin helper

angeloashmore:
## Types of changes

## Description

This PR adds a validation step to the `getProject()` plugin helper.

> **Warning**
>
> This introduces some code duplication between `@slicemachine/manager` and `@slicemachine/plugin-kit`. Note that `SliceMachineConfig` is a codec and TypeScript type that MUST be updated
> in both files. Both packages cannot reference each other because that would cause an invalid circular dependency.
> 
> `@slicemachine/plugin-kit` could export its `SliceMachineConfig` type and codec, but `@slicemachine/plugin-kit`'s purpose is provide a plugin runner and SDK, not types and helpers for reuse elsewhere. We could still do this, but we should consider if we are making the codebase messy by doing so.

**Aside**: a small refactor changing `any` to `unknown` was made in this file: https://github.com/prismicio/slice-machine/pull/848/files#diff-0756541576108a488ddea0aa5cfb63b47fe61dd716a8d63875b3c98319f0c995

## Checklist:

🐱


packages/plugin-kit/package.json
```diff
@@ -51,6 +51,8 @@
 		"@prismicio/types": "^0.2.7",
 		"common-tags": "^1.8.2",
 		"defu": "^6.1.1",
+		"io-ts": "^2.2.20",
```
bapmrl, line 57: 💡 The `packages/slice-machine/package.json` file uses version `^2.2.16`. I would use the same to simply searching and replacing.

packages/plugin-kit/src/lib/DecodeError.ts
```diff
@@ -0,0 +1,22 @@
+import * as t from "io-ts";
```
bapmrl, line 1: 💡 This is a duplicate of `packages/manager/src/lib/DecodeError.ts`. What do you think of adding a `packages/support` workspace to encourage sharing utilities between packages? Would this be useful for other parts of the code?

packages/plugin-kit/src/lib/decode.ts
```diff
@@ -0,0 +1,36 @@
+import * as t from "io-ts";
```
bapmrl, line 1: 💡 This is a duplicate of `packages/manager/src/lib/decode.ts`.

packages/plugin-kit/src/lib/decodeSliceMachineConfig.ts
```diff
@@ -0,0 +1,43 @@
...
+		t.type({
+			resolve: t.string,
+		}),
+		t.partial({
+			options: t.UnknownRecord,
+		}),
+	]),
+]);
+
+const SliceMachineConfigCodec = t.intersection([
```
bapmrl, line 19: 💡 What do you think of adding a `packages/domain` or `packages/models` workspace? Would this help you with the circular references that would be introduced by sharing `SliceMachineConfigCodec`?

packages/plugin-kit/src/lib/decodeSliceMachineConfig.ts
```diff
@@ -0,0 +1,43 @@
...
+		plugins: t.array(SliceMachineConfigPluginRegistrationCodec),
+	}),
+]);
+
+// TODO: Maybe rename "decode" to "validate". "decode" exposes the `io-ts`
+// internals.
+export const decodeSliceMachineConfig = (
+	input: unknown,
+): DecodeReturnType<SliceMachineConfig, SliceMachineConfig, unknown> => {
+	return decode(SliceMachineConfigCodec, input) as DecodeReturnType<
```
bapmrl: ❓ Why do we have to cast?

bapmrl (commented):
Nice! I've added some questions/ideas.

packages/plugin-kit/src/lib/DecodeError.ts
angeloashmore: Maybe a general-purpose "support" package is inevitable, but I think it could be a dangerous move. By creating a "support" package, we make it easy for us to prematurely abstract code to be shared across packages. We may then run into technical problems, like cross-platform incompatibilities (e.g. importing Node.js packages into browser code) or making non-obvious breaking changes.

`@slicemachine/core` suffers from this problem today.

As much as possible, I think we should contain all code that a package uses within itself, except for very deliberate exports (e.g. `createSliceMachineManager()` from `@slicemachine/manager` is a deliberate export).

I can't think of another solution other than creating an additional package, but maybe we can refine the idea of a "support" package. Rather than making it generic in the same way that a "utils," "helpers," or "core" package would be, we can be deliberate about what the package is.

For example, maybe the package only contains codecs/validators. What is allowed in the package becomes more obvious and discourages bad practices. Such a package could contain `decode()`, `DecodeError`, and shared codecs.

Some ideas:
- `@slicemachine/validators`
- `@slicemachine/io-ts-validators`
- `@slicemachine/io-ts-codecs`
- `@slicemachine/io-ts-decoders`

What do you think?

packages/plugin-kit/src/lib/decode.ts
angeloashmore: Replied here: https://github.com/prismicio/slice-machine/pull/848#discussion_r1114889135

packages/plugin-kit/src/lib/decodeSliceMachineConfig.ts
angeloashmore: Replied here: https://github.com/prismicio/slice-machine/pull/848#discussion_r1114889135

packages/plugin-kit/src/lib/decodeSliceMachineConfig.ts
angeloashmore: It turns out we don't need to use `as`. Thanks for catching that!

I removed it in this commit: https://github.com/prismicio/slice-machine/pull/848/commits/f1385a44ac879d923ce02400c2f145067dd24b38

It was added when the function was first written because there was some kind of type incompatibility that required asserting the type. Note that we purposely use the `SliceMachineConfig` type rather than the inferred type from `SliceMachineConfigCodec`. This allows us to include TSDoc comments, which is not possible with `io-ts` codecs.

packages/plugin-kit/src/lib/DecodeError.ts
bapmrl: I think a general "support" package is inevitable to avoid duplicating code, but you're right and we could fall into the technical problems you mention. Would specifying [ESLint's env](https://eslint.org/docs/latest/use/configure/language-options#specifying-environments) to `shared-node-browser`, [TypeScript's lib](https://www.typescriptlang.org/tsconfig#lib) to `ES2020` and clearly indicating this issue in a `README.md` file be strong enough incentives? Could some other configs or tools help us?

Otherwise, let's create a `@slicemachine/validators` or `@slicemachine/codecs` package, but it has the drawback of "virtually" increasing the number of things we have to publish to NPM. I don't see how we could alleviate that though.

packages/plugin-kit/package.json
MarcMcIntosh: There should be an update script in the top level package.json that uses `lerna-wizard` it makes it really easy to update a dep across projects :) 
`"lerna:update": "npx --yes lerna-update-wizard"`

packages/plugin-kit/src/createSliceMachineHelpers.ts
```diff
@@ -44,12 +46,31 @@ export class SliceMachineHelpers {
...
+		let rawConfig: unknown | undefined;
+		try {
+			const contents = await fs.readFile(configFilePath, "utf8");
+			rawConfig = JSON.parse(contents);
+		} catch {
+			// noop
+		}
+
+		if (!rawConfig) {
+			// TODO: Write a more friendly and useful message.
```
MarcMcIntosh, line 59: That's got me thinking, should we ask Sam about error messages?

MarcMcIntosh: ...add the path maybe?

packages/plugin-kit/src/lib/DecodeError.ts
MarcMcIntosh: I guess codecs, would be limted to io-ts types, while validators could theoretically include form validators.... codecs might be better :)

packages/plugin-kit/src/lib/DecodeError.ts
```diff
@@ -0,0 +1,22 @@
...
+import { formatValidationErrors } from "io-ts-reporters";
+
+type DecodeErrorConstructorArgs<TInput = unknown> = {
+	input: TInput;
+	errors: t.Errors;
+};
+
+export class DecodeError<TInput = unknown> extends Error {
```
MarcMcIntosh, line 9: Odd idea (optional), but i see that vite has insource testing. https://vitest.dev/guide/in-source.html

would it be worth adding an example so we know what the errors look like?

packages/plugin-kit/src/lib/decode.ts
```diff
@@ -0,0 +1,36 @@
...
+	| {
+			value: A;
+			error?: never;
+	  }
+	| {
+			value?: never;
+			error: DecodeError<I>;
+	  };
+
+export const decode = <A, O, I>(
```
MarcMcIntosh, line 17: neat

packages/plugin-kit/src/lib/decodeSliceMachineConfig.ts
```diff
@@ -0,0 +1,39 @@
...
+	}),
+	t.partial({
+		apiEndpoint: t.string,
+		libraries: t.array(t.string),
+		localSliceSimulatorURL: t.string,
+		plugins: t.array(SliceMachineConfigPluginRegistrationCodec),
+	}),
+]);
+
+// TODO: Maybe rename "decode" to "validate". "decode" exposes the `io-ts`
```
MarcMcIntosh, line 33: It would make it more user friendly

MarcMcIntosh (commented):
Left a few comments and I've asked on slack about throwing more friendly error messages when searching for the config.... would adding the path help for that part?

packages/plugin-kit/src/lib/DecodeError.ts
lihbr: Source testing is interesting, my opinion however is that I'd prefer to keep all test at the same location and not start spreading out tests everywhere, happy to reopen if needed!

packages/plugin-kit/src/lib/decodeSliceMachineConfig.ts
lihbr: Since it's not a public API, I think it's fine to preserve `io-ts` semantic, the manager uses `decode*` too, happy to reopen if needed!

packages/plugin-kit/src/lib/DecodeError.ts
lihbr: That's an interesting idea and discussion, I think we don't want to include yet another package in Milestone 1/2, I tracked it for later in this ticket: https://linear.app/prismic/issue/SMX-118/consider-a-slicemachinevalidators-package-to-group-and-share-type

packages/plugin-kit/package.json
lihbr: Everything is on `2.2.20` now

packages/plugin-kit/src/createSliceMachineHelpers.ts
lihbr: I'll plan a pass with Sam regarding error messages (all tracked with TODO comments), those are the same as the one in `packages/manager` for now, happy to reopen if needed!

packages/plugin-kit/src/lib/DecodeError.ts
bapmrl: Thank you!

bapmrl (approved):

PR #849, opened 2023-02-24
refactor(init, manager): move package installation from `init` to `manager`

angeloashmore:
## Types of changes

- [x] Chore (a non-breaking change which is related to package maintenance)

## Description

> **Note**: This PR is a work in progress and may be rejected.

This PR refactors `@slicemachine/init` and `@slicemachine/manager` to move package installation logic from `init` to `manager`.

The refactor was done for a few reasons:

1. Package manager detection is project-level metadata that could be used elsewhere in Slice Machine. As such, it makes sense to make it a method in `ProjectManager` (`manager.project.detectPackageManager()`).
2. We may want to handle package installation in other locations, such as lazily in plugins (think: only install a dependency once it is needed).
3. Project initialization via `@slicemachine/init` requires a few steps, including installing core dependencies (SM, an adapter). It also involves installing a plugin's dependencies and letting a plugin bootstrap any files it needs.
   - We currently couple this step with `@slicemachine/init` using the `command:init` plugin hook.
   - However, we could make it more general by naming it `project:init` and make it something `@slicemachine/manager` handles. That allows us to run the initialization again outside the process of `@slicemachine/init`.

This refactor is mostly complete except that `@slicemachine/init`'s tests don't pass (so maybe it's not mostly done). `execa` doesn't seem to be mocked in `@slicemachine/manager`, which causes package installation to actually happen on the filesystem. If you run the tests, be aware that a `foo` package will be installed into `@slicemachine/init`.

## Checklist:

🦬


lihbr (approved):
LGTM, let's get those tests to work

PR #850, opened 2023-02-24
fix(init): add error handler to segment

MarcMcIntosh:
## Context
https://linear.app/prismic/issue/SM-1035/spike-1day-aauser-i-am-informed-about-the-tracking-policy-when

- [ ] The CI is successful.


danigo99 (approved):
1. Should we create a ticket to back port this to the new plugins architecture also so no one forgets about it? 
2. Maybe you should target the dev branch so that Mathieu can test it 🙂

PR #852, opened 2023-02-25
chore: improve dev environment

angeloashmore:
## Types of changes

- [x] Chore (a non-breaking change which is related to package maintenance)

## Description

This PR makes the following changes to the repository:

- **`start-slicemachine`**: During development (started via `npm run dev`), the `start-slicemachine` command will restart if any file changes are detected. It watches itself, `@slicemachine/manager`, and `@slicemachine/plugin-kit` for changes.
- **`start-slicemachine`**: Supports HMR when working on Slice Machine UI. The previous h3 proxy was replaced with an express proxy middleware that supports WebSockets.
- **Turborepo**: Adds Turborepo to run tasks in parallel in the correct topological order. Tasks are smartly cached.

The Turborepo integration is _mostly_ working with the following exceptions:

- Root-level `npm run dev` still has build order issues. Since `vite build --watch` is a running process that never ends, Turborepo can only run each process in parallel with no dependencies; it does not currently have the capability of running watchers for long-running tasks.
- GitHub Actions has not been updated to cache the Turborepo cache. If we move forward with Turborepo, the cache should be persisted in CI to speed up tasks.

## Checklist:

🐰


.depcheckrc
```diff
@@ -0,0 +1,5 @@
+ignores:
+  [
+    "husky",
```
bapmrl, line 3: ❓ Why do we need to ignore `husky` now?

package.json
```diff
@@ -5,44 +5,31 @@
...
-    "build-essential": "npm --workspace @slicemachine/plugin-kit run build && npm --workspace @slicemachine/manager run build",
-    "audit": "lerna run audit",
-    "lint": "npm --workspaces --if-present run lint",
+    "dev": "turbo run dev",
+    "build": "turbo run build --filter=!e2e-project-next2",
+    "audit": "turbo run audit",
+    "lint": "turbo run lint",
     "lint-staged": "lint-staged",
-    "pretty": "prettier --write .",
-    "prettier:check": "prettier --check .",
```
bapmrl, line 25: ⚠️ The `prettier:check` script is used in `.github/workflows/test-and-deploy-alpha.yml`.

package.json
```diff
@@ -5,44 +5,31 @@
...
   "scripts": {
-    "dev":"concurrently --prefix-colors auto \"npm:dev:*\"",
-    "dev:manager":"npm --workspace @slicemachine/manager run dev",
-    "dev:start-slicemachine":"npm --workspace start-slicemachine run dev",
-    "dev:plugin-kit":"npm --workspace @slicemachine/plugin-kit run dev",
-    "dev:slice-machine-ui":"npm --workspace slice-machine-ui run dev",
-    "dev:e2e-next-2":"cd ./e2e-projects/next2 && npm run dev -- --port 8000",
-    "cypress-setup": "./cypress-setup.sh",
-    "clean-e2e-projects": "git checkout e2e-projects/ && git clean -f e2e-projects/",
-    "prepare": "husky install",
```
bapmrl, line 16: ❓ Why has `husky install` been removed?

packages/start-slicemachine/tsconfig.json
```diff
@@ -14,7 +14,7 @@
 		"forceConsistentCasingInFileNames": true,
 		"jsx": "preserve",
 		"lib": ["esnext", "dom"],
-		"types": ["node"]
+		"types": ["node", "vite/client"]
```
bapmrl, line 16: ❓ Why do we need the `vite/client` types?

bapmrl (commented):
Nice, but I'm not sure it's the right time to add Turbo. What problem(s) does it solve apart from executing the `build` scripts in topological order? It apparently doesn't help for development and I think we should try and reduce the number of new tools we add in the `dev-plugins` Pull Request (which is already huge). No one in the team has experience with Turbo, so it should bring clear value if we want to use it. Is that the case?

packages/init/vite.config.ts
```diff
@@ -11,17 +17,15 @@ export default defineConfig({
 			},
 		},
 		rollupOptions: {
+			output: {
+				exports: "named",
+			},
```
lihbr, line 22: We're not moving forward with Turbo for now, do we need to backport this though @angeloashmore?

packages/start-slicemachine/tsconfig.json
lihbr: It's to have types for `import.meta.*` inside the code :)

lihbr:
Closing in favor of #879 for now

packages/init/vite.config.ts
angeloashmore: If you still see the default/named exports warning in the build output from Rollup, then yes, it could be ported. However, it isn't necessary since it just suppresses the warning.

PR #853, opened 2023-02-27
Deprecate link & date fields mock configs

dani-mp:
[Linear ticket](https://linear.app/prismic/issue/SM-1040/[update-editor]-aauser-i-have-access-to-new-fields-in-the-slice).

Most of the changes come from updating the following libraries:
- "@prismicio/mocks": "2.0.0-alpha.1"
- "@prismicio/types-internal": "2.0.0-alpha.7"

We needed these to update the editor:
- "@prismicio/editor-fields": "0.3.0"
- "@prismicio/editor-ui": "0.3.0"

these have been already reviewed by @arnaudlewis, who knows his types well.

For a quicker review, check for the changes made in:
- Link field widget
- Date field widget

as their mocks have been removed.


packages/slice-machine/lib/models/common/widgets/Widget.ts
```diff
@@ -1,9 +1,9 @@
-import { WidgetTypes } from "@prismicio/types-internal/lib/customtypes/widgets";
 import { IconType } from "react-icons";
 import { AnyObjectSchema } from "yup";
 import { TabField } from "@slicemachine/core/build/models/CustomType";
+import { NestableFieldTypes } from "@prismicio/types-internal/lib/customtypes";
 export interface Widget<F extends TabField, S extends AnyObjectSchema> {
-  TYPE_NAME: WidgetTypes;
+  TYPE_NAME: NestableFieldTypes;
```
arnaudlewis:
```suggestion
  TYPE_NAME: FieldType;
```

packages/slice-machine/lib/utils/index.ts
```diff
@@ -38,7 +38,7 @@ export const ensureWidgetTypeExistence = (
   return false;
 };
 
-export const createDefaultWidgetValues = (TYPE_NAME: WidgetTypes) => ({
+export const createDefaultWidgetValues = (TYPE_NAME: NestableFieldTypes) => ({
```
arnaudlewis: are you sure it is not `FieldType` here instead?

arnaudlewis (changes requested):

arnaudlewis (approved):

packages/client/package.json
```diff
@@ -23,7 +23,7 @@
     "depcheck": "depcheck --config=.depcheckrc"
   },
   "dependencies": {
-    "@prismicio/types-internal": "^1.5.3",
+    "@prismicio/types-internal": "2.0.0-alpha.7",
```
mathieud, line 26: `alpha`? Is there a stable version for the release?

packages/client/src/index.ts
```diff
@@ -3,8 +3,10 @@ import * as t from "io-ts";
 import { fold } from "fp-ts/Either";
 
 import { UserProfile } from "@slicemachine/core/build/models";
-import { SharedSlice } from "@prismicio/types-internal/lib/customtypes/widgets/slices";
-import { CustomType } from "@prismicio/types-internal/lib/customtypes/CustomType";
+import {
+  CustomType,
+  SharedSlice,
+} from "@prismicio/types-internal/lib/customtypes";
```
mathieud, line 9: Did you check if this prevented tree-shaking?

packages/slice-machine/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
```diff
@@ -36,7 +35,7 @@ const mapAvailableAndSharedSlices = (
   }>(
     (acc, { key, value }) => {
       // Shared Slice
-      if (value.type === SlicesTypes.SharedSlice) {
+      if (value.type === "SharedSlice") {
```
mathieud, line 38: What triggered this change? Should this be our new standard? Is it checked by eslint/typescript?

mathieud (approved):

packages/client/package.json
dani-mp: Not yet, but it won't change much, if any. @arnaudlewis can explain more about why it is still considered an alpha.

packages/client/src/index.ts
dani-mp: That's a good question. AFAIK, we don't check this at the moment.

packages/slice-machine/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
dani-mp: It is type-safe, yes. Having to use the constant all the time instead of the string literal was unnecessary indirection when TypeScript protects us in both cases anyway. It's not only the new standard, it's also the only way now because the library doesn't expose those constants anymore.

packages/client/package.json
mathieud: I was worried an alpha (eg unstable) might not install on some cases, but seems marked at stable on npm (https://www.npmjs.com/package/@prismicio/types-internal?activeTab=versions) so we should be ok

PR #856, opened 2023-02-28
Sm 1053/tracking bug sm version

MarcMcIntosh:
## Context
[https://linear.app/prismic/issue/SM-1053/[tracking-bug]-aapm-i-would-like-to-know-which-version-of-sm-users-are](https://linear.app/prismic/issue/SM-1053/%5Btracking-bug%5D-aapm-i-would-like-to-know-which-version-of-sm-users-are)

- [ ] The CI is successful.


cypress/e2e/user-flows/scenario_custom_screenshots.cy.js
```diff
@@ -37,10 +35,14 @@ describe("I am an existing SM user and I want to upload screenshots on variation
 
     screenshotModal
       .verifyImageIsEmpty()
-      .uploadImage(wrongScreenshot)
+      .dragAndDropImage(wrongScreenshot)
```
MarcMcIntosh, line 38: order matters here, because drag and drop doesn't work that well

packages/slice-machine/server/src/api/tracking.ts
```diff
@@ -37,14 +46,23 @@ export function sendEvents(
       });
     }
   } else if (isTrackingEvent(event)) {
+    const maybeNodeVersion =
+      event.name === EventNames.PageView ? { nodeVersion } : {};
```
Heyrows, line 61: 😔 sad that we have to do it that way, but i understand why.
Could you just add a comment. to explain
- why we chose this events
- what kind of extra properties would be tracked in this one ?

Heyrows: Also
- Why do we do that back-end and not frontend, why here ?

packages/slice-machine/server/src/api/tracking.ts
```diff
@@ -37,14 +46,23 @@ export function sendEvents(
...
       properties: {
         ...event.props,
-        nodeVersion: process.versions.node,
+        ...maybeNodeVersion,
       },
       ...(userId !== undefined ? { userId } : { anonymousId }),
-      context: { groupId: { Repository: repositoryName } },
+      context: {
+        app: {
+          name,
```
Heyrows, line 72: would it be possible to add this to group and identify calls as well so the APP is registered on everything we send to Segment. 🙏

Heyrows (commented):
Sounds good, before I approve, could you
- add the comment
- add app version to other calls if possible
- share with us a list of Events that needs to be updated on Segment that i'm sure you updated to test your development. So we don't forget any when going to prod. 👍

PR #861, opened 2023-03-08
port sentry tunnel to plugins

mathieud:
## Context

Current implementation: https://github.com/prismicio/slice-machine/blob/6fcf51a1fbc478357318f7109c7c0ad3a3374963/packages/slice-machine/server/src/api/sentry.ts

- [ ] The CI is successful.

### DIFF
--- a/packages/start-slicemachine/src/lib/createSliceMachineServer.ts
+++ b/packages/start-slicemachine/src/lib/createSliceMachineServer.ts
@@ -18,7 +18,7 @@ import serveStatic from "serve-static";
 import cors from "cors";
 import fetch from "node-fetch";
 import { createStaticFileEventHandler } from "./createStaticFileEventHandler";
-import { default as sentryHandler } from "./sentry";
+import { default as sentryHandler } from "./sentryHandler";
 
 type CreateSliceMachineServerArgs = {
 	sliceMachineManager: SliceMachineManager;
--- a/packages/start-slicemachine/src/lib/createSliceMachineServer.ts
+++ b/packages/start-slicemachine/src/lib/createSliceMachineServer.ts
@@ -18,6 +18,7 @@ import serveStatic from "serve-static";
 import cors from "cors";
 import fetch from "node-fetch";
 import { createStaticFileEventHandler } from "./createStaticFileEventHandler";
+import { default as sentryHandler } from "./sentry";
 
 type CreateSliceMachineServerArgs = {
 	sliceMachineManager: SliceMachineManager;
@@ -75,6 +76,8 @@ export const createSliceMachineServer = async (
 			return {};
 		}),
 	);
+
+	router.add("/api/t", eventHandler(sentryHandler));
 
 	if (process.env.NODE_ENV === "development") {
 		router.get(
--- a/packages/start-slicemachine/src/lib/createSliceMachineServer.ts
+++ b/packages/start-slicemachine/src/lib/createSliceMachineServer.ts
@@ -18,7 +18,7 @@ import serveStatic from "serve-static";
 import cors from "cors";
 import fetch from "node-fetch";
 import { createStaticFileEventHandler } from "./createStaticFileEventHandler";
-import { default as sentryHandler } from "./sentryHandler";
+import { handler as sentryHandler } from "./sentryHandler";
 
 type CreateSliceMachineServerArgs = {
 	sliceMachineManager: SliceMachineManager;
--- a/packages/start-slicemachine/src/lib/sentry.ts
+++ b/packages/start-slicemachine/src/lib/sentry.ts
@@ -1,0 +1,34 @@
+// We tunnel the Sentry calls via the express server to avoid add blockers
+// adapted from https://github.com/getsentry/examples/blob/master/tunneling/nextjs/pages/api/tunnel.js
+
+import * as url from "url";
+const axios = require("axios");
+import { H3Event, readRawBody } from "h3";
+
+async function handler(event: H3Event): Promise<Record<string, never>> {
+	try {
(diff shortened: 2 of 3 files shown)


packages/start-slicemachine/src/lib/sentry.ts
```diff
@@ -0,0 +1,34 @@
+// We tunnel the Sentry calls via the express server to avoid add blockers
+// adapted from https://github.com/getsentry/examples/blob/master/tunneling/nextjs/pages/api/tunnel.js
+
+import * as url from "url";
+const axios = require("axios");
```
hypervillain: Is there a reason to use `require` over `import`?

hypervillain (changes requested):

packages/start-slicemachine/src/lib/sentry.ts
mathieud: Got an error similar to `https://github.com/axios/axios/issues/4231`, and one comment mentioned using `require` instead would work. I tried and it did.
Not sure what the root cause is though

hypervillain (approved):

packages/start-slicemachine/src/lib/createSliceMachineServer.ts
```diff
@@ -18,6 +18,7 @@ import serveStatic from "serve-static";
 import cors from "cors";
 import fetch from "node-fetch";
 import { createStaticFileEventHandler } from "./createStaticFileEventHandler";
+import { default as sentryHandler } from "./sentryHandler";
```
bapmrl: 💡 I'm not sure Lucie and Angelo used default exports in the plugins PR and I personally think it's not good to use them. Named exports provide one consistent way of importing modules in TypeScript.

packages/start-slicemachine/src/lib/sentryHandler.ts
```diff
@@ -0,0 +1,37 @@
+// We tunnel the Sentry calls via the express server to avoid add blockers
```
bapmrl: ❓ Is it Express or `h3`?

bapmrl: ⚠️ "ad blockers" rather than "add blockers".

packages/start-slicemachine/src/lib/sentryHandler.ts
```diff
@@ -0,0 +1,37 @@
...
+// adapted from https://github.com/getsentry/examples/blob/master/tunneling/nextjs/pages/api/tunnel.js
+
+import * as url from "url";
```
bapmrl: 💡 Let's not use wildcard imports to avoid importing unnecessary functions (which could lead to increased bundle sizes).

packages/start-slicemachine/src/lib/sentryHandler.ts
```diff
@@ -0,0 +1,37 @@
...
+import { H3Event, readRawBody } from "h3";
+import fetch from "node-fetch";
+
+async function handler(event: H3Event): Promise<Record<string, never>> {
```
bapmrl: ❓ Would `Promise<void>` or `Promise<unknown>` work to simplify the return type?

packages/start-slicemachine/src/lib/sentryHandler.ts
```diff
@@ -0,0 +1,37 @@
...
+
+		const projectId = (path?.endsWith("/") ? path.slice(0, -1) : path) ?? "";
+
+		const sentryUrl = `https://${host}/api/${projectId}/envelope/`;
+		await fetch(sentryUrl, {
+			method: "POST",
+			body: envelope,
+		});
+	} catch (e) {
+		// TODO add this back when we have the express
```
bapmrl, line 32: ❓ What should be done with this `TODO`?

packages/start-slicemachine/src/lib/sentryHandler.ts
```diff
@@ -0,0 +1,37 @@
+// We tunnel the Sentry calls via the express server to avoid add blockers
+// adapted from https://github.com/getsentry/examples/blob/master/tunneling/nextjs/pages/api/tunnel.js
```
bapmrl: ⚠️ That URL is broken. You could use the `Y` shortcut mentioned by Angelo to create a canonical URL ^^

packages/start-slicemachine/src/lib/sentryHandler.ts
```diff
@@ -0,0 +1,37 @@
...
+
+		const pieces = envelope.split("\n");
+
+		const header = JSON.parse(pieces[0]);
+
+		const { host, path } = url.parse(header.dsn);
+
+		const projectId = (path?.endsWith("/") ? path.slice(0, -1) : path) ?? "";
+
+		const sentryUrl = `https://${host}/api/${projectId}/envelope/`;
```
bapmrl, line 24: ❓ What happens if `projectId` is `""`? It would go inside the `catch` and do nothing?

packages/start-slicemachine/src/lib/sentryHandler.ts
mathieud: Sentry returns `{}` on success, wanted to keep the same behavior

packages/start-slicemachine/src/lib/sentryHandler.ts
mathieud: I'm working on it :-)

packages/start-slicemachine/src/lib/sentryHandler.ts
mathieud: @bapmrl Will handle this with the server part of Sentry

bapmrl (approved):

PR #863, opened 2023-03-09
Add transactional push happy path back

danigo99:
## Context

Ticket: https://linear.app/prismic/issue/SMX-97/aauser-on-the-new-sm-with-plugins-i-can-use-transactional-push

## The Solution

Create TransactionalMerge manager.
Backport apiClient + pushChanges saga to use this new manager.

## Impact / Dependencies

There is one major TODO to update customTypes api client and move the API call there. It is temporarily in the manager for now.

- [ ] The CI is successful.


packages/slice-machine/pages/changes.tsx
```diff
@@ -13,16 +13,19 @@ import {
...
+import { unSyncStatuses, useUnSyncChanges } from "@src/hooks/useUnSyncChanges";
 import { isLoading } from "@src/modules/loading";
 import { LoadingKeysEnum } from "@src/modules/loading/types";
 import useSliceMachineActions from "@src/modules/useSliceMachineActions";
-import {
-  SoftDeleteDocumentsDrawer,
-  HardDeleteDocumentsDrawer,
-  ReferencesErrorDrawer,
-} from "@components/DeleteDocumentsDrawer";
+// import {
```
mathieud, line 20: Are these coming back with the `delete` feature?

packages/slice-machine/pages/changes.tsx
```diff
@@ -32,6 +35,27 @@ const Changes: React.FunctionComponent = () => {
     authStatus,
     isOnline,
   } = useUnSyncChanges();
+
+  type ChangedSlice = { status: ChangesStatus; c: ComponentUI };
```
mathieud: Can we define these somewhere else? And rename `c` (to `slice` and `customType` below, unless it makes sense to have a common type in which case probably `component`)

packages/slice-machine/pages/changes.tsx
```diff
@@ -32,6 +35,27 @@ const Changes: React.FunctionComponent = () => {
...
+  type ChangedCustomType = { status: ChangesStatus; c: CustomTypeSM };
+  const { changedSlices, changedCustomTypes } = useMemo(() => {
+    const changedSlices = unSyncedSlices
+      .map((s) => ({
+        c: s,
+        status: modelsStatuses.slices[s.model.id],
+      }))
+      .filter((s): s is ChangedSlice => unSyncStatuses.includes(s.status)); // FIXME can we sync unSyncStatuses and ChangedSlice?
```
mathieud: Did you look into this one further?

packages/slice-machine/src/apiClient.ts
```diff
@@ -13,17 +13,12 @@ import { ComponentUI } from "@lib/models/common/ComponentUI";
...
-//   PushChangesPayload,
-// } from "@lib/models/common/TransactionalPush";
-// import { Limit } from "@slicemachine/client/build/models/BulkChanges";
 
 import { managerClient } from "./managerClient";
 import { CustomTypeMockConfig } from "@lib/models/common/MockConfig";
 import { Frameworks } from "@lib/models/common/Framework";
 
-/** State Routes **/
+/** State Routes * */
```
mathieud, line 21: These are annoying. Looks like a prettier config issue

packages/slice-machine/src/apiClient.ts
```diff
@@ -239,13 +234,17 @@ export const pushSliceApiClient = async (
   });
 };
 
-// export const pushChanges = (
-//   payload: PushChangesPayload
-// ): Promise<AxiosResponse<InvalidCustomTypeResponse | Limit | null>> => {
-//   return axios.post("/api/push-changes", payload, defaultAxiosConfig);
-// };
+export const pushChanges = async (
+  payload: Parameters<
```
mathieud: You can type the entire function which looks cleaner I think
```
export const pushChanges: SliceMachineManagerClient["transactionalMerge"]["pushChanges"] = ...
```

packages/slice-machine/test/src/modules/pushChangesSaga.test.ts
```diff
@@ -1,562 +1,243 @@
...
+          content:
+            "Something went wrong when pushing your changes. Check your terminal logs.",
+          type: ToasterType.ERROR,
+        })
+      );
+
+      saga.next().isDone();
+    });
+
+    // TODO: unskip when auth error is handled
```
mathieud, line 194: Do we have a clear list of things left to do somewhere?

packages/manager/src/managers/transactionalMerge/TransactionalMerge.ts
```diff
@@ -0,0 +1,190 @@
...
+import * as t from "io-ts";
+import { fold } from "fp-ts/Either";
+
+export class TransactionalMergeManager extends BaseManager {
+	constructor(sliceMachineManager: SliceMachineManager) {
+		super(sliceMachineManager);
+
+		if (!sliceMachineManager.getPrismicAuthManager()) {
+			throw new SliceMachineError(
+				"SliceMachineManager._prismicAuthManager must be set with a PrismicAuthManager instance before instantiating TransactionalMergeManager.",
```
mathieud: I wouldn't mention `SliceMachineManager._prismicAuthManager` in here as it's an implementation detail that could change and become confusing

packages/manager/src/managers/transactionalMerge/TransactionalMerge.ts
```diff
@@ -0,0 +1,190 @@
...
+		args: TransactionalMergeArgs,
+	): Promise<TransactionalMergeReturnType> {
+		try {
+			const allChanges: AllChangeTypes[] = await Promise.all(
+				args.changes.map(async (change) => {
+					if (change.type === "Slice") {
+						switch (change.status) {
+							case "NEW": {
+								const { model } = await this.slices.readSlice({
+									libraryID: "./slices",
```
mathieud: Is this correct?

packages/slice-machine/src/modules/pushChangesSaga/index.ts
```diff
@@ -1,281 +1,186 @@
...
+} from "@lib/models/common/TransactionalPush";
+import { ComponentUI } from "@lib/models/common/ComponentUI";
+import { CustomTypeSM } from "@lib/models/common/CustomType";
+import { trackPushChangesSuccess } from "./trackPushChangesSuccess";
+import Tracker from "@src/tracking/client";
+import { SliceMachineManagerClient } from "@slicemachine/manager/client";
+import { ChangesStatus } from "@lib/models/common/ModelStatus";
+
+export type ChangesPushSagaPayload = PushChangesPayload & {
+  changedSlices: ReadonlyArray<{ c: ComponentUI; status: ChangesStatus }>;
```
mathieud: We can share the type from the `changes` page here

packages/slice-machine/src/modules/pushChangesSaga/index.ts
```diff
@@ -1,281 +1,186 @@
...
+  ["HARD"]: ModalKeysEnum.HARD_DELETE_DOCUMENTS_DRAWER,
+};
+
+export function* changesPushSaga({
+  payload,
+}: ReturnType<typeof changesPushCreator.request>): Generator {
+  const startTime = Date.now();
+  const { changedSlices, changedCustomTypes } = payload;
+
+  const sliceChanges: Parameters<typeof pushChanges>[0]["changes"] =
```
mathieud: The types don't look great.`Parameters<typeof pushChanges>[0]["changes"]`. Could we let Typescript infer them?

mathieud:
Can we remove the `push` functions from the Slice and CustomType manager now?

packages/slice-machine/pages/changes.tsx
danigo99: Yep!

packages/slice-machine/pages/changes.tsx
danigo99: I'll move them somewhere more common. I don't think a shared _"component"_ is a good idea though since this is the name given to slice-like objects

packages/slice-machine/pages/changes.tsx
danigo99: No, good catch. I only searched for _TODO_ 😅

packages/manager/src/managers/transactionalMerge/TransactionalMerge.ts
danigo99: I just copied this auth handling from other managers. Seems to be the common way of doing it

packages/slice-machine/test/src/modules/pushChangesSaga.test.ts
danigo99: I'm creating a ticket for this one, and have updated the delete tickets too

packages/manager/src/managers/transactionalMerge/TransactionalMerge.ts
danigo99: Nope, thanks. Will fix

packages/slice-machine/pages/changes.tsx
danigo99: I'm not ignoring this btw. Spent some time trying to improve it but couldn't find anything better. Have changed this to a TODO since we also spoke about coming back to how we pass model status around in general, if that is the case later then we will re-write this anyway

lihbr (changes requested):
Looking good to me! Just some changes I'd like to ask, see below.

---

I think the new Transactional Merge manager might be too specific of a manager. Will this manager ever have other methods? This new manager might be too tied to the feature itself. Creating a manager per feature doesn't seem scalable/relatable to other managers.

I think it'll make more sense for the `pushChanges` method to be moved to the existing Prismic Repository manager.

- ✔️ Prismic Repository manager already has a similar `pushDocuments` method
- ✔️ "Pushing changes" is ultimately related to the Prismic repository, so I think it makes sense for this method to leave there

---

I think dd3852c58022373590cf1411064a36917d41b7c2 (`Remove push methods from slice and CT managers + unit test`) should be reverted.

`pushSlice` and `pushCustomType` can still prove useful (e.g. we want to have an instance "push" button directly on the slice edition page)

The init command also relies on them today (maybe it can be refactored to use transactional merge but that doesn't seem needed today/can be part of Milestone 3+).

Just for reference, this is also typically the kind of changes that we'll need to be mindful of and respect SemVer in the future to not break user applications.

packages/slice-machine/src/apiClient.ts
```diff
@@ -239,13 +234,12 @@ export const pushSliceApiClient = async (
   });
 };
 
-// export const pushChanges = (
-//   payload: PushChangesPayload
-// ): Promise<AxiosResponse<InvalidCustomTypeResponse | Limit | null>> => {
-//   return axios.post("/api/push-changes", payload, defaultAxiosConfig);
-// };
+export const pushChanges: SliceMachineManagerClient["transactionalMerge"]["pushChanges"] =
```
mathieud: There seem to be a couple of old-push related functions in here (flagged  by a github action)

mathieud (approved):

lihbr (approved):
LGTM should tests pass ☺️

PR #864, opened 2023-03-10
add server side Sentry

mathieud:
- [ ] The CI is successful.


packages/start-slicemachine/src/lib/sentrySetup.ts
```diff
@@ -0,0 +1,57 @@
+import * as Sentry from "@sentry/node";
+import { SliceMachineManager } from "@slicemachine/manager/*";
```
angeloashmore: `"@slicemachine/manager/*"` can be replaced with `"@slicemachine/manager". I'm not sure why `tsserver` uses the `/*` path when auto-importing. Any idea why?

packages/start-slicemachine/src/lib/sentrySetup.ts
```diff
@@ -0,0 +1,57 @@
...
+// TODO disable this in dev, or set env in dev
+// Could become "isProd() || enableSentry()"
+export const enableSentry = (): boolean => !process.env.NO_SENTRY;
+
+export const initSentry = async (
+	manager: SliceMachineManager,
+): Promise<void> => {
+	if (!enableSentry()) {
+		return;
+	}
```
angeloashmore: Is there a reason we need to use a function? Could we determine this condition statically?

For example:

```
if (!process.env.NO_SENTRY) {
  return
}
```

---

If we plan to make `enableSentry()` return a value based on other conditions (e.g. a user's settings that may disable error tracking), we may want to convert that function into a method in the manager. `manager.telemetry.checkIsErrorTrackingAllowed()` could return a boolean.

---

`!process.env.NO_SENTRY` is a double negative, which gets negated again when we run `if (!enableSentry())`. It can be confusing to follow compared to a more explicit `SENTRY_ENABLED` constant. I can't find a better source right now, but [this article briefly describes why double negatives should be avoided](https://dev.to/mcsee/code-smell-51-double-negatives-4fh0).

---

`enableSentry` is a verb that makes me think I can call this function to enable Sentry. Something like `checkIsSentryEnabled()` more closely describes what the function does.

`checkIsSentryEnabled()` is preferred over `isSentryEnabled()` because the former lets you declare a variable named `isSentryEnabled`:

```ts
const isSentryEnabled = checkIsSentryEnabled()
```

packages/start-slicemachine/src/lib/sentrySetup.ts
```diff
@@ -0,0 +1,57 @@
...
+
+	const version = await manager.versions.getRunningSliceMachineVersion();
+	let userId: string | undefined;
+	try {
+		const user = await manager.getPrismicAuthManager().getProfile();
+		userId = user.shortId;
+	} catch {
+		userId = undefined;
```
angeloashmore: This assignment is unnecessary since `userId` defaults to `undefined`.

packages/start-slicemachine/src/lib/sentrySetup.ts
```diff
@@ -0,0 +1,57 @@
...
+): Promise<void> => {
+	if (!enableSentry()) {
+		return;
+	}
+
+	const version = await manager.versions.getRunningSliceMachineVersion();
+	let userId: string | undefined;
+	try {
+		const user = await manager.getPrismicAuthManager().getProfile();
+		userId = user.shortId;
```
angeloashmore: I recognize that this is a subjective opinion, but using `user.shortId` throughout the code is better than saving it in a `userId` variable for a couple of reasons:

- `userId` is a different name than `shortId` and could miscommunicate what the variable holds. Consider that `user` contains a `userId` property, which is different from its `shortId`.
- `user.shortId` shows us exactly where the `shortId` value comes from: the `user` variable. In contrast, `userId` is disconnected from `user` and could theoretically contain any string.

packages/start-slicemachine/src/lib/sentrySetup.ts
```diff
@@ -0,0 +1,57 @@
...
+	manager: SliceMachineManager,
+): Promise<void> => {
+	if (!enableSentry()) {
+		return;
+	}
+
+	const version = await manager.versions.getRunningSliceMachineVersion();
+	let userId: string | undefined;
+	try {
+		const user = await manager.getPrismicAuthManager().getProfile();
```
angeloashmore: `manager.getPrismicAuthManager()` is for low-level use when you need access to the whole `PrismicAuthManager` instance. In this case, we only need one of its methods (`getProfile()`), which is exposed via the `UserManager`.

This version should be used instead:

```ts
manager.user.getProfile()
```

(I know we're missing documentation on these methods. Sorry for the confusion!)

packages/start-slicemachine/src/lib/sentrySetup.ts
```diff
@@ -0,0 +1,57 @@
...
+
+	const config = await manager.project.getSliceMachineConfig();
+
+	const isStableVersion =
+		/^\d+\.\d+\.\d+$/.test(version) && semver.lte("0.1.0", version);
+
+	Sentry.init({
+		dsn:
+			process.env.SENTRY_EXPRESS_DSN ??
+			"https://b673ba8b041d4449a0fb0a38691882dd@o146123.ingest.sentry.io/4504179268845568",
```
angeloashmore: This URL is a constant that should live in a dedicated constants file. Because `start-slicemachine` is relative small, we could just put it in a new `src/constants.ts` file:

```ts
// src/constants.ts

export const SENTRY_EXPRESS_DSN = "https://b673ba8b041d4449a0fb0a38691882dd@o146123.ingest.sentry.io/4504179268845568";
```

packages/start-slicemachine/src/lib/sentrySetup.ts
```diff
@@ -0,0 +1,57 @@
...
+export const enableSentry = (): boolean => !process.env.NO_SENTRY;
+
+export const initSentry = async (
+	manager: SliceMachineManager,
+): Promise<void> => {
+	if (!enableSentry()) {
+		return;
+	}
+
+	const version = await manager.versions.getRunningSliceMachineVersion();
```
angeloashmore: `version` isn't clear about what the version is from, which could be confusing further down in the code. For example, `version` could mean SM version, Sentry version, or `start-slicemachine` version.

`const sliceMachineVersion` is a good alternative.

packages/start-slicemachine/src/lib/sentrySetup.ts
```diff
@@ -0,0 +1,57 @@
...
+
+	const isStableVersion =
+		/^\d+\.\d+\.\d+$/.test(version) && semver.lte("0.1.0", version);
+
+	Sentry.init({
+		dsn:
+			process.env.SENTRY_EXPRESS_DSN ??
+			"https://b673ba8b041d4449a0fb0a38691882dd@o146123.ingest.sentry.io/4504179268845568",
+		release: version,
+		environment: isStableVersion ? process.env.NODE_ENV : "alpha",
```
angeloashmore: When happens if `process.env.NODE_ENV` is not defined? I'm not sure it would be defined in production since developers aren't specifying `NODE_ENV` in their shell.

packages/start-slicemachine/src/lib/sentrySetup.ts
```diff
@@ -0,0 +1,57 @@
...
+	if (enableSentry()) {
+		Sentry.withScope(function (scope) {
+			// TODO: we need the `procedurePath` fron payload here
+			scope.setTransactionName(event.path);
+			Sentry.captureException(error);
+		});
+	}
+
+	sendError(event, error);
+};
```
angeloashmore: This `onError()` function could live directly in `createSliceMachineServer()`. Moving it there more clearly shows the connection between an error thrown in `h3` and the data sent to Sentry. This is especially true since `onError()` is a generic name that does not convey that it is related to Sentry.

```ts
const app = createApp({
  onError: (error, event) => {
    if (args.manager.telemetry.checkIsErrorTrackingAllowed()) {
      // The rest of the Sentry code...
    }
  }
})
```

---

In general, files in `lib` should only export one function. The name of the file should match the name of the export so it is clear where to find functions throughout the project.

In this case, we should rename the file to `lib/initSentry.ts`.

packages/start-slicemachine/src/lib/createSliceMachineServer.ts
```diff
@@ -79,6 +85,22 @@ export const createSliceMachineServer = async (
...
+	router.add(
+		"/api/error",
+		eventHandler(() => {
+			return createError({
+				status: 500,
+				message: "Test Error",
+			});
+		}),
+	);
+
```
angeloashmore: Is this endpoint still needed? `"Test Error"` sounds like something used during development.

packages/start-slicemachine/src/lib/createSliceMachineServer.ts
```diff
@@ -79,6 +85,22 @@ export const createSliceMachineServer = async (
 
 	router.add("/api/t", eventHandler(sentryHandler));
 
+	router.add(
+		"/api/crash",
+		eventHandler(() => {
+			throw new Error("Uncaught Error");
+		}),
+	);
```
angeloashmore: What is this endpoint used for? If we are catching errors and re-throwing this custom "Uncaught Error", are we losing metadata from the originally thrown error somewhere in the process?

angeloashmore (changes requested):
Hey @mathieud, I left a bunch of comments through the review. Most of them are about naming, with others commenting about the locations of logic.

Let me know if you have any questions about what I posted! 🙂

packages/start-slicemachine/src/lib/createSliceMachineServer.ts
mathieud: I used these urls to simulate a manager returning an error or throwing one. Will be deleted before we merge

packages/start-slicemachine/src/lib/sentrySetup.ts
mathieud: I made a change keeping all the Sentry code in its file (modifying the app's `onError` instead).
It allows to keep all the Sentry code in its file and not leak anywhere else in the app.

What do you think?

packages/start-slicemachine/src/lib/sentrySetup.ts
mathieud: Made some changes, let me know what you think.

Also I found [this](https://github.com/prismicio/slice-machine/blob/71f0fd828d77e43344970731cff0ce54ff81ee10/packages/manager/src/constants/API_TOKENS.ts#L8). Could I use it to check for production env? Or is based on the remote endpoint I will have false positives for prod?

packages/start-slicemachine/src/lib/sentrySetup.ts
mathieud: About the named export matching the filename, is it something we would want to automate? Found https://github.com/ekwoka/eslint-plugin-filename-export with a quick search, it looks like it does what we want. Should be relatively easy to adapt if we want a separate behaviour

packages/start-slicemachine/src/lib/setupSentry.ts
```diff
@@ -0,0 +1,70 @@
...
+import semver from "semver";
+import { App, H3Event, sendError } from "h3";
+import { SliceMachineManager } from "@slicemachine/manager";
+import { SENTRY_EXPRESS_DSN } from "../constants";
+
+export const setupSentry = async (
+	manager: SliceMachineManager,
+	app: App,
+): Promise<void> => {
+	!isDevEnv() || process.env.SENTRY_AUTH_TOKEN !== undefined;
```
angeloashmore: Should this line be removed?

packages/start-slicemachine/src/lib/setupSentry.ts
```diff
@@ -0,0 +1,70 @@
...
+	if (!isSentryEnabled()) {
+		return;
+	}
+
+	const sliceMachineVersion =
+		await manager.versions.getRunningSliceMachineVersion();
+
+	let user;
```
angeloashmore: We're missing a type annotation here:

```ts
let user: PrismicUserProfile | undefined;
```

The type is exported from `@slicemachine/manager`:

```ts
import { PrismicUserProfile } from "@slicemachine/manager";
```

(Small nitpick, but since the value is the user's profile, not necessarily the "user", could we rename the variable to `profile` or `userProfile`?)

packages/start-slicemachine/src/lib/setupSentry.ts
```diff
@@ -0,0 +1,70 @@
...
+
+	Sentry.init({
+		dsn: process.env.SENTRY_EXPRESS_DSN ?? SENTRY_EXPRESS_DSN,
+		release: sliceMachineVersion,
+		environment: isStableVersion
+			? process.env.NODE_ENV ?? "production"
+			: "alpha",
+	});
+
+	Sentry.setUser({ id: user?.shortId });
```
angeloashmore: This could be slightly more readable if we wrapped it in an `if` statement instead of using optional chaining. The `if` statement more clearly states that the user is set only if `user` is populated. This is just my personal opinion, so feel free to ignore if you feel otherwise.

```ts
if (user) {
  Sentry.setUser({ id: user.shortId })
}
```

packages/start-slicemachine/src/lib/setupSentry.ts
```diff
@@ -0,0 +1,70 @@
...
+};
+
+const isDevEnv = () =>
+	["development", "test"].includes(process.env.NODE_ENV ?? "");
+
+// We want Sentry enabled
+// - automatically in production (not dev or test)
+// - or if explicitely set in the environment
+const isSentryEnabled = (): boolean =>
+	!isDevEnv() || process.env.SENTRY_AUTH_TOKEN !== undefined;
```
angeloashmore: It looks like `isDevEnv()` is only used in `isSentryEnabled()`. We could clean up the code a bit and inline the development check:

```ts
["development", "test"].includes(process.env.NODE_ENV ?? "") || process.env.SENTRY_AUTH_TOKEN !== undefined;
```

Again, a stylistic choice, so I'll leave it up to you if you want to make this change. 🙂

packages/start-slicemachine/src/lib/setupSentry.ts
```diff
@@ -0,0 +1,70 @@
...
+
+const onError = (error: Error, event: H3Event): void => {
+	if (isSentryEnabled()) {
+		Sentry.withScope(function (scope) {
+			// TODO: we need the `procedurePath` fron payload here
```
angeloashmore: Could we check the error that is thrown instead? It should contain some metadata in the Error's `name` and `message` properties. Checking `error` would make this a more universal `onError` handler that also handles errors not thrown by r19.

`procedurePath` is an implementation detail of r19 that could change at any time. r19 does not expect developers to read the raw network request.

If the error does _not_ contain the metadata we need, let me know and we can try to make adjustments to r19 so it contains the metadata.

angeloashmore (commented):
Hey Mathieu, the changes look good! I have a few new comments that should be easy to address. After those fixes, I think we're good to merge.

The r19/`procedurePath` TODO could be quick if the thrown `error` contains the metadata we need. If it doesn't, let me know and we can update the library. We're most likely going to change the `/_manager` endpoint to include the full procedure path in the URL, like: `/_manager/users/getProfile`, which would make it easy to log which URL threw an error.

packages/start-slicemachine/src/lib/setupSentry.ts
```diff
@@ -0,0 +1,70 @@
...
+		user = await manager.user.getProfile();
+	} catch {
+		// not logged in
+		// user stays undefined
+	}
+
+	const config = await manager.project.getSliceMachineConfig();
+
+	const isStableVersion =
+		/^\d+\.\d+\.\d+$/.test(sliceMachineVersion) &&
```
lihbr: Maybe this could use `semver.prerelease` instead to know if any prerelease is present

packages/start-slicemachine/src/lib/setupSentry.ts
```diff
@@ -0,0 +1,70 @@
...
+		semver.lte("0.1.0", sliceMachineVersion);
+
+	Sentry.init({
+		dsn: process.env.SENTRY_EXPRESS_DSN ?? SENTRY_EXPRESS_DSN,
```
lihbr: Maybe this logic could be done in the constant file (similar to https://github.com/prismicio/slice-machine/blob/71f0fd828d77e43344970731cff0ce54ff81ee10/packages/manager/src/constants/API_TOKENS.ts, but simpler), this way if this token is used elsewhere, then it's consistent

packages/start-slicemachine/src/lib/setupSentry.ts
```diff
@@ -0,0 +1,70 @@
...
+import { SENTRY_EXPRESS_DSN } from "../constants";
+
+export const setupSentry = async (
+	manager: SliceMachineManager,
+	app: App,
+): Promise<void> => {
+	!isDevEnv() || process.env.SENTRY_AUTH_TOKEN !== undefined;
+	if (!isSentryEnabled()) {
+		return;
+	}
```
lihbr: Maybe we should also add an additional `enabled` getter as part of the `SliceMachineManager.telemetry` manager so that people preference on telemetry is also honored for Sentry

packages/start-slicemachine/src/lib/createSliceMachineServer.ts
lihbr: I assume we can delete it now? @mathieud

packages/start-slicemachine/src/lib/createSliceMachineServer.ts
lihbr: Same as above @mathieud?

lihbr (approved):

PR #865, opened 2023-03-13
fix iframe connection

scvsoft-ezequielsuthering:
## Context

We implemented the `slice-simulator` page on a `Next.JS` app that loads global app data on each page load to configure some internal resources. The issue is that this data loading takes some time and when we try to simulate a slice on the `slice-machine-ui`, it redirects to the page that loads the `slice-simulator` Iframe and it always fails to connect the iframe.

## The Solution

Reviewing the code, we found that the `connectToSimulatorIframe` function inside `packages/slice-machine/src/modules/simulator/index.ts` has a static timeout value, and it put to race 3 events (timeout, iframeCheckKO, and iframeCheckOk) and it keeps only the first winner to run the remaining function code. If the winner is `timeout` or `iframeCheckKO` it put the iframe connection in what it looks like a `failure status` if none of these is the winner then put the iframe connection in what it looks like a `success status`. 

_**What we changed:**_

1) We increased the static timeout value from 5000 to 20000. In addition we think it will be good to have this timeout value available in the sm.json file to be configured.

2) We removed the piece that receives the `timeout` and `iframeCheckKO` events if some of these wins the race.

3) We added the piece that receives the `iframeCheckOk` event if it wins the race.

4) We changed the condition that was saying ` if timeout or iframeCheckKO lets fail else lets success ` to ` if iframeCheckOk lets success else lets fail `

5) We removed the unused import `CallEffect`

After these changes the `slice-simulator` worked consistently for all our team members.

- [ ] The CI is successful.

### DIFF
--- a/packages/slice-machine/src/modules/simulator/index.ts
+++ b/packages/slice-machine/src/modules/simulator/index.ts
@@ -12,7 +12,6 @@ import {
   race,
   take,
   delay,
-  CallEffect,
   SagaReturnType,
 } from "redux-saga/effects";
 import {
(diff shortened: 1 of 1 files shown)


hypervillain:
Thanks a lot Ezequiel! This definitely makes sense. We're in the process of releasing a large architecture change in our codebase at the moment and will integrate your fix with it

lihbr:
Hey there, just letting you know we backported your changes in this commit (with credits): [`2b673f4` (#911)](https://github.com/prismicio/slice-machine/pull/911/commits/2b673f4ababf51760f13d1cc809732cb8670a69c)

It will be part of our next release (early May)

Thank you so much for your contribution and patience again! Let us know if anything~

PR #868, opened 2023-03-17
remove mockConfig globally + reset MockSlice

hypervillain:
https://linear.app/prismic/issue/SMX-82/[bigmerge]-aadev-i-want-to-get-rid-of-mocksconfig-and-display-the

- remove occurrences of mock config everywhere
- remove mock forms in field editor and replace with deprecation component
- rewrite call to MockConfig to ignore mockConfig arguments


lihbr (approved):
LGTM, however, it'll be nice for someone with more experience than me to also have a look at it as I likely missed a few things~

PR #872, opened 2023-03-20
expect all existing TS errors and restore build + CI type checking

danigo99:
## Context

Ticket: https://linear.app/prismic/issue/SMX-93/[bigmerge]-remove-the-ignore-type-error-on-build

## The Solution

Add ts-expect-error to all existing errors to unblock type-checking on the rest of the codebase.
These are mostly in tests and are likely to not be fixed any time soon
The rest of them can be incrementally adjusted.
This solution also blocks any more new errors (instead of ignoring errors at a global level somewhere).

- [ ] The CI is successful.

### DIFF
--- a/packages/slice-machine/pages/_app.tsx
+++ b/packages/slice-machine/pages/_app.tsx
@@ -117,6 +117,7 @@ function MyApp({
               <LoadingPage />
             ) : (
               <Provider store={smStore.store}>
+                {/** @ts-expect-error TS(2322) FIXME: Type 'SingletonRouter' is not assignable to type '... Remove this comment to see the full error message */}
                 <ConnectedRouter Router={Router}>
                   <PersistGate loading={null} persistor={smStore.persistor}>
                     <ComponentLayout>
--- a/packages/slice-machine/src/apiClient.ts
+++ b/packages/slice-machine/src/apiClient.ts
@@ -27,6 +27,7 @@ export const getState = async (): Promise<ServerState> => {
   // transform the data to something SM recognizes.
   const state: ServerState = {
     ...rawState,
+    // @ts-expect-error TS(2322) FIXME: Type '{ framework: Frameworks; mockConfig: CustomT... Remove this comment to see the full error message
     env: {
       ...rawState.env,
       framework: rawState.env.framework as Frameworks,
--- a/packages/slice-machine/test/components/CreateCustomTypeModal.test.tsx
+++ b/packages/slice-machine/test/components/CreateCustomTypeModal.test.tsx
@@ -1,9 +1,6 @@
 // @vitest-environment jsdom
 
 import { describe, test, afterEach, expect, vi } from "vitest";
-import React from "react";
-// @ts-expect-error TS(6133) FIXME: 'rest' is declared but its value is never read.
-import { rest } from "msw";
 import SegmentClient from "analytics-node";
 
 import { render, fireEvent, act } from "../__testutils__";
@@ -22,8 +19,7 @@ describe("CreateCustomTypeModal", () => {
   div.setAttribute("id", "__next");
   document.body.appendChild(div);
 
-  // @ts-expect-error TS(6133) FIXME: 'ctx' is declared but its value is never read.
-  test("when a slice is created the tracker should be called", async (ctx) => {
+  test("when a slice is created the tracker should be called", async () => {
     const fakeId = "testing_id";
     const fakeName = "testing-name";
     const fakeRepo = "foo";
@@ -45,18 +41,21 @@ describe("CreateCustomTypeModal", () => {
     const submitButton = document.querySelector('[type="submit"]');
 
     await act(async () => {
-      // @ts-expect-error TS(2345) FIXME: Argument of type 'Element | null' is not assignabl... Remove this comment to see the full error message
-      fireEvent.change(labelInput, { target: { value: fakeName } });
+      if (labelInput) {
+        fireEvent.change(labelInput, { target: { value: fakeName } });
+      }
     });
 
     await act(async () => {
-      // @ts-expect-error TS(2345) FIXME: Argument of type 'Element | null' is not assignabl... Remove this comment to see the full error message
-      fireEvent.change(idInput, { target: { value: fakeId } });
+      if (idInput) {
+        fireEvent.change(idInput, { target: { value: fakeId } });
+      }
     });
 
     await act(async () => {
-      // @ts-expect-error TS(2345) FIXME: Argument of type 'Element | null' is not assignabl... Remove this comment to see the full error message
-      fireEvent.click(submitButton);
+      if (submitButton) {
+        fireEvent.click(submitButton);
+      }
     });
 
     expect(SegmentClient.prototype.track).toHaveBeenCalledOnce();
--- a/packages/slice-machine/test/components/CreateCustomTypeModal.test.tsx
+++ b/packages/slice-machine/test/components/CreateCustomTypeModal.test.tsx
@@ -2,6 +2,7 @@
 
 import { describe, test, afterEach, expect, vi } from "vitest";
 import React from "react";
+// @ts-expect-error TS(6133) FIXME: 'rest' is declared but its value is never read.
 import { rest } from "msw";
 import SegmentClient from "analytics-node";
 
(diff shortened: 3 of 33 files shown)


lihbr (approved):
LGTM, nice job!

packages/slice-machine/pages/_app.tsx
```diff
@@ -117,6 +117,7 @@ function MyApp({
               <LoadingPage />
             ) : (
               <Provider store={smStore.store}>
+                {/** @ts-expect-error TS(2322) FIXME: Type 'SingletonRouter' is not assignable to type '... Remove this comment to see the full error message */}
                 <ConnectedRouter Router={Router}>
```
bapmrl, line 121: I think this error can be fixed by removing `Router={Router}`, but Marc told me it might re-introduce a very hard to track issue related to routing (however, I don't understand why as `Router` is already the default value of `ConnectedRouter`).

packages/slice-machine/test/components/CreateCustomTypeModal.test.tsx
```diff
@@ -2,6 +2,7 @@
 
 import { describe, test, afterEach, expect, vi } from "vitest";
 import React from "react";
+// @ts-expect-error TS(6133) FIXME: 'rest' is declared but its value is never read.
 import { rest } from "msw";
```
bapmrl: This import can be removed if it's not being used.

packages/slice-machine/test/components/CreateCustomTypeModal.test.tsx
```diff
@@ -21,17 +22,20 @@ describe("CreateCustomTypeModal", () => {
   div.setAttribute("id", "__next");
   document.body.appendChild(div);
 
+  // @ts-expect-error TS(6133) FIXME: 'ctx' is declared but its value is never read.
   test("when a slice is created the tracker should be called", async (ctx) => {
```
bapmrl: `ctx` can be removed if it's not being used.

packages/slice-machine/test/components/CreateCustomTypeModal.test.tsx
```diff
@@ -41,14 +45,17 @@ describe("CreateCustomTypeModal", () => {
     const submitButton = document.querySelector('[type="submit"]');
 
     await act(async () => {
+      // @ts-expect-error TS(2345) FIXME: Argument of type 'Element | null' is not assignabl... Remove this comment to see the full error message
```
bapmrl: To fix the various `TS(2345)` errors we could simply conditionally call `fireEvent.change` (as `submitButton` can be `null` if the selector doesn't match any button).

packages/slice-machine/test/lib/mock/Slice.test.ts
```diff
@@ -334,10 +334,12 @@ describe.skip("MockSlice", () => {
       },
     ];
     const sliceDiff: SliceDiff = {
+      // @ts-expect-error TS(2322) FIXME: Type '"updated"' is not assignable to type 'DiffOp... Remove this comment to see the full error message
```
bapmrl: This error can probably be fixed with `as const` or an import to the enum if `op` is an enum.

packages/slice-machine/test/pages/cts.test.tsx
```diff
@@ -12,6 +12,7 @@ import {
 import React from "react";
 import Router from "next/router";
 import mockRouter from "next-router-mock";
+// @ts-expect-error TS(6133) FIXME: 'rest' is declared but its value is never read.
 import { rest } from "msw";
```
bapmrl: The import to `rest` can be removed if it's not being used anywhere.

packages/slice-machine/test/pages/cts.test.tsx
```diff
@@ -131,6 +132,7 @@ describe("Custom Type Builder", () => {
     },
   ];
 
+  // @ts-expect-error TS(6133) FIXME: 'ctx' is declared but its value is never read.
```
bapmrl: `ctx` can be removed.

packages/slice-machine/test/pages/cts.test.tsx
```diff
@@ -226,6 +231,7 @@ describe("Custom Type Builder", () => {
     );
   });
 
+  // @ts-expect-error TS(6133) FIXME: 'ctx' is declared but its value is never read.
   test("should send a tracking event when the user adds a slice", async (ctx) => {
```
bapmrl: `ctx` can be removed.

packages/slice-machine/test/pages/simulator.test.tsx
```diff
@@ -2,6 +2,7 @@
 
 import { describe, test, afterEach, expect, beforeAll, vi } from "vitest";
 import { render, fireEvent, act, waitFor } from "../__testutils__";
+// @ts-expect-error TS(6133) FIXME: 'rest' is declared but its value is never read.
 import { rest } from "msw";
```
bapmrl: `import { rest } from "msw"` can be removed.

packages/slice-machine/test/pages/slices.test.tsx
```diff
@@ -11,6 +11,7 @@ import {
 } from "vitest";
 import React from "react";
 import mockRouter from "next-router-mock";
+// @ts-expect-error TS(6133) FIXME: 'rest' is declared but its value is never read.
 import { rest } from "msw";
```
bapmrl: `msw` import can be removed.

packages/slice-machine/test/pages/slices.test.tsx
```diff
@@ -174,11 +177,13 @@ describe("slices", () => {
 
     const createOneButton = document.querySelector('[data-cy="create-slice"]');
     await act(async () => {
+      // @ts-expect-error TS(2345) FIXME: Argument of type 'Element | null' is not assignabl... Remove this comment to see the full error message
```
bapmrl: `fireEvent.change` can be conditionally called.

packages/slice-machine/test/src/modules/availableCustomTypes/index.test.ts
```diff
@@ -33,13 +33,15 @@ const dummyCustomTypesState: AvailableCustomTypesStoreType = {};
...
+      // @ts-expect-error TS(2345) FIXME: Argument of type '{}' is not assignable to paramet... Remove this comment to see the full error message
       expect(availableCustomTypesReducer(dummyCustomTypesState, {})).toEqual(
         dummyCustomTypesState
       );
     });
 
     it("should return the initial state if no matching action", () => {
       expect(
+        // @ts-expect-error TS(2322) FIXME: Type '"NO.MATCH"' is not assignable to type '"CUST... Remove this comment to see the full error message
         availableCustomTypesReducer(dummyCustomTypesState, { type: "NO.MATCH" })
```
bapmrl, line 45: Does this work with `as const`?

packages/slice-machine/src/apiClient.ts
```diff
@@ -27,6 +27,7 @@ export const getState = async (): Promise<ServerState> => {
   // transform the data to something SM recognizes.
   const state: ServerState = {
     ...rawState,
+    // @ts-expect-error TS(2322) FIXME: Type '{ framework: Frameworks; mockConfig: CustomT... Remove this comment to see the full error message
```
mathieud, line 30: Is this one real or due to a build issue on your machine?

packages/slice-machine/test/src/modules/selectedCustomType/reducer.test.ts
```diff
@@ -72,13 +76,15 @@ describe("[Selected Custom type module]", () => {
     });
     it("should update the custom type state given CUSTOM_TYPE/CREATE_TAB action", () => {
       const newTabId = "Tab1";
+      // @ts-expect-error TS(2531) FIXME: Object is possibly 'null'.
       const previousTabLength = dummyCustomTypesState.model.tabs.length;
```
bapmrl: Can we use `?.`

bapmrl (commented):
Nice, I've added some quick suggestions, but I don't think it should be blocking merging this pull request.

packages/slice-machine/test/src/modules/availableCustomTypes/index.test.ts
danigo99: No, because this is purposefully testing a wrong action. So the type is right.

packages/slice-machine/src/apiClient.ts
danigo99: It's real. The type in the manager comes from a library @antfu/ni, which contains a few more options than the one in SM UI. At some point these should be aligned, but I've seen a lot of TODO's around getState so didn't want to be too preemptive with the change.

Maybe @angeloashmore can provide more clarity at some point since those TODOs seem to be from him 🙂

PR #873, opened 2023-03-20
fix(slice-machine-ui): consistently display custom types and slices

lihbr:
## Context

SMX-106

## The Solution

This PR ensures custom types and slices are displayed consistently within the UI lists.

I think it makes more sense to keep the sorting on the component level, this way the component is in control of its own display ordering, that's why I implemented most of the sorting on each page rather than on the manager itself.

- [x] The CI is successful.


danigo99 (approved):
Looks good! 👍

PR #860, opened 2023-03-07
feat: track SM UI events

bapmrl:
## Context

Corresponding Linear issue: [[BigMerge] AAdev I want to send tracking events via Telemetry](https://linear.app/prismic/issue/SMX-81/[bigmerge]-aadev-i-want-to-send-tracking-events-via-telemetry).

## The Solution

1. Remove the `packages/slice-machine/src/tracking/client.ts` module.
2. Remove the `/api/s` route.
3. Use the new `TelemetryManager` for all tracking events.

## Impact / Dependencies

We still need to ensure our Segment tracking plans are unified and not split between `@slicemachine/init` and `slice-machine-ui`.

## Checklist before requesting a review
- [x] I hereby declare my code ready for review.
- [x] If it is a critical feature, I have added tests.
- [ ] The CI is successful.
- [ ] If there could backward compatibility issues, it has been discussed and planned.

PR #874, opened 2023-03-21
fix: identify and group

bapmrl:
## Context

This pull request fixes some changes introduced in [feat: track SM UI events](https://github.com/prismicio/slice-machine/pull/860) (which corresponds to the [[BigMerge] AAdev I want to send tracking events via Telemetry](https://linear.app/prismic/issue/SMX-81/[bigmerge]-aadev-i-want-to-send-tracking-events-via-telemetry) Linear issue).

## The Solution

1. Remove the `IdentifyUser` event and use the `TelemetryManager.identify` method instead.
2. Remove the `GroupLibraries` event and use the `TelemetryManager.group` method instead.
3. The `TelemetryManager.initTelemetry` method now expects `{ appName, appVersion }`.
4. Remove `slicemachineVersion` and `version` in all `SegmentEvent`s.
5. Add `nodeVersion` in the `properties` of the `SliceMachine Page View` event.
6. Refactor `mockSliceMachineUIDirectory`.
7. Namespace telemetry methods in the `packages/slice-machine/src/apiClient.ts` module.

- [ ] The CI is successful.


packages/manager/src/managers/telemetry/TelemetryManager.ts
```diff
@@ -92,6 +115,11 @@ export class TelemetryManager extends BaseManager {
 			payload.context.groupId.Repository = repository;
 		}
 
+		if (event === "page-view") {
+			payload.properties ||= {};
```
lihbr: Just double check, but I think `properties` is always defined (line 99), no need to initialize it again here

packages/manager/src/managers/telemetry/TelemetryManager.ts
```diff
@@ -117,20 +145,22 @@ export class TelemetryManager extends BaseManager {
...
 	identify(args: TelemetryManagerIdentifyArgs): Promise<void> {
+		const payload = {
+			userId: args.userID,
+			anonymousId: this._anonymousID,
+			integrations: {
+				Intercom: {
+					user_hash: args.intercomHash,
+				},
+			},
+			context: this._context,
```
lihbr: Maybe spread it like in the `track` method to avoid using the reference

packages/manager/src/managers/telemetry/TelemetryManager.ts
```diff
@@ -92,6 +115,11 @@ export class TelemetryManager extends BaseManager {
 			payload.context.groupId.Repository = repository;
 		}
 
+		if (event === "page-view") {
+			payload.properties ||= {};
+			payload.properties.nodeVersion = process.versions.node;
+		}
```
lihbr: I would avoid starting adding per-event exceptions here, can this be moved to where page view events are sent? Or added by default on all events? (since frontend might not have access to node version)

packages/start-slicemachine/src/StartSliceMachineProcess.ts
```diff
@@ -63,7 +63,12 @@ export class StartSliceMachineProcess {
 	 * Runs the process.
 	 */
 	async run(): Promise<void> {
-		await this._sliceMachineManager.telemetry.initTelemetry();
+		const appVersion =
+			await this._sliceMachineManager.versions.getRunningSliceMachineVersion();
+		await this._sliceMachineManager.telemetry.initTelemetry({
+			appName: "slice-machine-ui",
```
lihbr: Maybe just make it a constant in the constant file next to this file

lihbr (changes requested):
Just some small things and you're good to go ☺️

lihbr (approved):

PR #876, opened 2023-03-21
fix(start-slicemachine): let unauthorized users start slicemachine

lihbr:
## Context

SMX-105 & SMX-79

- [ ] The CI is successful.


packages/start-slicemachine/src/StartSliceMachineProcess.ts
```diff
@@ -115,12 +115,14 @@ export class StartSliceMachineProcess {
...
 		}
-		await Promise.all([
-			profile
-				? this._sliceMachineManager.user.refreshAuthenticationToken()
-				: undefined,
-			profile ? this._sliceMachineManager.screenshots.initS3ACL() : undefined,
-		]);
+		if (profile) {
+			await Promise.allSettled([
+				// noop - We'll try again when needed.
```
bapmrl, line 133: I don't understand these comments and why there's no `try`/`catch` (as I understood the `profile` may not be authorised to access the resource).

lihbr: Sure! Here we're performing two things:

1. Refreshing the user auth token to keep its session active longer
2. Initing its S3 ACL for the current repository

Both those operations yield promises.

If 1. fails, it's not a big deal, the user session might expire earlier

If 2. fails, it's not a big deal either, it might be because the user doesn't have access to this repository S3 ACL, in any case, this will be attempted again before uploading screenshots. The reason we're doing it here is to attempt to later speed up the screenshot upload process by having this step already performed.

---

There is no `try`/`catch` because `allSettled` catches and returns promise results whether they resolve or reject: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled#return_value

I used `allSettled` over `all` here in case we need to retrieve the promise results and analyze them later.

That thing said, I can refactor to `all` + `try`/`catch` if you prefer it this way :)

---

Let me know if anything :)

bapmrl: Thank you! It's clear and we can keep this implementation.

bapmrl (approved):

PR #879, opened 2023-03-23
build(start-slicemachine): support websocket for development (SMX-99)

lihbr:
## Context

SMX-99

- [ ] The CI is successful.


packages/adapter-next/package.json
```diff
@@ -82,6 +82,7 @@
 		"@vitejs/plugin-react": "^3.1.0",
 		"@vitest/coverage-c8": "^0.27.3",
 		"better-npm-audit": "^3.7.3",
+		"depcheck": "^1.4.3",
```
bapmrl, line 85: 💡 I think we can remove `depcheck` from the root `package.json` file.

packages/start-slicemachine/package.json
```diff
@@ -26,7 +26,7 @@
 	],
 	"scripts": {
 		"build": "vite build",
-		"dev": "vite build --watch",
+		"dev": "vite build --watch --mode development",
```
bapmrl, line 29: ❓ Should `--mode development` be added to all `dev` scripts?

packages/start-slicemachine/src/lib/sentryFrontendTunnel.ts
```diff
@@ -1,39 +1,35 @@
 // We tunnel the Sentry calls via the h3 server to avoid ad blockers
```
bapmrl: ⚠️ "Express" and not "h3" 🙂

packages/start-slicemachine/src/lib/createSliceMachineExpressApp.ts
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * The Slice Machine app is served from the project's Slice Machine module
+ * installation.
+ *
+ * If `NODE_ENV` is `development`, the Slice Machine app is served via a proxy
+ * to Next.js's development server. Slice Machine must be running on port 3000.
+ *
+ * @param args - Configuration for the server.
+ *
+ * @returns A standard `node:http` server.
```
bapmrl: ❓ Should we write "Express" instead of `node:http`?

bapmrl (approved):

packages/adapter-next/package.json
lihbr: Yes!

packages/start-slicemachine/package.json
lihbr: No, it's only for this one because we rely on `import.meta` in the `start-slicemachine` script to detect development mode, adding this flag here allows us to tell Vite to keep that part of the code while developing instead of the production one :)

packages/start-slicemachine/src/lib/sentryFrontendTunnel.ts
lihbr: Done

packages/start-slicemachine/src/lib/createSliceMachineExpressApp.ts
lihbr: Updated

packages/start-slicemachine/package.json
bapmrl: Okay, it's clear. Thank you 🙂

PR #882, opened 2023-03-24
chore: update Next.js to version 13.2.3

bapmrl:
## Context

Completes DT-1101 (previously SMX-98).

- [x] The CI is successful.


lihbr (approved):
LGTM, but do we need all those `legacyBehavior`? Cannot we migrate easily? (I'm not familiar enough with Next.js to answer that)

As discussed together, we'll release that in an upcoming version, not 1.0.0

xrutayisire:
@bapmrl, should we merge now?

PR #886, opened 2023-03-27
fix: output telemetry logs only in DEV

bapmrl:
## Context

Fixes an issue, related to SMX-81, where telemetry logs where always output.

## The Solution

1. Output telemetry logs only in development mode (using `import.meta.env.mode`).

- [ ] The CI is successful.

### DIFF
--- a/packages/start-slicemachine/package.json
+++ b/packages/start-slicemachine/package.json
@@ -26,7 +26,7 @@
 	],
 	"scripts": {
 		"build": "vite build",
-		"dev": "vite build --watch",
+		"dev": "vite build --watch --mode development",
 		"format": "prettier --write .",
 		"lint": "eslint --ext .js,.ts .",
 		"prepublishOnly": "npm run build",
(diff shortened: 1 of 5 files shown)


packages/start-slicemachine/package.json
```diff
@@ -26,7 +26,7 @@
 	],
 	"scripts": {
 		"build": "vite build",
-		"dev": "vite build --watch --mode development",
```
lihbr: From my understanding, we need to keep this (in both packages) because we're not using `vite dev` to develop but `vite build` with a watch flag, which uses `production` mode as a default: https://vitejs.dev/guide/env-and-mode.html#modes

lihbr (changes requested):

packages/start-slicemachine/package.json
bapmrl: You're right. Sorry, I went to fast and forgot we were using `vite build --watch` and not just `vite` 🙂

lihbr (approved):

PR #887, opened 2023-03-27
fix(manager, slice-machine-ui): compound M1 review

lihbr:
## Context

SMX-82, SMX-97, SMX-105

- [ ] The CI is successful.

### DIFF
--- a/packages/manager/src/managers/SliceMachineManager.ts
+++ b/packages/manager/src/managers/SliceMachineManager.ts
@@ -1,6 +1,9 @@
 import { CustomTypes } from "@prismicio/types-internal";
 import { SharedSliceContent } from "@prismicio/types-internal/lib/content";
-import { PrismicError } from "@prismicio/custom-types-client";
+import {
+	ForbiddenError,
+	UnauthorizedError,
+} from "@prismicio/custom-types-client";
 import {
 	SliceMachinePlugin,
 	SliceMachinePluginRunner,
@@ -164,7 +167,7 @@ export class SliceMachineManager {
 	async getState(): Promise<SliceMachineManagerGetStateReturnType> {
 		const [
 			{ sliceMachineConfig, libraries },
-			{ profile, remoteCustomTypes, remoteSlices, unauthorizedError },
+			{ profile, remoteCustomTypes, remoteSlices, authError },
 			customTypes,
 			packageManager,
 		] = await Promise.all([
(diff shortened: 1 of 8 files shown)


packages/manager/src/managers/SliceMachineManager.ts
```diff
@@ -173,42 +174,60 @@ export class SliceMachineManager {
...
 			this.project.detectPackageManager(),
 		]);
 
-		// TODO: SM UI detects if a user is logged out by looking at
+		// SM UI detects if a user is logged out by looking at
 		// `clientError`. Here, we simulate what the old core does by
-		// returning an `ErrorWithStatus`-like object if the user is
-		// not logged in.
+		// returning an `ErrorWithStatus`-like object if the user does
+		// not have acces to the repository or is not logged in.
```
bapmrl: ⚠️ Small typo.

packages/manager/src/managers/SliceMachineManager.ts
```diff
@@ -173,42 +174,60 @@ export class SliceMachineManager {
...
+							remoteSlices,
+							unauthorizedError,
+						};
+					} catch (error) {
+						// Non-Prismic error
+						if (!(error instanceof PrismicError)) {
+							throw error;
+						}
+
+						unauthorizedError = {
```
bapmrl: ❓ Does it really always correspond to an unauthorized error? What if it's an internal server error for instance?

packages/manager/src/managers/SliceMachineManager.ts
lihbr: Nice catch!

packages/manager/src/managers/SliceMachineManager.ts
lihbr: Basically, the API doesn't return a 401 upon unauthorized calls, it returns a 500, which makes triaging errors quite tricky :c

Here are the case we handle: https://github.com/prismicio/prismic-custom-types-client/blob/main/src/client.ts#L430-L497

Let me know if you have an idea on how to improve that 🤔

lihbr: OK, now it actually returns 401, I'll check to update that on the remote client 👌

lihbr: Done in [d252b4e](https://github.com/prismicio/slice-machine/pull/887/commits/d252b4e1778abb413fa5b4248744cdaf384e5725), let me know!

bapmrl (approved):

PR #890, opened 2023-03-28
feat(manager,slice-machine-ui): detect if manager supports simulator (SMX-94)

lihbr:
## Context

SMX-94

- [ ] The CI is successful.

### DIFF
--- a/packages/manager/src/managers/SliceMachineManager.ts
+++ b/packages/manager/src/managers/SliceMachineManager.ts
@@ -171,7 +171,6 @@ export class SliceMachineManager {
 			{ profile, remoteCustomTypes, remoteSlices, authError },
 			customTypes,
 			packageManager,
-			supportsSliceSimulator,
 		] = await Promise.all([
 			this.project.getSliceMachineConfig().then(async (sliceMachineConfig) => {
 				const libraries = await this._getLibraries(sliceMachineConfig);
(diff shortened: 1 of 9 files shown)


packages/manager/src/managers/SliceMachineManager.ts
```diff
@@ -218,6 +220,15 @@ export class SliceMachineManager {
 			}),
 			this._getCustomTypes(),
 			this.project.detectPackageManager(),
+			(async () => {
```
bapmrl: 💡 Maybe it would be clearer and future-proof to add a `supportsSliceSimulator` method in the `SimulatorManager`? It would enable direct usage in the front end when we'll remove `getState`.

packages/manager/src/managers/SliceMachineManager.ts
```diff
@@ -218,6 +220,15 @@ export class SliceMachineManager {
...
+				try {
+					await this.simulator.readSliceSimulatorSetupSteps();
+
+					return true;
+				} catch (error) {
```
bapmrl: ❓ Is there a way to avoid relying on catching an exception? I'm not found of this kind of logic because it introduces two opposite execution streams (⬇️ then ⬆). Would a `supportsSliceSimulator` method in `SimulatorManager` work?

bapmrl (approved):

packages/manager/src/managers/SliceMachineManager.ts
lihbr: Good call!

packages/manager/src/managers/SliceMachineManager.ts
lihbr: I agree, got an idea for a middleground.

bapmrl (approved):
Perfect! Thank you 😊

PR #891, opened 2023-03-29
feat: automatically clean hosted screenshots

bapmrl:
## Context

Completes SMX-126: Delete hosted screenshots of deleted slices.

## The Solution

1. Add `deleteScreenshotFolder` in `ScreenshotsManager`.
2. Use `ScreenshotsManager.deleteScreenshotFolder` in `PrismicRepositoryManager.pushChanges`.

- [ ] The CI is successful.

### DIFF
--- a/packages/init/test/__testutils__/mockAWSACLAPI.ts
+++ b/packages/init/test/__testutils__/mockAWSACLAPI.ts
@@ -27,6 +27,7 @@ type MockAWSACLAPIConfig = {
 					contentType?: string;
 				}[];
 		  };
+	deleteFolderEndpoint?: { expectedSliceIDs: string[] };
 };
 
 type MockAWSACLAPIReturnType = {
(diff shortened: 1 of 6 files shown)


packages/manager/src/managers/slices/SlicesManager.ts
```diff
@@ -572,26 +572,37 @@ export class SlicesManager extends BaseManager {
...
+			([screenshot, variation]) =>
+				!screenshot ||
+				!variation.imageUrl.includes(createContentDigest(screenshot)),
+		);
+		if (hasSomeScreenshotChanged) {
+			await this.screenshots.deleteScreenshotFolder({ sliceID: args.model.id });
+		}
+
+		const sliceMachineConfig = await this.project.getSliceMachineConfig();
+		const variations = await Promise.all(
```
lihbr: If I get it right, this has the risk of deleting the folder, and then only uploading new screenshots, having unchanged screenshots being deleted as a result. Not sure this method should be updated, or we should only delete old screenshots that have changed individually in the next map.

lihbr (approved):
LGTM

lihbr (approved):
LGTM

PR #894, opened 2023-03-29
ci: disable git tag and push for alpha release

lihbr:
## Context

Right now, when publishing an alpha through the CI, lots of things happen. Among them:
- A git tag is created and pushed (e.g. https://github.com/prismicio/slice-machine/releases/tag/start-slicemachine%400.6.1-dev-plugins-m2-validation.0)
- A commit is made (e.g. https://github.com/prismicio/slice-machine/commit/acfe527c091dbc11c5e6a1539042b28a9cd36f31)

Those create some issues:
- The git repository is "spammed" with short-living version tags, making navigating git tags complex
- The git repository is "spammed" with release commits, sometimes corrupting `package-lock.json` files

## The Solution

Lerna allows us to disable the git tag and commit behavior using flags: https://github.com/lerna/lerna/tree/f6e7a13e60fefc523d701efddfcf0ed41a77749b/commands/version

By using them we're essentially creating "virtual" alpha releases from GitHub point of view, not polluting GitHub with them (only NPM)

However, this creates collision issues (thanks @bapmrl!) because Lerna would not be able to determine the prerelease number since we stop committing them.

Therefore we can instead fallback on the "canary" flag of the publish command, which does just that and is tailored for per-commit release: https://github.com/lerna/lerna/blob/f6e7a13e60fefc523d701efddfcf0ed41a77749b/commands/publish/README.md#--canary, this is what this PR implements.

- [ ] The CI is successful.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,5 +207,5 @@ jobs:
           git config user.name "${{ github.actor }}"
           git config user.email "${{ github.actor }}@users.noreply.github.com"
           git checkout .
-          npm config set '//registry.npmjs.org/:_authToken' "${{NPM_TOKEN}}"
+          npm config set '//registry.npmjs.org/:_authToken' "${NPM_TOKEN}"
           npm run publish:alpha -- --preid=dev-ci.$(git rev-parse --short ${{ github.sha }})+0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,6 +207,6 @@ jobs:
           git config user.name "${{ github.actor }}"
           git config user.email "${{ github.actor }}@users.noreply.github.com"
           git checkout .
-          npm run bump:alpha --preid="${{ github.ref_name }} --no-push"
+          npm run bump:alpha --preid=dev-ci --no-push
           npm config set '//registry.npmjs.org/:_authToken' "${NPM_TOKEN}"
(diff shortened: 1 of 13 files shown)


bapmrl (approved):
Thank you! However, I don't know Lerna and wonder how it knows which versions already exist? Does it query `npm` before bumping?

bapmrl (approved):

lihbr:
Alright, I'm almost there but I think it's related to Lerna version where the canary flag tries to publish versions that can't be published (`1.0.0-alpha.0+123456` is sanitized to `1.0.0-alpha.0` which is not what we want):

---

image

---

image

---

Related: https://github.com/lerna/lerna/issues/2060#issuecomment-569968758, https://github.com/lerna/lerna/issues/3554

tl;dr; Lerna canary appears to have some issue, even in recent versions of Lerna

lihbr:
OK, I got it to work using this fix: https://github.com/lerna/lerna/issues/2060#issuecomment-658182391, see this job's result: https://github.com/prismicio/slice-machine/actions/runs/4566429321/jobs/8058917521

However, this still references a corrupted package version in package jsons, see the code here for example: https://www.npmjs.com/package/start-slicemachine/v/0.6.1-dev-ci.00aa6f7.0?activeTab=code

So I'm out of ideas on how to run that now :(

lihbr:
Closing until we get a better idea o/

PR #896, opened 2023-03-31
Create mock file on slice:create (SMX-136)

hypervillain:
- moved `mockSlice` function to manager
- call updateSliceMock on slice create
- return errors from both hooks


packages/manager/src/lib/mockSlice.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { SharedSliceContent } from "@prismicio/types-internal/lib/content";
+import { SliceDiff } from "@prismicio/types-internal/lib/customtypes/diff";
+import { SharedSlice } from "@prismicio/types-internal/lib/customtypes/widgets/slices";
+
+export const mockSlice = (
+	model: SharedSlice,
+	previousMocks?: SharedSliceContent[],
+	sliceDiff?: SliceDiff,
+): SharedSliceContent[] => {
```
lihbr: I'd suggest refactoring arguments to use an `args` object, this way, should the method evolve we're not relying on positional arguments (e.g. what if we need an additional mandatory argument?)

See an example of it here (from another lib file): https://github.com/prismicio/slice-machine/blob/06c0bee631fc20edf2855ca8f7abd180f1a01910/packages/manager/src/lib/installDependencies.ts#L30-L32

It's also a bit more consistent with the rest of the codebase~

packages/manager/src/lib/mockSlice.ts
```diff
@@ -0,0 +1,34 @@
+import { SharedSliceMock } from "@prismicio/mocks";
+
+import { SharedSliceContent } from "@prismicio/types-internal/lib/content";
+import { SliceDiff } from "@prismicio/types-internal/lib/customtypes/diff";
+import { SharedSlice } from "@prismicio/types-internal/lib/customtypes/widgets/slices";
+
+export const mockSlice = (
+	model: SharedSlice,
+	previousMocks?: SharedSliceContent[],
```
lihbr: Any thoughts on renaming the argument to "mocks"? I think it'd make more sense

packages/manager/test/SliceMachineManager-slices-createSlice.test.ts
```diff
@@ -12,6 +12,7 @@ it("calls plugins' `slice:create` hook", async (ctx) => {
 	const adapter = createTestPlugin({
 		setup: ({ hook }) => {
 			hook("slice:create", hookHandler);
+			hook("slice:asset:update", hookHandler);
```
lihbr: Since we're spying on this hook later, using the same spy here could have unexpected side effects later on, instead you could just use `vi.fn()` inline directly here.

packages/manager/test/mockSlice.test.ts
```diff
@@ -0,0 +1,400 @@
...
+			};
+		}),
+	};
+});
+
+vi.mock("@prismicio/mocks/lib/generators/utils/slug", () => {
+	return vi.fn().mockReturnValue("Woo");
+});
+
+describe.skip("mockSlice", () => {
```
lihbr: Is there still a reason to skip these tests? (not sure if it was something from React blocking before)

packages/manager/src/lib/mockSlice.ts
```diff
@@ -0,0 +1,34 @@
...
+	sliceDiff?: SliceDiff,
```
lihbr: What's `sliceDiff` exactly? Feels like it's never used 🤔

lihbr (changes requested):
LGTM, mainly some stylistic changes.

Do we need to also update mocks when the slice gets updated?

+ You might need to pull `dev-plugins-m2` to fix tests

packages/manager/src/lib/mockSlice.ts
```diff
@@ -0,0 +1,36 @@
+import { SharedSliceMock } from "@prismicio/mocks";
+
+import { SharedSliceContent } from "@prismicio/types-internal/lib/content";
+import { SliceDiff } from "@prismicio/types-internal/lib/customtypes/diff";
+import { SharedSlice } from "@prismicio/types-internal/lib/customtypes/widgets/slices";
+
+type mockSliceArgs = {
+	model: SharedSlice;
+	mocks?: SharedSliceContent[];
+	sliceDiff?: SliceDiff;
```
lihbr: Don't want to nitpick, but since we're already providing a (slice) model, (slice) mocks, maybe we can align `sliceDiff` with that and rename it to (slice) `diff`?

Up to you!

packages/plugin-kit/src/hooks/slice-update.ts
```diff
@@ -12,12 +14,16 @@ import type {
 export type SliceUpdateHookData = {
 	libraryID: string;
 	model: SharedSliceModel;
+	mocks?: SharedSliceContent[];
```
lihbr: Why do mocks have to be passed to `slice:update` hook? This sounds unneeded/introducing a side effect/business logic to the hook

packages/plugin-kit/src/hooks/slice-update.ts
```diff
@@ -12,12 +14,16 @@ import type {
...
 };
 
 /**
  * Return value for `slice:update` hook handlers.
  */
-export type SliceUpdateHookReturnType = void;
+export type SliceUpdateHookReturnType = {
+	mocks: SharedSliceContent[] | undefined;
```
lihbr: Same as above 🤔 If the manager needs to read mocks it better read them using its read mocks function

lihbr: Ok I think I got it, if you need a manager method to return different data, then you can create your own type there, we do that for multiple methods already. These types are primarily meant to type hooks API used by plugins.

packages/manager/src/managers/slices/SlicesManager.ts
```diff
@@ -293,16 +305,40 @@ export class SlicesManager extends BaseManager {
...
+			mocks: updatedMocks,
+		};
+
+		const { errors: updateSliceMocksHookResult } = await this.updateSliceMocks(
+			updateSliceMocksArgs,
+		);
+
 		return {
-			errors: hookResult.errors,
+			mocks: updatedMocks,
```
lihbr: Do we need to return mocks here? Just wondering if it'd make more sense to just call the `readMock` method after if they are needed by the caller

lihbr (approved):
LGTM, nice work! Just some types things to sort out, and tests to fix, let me know if you need any help!

lihbr (approved):
LGTM, just the CI to pass now :)

lihbr (approved):

PR #897, opened 2023-03-31
delete slices and custom types (SMX-125)

danigo99:
## Context

Ticket: https://linear.app/prismic/issue/SMX-125/delete-slices-and-custom-types-with-plugins

- [ ] The CI is successful.


hypervillain (commented):
I _think_ both hooks should be added to REQUIRED_ADAPTER_HOOKS. Can you confirm @lihbr?

packages/slice-machine/package.json
```diff
@@ -31,7 +31,6 @@
     "@prismicio/mocks": "1.6.4",
     "@prismicio/types-internal": "^1.5.3",
     "@slicemachine/manager": "0.1.1-dev-plugins-testing.0",
-    "axios": "^0.26.1",
```
lihbr, line 34: Thanks for cleaning that up!

packages/manager/src/managers/slices/SlicesManager.ts
```diff
@@ -623,4 +651,55 @@ export class SlicesManager extends BaseManager {
...
+					);
+
+					return { ...tabAccumulator, [tabKey]: updatedTabFields };
+				},
+				{},
+			);
+
+			this.customTypes.updateCustomType({
+				model: { ...customType.model, json: updatedJsonModel },
+			});
```
lihbr: I think this needs to be awaited in case there's an error, you might want to use a `models.map(...)` + `Promise.all()` pattern like here: https://github.com/prismicio/slice-machine/blob/7e91e377ae4601e5584360b785b85273179d9c05/packages/manager/src/managers/SliceMachineManager.ts#L282-L283

lihbr (changes requested):
LGTM, super clean!

I think there's just a small bug with slice removal from custom types.

Also, could you add a test in `slices-deleteSlice.test.ts` like `removes deleted slice from custom types referencing it` to test the mandatory side-effect?

lihbr:
> I _think_ both hooks should be added to REQUIRED_ADAPTER_HOOKS. Can you confirm @lihbr?

I thought they were already, but yes, definitely! Thanks for checking~

packages/manager/src/managers/slices/SlicesManager.ts
```diff
@@ -623,4 +651,58 @@ export class SlicesManager extends BaseManager {
 			variations,
 		};
 	}
+
+	private async _removeSliceFromCustomTypes(sliceID: string) {
+		const { models, errors: customTypeReadErrors } =
+			await this.customTypes.readAllCustomTypes();
+
+		// Successfully update all custom types or throw
+		Promise.all(
```
lihbr: `Promise.all` returns a promise, you need to await it :)

lihbr (approved):
One last small thing on the `Promise.all`, but LGTM :)

PR #898, opened 2023-04-03
Remove Storybook from slice-machine-ui

hypervillain:
As we're moving Storybook to its own plugin, it does not make sense to store storybook local URL in manifest, nor check for its presence in the SliceBuilder. If we ever need to display a link to an external page from the slice builder, the feature will have to be build independently via hooks


bapmrl (commented):
Hello @hypervillain,

Searching `storybook` I found references in `packages/slice-machine/test/src/modules/__fixtures__/serverState.ts` and `packages/slice-machine/test/src/utils/storybook.test.ts`. Does it make sense to keep them?

hypervillain:
Thanls @bapmrl I was excluding `**/tests/**` from my search 😬

bapmrl (approved):

PR #902, opened 2023-04-04
e2e: update e2e tests following plugin changes

mdeclercq:
## Context

Ticket: https://linear.app/prismic/issue/SMX-139/aadev-i-dont-want-the-e2e-tests-to-fail-in-the-ci

## The Solution

**Code changes**
- Building the endpoint url of `/bulk` was not working for api endpoints that have a subpath.
    - In the CI, the endpoint is https://something-private.com/dev, and SM was trying to push to `https://something-private.com/bulk` instead of `https://something-private.com/dev/bulk` 
- removed the `/customtypes/` suffix in the PrismicModels default stage/prod api endpoint: `https://customtypes.wroom.io/customtypes/` - I’m not sure why it was there, it doesn’t seem to be used - **to be checked with a dev**

**Test changes**
- made a few tests retriable
    - In most tests, the slice id is set once for the test. If the first attempt fails, the 2nd attempt usually fails because the slice already exists (locally or on the remote repo).
    - instead we create new unique slices in the `beforeEach`
- redo a `npm i` after SM is installed to make Puppeteer install with the right version.
- skip transactional push test until mocking network requests is available.
- removed testing of the onboarding screen within `scenario_001` there is a dedicated test for that.

- [x] The CI is successful.

image


packages/manager/src/constants/API_ENDPOINTS.ts
```diff
@@ -43,7 +43,7 @@ export const API_ENDPOINTS: APIEndpoints = (() => {
 			return {
 				PrismicWroom: "https://wroom.io/",
 				PrismicAuthentication: "https://auth.wroom.io/",
-				PrismicModels: "https://customtypes.wroom.io/customtypes/",
```
danigo99, line 46: I'm not sure how this affects the `@prismicio/custom-types-client` since there were some notes there about backwards compatibility. From what I can see it shouldn't be a problem but not sure if @lihbr or @angeloashmore want to take a closer look.

See here: https://github.com/prismicio/prismic-custom-types-client/blob/main/src/client.ts#L132-L143

danigo99 (commented):
Looks good for me, just have one point to clarify but otherwise great job 👍 

Looking forward to seeing green CI again 😄

packages/manager/src/constants/API_ENDPOINTS.ts
lihbr: Yes, it's fine!

danigo99 (approved):

packages/manager/src/constants/API_ENDPOINTS.ts
angeloashmore: @danigo99 Yep, looks good to me. In case you're curious why that `if` statement was added, this PR shares some details: https://github.com/prismicio/prismic-custom-types-client/pull/6

PR #905, opened 2023-04-05
fix: NewField being unexpectedly reset

bapmrl:
## Context

Resolves SMX-114: AAUser, I don't want to get Slice/CT field inputs reset while I type.

## The Solution

1. Stop creating a new component during each render.

- [ ] The CI is successful.


lihbr (approved):
LGTM, nice catch!

PR #911, opened 2023-04-07
Backport types internal (SMX-120,SMX-143)

hypervillain:
- Use alpha (latest) version of @prismicio/types-internal, @prismicio/mocks editor-fields and editor-ui everywhere
- Replace types with types-internal everywhere


arnaudlewis:
Don't forget the backport the rest regarding the mock config of date and link that are new in editor 0.3.0.

package.json
```diff
@@ -42,6 +42,7 @@
     "prettier": "^2.8.4"
   },
   "dependencies": {
+    "@prismicio/mocks": "^2.0.0-alpha.2",
```
lihbr: Why does `@prismicio/mocks` needs to be installed top level?

packages/manager/test/lib-mockSlice.test.ts
```diff
@@ -31,13 +28,13 @@ describe("mockSlice", () => {
 		};
 		const geoR = GeoPointContent.decode(geo.key);
 
-		expect(isRight(geoR)).toBeTruthy();
+		expect(isRight(geoR)).toBeFalsy();
```
lihbr: This should be truthy no? You likely need to uncomment line 25

Left as is (as I understand it) it defeats the purpose of the test (which also has no reason to be a test in first place here, we shouldn't have to test types-internal here)

packages/slice-machine/components/Select/index.tsx
```diff
@@ -13,18 +13,18 @@
...
+//   }
 
-  // @ts-expect-error - `default` is added by Rollup during tests since
-  // `react-select` has default and named exports. It does not exist normally.
-  Select = _Select.default;
-}
+//   // @ts-expect-error - `default` is added by Rollup during tests since
+//   // `react-select` has default and named exports. It does not exist normally.
+//   Select = _Select.default;
+// }
```
lihbr: If this is not needed anymore, maybe you can just get rid of it?

lihbr (changes requested):
LGTM, just some small things.

Could you have a look at backporting the date & link from editor 0.3.0 as Arnaud mentioned? (this PR or another, whatever works best for you)

lihbr (approved):

PR #914, opened 2023-04-09
Upgrade axios to v1.3.5

igitur:
Hi,

Please see discussion at https://community.prismic.io/t/slice-machine-init-behind-corporate-proxy/11885/4 . Slice Machine is struggling behind certain corporate proxies that add SSL inspection. I'm not 100% sure, but I suspect upgrading axios to the latest version could solve this problem. I don't see any disadvantage to upgrading some/all of the libraries, so hopefully you're OK with this.

I notice that `package-lock.json` is set to be treated as a binary file, so I left it as is and as generated by npm.


lihbr:
Hey @igitur, thank you so much for your contribution!

We stopped using `axios` in our next release and will now only rely on vanilla `fetch` API (via `node-fetch` or browser `fetch`)

Do you think this will be enough to fix the SSL inspection issue?

igitur:
Thanks, in which commit did you do this change?

lihbr:
It's part of #799 which refactors our whole backend to introduce better framework support :)

igitur:
Ah, thanks. I'm in no rush, so I'll just wait for that to be released and test again from there. Thanks again.

lihbr:
Thank you for your comprehension and patience!

PR #919, opened 2023-04-13
feat: support alpha version of the init command (SMX-146)

lihbr:
Updates the init command to install alpha versions of packages when running an alpha version of the init command.

### DIFF
--- a/packages/init/src/lib/framework.ts
+++ b/packages/init/src/lib/framework.ts
@@ -3,6 +3,8 @@ import { join } from "node:path";
 
 import semver from "semver";
 
+import { version as pkgVersion } from "../../package.json";
+
 export type Framework = {
 	/**
 	 * Framework's human readable name.
@@ -44,8 +46,11 @@ export type Framework = {
 	devDependencies: Record<string, string>;
 };
 
+const isPrerelease =
+	semver.prerelease(pkgVersion) !== null && process.env.NODE_ENV !== "test";
+
 const DEFAULT_DEV_DEPENDENCIES: Record<string, string> = {
-	"slice-machine-ui": "<1.0.0",
+	"slice-machine-ui": isPrerelease ? "alpha" : "latest",
 };
 
 /**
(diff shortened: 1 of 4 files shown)


packages/init/src/lib/framework.ts
```diff
@@ -44,8 +46,11 @@ export type Framework = {
 	devDependencies: Record<string, string>;
 };
 
+const isPrerelease =
```
hypervillain, line 49: why do we need to exclude test env from this?

packages/init/src/lib/framework.ts
```diff
@@ -44,8 +46,11 @@ export type Framework = {
...
+	semver.prerelease(pkgVersion) !== null && process.env.NODE_ENV !== "test";
+
 const DEFAULT_DEV_DEPENDENCIES: Record<string, string> = {
-	"slice-machine-ui": "<1.0.0",
+	"slice-machine-ui": isPrerelease ? "alpha" : "latest",
```
hypervillain, line 53: I guess the tag could be created once on top and reused in all objects below. It could prevent possible typos. Also it could also be typed with a union

packages/init/src/lib/framework.ts
lihbr: Tests test for package detection and version to be installed depending on the framework.

This exception makes `isPrerelease` always false, allowing tests to be consistent whether the package was last published as latest (`0.1.0`) or as an alpha (`0.1.0-dev-plugins`)

Let me know if you have a better idea ☺️

packages/init/src/lib/framework.ts
lihbr: I'm not sure I get it, I first had something like `prefersAlpha || "latest"` on all definition, but I didn't like the implicitness of the assignment.

I think using the ternary makes the assignment clearer while also allowing us to specify another tag (e.g. `canary`) should the installed package be published under another tag than `alpha`

I think ultimately we won't have `latest` specified in this file but a regular SemVer range (e.g. `"slice-machine-ui": "^1.0.0"`) allowing the init command to not break itself over time.

I indeed reverted, here, the `<1.0.0` range because that's the range we need now (`slice-machine-ui` version `1.0.0` or greater doesn't exist yet) but since we'll likely publish it as `1.0.0` this would have broke, so at least now it's safe to publish(?) If that makes sense.

PR #923, opened 2023-04-17
fix: deleteScreenshotFolder

bapmrl:
## Context

Resolves SMX-126: Delete hosted screenshots of deleted slices.

## The Solution

1. Change `sliceId` to `sliceName` as the ACL Provider route to delete screenshots is expecting a slice ID in a `sliceName` field.

- [x] The CI is successful.


packages/manager/src/managers/screenshots/ScreenshotsManager.ts
```diff
@@ -296,7 +296,7 @@ export class ScreenshotsManager extends BaseManager {
 		args: ScreenshotsManagerDeleteScreenshotFolderArgs,
 	): Promise<void> {
 		const res = await this._fetch({
-			body: { sliceId: args.sliceID },
+			body: { sliceName: args.sliceID },
```
lihbr, line 301: Maybe add a small comment that this is because it's the argument the ACL expects but that we do want a Slice ID here (because those aren't sensible to renaming)?

lihbr (approved):
LGTM, just a small one to consider + tests to fix

PR #935, opened 2023-04-26
Allow user to reset slice image field dimensions to auto

### LINKED ISSUE #672: Cannot reset Slice image field dimensions to "auto“
### Versions
- slice-machine-ui: `0.5.2`
- node: `v16.17.1`

### Steps to reproduce
- Launch Slicemachine
- Create a new Slice named Hero
- Add a new Image field named Image in the Non-Repeatable Zone
- Open the Image settings and set the `height` and `width` to `500`
- Save the image settings. Save the slice. 
- Open the Image settings and erase the height and width input values. The dimensions switches to `auto`
- Save the image settings
- Open the Image settings again

### What is expected?
- The image dimensions is still set to `auto`
- After saving the slice, the `slices/Hero/model.json` file contains empty dimension settings: `”constraint":{}`

- The image dimension fields are still set to 500
- The slice cannot be saved since no changes are detected
- The only way to set the image dimension to `auto` in the UI is to set the width and height to `0`. But this translates to `”constraint":{"width": 0, "height": 0}` in the model file. There should be another way.

https://user-images.githubusercontent.com/6433957/199792836-0c31cb63-4b28-4cd4-8594-cc2700b5b8b9.mov

xrutayisire:
## Context

- [AAUser, I can reset Slice image field dimensions to "auto“](https://linear.app/prismic/issue/DT-815/aauser-i-can-reset-slice-image-field-dimensions-to-auto)
- Fixes #672 

## The Solution

- Adding a transform function with Yup let us transform the empty string to null to have "auto" dimensions

## Impact / Dependencies

- The model doesn't contain: `”constraint":{}` but `"constraint": { "width": null, "height": null }`
There should not be an impact from that and the result is still auto dimensions, but reviewer could help me confirm that.

- [x] The CI is successful.

## Preview

https://user-images.githubusercontent.com/19946868/234526066-ca0b17bd-0f49-4596-87e3-9544b18cea83.mov

### DIFF
--- a/packages/slice-machine/lib/models/common/widgets/Image/Form.tsx
+++ b/packages/slice-machine/lib/models/common/widgets/Image/Form.tsx
@@ -17,14 +17,24 @@ import {
 } from "./components";
 import { TabFields } from "@lib/models/common/CustomType";
 
+const nullableNumberSchema = () => {
+  return yup
+    .number()
+    .nullable()
+    .transform((value: string | number, originalValue: string | number) =>
+      // When the user empties the field, it should convert it to null so that "auto" dimensions are set
+      originalValue === "" ? null : value
+    );
+};
+
 const FormFields = {
   label: DefaultFields.label,
   id: DefaultFields.id,
   constraint: {
     validate: () =>
       yup.object().defined().shape({
-        width: yup.number().nullable(),
-        height: yup.number().nullable(),
+        width: nullableNumberSchema(),
+        height: nullableNumberSchema(),
       }),
   },
   thumbnails: {
--- a/packages/slice-machine/test/lib/models/common/widgets/index.test.ts
+++ b/packages/slice-machine/test/lib/models/common/widgets/index.test.ts
@@ -60,4 +60,25 @@ describe("Widgets", () => {
       expect(field).toEqual(fieldExpected);
     }
   );
+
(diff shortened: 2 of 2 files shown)


packages/slice-machine/lib/models/common/widgets/Image/Form.tsx
```diff
@@ -17,14 +17,23 @@ import {
 } from "./components";
 import { TabFields } from "@lib/models/common/CustomType";
 
+const nullableNumberSchema = () => {
+  return yup
+    .number()
+    .nullable()
+    .transform((value: string | number, originalValue: string | number) =>
+      originalValue === "" ? null : value
```
bapmrl, line 26: 💡 We could add a short comment to explain that it helps supporting the `auto` constraint.

bapmrl (approved):

packages/slice-machine/lib/models/common/widgets/Image/Form.tsx
xrutayisire: Thanks, done

bapmrl: Small typo. What do you think about:
> When the user empties the field, it should convert it to null so that "auto" dimensions are set.

Otherwise, you can merge!

bapmrl (approved):

PR #940, opened 2023-04-27
Fix code snippet for GeoPoint field type in Next

### LINKED ISSUE #790: GeoPoint code snippet for React is invalid
### Versions

- slice-machine-ui: `0.6.0`
- node: `v18.12.1`

### Steps to reproduce
- Create a `Next.js` project and install SliceMachine
- Run SliceMachine
- Create a Slice with a `GeoPoint` field
- Show the code snippets
- Notice the code snippet of the GeoPoint field is `<span>{ slice.primary.geopoint }</span>`

### What is expected?
- The code snippets should work when copy/pasted as such, maybe something like
- `<span>{slice.primary.geopoint.latitude},{slice.primary.geopoint.longitude}</span>`

- Since `slice.primary.geopoint` is an object, using the code as such generates an error in the React component: 
> `Objects are not valid as a React child (found: object with keys {latitude, longitude}).`

xrutayisire:
## Context

- [AAUser, I want to copy/paste a valid React GeoPoint code snippet](https://linear.app/prismic/issue/DT-960/aauser-i-want-to-copypaste-a-valid-react-geopoint-code-snippet)
- Fixes #790 

## The Solution

- Add a specific code snippet in Next adapter with a simple stringify
- Stringify solution let us have the object display and looks the same as with Nuxt adapter

## Impact / Dependencies

- None that I can think of

- [x] If it is a critical feature, I have added tests.~~
- [x] The CI is successful.

### Next before

Screenshot 2023-04-27 at 10 09 22

### Next after

Screenshot 2023-04-27 at 10 09 11

### And for reference in Nuxt 

Screenshot 2023-04-27 at 09 45 27

### DIFF
--- a/packages/adapter-next/src/hooks/snippet-read.ts
+++ b/packages/adapter-next/src/hooks/snippet-read.ts
@@ -118,6 +118,21 @@ export const snippetRead: SnippetReadHook<PluginOptions> = async (
 			};
 		}
 
+		case "GeoPoint": {
+			const code = await format(
+				stripIndent`
+					<>{JSON.stringify(${dotPath(fieldPath)})}</>
+				`,
+				helpers,
+			);
+
+			return {
+				label,
+				language: "tsx",
+				code,
+			};
+		}
+
 		default: {
 			return {
 				label,
--- a/packages/adapter-next/test/plugin-snippet-read.test.ts
+++ b/packages/adapter-next/test/plugin-snippet-read.test.ts
@@ -84,7 +84,7 @@ testSnippet("date", `<>{${model.id}.data.date}</>`);
 
 testSnippet("embed", `<>{${model.id}.data.embed}</>`);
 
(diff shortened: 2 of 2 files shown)


packages/adapter-next/src/hooks/snippet-read.ts
```diff
@@ -118,6 +118,21 @@ export const snippetRead: SnippetReadHook<PluginOptions> = async (
 			};
 		}
 
+		case "GeoPoint": {
+			const code = await format(
+				stripIndent`
+					<>{JSON.stringify(${dotPath(fieldPath)})}</>
```
bapmrl, line 124: ❓ This fix works, but have you checked what happens for all the other fields? For instance, will the problem occur for the Embed field? If so, does it make sense to call `JSON.stringify` in the default case?

hypervillain: Then you would have to add checks for Text and others probably, I'll merge let's discuss this later

PR #943, opened 2023-04-27
Prevent renaming a slice with a name starting with a number

### LINKED ISSUE #791: Renaming a Slice accepts strings starting with number
### Versions

- slice-machine-ui: `0.6.0`
- node: `v18.12.1`

### Steps to reproduce

- Run SliceMachine
- Create a Slice 
- Open the "rename slice" modal 
- Type a name starting with a number, for example `3Columns`

### What is expected?
- The submit button is disabled, just like in the Create slice modal

- The submit button is enabled, we can rename the slice with an invalid name
- The breaks the Slice simulator since the slice cannot be imported

image

image

xrutayisire:
## Context

- [AAUser, I cannot rename a slice name with a name starting with a number](https://linear.app/prismic/issue/DT-1109/aauser-i-cannot-rename-a-slice-name-with-a-name-starting-with-a-number)
- Fixes #791

## The Solution

- Mutualize code between the CreateSliceModal and RenameSliceModal, so they share the same validation function, and it's not possible to correctly validate in one ase and not the other
- Added some tests to understand your Cypress setup

## Impact / Dependencies

- Simple refactor since I didn't want to change too much change for this simple fix
- Don't know if you have some best practices for the code organization, example the file "types.ts" and "validator.ts"

- [x] The CI is successful.

## Preview

Screenshot 2023-04-27 at 14 22 01

### DIFF
--- a/cypress/e2e/slices/00-create.cy.js
+++ b/cypress/e2e/slices/00-create.cy.js
@@ -1,6 +1,9 @@
 import { SLICE_MOCK_FILE } from "../../consts";
+import { sliceRenameModal } from "../../pages/RenameModal";
 import { simulatorPage } from "../../pages/simulator/simulatorPage";
+import { createSliceModal } from "../../pages/slices/createSliceModal";
 import { sliceBuilder } from "../../pages/slices/sliceBuilder";
+import { slicesList } from "../../pages/slices/slicesList";
 
 const sliceName = "TestSlice";
 const editedSliceName = "EditedSliceName";
@@ -142,5 +145,28 @@ describe("Create Slices", () => {
     cy.get('ul[data-cy="slice-non-repeatable-zone"] > li')
       .eq(0)
       .contains("Description");
+  });
+
+  // See: #599
+  it("A user cannot create a slice with a name starting with a number", () => {
+    slicesList.goTo();
+    slicesList.emptyStateButton.click();
+
+    createSliceModal.nameInput.type("42Test");
+    createSliceModal.submitButton.should("be.disabled");
+  });
+
+  // See: #791
+  it("A user cannot rename a slice with a name starting with a number", () => {
+    const sliceName = "SliceName";
+
+    cy.createSlice(lib, sliceId, sliceName);
+    slicesList.goTo();
+
+    slicesList.getOptionDopDownButton(sliceName).click();
+    slicesList.renameButton.click();
+
+    sliceRenameModal.input.clear().type("42Test");
+    sliceRenameModal.submitButton.should("be.disabled");
   });
 });
--- a/packages/slice-machine/components/Forms/RenameSliceModal/RenameSliceModal.tsx
+++ b/packages/slice-machine/components/Forms/RenameSliceModal/RenameSliceModal.tsx
@@ -6,10 +6,9 @@ import { useSelector } from "react-redux";
 import { SliceMachineStoreType } from "@src/redux/type";
 import { isModalOpen } from "@src/modules/modal";
 import { ModalKeysEnum } from "@src/modules/modal/types";
-import { RESERVED_SLICE_NAME, API_ID_REGEX } from "@lib/consts";
-import camelCase from "lodash/camelCase";
 import { getLibraries, getRemoteSlices } from "@src/modules/slices";
-import startCase from "lodash/startCase";
+import { SliceModalValues } from "../formsTypes";
+import { validateSliceModalValues } from "../formsValidator";
 
 interface RenameSliceModalProps {
   sliceName: string;
@@ -31,12 +30,12 @@ export const RenameSliceModal: React.FC<RenameSliceModalProps> = ({
     })
   );
 
-  const handleOnSubmit = (values: { sliceName: string }) => {
+  const handleOnSubmit = (values: SliceModalValues) => {
     renameSlice(libName, sliceId, values.sliceName);
   };
 
   return (
-    <ModalFormCard
+    <ModalFormCard<SliceModalValues>
       dataCy="rename-slice-modal"
       isOpen={isRenameSliceModalOpen}
       widthInPx="530px"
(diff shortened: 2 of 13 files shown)


cypress/e2e/slices/00-create.cy.js
```diff
@@ -143,4 +146,24 @@ describe("Create Slices", () => {
...
+    slicesList.goTo();
+    slicesList.emptyStateButton.click();
+
+    createSliceModal.nameInput.type("42Test");
+    createSliceModal.submitButton.should("be.disabled");
+  });
+
+  // See: #791
+  it("A user cannot rename a slice with a name starting with a number", () => {
+    const correctSliceName = "CorrectSliceName";
```
bapmrl: 💡 Maybe a better name would be `sliceName`? In fact, upon first read, I thought we were going to compare this to some other value (as it starts with `correct`).

cypress/e2e/slices/00-create.cy.js
```diff
@@ -143,4 +146,24 @@ describe("Create Slices", () => {
...
+    cy.createSlice(lib, sliceId, correctSliceName);
+    slicesList.getOptionDopDownButton(correctSliceName).click();
```
bapmrl: ⚠️ Apparently this fails in the CI.

packages/slice-machine/components/Forms/RenameSliceModal/RenameSliceModal.tsx
```diff
@@ -6,11 +6,9 @@ import { useSelector } from "react-redux";
 import { SliceMachineStoreType } from "@src/redux/type";
 import { isModalOpen } from "@src/modules/modal";
 import { ModalKeysEnum } from "@src/modules/modal/types";
-import { RESERVED_SLICE_NAME, API_ID_REGEX } from "@lib/consts";
-import camelCase from "lodash/camelCase";
 import { getLibraries, getRemoteSlices } from "@src/modules/slices";
-import startCase from "lodash/startCase";
-
+import { SliceModalValues } from "../types";
+import { validateSliceModalValues } from "../validator";
```
bapmrl: ⚠️ Let's add an empty line between the `interface` and imports.

packages/slice-machine/components/Forms/types.ts
```diff
@@ -0,0 +1,3 @@
+export interface SliceModalValues {
```
bapmrl: ❓ How do you think we should name files in the project? I'd personnally advocate for having globally unique names when it's possible as I find it easier to quickly find modules (with CMD+P in VS Code). In this case, would `formTypes` be better in your opinion?

packages/slice-machine/components/Forms/validator.ts
```diff
@@ -0,0 +1,42 @@
+import { SliceModalValues } from "./types";
```
bapmrl: ❓ Same question regarding naming.

packages/slice-machine/components/Forms/validator.ts
```diff
@@ -0,0 +1,42 @@
...
+import { LibraryUI } from "@lib/models/common/LibraryUI";
+import { SliceSM } from "@lib/models/common/Slice";
+import camelCase from "lodash/camelCase";
+import startCase from "lodash/startCase";
+
+export function validateSliceModalValues(
+  { sliceName }: SliceModalValues,
+  localLibs: ReadonlyArray<LibraryUI>,
+  remoteLibs: ReadonlyArray<SliceSM>
+) {
```
bapmrl: 💡 I think it's better to explicitely state what is the return type of functions, especially when they are exported.

bapmrl (changes requested):
Nice work, thank you!

cypress/e2e/slices/00-create.cy.js
xrutayisire: Agreed, done 👍

packages/slice-machine/components/Forms/RenameSliceModal/RenameSliceModal.tsx
xrutayisire: Good catch, done 👍

packages/slice-machine/components/Forms/types.ts
xrutayisire: Hum, I wanted to discuss that, I like your idea. Maybe even more, we could just use the folder name as the prefix. Here it's Forms with an s (not the best I guess for our futur clean Form), so it would be "formsTypes.ts" and "formsValidator.ts", what do you think?

packages/slice-machine/components/Forms/validator.ts
xrutayisire: Done 👍 
We could even add that rule in eslint if we want

packages/slice-machine/components/Forms/types.ts
bapmrl: It works for me in this case 🙂

packages/slice-machine/components/Forms/validator.ts
bapmrl: Yes, maybe in a separate Pull Request if you have time or you could just create a Linear issue?

cypress/e2e/slices/00-create.cy.js
```diff
@@ -143,4 +146,28 @@ describe("Create Slices", () => {
...
+  });
+
+  // See: #791
+  it("A user cannot rename a slice with a name starting with a number", () => {
+    const sliceName = "SliceName";
+
+    cy.createSlice(lib, sliceId, sliceName);
+    slicesList.goTo();
+
+    cy.waitUntil(() => cy.get("[data-cy=slice-action-icon]"));
```
bapmrl: ❓ Nice, but is there a way to use a selector from the Page Object Model?

packages/slice-machine/components/Forms/validator.ts
xrutayisire: I think best is a separate issue, because it will throw error on a lot of functions I guess

xrutayisire: I will create it ;)

packages/slice-machine/components/Forms/formsValidator.ts
```diff
@@ -0,0 +1,46 @@
...
+import { SliceSM } from "@lib/models/common/Slice";
+import camelCase from "lodash/camelCase";
+import startCase from "lodash/startCase";
+
+export function validateSliceModalValues(
+  { sliceName }: SliceModalValues,
+  localLibs: ReadonlyArray<LibraryUI>,
+  remoteLibs: ReadonlyArray<SliceSM>
+):
+  | {
```
bapmrl: 💡 It's better, but I find it hard to read. Maybe defining the `type` just on top of the function would be more readable? It could be named `SliceModalValuesValidity` or `ValidateSliceModalValuesReturnType` for instance.

bapmrl (approved):

packages/slice-machine/components/Forms/formsValidator.ts
xrutayisire: Yes I agree, better to export it if we want one day 👍

xrutayisire: Done

cypress/e2e/slices/00-create.cy.js
xrutayisire: I was able to even remove this line.

slicesList.getOptionDopDownButton(sliceName).click(); already access "[data-cy=slice-action-icon]" and so the wait is automatically done there

PR #944, opened 2023-04-27
Fix default value checkbox for Select Field

xrutayisire:
## Context

- [AAUser, I want to set the default value of a Select field](https://linear.app/prismic/issue/DT-1059/aauser-i-want-to-set-the-default-value-of-a-select-field)

## The Solution

- Without TypeScript, hard to see this kind of problems, but basically there is a "buildControlledValue" that receive 2 params. The second param is if the first value should be considered the default one, and it was never given to the function.
The "buildControlledValue" function is located [here](https://github.com/prismicio/slice-machine/blob/master/packages/slice-machine/lib/models/common/widgets/Select/FormFields.jsx#L17) 

## Impact / Dependencies

- It was introduced in this [tag](https://github.com/prismicio/slice-machine/compare/slice-machine-ui%400.6.0...slice-machine-ui%400.6.1#diff-055a1d388300e2493dc0a9089422a4a349b835b2290b45ace19ace8f85a0e682R34) for the 0.6.1 of slice-machine-ui
Don't look like it could be an impact on my fix
Didn't want to change this code because it will be refactored anyway, so I just provided a fix, so user can use the default value feature again

- ~~If it is a critical feature, I have added tests.~~
- [x] The CI is successful.

## Preview

Screenshot 2023-04-27 at 18 09 02

Screenshot 2023-04-27 at 18 09 20

### DIFF
--- a/packages/slice-machine/components/FormFields/CheckboxControl.jsx
+++ b/packages/slice-machine/components/FormFields/CheckboxControl.jsx
@@ -3,14 +3,15 @@ import { useState, useEffect } from "react";
 import { FormFieldCheckbox } from "./";
 
 /**
- * This components allows to set/unset the value of an arbitrary field via checking/unchecking the box
+ * This components allows to set/unset the value of an arbitrary field via
+ * checking/unchecking the box
  *
- * field: the controlled Formik field that we want to manipulate
- * label: a function of text displayed next to the checkbox. Can be either a string or a function
- * controlledValue: the value of the field being controlled
- * setControlledValue: function to update the value of the controlled value in Formik
- * buildControlledValue: function to build the new value based on the current controlled value and the state of the checkbox
- *
+ * Field: the controlled Formik field that we want to manipulate label: a
+ * function of text displayed next to the checkbox. Can be either a string or a
+ * function controlledValue: the value of the field being controlled
+ * setControlledValue: function to update the value of the controlled value in
+ * Formik buildControlledValue: function to build the new value based on the
+ * current controlled value and the state of the checkbox
  */
 const CheckboxControl = ({
   field,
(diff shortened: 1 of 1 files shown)


packages/slice-machine/components/FormFields/CheckboxControl.jsx
```diff
@@ -3,14 +3,15 @@ import { useState, useEffect } from "react";
...
- *
- * field: the controlled Formik field that we want to manipulate
- * label: a function of text displayed next to the checkbox. Can be either a string or a function
- * controlledValue: the value of the field being controlled
- * setControlledValue: function to update the value of the controlled value in Formik
- * buildControlledValue: function to build the new value based on the current controlled value and the state of the checkbox
+ * This components allows to set/unset the value of an arbitrary field via
+ * checking/unchecking the box
  *
+ * Field: the controlled Formik field that we want to manipulate label: a
```
bapmrl, line 9: ⚠️ Your formatter has mixed together some params 😛

packages/slice-machine/components/FormFields/CheckboxControl.jsx
```diff
@@ -31,7 +32,7 @@ const CheckboxControl = ({
     // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
     buildControlledValue
       ? // eslint-disable-next-line @typescript-eslint/no-unsafe-call
-        setControlledValue(buildControlledValue(controlledValue))
+        setControlledValue(buildControlledValue(controlledValue, isChecked))
       : // eslint-disable-next-line @typescript-eslint/no-unsafe-call
         setControlledValue(buildControlledValue);
```
bapmrl, line 37: ❓ This will call `setControlledValue` with an `undefined` parameter. What does it imply? Maybe `setControlledValue(controlledValue)` wouldn't make sense and would create an infinite loop?

bapmrl (approved):

packages/slice-machine/components/FormFields/CheckboxControl.jsx
xrutayisire: --', thanks good catch!

packages/slice-machine/components/FormFields/CheckboxControl.jsx
xrutayisire: Honestly, I thought the same 😄 
For me, the whole useEffect could be improved and maybe even removed, but since it's jsx with old forms I decided to only fix the bug and assuring myself that I will not break something else.
Do you think I should refactor this now?

bapmrl: Okay, I understand 😊 Let's keep it this way.

PR #946, opened 2023-04-27
Remove link to simulator from Custom Types page

xrutayisire:
## Context

- [AAUser, I don't need a link to the Simulator on the Custom Types page](https://linear.app/prismic/issue/DT-991/aauser-i-dont-need-a-link-to-the-simulator-on-the-custom-types-page)

## The Solution

- I identified the different zones that we have in two: custom_type or slice, with that I was able to remove the tab on custom type context

## Impact / Dependencies

- Should not have any impact outside of the removal of the tab. 
- For me the best way was to use a new prop because all the other ones were not specific enough or too risky to use for the maintenance
- I also changed the way to pass down the card children because it requires an array if I don't want to refactor everything, and it should not pass down even null or false as JSX to prevent Tabs to have two children. With just a simple condition, I end up with this:
```
warning: Failed prop type: There should be an equal number of 'Tab' and 'TabPanel' in `Tabs`. Received 1 'Tab' and 2 'TabPanel'.
```

- ~~If it is a critical feature, I have added tests.~~
- [x] The CI is successful.

## Preview

Before, on the custom type context, I have the Mock tab:

Screenshot 2023-04-27 at 19 41 14

On the custom type context, I now don't have the Mock tab:

Screenshot 2023-04-27 at 19 37 44

### DIFF
--- a/packages/slice-machine/components/FormFields/CheckboxControl.jsx
+++ b/packages/slice-machine/components/FormFields/CheckboxControl.jsx
@@ -6,12 +6,14 @@ import { FormFieldCheckbox } from "./";
  * This components allows to set/unset the value of an arbitrary field via
  * checking/unchecking the box
  *
- * Field: the controlled Formik field that we want to manipulate label: a
- * function of text displayed next to the checkbox. Can be either a string or a
- * function controlledValue: the value of the field being controlled
- * setControlledValue: function to update the value of the controlled value in
- * Formik buildControlledValue: function to build the new value based on the
- * current controlled value and the state of the checkbox
+ * @param field - The controlled Formik field that we want to manipulate
+ * @param label - A function of text displayed next to the checkbox. Can be
+ *   either a string or a function
+ * @param controlledValue - The value of the field being controlled
+ * @param setControlledValue - Function to update the value of the controlled
+ *   value in
+ * @param buildControlledValue - Formik function to build the new value based on
+ *   the current controlled value and the state of the checkbox
  */
 const CheckboxControl = ({
   field,
--- a/packages/slice-machine/lib/builders/common/EditModal/index.jsx
+++ b/packages/slice-machine/lib/builders/common/EditModal/index.jsx
@@ -26,7 +26,7 @@ if (process.env.NODE_ENV !== "test") {
 
 const FORM_ID = "edit-modal-form";
 
-const EditModal = ({ close, data, fields, onSave }) => {
+const EditModal = ({ close, data, fields, onSave, zoneType }) => {
   const { theme } = useThemeUI();
 
   // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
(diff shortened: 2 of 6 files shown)


packages/slice-machine/lib/builders/common/Zone/index.jsx
```diff
@@ -13,6 +13,7 @@ import ZoneHeader from "./components/ZoneHeader";
 import EmptyState from "./components/EmptyState";
 
 const Zone = ({
+  zoneType /* type of the zone: custom_type or slice */,
```
bapmrl: 💡 I don't think we're using snake casing in a lot of places currently. Would camel casing work for you? What do you think?

packages/slice-machine/lib/builders/common/EditModal/index.jsx
```diff
@@ -161,11 +161,53 @@ const EditModal = ({ close, data, fields, onSave }) => {
             initialValues,
           } = props;
 
+          const fieldModelTabContent = CustomForm ? (
```
bapmrl, line 164: ⚠️ You're testing for `zoneType === "slice"` in 3 different places. It makes the code a bit hard to read in my opinion. What about something like this?

```
const tabs = ["Field Model"]
const cardContent = [CustomForm ? /* ... */]

if (zoneType === "slice") {
  tabs.push("Mock Data")
  cardContent.push(
    <Box>
      <DeprecatedMockConfigMessage />
    </Box>
  )
}
```

bapmrl (changes requested):

packages/slice-machine/lib/builders/common/Zone/index.jsx
xrutayisire: Better, you're right! Done

packages/slice-machine/lib/builders/common/EditModal/index.jsx
xrutayisire: Lot simpler thanks, updated it

packages/slice-machine/components/FormFields/CheckboxControl.jsx
```diff
@@ -6,12 +6,13 @@ import { FormFieldCheckbox } from "./";
  * This components allows to set/unset the value of an arbitrary field via
  * checking/unchecking the box
  *
- * Field: the controlled Formik field that we want to manipulate label: a
- * function of text displayed next to the checkbox. Can be either a string or a
- * function controlledValue: the value of the field being controlled
- * setControlledValue: function to update the value of the controlled value in
- * Formik buildControlledValue: function to build the new value based on the
- * current controlled value and the state of the checkbox
+ * - "field": the controlled Formik field that we want to manipulate
```
xrutayisire: Also update comment format from last MR since it was on master ;)

bapmrl: 💡 Okay, but I would either not change anything or use something closer to the [TSDoc](https://tsdoc.org) format introduced by DevX in other parts of the code. It would give something like:
```
/**
 * This components allows to set/unset the value of an arbitrary field via
 * checking/unchecking the box
 *
 * @param field - the controlled Formik field that we want to manipulate
 * @param label - a function of text displayed next to the checkbox. Can be either a
 *   string or a function
 * @param controlledValue - the value of the field being controlled
 * @param setControlledValue - function to update the value of the controlled value in
 * @param buildControlledValue - Formik function to build the new value based on the
 *   current controlled value and the state of the checkbox
 */
```

packages/slice-machine/lib/builders/common/EditModal/index.jsx
```diff
@@ -161,11 +161,52 @@ const EditModal = ({ close, data, fields, onSave }) => {
...
+                    fields={fields}
+                    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+                    initialValues={initialValues}
+                  />
+                </Col>
+              ))}
+            </FlexGrid>
+          );
+
+          const mockDataTabContent = zoneType === "slice" && (
```
bapmrl: 💡 Not a big deal, but why declare this `const` outside of the `if (zoneType === "slice")` as the current expression introduces an extraneous `zoneType === "slice"` expression?

bapmrl (approved):
I've added two additional comments, but feel free to apply them @xrutayisire.

packages/slice-machine/lib/builders/common/EditModal/index.jsx
xrutayisire: My bad honestly, just copy and paste from old code, will fix it

packages/slice-machine/components/FormFields/CheckboxControl.jsx
xrutayisire: Ah nice, I will take the new way to do then. If I don't touch anything prettier break the comment, and it's not really readable

PR #948, opened 2023-04-28
Don't update slice with screenshots on pushChanges

hypervillain:
~~`upsertGlobalContentTypes` is called on each slice:update, which happens once per change in `PrismicRepositoryManager.pushChanges`. This PR fixes the problem by debouncing calls to `upsertGlobalContentTypes`~~

you don't need to update local model when pushing changes with new screenshots, this PR removes all calls to `updateSlice`, making push much faster. It also makes sure ACL is initialised before pushing screenshots.

A second PR will optimise updateSlice call to upsertGlobalContentTypes only if a change to the model was made


packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
```diff
@@ -279,15 +279,10 @@ export class PrismicRepositoryManager extends BaseManager {
 										model,
 									});
 
-								await this.slices.updateSlice({
-									libraryID: change.libraryID,
-									model: modelWithScreenshots,
-								});
```
lihbr, line 285: I don't think this can be removed. Otherwise, from my understanding, the slice will stay on "changed" status because its model image URL will always be different from the local screenshot.

packages/manager/src/managers/screenshots/ScreenshotsManager.ts
```diff
@@ -253,6 +253,9 @@ export class ScreenshotsManager extends BaseManager {
 	async uploadScreenshot(
 		args: ScreenshotsManagerUploadScreenshotArgs,
 	): Promise<ScreenshotsManagerUploadScreenshotReturnType> {
+		if (this._s3ACL === undefined) {
+			await this.initS3ACL();
+		}
```
lihbr: I don't get why this need to appear here, the ACL is inited upon start and/or login of user. This could cause issues if the user is not logged in.

ACL assertion is made after to ensure the ACL is ready.

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
```diff
@@ -306,15 +301,10 @@ export class PrismicRepositoryManager extends BaseManager {
...
```
lihbr, line 312: Same as above, I don't think we can get rid of that.

lihbr (changes requested):
~~I'm sorry, I don't understand what is made in this PR, to me this creates additional issues.~~

~~Maybe I get the issue wrong completely 🤷‍♀️ Shared additional details with you on Slack~~

LGTM
1. Let's resolve the ACL init discussion (see below)
2. Let's also remove the slice update inside `sliceManager.pushSlice` to make it consistent with `pushChanges`

packages/manager/src/managers/screenshots/ScreenshotsManager.ts
lihbr: Just to add some context, the reason we prefer to init the ACL context separately (similarly to the reason we init the plugin runner separately and not in each function), is because it makes the app flow much simpler to debug.

- **ACL/plugin runner are like car engines, you start them once, and then leverage them to drive**
   *In the case of the ACL, we start it once when we can (when user is logged in or just logged in), you start the car engine when you have the key for it.*
- **Decoupling the engine start from engine usage allows for easier troubleshooting**
   *If you were checking that the engine is started, start it if not, to switch gear, you wouldn't know if the car engine startup failed or the switch gear failed if both are done at the same time.*
   *That's why we just check that the engine is started with `assert...()` functions, so that if this fails, it indicates a clear fail in the way the manager is used (i.e. a bug in the way the code is written, not in the runtime), rather than maybe a bug that happened during the engine startup process.*
- **Decoupling the engine start from engine usage allows for easier implementation of functions using the engine**
   *To keep on going with the car analogy, you don't want to check if the engine is started, and start it if not, whenever you switch gears, accelerate, turn on the windscreen wiper, or turn on the lights. It's better to let it fail gracefully if so, hence the `assert...()` functions indicating a bug in the way the code is written.*

lihbr (approved):
LGTM

Maybe you can just quickly add that?

> Let's also remove the slice update inside `sliceManager.pushSlice` to make it consistent with `pushChanges`

hypervillain:
So fast @lihbr

lihbr (approved):

Issue #958, opened 2023-05-02
Error: "An S3 ACL has not be initialized" when pushing changes


angeloashmore:
### Versions

- slice-machine-ui: `1.0.2`
- node: [unknown]

### Reproduction

N/A

### Steps to reproduce

Push changes to Prismic containing a new screenshot.

### What is expected?

Changes are pushed to Prismic.

### What is actually happening?

An error occurs when uploading the screenshots. See the following screenshot of the error.

![](https://user-images.githubusercontent.com/38848268/235800561-16af62fc-9c69-4ede-aa14-fffc50db85b2.png)

---

This error was originally reported by @draykefriesen and @tjkohli here: https://github.com/prismicio/slice-machine/issues/939#issuecomment-1532242808.

It's possible the error reported in #939 is related to the error reported here.

angeloashmore:
I suggest changing `ScreenshotsManager.prototype.uploadScreenshot()` to initialize an S3 ACL rather than asserting that it was initialized. `initS3ACL()` is idempotent, so we can initialize it early _and_ just in time in case it wasn't already initialized.

https://github.com/prismicio/slice-machine/blob/25991210dccfe334f0a41ef31824e552f1025840/packages/manager/src/managers/screenshots/ScreenshotsManager.ts#L256

That that line can become:

```typescript
async uploadScreenshot(
	args: ScreenshotsManagerUploadScreenshotArgs,
): Promise<ScreenshotsManagerUploadScreenshotReturnType> {
	await this.initS3ACL();

	// ...
}
```

hypervillain:
This should be fixed with version `1.0.3`. Thanks for your report!

Issue #953, opened 2023-05-02
Error "Unable to upload screenshot with status code: 403" when pushing changes


mdeclercq:
From https://github.com/prismicio/slice-machine/issues/939#issuecomment-1528790787

Users are reporting that they get an error 403 when trying to push changes. We are working on the issue.
I updated slice-machine-ui to 1.0.2 and also got an error when trying to push changes.

- `node v18.16.0`
- `npm v8.1.2`

```
    "@prismicio/client": "^6.7.3",
    "@prismicio/helpers": "^2.3.9",
    "@prismicio/next": "^1.0.3",
    "@prismicio/react": "^2.5.2",
    "@prismicio/slice-simulator-react": "^0.2.3",
    "next": "13.3.1",
    "react": "18.2.0",
    "@slicemachine/adapter-next": "^0.1.2",
    "slice-machine-ui": "^1.0.2"
```

```
An error happened while pushing your changes
Error: Unable to upload screenshot with status code: 403
    at ScreenshotsManager.uploadScreenshot (D:\GitHub\prismic-next-starter\node_modules\@slicemachine\manager\dist\managers\screenshots\ScreenshotsManager.cjs:220:13)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async D:\GitHub\prismic-next-starter\node_modules\@slicemachine\manager\dist\managers\slices\SlicesManager.cjs:394:34      
    at async Promise.all (index 0)
    at async SlicesManager.updateSliceModelScreenshotsInPlace (D:\GitHub\prismic-next-starter\node_modules\@slicemachine\manager\dist\managers\slices\SlicesManager.cjs:371:24)
    at async D:\GitHub\prismic-next-starter\node_modules\@slicemachine\manager\dist\managers\prismicRepository\PrismicRepositoryManager.cjs:217:44
    at async Promise.all (index 0)
    at async PrismicRepositoryManager.pushChanges (D:\GitHub\prismic-next-starter\node_modules\@slicemachine\manager\dist\managers\prismicRepository\PrismicRepositoryManager.cjs:188:26)
    at async Module.handleRPCRequest (D:\GitHub\prismic-next-starter\node_modules\@slicemachine\manager\dist\_node_modules\r19\dist\handleRPCRequest.cjs:58:11)
[...]

PR #959, opened 2023-05-03
fix: update AWS ACL before uploading screenshots

bapmrl:
## Context

Resolves SM-1272/DT-1272 (#953 and #958): Cannot push changes because screenshot upload returns 403.

## The Solution

1. Always call `ScreenshotsManager.initS3ACL` in `SlicesManager.updateSliceModelScreenshotsInPlace`.

- [x] The CI is successful.

### DIFF
--- a/packages/manager/src/managers/screenshots/ScreenshotsManager.ts
+++ b/packages/manager/src/managers/screenshots/ScreenshotsManager.ts
@@ -118,9 +118,10 @@ export class ScreenshotsManager extends BaseManager {
 	}
 
 	async initS3ACL(): Promise<void> {
-		if (this._s3ACL) {
-			return;
-		}
+		// TODO: we need to find a way to create a new AWS ACL only when necessary (e.g., when it has expired).
+		// if (this._s3ACL) {
+		// 	return;
+		// }
 
 		const awsACLURL = new URL("create", API_ENDPOINTS.AwsAclProvider);
 		const awsACLRes = await this._fetch({ url: awsACLURL });
--- a/packages/manager/src/managers/slices/SlicesManager.ts
+++ b/packages/manager/src/managers/slices/SlicesManager.ts
@@ -641,8 +641,7 @@ export class SlicesManager extends BaseManager {
 	): Promise<SharedSlice> {
 		const sliceMachineConfig = await this.project.getSliceMachineConfig();
 
-		// Update the AWS ACL before uploading screenshots as it might have expired
-		await this.screenshots.initS3ACL();
+		let initS3ACL = false;
 
 		const variations = await Promise.all(
 			args.model.variations.map(async (variation) => {
@@ -667,6 +666,12 @@ export class SlicesManager extends BaseManager {
 				// If screenshot hasn't changed, do nothing
 				if (!hasScreenshotChanged) {
 					return variation;
+				}
+
+				// Update the AWS ACL before uploading the first screenshot as it might have expired
+				if (!initS3ACL) {
+					await this.screenshots.initS3ACL();
+					initS3ACL = true;
 				}
 
 				const keyPrefix = [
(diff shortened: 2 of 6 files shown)


packages/manager/src/managers/slices/SlicesManager.ts
```diff
@@ -641,6 +641,9 @@ export class SlicesManager extends BaseManager {
 	): Promise<SharedSlice> {
 		const sliceMachineConfig = await this.project.getSliceMachineConfig();
 
+		// Update the AWS ACL before uploading screenshots as it might have expired
+		await this.screenshots.initS3ACL();
+
```
lihbr: I'd maybe move that down the stream right to the `uploadScreenshot` method, this way if there's no new screenshot we don't even need to bother with initing the ACL and other methods relying on `uploadScreenshot` would not fail too.

packages/manager/src/managers/screenshots/ScreenshotsManager.ts
```diff
@@ -118,9 +118,10 @@ export class ScreenshotsManager extends BaseManager {
 	}
 
 	async initS3ACL(): Promise<void> {
-		if (this._s3ACL) {
-			return;
-		}
+		// TODO: we need to find a way to create a new AWS ACL only when necessary (e.g., when it has expired).
+		// if (this._s3ACL) {
+		// 	return;
+		// }
```
lihbr, line 124: I assume this would require an endpoint on our S3 API to validate the session, maybe we could add a small cache/debounce here (like a 60s cache) in the meantime to prevent spamming the S3 ACL init endpoint when we're uploading all screenshots at once.

We're doing something similar on the client to dedupe `/api/v2` request (a bit hard to follow at first I admit): https://github.com/prismicio/prismic-client/blob/v7/src/createClient.ts#L1858-L1910

lihbr (approved):
LGTM, just some suggestions if you want to give it a shot

packages/manager/src/managers/slices/SlicesManager.ts
xrutayisire: I propose this version:

Screenshot 2023-05-03 at 14 44 18

The goal is to only do it if necessary one time for each "updateSliceModelScreenshotsInPlace" function.
It also prevents us to do the cache management since it will be done only one time for all screenshots.

We will need anyway to improve the check at the TODO mentioned.

packages/manager/src/managers/screenshots/ScreenshotsManager.ts
bapmrl: Hello @lihbr,

I think @xrutayisire's solution is the simplest for now.

I'll create an issue in Linear to implement the endpoint to validate the session 🙂

Thank you!

packages/manager/src/managers/slices/SlicesManager.ts
bapmrl: Thanks!

lihbr: Sounds good to me, just a heads up though that this *might* not work because this method is called in parallel (concurrently, not sequentially) inside `pushChanges`: https://github.com/prismicio/slice-machine/blob/bmo/fix-init-s3-acl/packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts#L262

If you want to be sure the ACL is inited only once, you could instead move it there, before the line I linked to.

xrutayisire: If we put the init function at the top of "pushChanges" function, "updateSliceModelScreenshotsInPlace" it's good because it will be called only once for sure BUT we will not be able to check that we do it only if there is a screenshot to upload :/ Right?

bapmrl: Thanks @lihbr, it's better indeed. I've updated the code.

lihbr: You're right @xrutayisire, that's a compromise

xrutayisire (approved):

packages/manager/src/managers/slices/SlicesManager.ts
bapmrl: At least it works like the init process now 😊 The init process first calls `initS3ACL` and then calls `pushSlice` (in parallel) on each slice.

PR #960, opened 2023-05-03
fix(http-connect): try wrapping fetch so that it's passed and agent t…

MarcMcIntosh:
https://linear.app/prismic/issue/DT-1271/pushing-changes-sometimes-fails-with-error-getaddrinfo-enotfound

Looks like enotfound errors could be coming from node-fetch
https://github.com/node-fetch/node-fetch/issues/449
and
https://github.com/node-fetch/node-fetch/issues/79
seems that lots of requests, looks like people are fixing it by passing a user agent https://github.com/node-fetch/node-fetch/issues/1735

- [ ] The CI is successful.

### DIFF
--- a/packages/manager/src/lib/fetch.ts
+++ b/packages/manager/src/lib/fetch.ts
@@ -1,5 +1,5 @@
 // This temporary wrapper around `node-fetch` fixes an issue where quick
-// consecutive network requests causes failed requests.
+// consecutive network requests cause failed requests.
 //
 // See https://github.com/node-fetch/node-fetch/issues/1735 for more details.
 //
--- a/packages/manager/src/lib/fetch.ts
+++ b/packages/manager/src/lib/fetch.ts
@@ -12,8 +12,8 @@ const httpsAgent = new https.Agent({
 });
 
 const options: RequestInit = {
-	agent: function (_parsedURL) {
-		if (_parsedURL.protocol == "http:") {
+	agent: function (parsedURL) {
+		if (parsedURL.protocol == "http:") {
 			return httpAgent;
 		} else {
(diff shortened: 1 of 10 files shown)


packages/manager/src/lib/fetch.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const httpAgent = new http.Agent({
+	keepAlive: true,
+});
+const httpsAgent = new https.Agent({
+	keepAlive: true,
+});
+
+const options: RequestInit = {
+	agent: function (_parsedURL) {
```
bapmrl: 💡 `_parseUrl` is used so there's no need to append `_`.

bapmrl (commented):
⚠️ This fix might work, but it would be reassuring if we could find a way to reproduce the issue 🙂

Also, I think the manager's tests should be using the same `packages/manager/src/lib/fetch.ts` module for consistency.

Note that I found more explanations here: https://github.com/node-fetch/node-fetch/issues/1735.

packages/manager/src/lib/fetch.ts
MarcMcIntosh: done

angeloashmore (commented):
Inline with @bapmrl's comment, it would be good to have a reproduction of the issue or a test that confirms this fix is effective. However, I understand that would be tricky to implement.

The implementation could be simplified, in my opinion, as well as benefit from some context around why the wrapper exists.

I opened a PR with some changes (it was quicker to write the changes than to explain them): https://github.com/prismicio/slice-machine/pull/961

Feel free to use it or not. 🙂

PR #964, opened 2023-05-04
fix(init): when the plugin fails, the init should continue

MarcMcIntosh:
## Context
When running the init and the configuration step fails the init should continue.

- [ ] The CI is successful.


xrutayisire (approved):
LGTM 👍

PR #968, opened 2023-05-09
chore: write components using @prismicio/editor-ui

bapmrl:
Completes DT-1098: AADev, I can write UI components in SM using the Editor's stack.


packages/slice-machine/.storybook/main.ts
```diff
@@ -0,0 +1,10 @@
+import type { StorybookConfig } from "@storybook/react-vite";
+
+const config: StorybookConfig = {
+  addons: ["@storybook/addon-essentials"],
+  docs: { autodocs: "tag" },
+  framework: "@storybook/react-vite",
+  stories: ["../src/**/*.stories.ts?(x)"],
```
xrutayisire: ❓ Will we have a story .ts and not .tsx? I say that because I don't think I created a one without .tsx in the past

packages/slice-machine/package.json
```diff
@@ -16,6 +16,8 @@
     "dev-cypress": "concurrently \"next dev\" \"npm run dev --prefix ../../e2e-projects/cypress-next-app\"",
     "lint": "eslint --ext .ts,.tsx,.js,.jsx .",
     "prepublishOnly": "npm run build",
+    "storybook": "STORYBOOK_DISABLE_TELEMETRY=1 storybook dev --no-open --port 6006",
```
xrutayisire: ⚠️ To make the package.json cleaner, I would but the telemetry param in the .storybook/main.js file: https://storybook.js.org/docs/react/configure/telemetry#how-to-opt-out

packages/slice-machine/package.json
```diff
@@ -31,6 +33,8 @@
     "@prismicio/mocks": "2.0.0-alpha.2",
     "@prismicio/types-internal": "2.0.0-alpha.9",
     "@slicemachine/manager": "0.2.2-dev-page-types.1",
+    "@vanilla-extract/css": "1.11.0",
```
xrutayisire, line 36: ❓ Do you have a preference between "1.11.0" and "^1.11.0"?
On my side, I always use the "1.11.0" form because I like to know the exact version directly from package.json

packages/slice-machine/src/components/PageLayout/PageLayout.tsx
```diff
@@ -0,0 +1,18 @@
+import type { FC, ReactNode } from "react";
+import { Text } from "@prismicio/editor-ui";
+
+import * as styles from "./PageLayout.css";
+
+type PageLayoutProps = Readonly<{
+  children?: ReactNode;
+}>;
+
+export const PageLayout: FC<PageLayoutProps> = ({
```
xrutayisire, line 12: ❓ To create component you prefer this way with FC? 
To have a coherent way to do in the future. Personally I don't really care, function, arrow function, arrow function with FC 😄

xrutayisire (changes requested):
Nice work!

xrutayisire (changes requested):
Nice work!

packages/slice-machine/.storybook/main.ts
bapmrl: It's true that I've never created a story without JSX inside it. Do you think we should only allow `.tsx` stories? By default, Storybook allows `.ts` stories.

packages/slice-machine/package.json
bapmrl: Okay, I'll change that 🙂 I didn't want the `boot` event to be sent, but it's fine as it doesn't contain any data.

packages/slice-machine/package.json
bapmrl: I think we should only use strict versions. Otherwise, when we need to recreate the `package-lock.json` file (for some technical issue), every dependency may change and then we need to do a full QA round.

packages/slice-machine/src/components/PageLayout/PageLayout.tsx
bapmrl: It's an habit I have, but we can change if you think another approach is better.

xrutayisire: No problem honestly on my side, again I will do the same to try not having 1000 ways to do :)

packages/slice-machine/package.json
xrutayisire: Nice, I like this also then!

packages/slice-machine/.storybook/main.ts
xrutayisire: I mean not really critical, it was more a question like I don't know why we would use that. Not a problem to keep it if it's the default. As you want ;)

packages/slice-machine/src/components/PageLayout/PageLayout.tsx
bapmrl: Okay, let's use `FC` and we'll see what happens ^^

packages/slice-machine/.storybook/main.ts
bapmrl: Let's keep it to avoid confusion if at some point we need a `.ts` file 🙂 And it's also the default (I've just removed the `.js` and `.jsx` extensions).

xrutayisire (approved):

xrutayisire:
What do you think about adding a build storybook GitHub action? It could be sure in the future to deploy the storybook, but currently it could help us to know when someone will break storybook because of the difficult deps management.

xrutayisire:
LGTM otherwise, good to merge, super nice to have this now, thanks!

PR #974, opened 2023-05-16
feat(page types): update Custom Types creation and edition with default values and restrictions

xrutayisire:
## Context

Completes DT-1323: [AAUser, I can create a Custom Type with default fields and restrictions](https://linear.app/prismic/issue/DT-1323/aauser-i-can-create-a-custom-type-with-default-fields-and-restrictions)

## The Solution

- Faking a customTypeFormat const that will come from custom type value later on
- Passing it to be sure we don't allow removing the UID default field (discussed with Côme about that, and we want to let the possibility to old Custom Type before format) 
- Update e2e tests with changes
- Add new e2e tests to protect changes
- Fix unit tests with changes

## Impact / Dependencies

- Don't add unit tests because it will be simpler with the new version of prismic-types-internal
- With these changes it's not possible to link slices to a Custom Types since Page Types is not available yet

- [x] The CI is successful.

## Preview

https://github.com/prismicio/slice-machine/assets/19946868/9479e7ec-f0ec-49b5-9f58-6bc400a93168

### DIFF
--- a/cypress/e2e/user-flows/scenario_slice_association.cy.js
+++ b/cypress/e2e/user-flows/scenario_slice_association.cy.js
@@ -11,7 +11,8 @@ const sliceName = `TestSlice${random}`;
 const sliceId = `test_slice${random}`; // generated automatically from the slice name
 const sliceLib = ".--slices";
 
-describe("I am an existing SM user (Next) and I want to associate a Slice to a CT and review my experience.", () => {
+// TODO: Use this test suite for Page Types since new Custom Type don't have Slice Zone anymore
+describe.skip("I am an existing SM user (Next) and I want to associate a Slice to a CT and review my experience.", () => {
   before(() => {
     cy.clearProject();
   });
@@ -23,13 +24,11 @@ describe("I am an existing SM user (Next) and I want to associate a Slice to a C
   it("Create a Custom type with multiple fields", () => {
     cy.createCustomType(customTypeId, customTypeName);
 
-    cy.addFieldToCustomType("UID", "ID Field", "uid");
     cy.addFieldToCustomType("Key Text", "Key Text Field", "key_text_id");
     cy.addFieldToCustomType("Rich Text", "Rich Text Field", "rich_text_id");
     customTypeBuilder.save();
 
     cy.reload();
-    cy.contains("ID Field");
     cy.contains("Key Text Field");
     cy.contains("Rich Text Field");
   });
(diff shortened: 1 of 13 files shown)


bapmrl:
DT-1323

cypress/e2e/user-flows/scenario_slice_association.cy.js
```diff
@@ -11,7 +11,8 @@ const sliceName = `TestSlice${random}`;
 const sliceId = `test_slice${random}`; // generated automatically from the slice name
 const sliceLib = ".--slices";
 
-describe("I am an existing SM user (Next) and I want to associate a Slice to a CT and review my experience.", () => {
+// TODO: Use this test suite for Page Types since new Custom Type don't have Slice Zone anymore
```
MarcMcIntosh: Is this mentioned in a ticket somewhere so we know when to come back to it?

xrutayisire: https://linear.app/prismic/issue/DT-1316/aauser-i-can-edit-custom-types-with-features-limited-by-the-format

I just added it in this issue, better you're right, so we don't forget.

packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
```diff
@@ -119,6 +119,7 @@ const FieldZones: React.FunctionComponent<FieldZonesProps> = ({
     <>
       <Zone
         zoneType="slice"
+        customTypeFormat={undefined}
```
MarcMcIntosh, line 122: does undefined have to explicitly passed?

MarcMcIntosh (approved):

packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
xrutayisire: So sad... Yes because of JSX. If you can try another way, I take it!

packages/slice-machine/lib/builders/CustomTypeBuilder/TabZone/index.tsx
```diff
@@ -156,15 +159,23 @@ const TabZone: React.FC<TabZoneProps> = ({ tabId, fields, sliceZone }) => {
...
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-        tabId={tabId}
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-        sliceZone={sliceZone}
-        onRemoveSharedSlice={onRemoveSharedSlice}
-        onCreateSliceZone={onCreateSliceZone}
-        onSelectSharedSlices={onSelectSharedSlices}
-      />
+      {
+        // TODO: DT-1316 - Add condition to also display the slice zone for the Main tab of "page" format
```
MarcMcIntosh, line 163: Did this a different way :)

PR #978, opened 2023-05-19
feat(slice-machine-ui): New table component

xrutayisire:
## Context

Completes DT-1106: [AADev, I need a redesigned Table component](https://linear.app/prismic/issue/DT-1106/aadev-i-need-a-redesigned-table-component) 

## The Solution

- [x] Write the Table, TableHead, TableBody, TableRow and TableCell components
- [x] Create a Table story with the new components
  - [x] The story should look like the Figma wireframe so we now if we miss something
- [x] Create a TableRowContext that will handle the scope of the cell, "td" or "th"
- [x] Create a new temporary component for the Icon and IconButton that is basically a copy of editor one (new issue needed to properly handle that because we have different icons, sizes…)

## Impact / Dependencies

- No ellipsis management done for the responsiveness, we need a tooltip and this should be done after [this issue](https://linear.app/prismic/issue/DT-1339/aadev-i-need-a-tooltip-component-in-the-editor-ui-package).
- I think we need an issue for our Icon management, using the one from Editor require changes, not a blocker for the first Milestone

- [ ] The CI is successful.

## Preview

https://github.com/prismicio/slice-machine/assets/19946868/ae4a9a4b-1ea7-4ae1-bf53-1a5c4960669b


benjammartin:
1. There is an issue with the placement of the context menu.
2. It is necessary to include ellipses when labels are too lengthy.
3. It appears that the font weight of the name is incorrect.

xrutayisire:
> 1. There is an issue with the placement of the context menu.
> 2. It is necessary to include ellipses when labels are too lengthy.
> 3. It appears that the font weight of the name is incorrect.

Thank you @benjammartin !

1. Done
2. If we want the full ellipsis + tooltip management it will cost more, I will see with Baptiste about that.
3. Done

xrutayisire:
Approved with Baptiste during a pair review

PR #982, opened 2023-05-24
Mm/page default fields

MarcMcIntosh:
## Context
https://linear.app/prismic/issue/DT-1316/aauser-i-can-create-and-edit-page-types-with-defaults-fields-and

- [ ] The CI is successful.

### DIFF
--- a/packages/slice-machine/src/modules/availableCustomTypes/factory.ts
+++ b/packages/slice-machine/src/modules/availableCustomTypes/factory.ts
@@ -1,36 +1,75 @@
-import { CustomTypeSM, TabFields } from "@lib/models/common/CustomType";
+import { CustomTypeSM, TabSM } from "@lib/models/common/CustomType";
+
+const DEFAULT_SEO_TAB: TabSM = {
+  key: "Metadata",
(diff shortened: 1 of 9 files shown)


packages/slice-machine/src/modules/availableCustomTypes/factory.ts
```diff
@@ -1,35 +1,74 @@
...
+            width: 2400,
+            height: 1260,
+          },
+        },
+      },
+    },
+  ],
+};
 
 export const createCustomType = (
```
xrutayisire, line 84: We need to add the condition for format "page" or "custom" here. Your work is only for Page types not Custom types. With new Prismic types internal 2.0.0.beta.1, you should be able to have the condition.

Then you can property manage it. Sorry if the issue was not clear

xrutayisire (changes requested):
You didn't handle the restrictions in this PR, you want to do it in another one? I'm talking about the display or not of the Static Zone ect, the creation modal when it's page or custom.

xrutayisire (approved):

PR #989, opened 2023-05-30
test: create-custom-type: return custom-type from types-internal

MarcMcIntosh:
## Context
Linear ticket 
https://linear.app/prismic/issue/DT-1376/createcustomtype-should-use-types-from-prismiciotypes-internal

- [x] The CI is successful.

### DIFF
--- a/packages/slice-machine/test/src/modules/availableCustomTypes/factory.test.ts
+++ b/packages/slice-machine/test/src/modules/availableCustomTypes/factory.test.ts
@@ -108,7 +108,7 @@ describe("[Custom types factory]", () => {
     });
   });
 
-  it("when non repeatable page type", () => {
+  it("when non repeatable page type is should contain Main with a slice-zone, no uid, and a SEO tab", () => {
     const result = createCustomType("foo", "bar", false, "page");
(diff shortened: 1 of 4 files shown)


packages/slice-machine/test/src/modules/availableCustomTypes/factory.test.ts
```diff
@@ -6,130 +6,156 @@ describe("[Custom types factory]", () => {
...
+          },
+          "label": "label",
+          "repeatable": true,
+          "status": true,
         }
       `);
     });
   });
+
+  it("when non repeatable page type", () => {
```
xrutayisire: ❓ What do you prefer for test title? I saw that the 3 tests above use "should ...." and this one "when ..."

I'm use to both but just wanna now, so maybe we can follow the same naming?

MarcMcIntosh: ah, yes i probally should add the extected result in the title, truth be told I wasn't quite sure what it should do so took a guess that the SEO tab should be there

xrutayisire (approved):
After test title change LGTM!

PR #990, opened 2023-05-30
fix(adapter-next,adapter-nuxt,adapter-nuxt2): use SemVer range

lihbr:
## Context

Not using SemVer ranges for dependencies can have undesired side effects when not explicitely required. For example users would not get patch releases when pinning a stable version.

Currently, this prevent users from getting a fix we released on `prismic-ts-codegen`, making pre-client v7 app types not working.

- [x] The CI is successful.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34812,9 +34812,9 @@
       }
     },
     "node_modules/prismic-ts-codegen": {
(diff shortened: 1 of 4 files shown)


xrutayisire (approved):
LGTM

PR #991, opened 2023-05-31
fix(init): Fix tracking for framework and starter when create a repository

xrutayisire:
## Context

- Completes DT-1379

## The Solution

- In SM changing "next-11-13", "nuxt-2" or "nuxt-3" to "next" or "nuxt" that is sent to wroom but keep the specific framework for SM tracking.
- Sending a starter id from SM (thanks to package.json) AND Set the starterId parameter in the createDomain function in wroom  (PR soon in wroom)

## Impact / Dependencies

- new init procedure to detect the starter

- [ ] The CI is successful.

### DIFF
--- a/packages/init/src/lib/starters.ts
+++ b/packages/init/src/lib/starters.ts
@@ -1,0 +1,30 @@
+import { readFile } from "node:fs/promises";
+import { join } from "node:path";
+
+import { StarterID } from "@slicemachine/manager";
+
+const STARTERS_REPOSITORY_NAME_TO_ID: Record<string, StarterID> = {
+	"nextjs-starter-prismic-multi-page": "next_multi_page",
+	"nextjs-starter-prismic-blog": "next_blog",
+	"nextjs-starter-prismic-multi-language": "next_multi_lang",
+	"nuxt-starter-prismic-multi-page": "nuxt_multi_page",
+	"nuxt-starter-prismic-blog": "nuxt_blog",
+	"nuxt-starter-prismic-multi-language": "nuxt_multi_lang",
+};
+
+export const detectStarterId = async (
+	cwd: string,
+): Promise<StarterID | undefined> => {
+	const path = join(cwd, "package.json");
+
+	try {
+		const pkg = JSON.parse(await readFile(path, "utf-8"));
+
+		return STARTERS_REPOSITORY_NAME_TO_ID[pkg.name];
+	} catch (error) {
+		throw new Error(
+			`Failed to read project's \`package.json\` at \`${path}\``,
+			{ cause: error },
+		);
+	}
+};
--- a/packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
+++ b/packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
@@ -22,8 +22,10 @@ import {
 	PrismicRepositoryUserAgent,
 	PrismicRepositoryUserAgents,
 	RawLimit,
+	StarterID,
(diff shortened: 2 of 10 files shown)


MarcMcIntosh:
Hmm, I wondering what adding a starter-id for non-starters and if it'll trick wroom into thinking they used a starter, and they'll see the starter instructions.

packages/init/src/lib/starters.ts
```diff
@@ -0,0 +1,35 @@
...
+export const detectStarterId = async (
+	cwd: string,
+): Promise<StarterId | undefined> => {
+	const path = join(cwd, "package.json");
+
+	let starterRepositoryName: string | undefined;
+	try {
+		const pkg = JSON.parse(await readFile(path, "utf-8"));
+
+		starterRepositoryName = pkg.name;
```
lihbr: You could return straight from here `return STARTERS_REPOSITORY_NAME_TO_ID[pkg.name]` which would yield `undefined` if not found anyway, and keep the catch to throw.

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
```diff
@@ -40,7 +42,8 @@ type PrismicRepositoryManagerCheckExistsArgs = {
 
 type PrismicRepositoryManagerCreateArgs = {
 	domain: string;
-	framework: string; // TODO: Type(?)
+	framework: FrameworkWroomTrackingId;
+	starterId: StarterId | undefined;
```
lihbr: Why not `starterID?: StarterID`, would allow for implicit undefined also.

lihbr (commented):
Hey, as you asked me, here's some comments, they are about style only, so up to you!

I'd rename "tracking" to "telemetry" to align with the rest of the codebase better (also sounds friendlier imo). Also up to you, but I know Angelo and I prefer to keep capitalized capitalized words, so "ID" instead of "Id".
- `sliceMachineTrackingId` > `sliceMachineTelemetryID`
- `wroomTrackingId` > `wroomTelemetryID`

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
xrutayisire: I like here to force dev to pass a starter because we need the value for tracking, by setting "| undefined" it forces the dev to think about it and not forget this.

What do you think? I'm also ok for implicit if it's more the idea here :)

packages/init/src/lib/starters.ts
xrutayisire: Thanks done :)

MarcMcIntosh:
What happens when running the init from an existing project (non-starter) ?

xrutayisire:
> What happens when running the init from an existing project (non-starter) ?

starterId is undefined and so no starterId is found and sent to wroom

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
lihbr: Up to you! I just find it odd that if you need to create a repo from a non-starter, you have to use

```typescript
create({ domain: "...", framework: "...", starterId: undefined })
```

xrutayisire: You're right, changed it :)

MarcMcIntosh (approved):

MarcMcIntosh:
Which branch are you targeting?

PR #995, opened 2023-06-01
Mm/dt 1333

MarcMcIntosh:
## Context
https://linear.app/prismic/issue/DT-1333/[simulator]-sm-uses-staging-unsplashwroomio-endpoint

Updated the simulator/editor configuration to use the correct endpoints

- [ ] The CI is successful.


packages/adapter-next/package.json
```diff
@@ -74,13 +74,13 @@
 		"@types/prettier": "^2.7.2",
 		"@types/react": "^18.0.28",
 		"@types/semver": "^7.3.13",
-		"@typescript-eslint/eslint-plugin": "^5.55.0",
+		"@typescript-eslint/eslint-plugin": "~5.55.0",
```
xrutayisire: I would not change that in this MR since the CI is flaky, best is to change how we are doing (and super soon) I think Baptiste will start to work on this.

xrutayisire: Your view on this @MarcMcIntosh?

bapmrl: Indeed, I'm starting work on this 😊

packages/slice-machine/components/Simulator/index.tsx
```diff
@@ -305,10 +311,26 @@ const Simulator: ComponentWithSliceProps = ({ slice, variation }) => {
...
 export default Simulator;
 
-// TODO(DT-1333): change this config depending on the environment.
-const SHARED_SLICE_EDITOR_CONFIG = {
-  embeds: {
-    url: "https://jntc6tzga1.execute-api.us-east-1.amazonaws.com/beta",
-  },
-  unsplash: { url: "https://unsplash.wroom.io" },
-};
+function makeEditorConfig(apiEndpoint: string): EditorConfig {
```
bapmrl: ⚠️ Why not use the same mechanism as the `@slicemachine/manager` package? We could directly import https://github.com/prismicio/slice-machine/blob/master/packages/manager/src/constants/API_ENDPOINTS.ts and rely on `SM_ENV`. What do you think? It would avoid introducing another way of specifying config values in the front end 🙂

bapmrl (changes requested):

packages/slice-machine/components/Simulator/index.tsx
MarcMcIntosh: I'll look into it

MarcMcIntosh: call `this.manager.getAPIEndpoints();` somewhere

MarcMcIntosh: updated :)

packages/adapter-next/package.json
MarcMcIntosh: should have been reverted

xrutayisire: What do you mean?

bapmrl: Hello @MarcMcIntosh,

Could you please explain what you mean by "should have been reverted"? These changes don't appear to be related to the Linear issue you're tackling and will conflict with my current work.

Thanks,

packages/manager/src/constants/API_ENDPOINTS.ts
```diff
@@ -7,6 +7,8 @@ export type APIEndpoints = {
 	PrismicModels: string;
 	PrismicUser: string;
 	AwsAclProvider: string;
+	Oembed: string;
```
bapmrl: 💡 Why not name them the same way as other endpoints? If we want to be consistent with what was done, it would be `PrismicOembed` and `PrismicUnsplash`.

packages/manager/src/constants/API_ENDPOINTS.ts
```diff
@@ -20,6 +22,8 @@ export const API_ENDPOINTS: APIEndpoints = (() => {
 				PrismicModels: addTrailingSlash(process.env.customtypesapi_endpoint),
 				PrismicUser: addTrailingSlash(process.env.user_service_endpoint),
 				AwsAclProvider: addTrailingSlash(process.env.acl_provider_endpoint),
+				Oembed: "https://oembed.wroom.io/",
```
bapmrl: ⚠️ Why not use `addTrailingSlash` and introduce overridable environment variables like what was done for `PrismicWroom`, `PrismicAuthentication`, etcetera?

packages/manager/src/types.ts
```diff
@@ -4,6 +4,7 @@ import { detect as niDetect } from "@antfu/ni";
 import { VERSION_KIND } from "./constants/VERSION_KIND";
 
 export type PackageManager = NonNullable<Awaited<ReturnType<typeof niDetect>>>;
+export type { APIEndpoints } from "./constants/API_ENDPOINTS";
```
bapmrl, line 7: 💡 This seems like an unnecessary indirection. Other types are directly exported from `packages/manager/src/index.ts` and this `packages/manager/src/types.ts` module only exports types that don't exist elsewhere. I would recommend sticking to the same practice.

packages/manager/test/SliceMachineManager-getState.test.ts
```diff
@@ -1,5 +1,43 @@
...
+import { createTestProject } from "./__testutils__/createTestProject";
+import { createTestPlugin } from "./__testutils__/createTestPlugin";
+import { mockPrismicUserAPI } from "./__testutils__/mockPrismicUserAPI";
 
-it.todo("returns global Slice Machine state", () => {
-	// TODO
+it("returns global Slice Machine state", async (ctx) => {
+	const adapter = await createTestPlugin();
+	const cwd = await createTestProject({ adapter });
+	mockPrismicUserAPI(ctx);
```
bapmrl: ❓ Nice test! I just wonder why we need to call `mockPrismicUserAPI`? The test wouldn't run otherwise?

packages/slice-machine/components/Simulator/index.tsx
```diff
@@ -53,13 +54,24 @@ const Simulator: ComponentWithSliceProps = ({ slice, variation }) => {
...
+    endpoints,
   } = useSelector((state: SliceMachineStoreType) => ({
     simulatorUrl: selectSimulatorUrl(state),
     iframeStatus: selectIframeStatus(state),
     manifestStatus: selectSetupStatus(state).manifest,
     isWaitingForIFrameCheck: selectIsWaitingForIFrameCheck(state),
+    endpoints: selectEndpoints(state),
   }));
 
+  const editorConfig: EditorConfig = {
```
bapmrl: ⚠️ I've noticed other `props` passed to `SharedSliceEditor` are memoized. I think we should apply the same principle and use `useMemo` for the `editorConfig`.

packages/slice-machine/src/modules/environment/index.ts
```diff
@@ -67,6 +67,12 @@ export const selectSimulatorUrl = (
 export const getRepoName = (store: SliceMachineStoreType): string =>
   store.environment.repo;
 
+export const selectApiEndpoint = (store: SliceMachineStoreType): string =>
```
bapmrl: ⚠️ This function is not used so let's remove it to avoid having too much dead code in the long run.

bapmrl (changes requested):
Nice work! We're nearly there 🙂

packages/manager/src/constants/API_ENDPOINTS.ts
MarcMcIntosh: I don't think oembed and unsplash are pat of the dev-env project.

packages/adapter-next/package.json
MarcMcIntosh: I thought those changes where reverted by now :/

packages/manager/test/SliceMachineManager-getState.test.ts
MarcMcIntosh: Look like the function makes a network call to that endpoint, removing it doesn't seem to break anything

packages/manager/src/types.ts
MarcMcIntosh: okay

packages/manager/src/constants/API_ENDPOINTS.ts
MarcMcIntosh: ok

packages/slice-machine/components/Simulator/index.tsx
MarcMcIntosh: ok

packages/slice-machine/src/modules/environment/index.ts
MarcMcIntosh: done

packages/slice-machine/components/Simulator/index.tsx
MarcMcIntosh: done

packages/manager/test/SliceMachineManager-getState.test.ts
MarcMcIntosh: done

packages/manager/src/types.ts
MarcMcIntosh: done

packages/manager/src/constants/API_ENDPOINTS.ts
MarcMcIntosh: done

packages/manager/src/constants/API_ENDPOINTS.ts
MarcMcIntosh: done

packages/adapter-next/package.json
MarcMcIntosh: done

packages/manager/src/constants/API_ENDPOINTS.ts
```diff
@@ -20,6 +22,12 @@ export const API_ENDPOINTS: APIEndpoints = (() => {
 				PrismicModels: addTrailingSlash(process.env.customtypesapi_endpoint),
 				PrismicUser: addTrailingSlash(process.env.user_service_endpoint),
 				AwsAclProvider: addTrailingSlash(process.env.acl_provider_endpoint),
+				PrismicOembed: addTrailingSlash(
+					process.env.oembed_endpoint ?? "https://oembed.wroom.io",
```
bapmrl, line 26: ❓ I don't know if the fallbacks are necessary. Why all environment variables would be mandatory, but not `oembed_endpoint` and `unsplash_endpoint`? It's fine however and I think we can merge as it is 🙂

bapmrl (commented):
⚠️ Have you deleted the `package-lock.json` file? It shouldn't have changed if you didn't modify any `package.json` file 🧐

packages/manager/src/constants/API_ENDPOINTS.ts
MarcMcIntosh: oembed and unsplash aren't part of the dev-env (last time i checked)

bapmrl (approved):

packages/manager/src/constants/API_ENDPOINTS.ts
MarcMcIntosh: Turns out the editor cant handle trailing slashes :/

PR #997, opened 2023-06-02
fix(init): Fix init for starters

xrutayisire:
## Context

- Completes DT-1386

## The Solution

- Fix init by detecting the starter used

## Impact / Dependencies

- Need PR from wroom to send starterId https://github.com/prismicio/wroom/pull/2470

- [x] The CI is successful.

### DIFF
--- a/packages/init/src/SliceMachineInitProcess.ts
+++ b/packages/init/src/SliceMachineInitProcess.ts
@@ -13,6 +13,7 @@ import {
 	PrismicRepository,
 	SliceMachineManager,
 	PackageManager,
+	StarterId,
 } from "@slicemachine/manager";
 
 import pkg from "../package.json";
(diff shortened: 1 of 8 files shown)


packages/init/src/SliceMachineInitProcess.ts
```diff
@@ -270,23 +271,22 @@ export class SliceMachineInitProcess {
...
+								this.context.starterID = await detectStarterId(
+									this.manager.cwd,
+								);
+
+								if (this.context.starterID) {
+									task.title = `Detected starter ${chalk.cyan(
+										this.context.starterID,
+									)}`;
+								} else {
+									task.title = "No starter detected";
```
MarcMcIntosh, line 286: hmm, they'll no what a starter is so I guess this won't cause any panic

MarcMcIntosh (approved):
One comment about wording, but it's nothing major :)

packages/init/src/SliceMachineInitProcess.ts
xrutayisire: You think I should change it to something less like a problem?

Maybe not even writing something?

xrutayisire: Discussed with Sam, it's ok like that 👍

lihbr: It appears for a split second anyway because the task drawer collapse right after

xrutayisire: Good one, true

PR #999, opened 2023-06-02
DT-1383 I can see the repository name and access it through the ui

MarcMcIntosh:
## Context
https://linear.app/prismic/issue/DT-1383/aauser-i-can-see-my-repository-name-and-access-it-through-the-slice

Screenshot 2023-06-06 at 14 56 37

- [ ] The CI is successful.


packages/slice-machine/components/AppLayout/Navigation/Menu/Logo.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+      environment: {
+        manifest: {
+          apiEndpoint: "https://foo.cdn.prismic.io/api/v2",
+        },
+      } as FrontEndEnvironment,
+    },
+  });
+
+  test("should display the repo name and domain from the api endpoint and a link the repo", async () => {
+    expect(await container.findByText("Foo")).toBeInTheDocument();
```
xrutayisire: It's recommended to use screen over container for all queries with react testing library: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

xrutayisire: It's better to check for toBeVisible that will check that the element is actually visible on the screen, when toBeInTheDocument only imply it's in the DOM. See: https://github.com/testing-library/jest-dom#tobevisible

packages/slice-machine/components/AppLayout/Navigation/Menu/Logo.tsx
```diff
@@ -1,27 +1,84 @@
 import React from "react";
-import Link from "next/link";
-import { Box, Flex, Heading, Link as ThemeLink, useThemeUI } from "theme-ui";
-import SliceMachineLogo from "../Icons/SliceMachineLogo";
+import { useSelector } from "react-redux";
+import { SliceMachineStoreType } from "@src/redux/type";
+import * as styles from "./Logo.css";
+
+const LogoIcon: React.FC<React.SVGProps<SVGSVGElement>> = (props) => {
```
xrutayisire: For the 2 new svgs, we have a new folder for that in src/icons

packages/slice-machine/components/AppLayout/Navigation/Menu/Logo.tsx
```diff
@@ -1,27 +1,84 @@
...
+}
 
 const Logo: React.FC = () => {
-  const { theme } = useThemeUI();
+  // TODO: reuse selector from DT-1333
+  const apiEndpoint = useSelector((state: SliceMachineStoreType) => {
+    return state.environment.manifest.apiEndpoint;
+  });
+
+  const repoDomain = new URL(apiEndpoint).hostname.replace(".cdn", "");
```
xrutayisire, line 18: Thanks to the login process and the fact we have an API, we can't ask of the repo URL or something?
repoNameOrNull feel like we could not have it? What are the use cases for that?

packages/slice-machine/components/AppLayout/Navigation/Menu/Logo.tsx
```diff
@@ -1,27 +1,84 @@
...
+        fillRule="evenodd"
+        clipRule="evenodd"
+        d="M9.5 9C9.22386 9 9 9.22386 9 9.5V22.4996C9 22.7758 9.22386 22.9996 9.5 22.9996H22.5C22.7761 22.9996 23 22.7758 23 22.4996V17.5C23 17.2239 23.2239 17 23.5 17C23.7761 17 24 17.2239 24 17.5V22.4996C24 23.3281 23.3284 23.9996 22.5 23.9996H9.5C8.67157 23.9996 8 23.3281 8 22.4996V9.5C8 8.67157 8.67157 8 9.5 8H14.5C14.7761 8 15 8.22386 15 8.5C15 8.77614 14.7761 9 14.5 9H9.5ZM24 8H18L20.6464 10.6464L15.6464 15.6464C15.4512 15.8417 15.4512 16.1583 15.6464 16.3536C15.8417 16.5488 16.1583 16.5488 16.3536 16.3536L21.3536 11.3536L24 14V8Z"
+        fill="#6F6E77"
+      />
+    </svg>
+  );
+};
+
+function capataliseFirstLetter(word: string | null): string | null {
```
xrutayisire: I feel this function could be exported in a util folder like src/utils, what do you think?

Also, I would not give null as a param, here the name imply we should have the value, if not just don't call the function no?

packages/slice-machine/components/AppLayout/Navigation/Menu/Logo.css.ts
```diff
@@ -0,0 +1,36 @@
+import { sprinkles } from "@prismicio/editor-ui";
+import { style } from "@vanilla-extract/css";
+
+export const container = style({ padding: "0 14px" });
+
+export const icon = style([
```
xrutayisire: No need for style of sprinkles if you have only sprinkles

xrutayisire: Same for the one after

packages/slice-machine/components/AppLayout/Navigation/Menu/Logo.css.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const flex = style([
+  sprinkles({
+    display: "flex",
+    justifyContent: "space-between",
+  }),
+]);
+
+export const title = {
+  // TODO: sprinkles doesn't do fonts or colors and style is over ridden by some global selector (probably theme-ui), so this needs to go inline
```
xrutayisire: With sprinkles or style you can always find a way, when you are blocked on that, call @bapmrl or me (even if Baptiste is the pro on that) so we could find a better way and still using sprinkles and style for the two elements here 😊 

It's important to share a common theme, color ect

And you're right, sprinkles with the editor at the moment don't let us handle: fontWeight, fontSize, lineHeight :/

packages/slice-machine/components/AppLayout/Navigation/Menu/Logo.tsx
```diff
@@ -1,27 +1,84 @@
...
-          </Flex>
-        </ThemeLink>
-      </Link>
-    </Box>
+    <div className={styles.container}>
+      <LogoIcon className={styles.icon} />
+
+      <div className={styles.flex}>
+        <div>
+          <h1 style={styles.title}>{repoName}</h1>
```
xrutayisire: Prefer `<Text component="h1"></Text>` component from the editor, same for the repo domain.

packages/slice-machine/components/AppLayout/Navigation/Menu/Logo.tsx
```diff
@@ -1,27 +1,84 @@
...
+
+          <h6 style={styles.repoUrl}>{repoDomain}</h6>
```
xrutayisire: Why h6? I'm not sure h6 is the most appropriate here.

For me, we should have one h1 per page and then without skipping heading go down levels. Not always easy in a React app, but here I guess we could handle it.

> Note: Only use one `<h1>` per page - this should represent the main heading/subject for the whole page. Also, do not skip heading levels - start with `<h1>`, then use `<h2>`, and so on.
See: https://www.w3schools.com/tags/tag_hn.asp

xrutayisire (changes requested):
This feature will help a lot of users 🤩
Don't hesitate to call someone from the team when you're stuck on sprinkles issues 🤝

Also, If you think some comments should be added in the code guidelines don't hesitate, I added the one for the tests 😊

packages/slice-machine/components/AppLayout/Navigation/Menu/Logo.tsx
MarcMcIntosh: Which login proess?
apiEndpoint comes from the slicemachine config file.

packages/slice-machine/components/AppLayout/Navigation/Menu/Logo.tsx
MarcMcIntosh: exec can return null, I can do a bit of refactoring and remove it

packages/slice-machine/components/AppLayout/Navigation/Menu/Logo.tsx
MarcMcIntosh: Doesn't support inline styles :/

MarcMcIntosh: tried using Text component but it doesn't support data attributes which has caused some issues with `BaseStyles` removing or reducing the foot print of `BaseStyles` has been made in to a tech-debt ticket/spike so it's impact can be messured

MarcMcIntosh: managed to use a class name but I had add a data-attribute, which isn't supported by <Text />

packages/slice-machine/components/AppLayout/Navigation/Menu/Logo.tsx
MarcMcIntosh: done

bapmrl:
Closing this as @MarcMcIntosh will reimplement the whole menu instead 🙂

PR #1000, opened 2023-06-05
test(page types): Add integration tests for page types

xrutayisire:
## Context

- Completes DT-1375

## The Solution

- Create a first set of integration tests, with each "test()" not relying on implementation details.
- Create helpers to centralize implementation details.
- Fix an error in console when deleting a custom type, ul should not be children of p

- [x] The CI is successful.

### DIFF
--- a/packages/slice-machine/src/features/customTypes/__tests__/CustomTypesTablePage.test.tsx
+++ b/packages/slice-machine/src/features/customTypes/__tests__/CustomTypesTablePage.test.tsx
@@ -1,0 +1,336 @@
+// @vitest-environment jsdom
+import { describe, vi, test, TestContext } from "vitest";
+import mockRouter from "next-router-mock";
+import { CustomType } from "@prismicio/types-internal/lib/customtypes";
+
+import {
+  CustomTypeFormat,
+  createSliceMachineManager,
+} from "@slicemachine/manager";
+import { createSliceMachineManagerMSWHandler } from "@slicemachine/manager/test";
+import { CustomTypeReadHookData } from "@slicemachine/plugin-kit";
+import { CustomTypes } from "@lib/models/common/CustomType";
+import { render, screen, waitFor, within } from "test/__testutils__";
+import { createTestPlugin } from "test/__testutils__/createTestPlugin";
+import { createTestProject } from "test/__testutils__/createTestProject";
+import { CustomTypesTablePage } from "../customTypesTable/CustomTypesTablePage";
+import { CUSTOM_TYPES_MESSAGES } from "../customTypesMessages";
+import pkg from "../../../../package.json";
+
+const formats = [
+  {
+    format: "custom",
+  },
+  {
+    format: "page",
+  },
+];
+
+describe.each(formats)(
+  "CustomTypesTablePage > All formats > $format type",
+  (args) => {
+    const format = args.format as CustomTypeFormat;
+
+    test(`should display the ${format} types table`, async (ctx) => {
+      await renderCustomTypesTablePage({ format, ctx });
+
+      // Check each custom type for the current format
+      for (const customTypeMock of customTypesMockByFormat[format]) {
+        // Scope the check for each row
+        const row = screen
+          .getByText(customTypeMock.id)
+          .closest("tr") as HTMLElement;
+        expect(within(row).getByText(customTypeMock.id)).toBeVisible();
+        expect(
+          within(row).getByText(customTypeMock.label as string)
+        ).toBeVisible();
+        expect(
+          within(row).getByText(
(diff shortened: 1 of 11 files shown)


packages/slice-machine/test/__testutils__/index.tsx
```diff
@@ -37,7 +44,10 @@ function render(
       </ThemeUIThemeProvider>
     );
   }
-  return rtlRender(ui, { wrapper: Wrapper, ...renderOptions });
+  return {
```
MarcMcIntosh, line 51: can we add a return type to this function?

packages/slice-machine/test/__testutils__/index.tsx
```diff
@@ -23,6 +24,12 @@ function render(
     ...renderOptions
   }: RenderArgs = {}
 ) {
+  if (!document.getElementById("__next")) {
```
MarcMcIntosh, line 31: guess that makes sense, as it should only effect some next dependent parts of the code base

packages/slice-machine/src/features/customTypes/__tests__/CustomTypesTablePage.test.tsx
```diff
@@ -0,0 +1,336 @@
...
+    cwd,
+  });
+
+  await manager.telemetry.initTelemetry({
+    appName: pkg.name,
+    appVersion: pkg.version,
+  });
+  await manager.plugins.initPlugins();
+
+  ctx.msw.use(
```
MarcMcIntosh, line 297: hmm, trying to figure out if there's a way to spy on these requests,
... maybe on the mock adapter

packages/slice-machine/src/features/customTypes/customTypesTable/convertCustomToPageType.ts
```diff
@@ -0,0 +1,41 @@
+import { toast } from "react-toastify";
+
+import { CustomType } from "@prismicio/types-internal/lib/customtypes";
+import { CustomTypeFormat } from "@slicemachine/manager";
+import { managerClient } from "@src/managerClient";
+import { CUSTOM_TYPES_MESSAGES } from "../customTypesMessages";
+
+export async function convertCustomToPageType(
```
MarcMcIntosh: This kind of looks like the sort of logic that would go into a saga ?
if so, I would recommend using `expectSaga` to test it, that way you can assert side-effects

...it kind of already uses one 🤔

MarcMcIntosh (commented):
Good effort,
I have a few questions and observations that I left some comments where appropriate, 
an other one would be that there's few assertions going on, which may or may not be easier to implement

packages/slice-machine/src/features/customTypes/__tests__/CustomTypesTablePage.test.tsx
xrutayisire: Don't get this comment, here it's kind of a way to spy on them, no? 
We are mocking what's happening normally here

packages/slice-machine/src/features/customTypes/customTypesTable/convertCustomToPageType.ts
xrutayisire: We don't want to directly test this function, that's the goal of the kind of test I'm bringing.
You don't test implementation details. This function is by design an implementation details. If you want to change this function into 2 or 3, users will not see any difference and so your test should not break.

packages/slice-machine/test/__testutils__/index.tsx
xrutayisire: Done 👍

packages/slice-machine/src/features/customTypes/__tests__/CustomTypesTablePage.test.tsx
MarcMcIntosh: msw is network interception, what I'm trying to figure out is what sort of side effects would we be able to assert actually happened. I'll probably need to take some time with code base to ind out what a viable approach would be.

packages/slice-machine/src/features/customTypes/customTypesTable/convertCustomToPageType.ts
MarcMcIntosh: ok

MarcMcIntosh (approved):

packages/slice-machine/src/features/customTypes/__tests__/CustomTypesTablePage.test.tsx
xrutayisire: Oh I see, yes In the future we could try to see that but for my use case I reuse Angelo work for that and it's working great :)

PR #1003, opened 2023-06-07
chore: stable deps install

bapmrl:
Completes DT-1391, DT-1356, DT-1359 and DT-1360.


packages/adapter-next/vite.config.ts
```diff
@@ -28,12 +28,14 @@ export default defineConfig({
 	test: {
 		testTimeout: 10000,
 		coverage: {
-			provider: "c8",
+			provider: "v8",
```
xrutayisire, line 31: Looks like it the default now, no?

packages/slice-machine/README.md
```diff
@@ -50,34 +50,6 @@ Whether you're helping us fix bugs, improve the docs, or spread the word, we'd l
 
 **Submitting code changes**: For small fixes, feel free to [open a PR][repo-pull-requests] with a description of your changes. For large changes, please first [open an issue][repo-feature-request] so we can discuss if and how the changes should be implemented.
 
-### Development scripts
```
xrutayisire, line 53: Thank you! Good removal here 💪

packages/adapter-next/vite.config.ts
bapmrl: Indeed, but `c8` was also the default provider used by the previously installed version of Vitest. I don't know if specifying it was intentional or not.

package.json
```diff
@@ -1,67 +1,77 @@
...
-    "::changeset": "npm run changeset add && git add .changeset && git commit -m '[release:changelog] update changeset'; echo 'We commited your changeset. You should defo push this master ✌️'",
-    "::release": "manypkg check && npm run build && lerna publish && npm run run changeset",
+    "test": "yarn workspaces foreach --parallel --topological-dev --verbose run test",
+    "test:e2e": "yarn cypress-setup && start-server-and-test 'npm run dev --prefix e2e-projects/cypress-next-app' http://localhost:3000 'SM_ENV=development ENABLE_SENTRY=false npm run slicemachine --prefix e2e-projects/cypress-next-app' http://localhost:9999 'cypress run'",
+    "test:e2e:dev": "yarn cypress-setup && start-server-and-test 'npm run dev --prefix e2e-projects/cypress-next-app' http://localhost:3000 'SM_ENV=development ENABLE_SENTRY=false npm run slicemachine --prefix e2e-projects/cypress-next-app' http://localhost:9999 'cypress open'",
```
xrutayisire: @mdeclercq remember we had to remove this `SM_ENV=development` to make it work. 
But I discovered that without this Segment is sent to [Prod] SliceMachine and it's not good right?

.github/actions/restore-cache/action.yml
```diff
@@ -11,7 +11,7 @@ runs:
           node_modules
           */*/node_modules
           ~/.cache/Cypress
-        key: ${{ runner.os }}-${{ hashFiles('package-lock.json', '**/package-lock.json') }}-dependencies
+        key: ${{ runner.os }}-${{ hashFiles('**/package-lock.json', '**/yarn.lock') }}-dependencies
```
MarcMcIntosh, line 14: Does yarn create lock files in each directory ?
Also with Plug and play doesn't caching become pointless?

.github/workflows/ci.yml
```diff
@@ -211,6 +210,5 @@ jobs:
           git config user.name "${{ github.actor }}"
           git config user.email "${{ github.actor }}@users.noreply.github.com"
           git checkout .
-          npm run bump:alpha --preid="${{ github.ref_name }}"
-          npm config set '//registry.npmjs.org/:_authToken' "${NPM_TOKEN}"
-          npm run publish:alpha
+          yarn bump:alpha --preid="${{ github.ref_name }}"
```
MarcMcIntosh, line 213: how will yarn publish with pout an access token?

package.json
```diff
@@ -1,67 +1,77 @@
...
     "bump:interactive": "lerna version prerelease --preid alpha --no-push --exact",
     "bump:alpha": "lerna version prerelease --preid $npm_config_preid --no-changelog --exact --yes",
     "bump:production": "lerna version --conventional-graduate --exact",
-    "publish:alpha": "lerna publish from-package --dist-tag alpha --no-verify-access --yes",
-    "publish:production": "PUBLIC_SM_UI_SEGMENT_KEY=cGjidifKefYb6EPaGaqpt8rQXkv5TD6P PUBLIC_SM_INIT_SEGMENT_KEY=ED3O1FVsa8DteNyRmAuDyMcPoftJ6VdP lerna publish from-package",
-    "lerna:update": "npx --yes lerna-update-wizard"
```
MarcMcIntosh, line 35: lena-update-wizard is pretty handy when updating packages across workspaces ;)

package.json
```diff
@@ -1,67 +1,77 @@
...
+    "cypress-localstorage-commands": "2.2.2",
+    "cypress-real-events": "1.7.6",
+    "cypress-wait-until": "1.7.2",
+    "husky": "8.0.3",
+    "lerna": "7.0.2",
+    "lint-staged": "13.2.0",
+    "prettier": "2.8.4",
+    "start-server-and-test": "1.15.5"
   },
   "lint-staged": {
```
MarcMcIntosh, line 63: why were the other packages removed from `lint-stage` ?

packages/adapter-next/vite.config.ts
MarcMcIntosh: what's the difference between the 2

packages/init/test/SliceMachineInitProcess-loginAndFetchUserData.test.ts
```diff
@@ -50,8 +50,7 @@ const loginAndFetchUserDataWithStdin = async (
 	await new Promise((res) => setTimeout(res, 50));
 
 	const port: number =
-		// @ts-expect-error - Mocked instance
-		spiedManager.user.getLoginSessionInfo.results[0][1].port;
+		spiedManager.user.getLoginSessionInfo.mock.results[0].value.port;
```
MarcMcIntosh, line 54: Is this one oof the small changes between jest and vitest ? `mock.calls` vs `mock.results`?

packages/manager/src/lib/installDependencies.ts
```diff
@@ -8,12 +8,12 @@ import {
 import { PackageManager } from "../types";
 
 const EXTRA_INSTALL_FLAGS: Record<PackageManager, string[]> = {
-	npm: ["--color=always", "--loglevel=info"],
-	pnpm: ["--color"], // TODO: Validate that flags are correct
-	yarn: ["--color"],
-	"yarn@berry": ["--color"], // TODO: Validate that flags are correct
-	"pnpm@6": ["--color"], // TODO: Validate that flags are correct
-	bun: ["--color"], // TODO: Validate that flags are correct
+	npm: ["--color=always", "--loglevel=info"], // Default: `--color=true --loglevel=notice`.
```
MarcMcIntosh, line 11: no color?

packages/slice-machine/package.json
```diff
@@ -8,10 +8,6 @@
     "url": "ssh://git@github.com/prismicio/slice-machine.git",
     "directory": "packages/slice-machine"
   },
-  "_moduleAliases": {
```
MarcMcIntosh, line 11: yeet

packages/slice-machine/package.json
```diff
@@ -20,125 +16,110 @@
...
-    "puppeteer": "^19.11.0",
-    "react": "^18.2.0",
-    "semver": "^7.3.8",
-    "start-slicemachine": "0.9.1-dev-next-release.5",
-    "yup": "^0.32.11"
+  "bin": {
+    "start-slicemachine": "./bin/start-slicemachine.cjs"
   },
-  "devDependencies": {
+  "dependencies": {
```
MarcMcIntosh, line 33: looks like a lot of the bundled deps will need to be re downloaded now?

MarcMcIntosh (commented):
Made a few comments and questions.

packages/adapter-next/vite.config.ts
bapmrl: https://github.com/vitest-dev/vitest/pull/3339 🙂

.github/actions/restore-cache/action.yml
bapmrl: >Does yarn create lock files in each directory ?

No, but I've left it because we might use Yarn in the `e2e-projects/cypress-next-app` project in the future and it could require adding an empty `yarn.lock` file to have it behave as a standalone project (not included as a workspace).

>Also with Plug and play doesn't caching become pointless?

Indeed, but I didn't enable it because it would require a bit of configuration for Plug'n'Play to work.

.github/workflows/ci.yml
bapmrl: It's a very good question.

The `actions/setup-node@v3` action automatically creates a `.npmrc` file that references the `NODE_AUTH_TOKEN` environment variable. Refer to https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry.

package.json
bapmrl: Let's not use Lerna for package management because of this: https://lerna.js.org/docs/legacy-package-management. Yarn has the [up](https://yarnpkg.com/cli/up) and [upgrade-interactive](https://yarnpkg.com/cli/upgrade-interactive) commands.

package.json
bapmrl: I don't understand your question Marc. I don't believe I've changed anything regarding `lint-staged`.

packages/init/test/SliceMachineInitProcess-loginAndFetchUserData.test.ts
bapmrl: I don't know ^^ Lucie helped me on this one.

packages/manager/src/lib/installDependencies.ts
bapmrl: If you're referring to `npm` and `pnpm`, they'll have colors applied. I've removed the arguments for other package managers because they don't necessarily exist and were preventing Yarn and pnpm users from initialising a Slice Machine project. Lucie has reviewed this change.

packages/slice-machine/package.json
bapmrl: What do you mean?

packages/slice-machine/package.json
bapmrl: This is an optimisation that wasn't applied consistently (some packages were supposed to be in `devDependencies`, but were in `dependencies` instead). Also, not all dependencies were listed and it prevented Yarn from installing `slice-machine-ui`.

I've removed this optimisation so that the decision is easier for us: either a dependency is only needed for our development purposes, or a dependency is needed at runtime.

We could consider putting it back at a later point, but the right strategy doesn't seem entirely clear when I asked around. I'll create a technical debt issue in Linear.

MarcMcIntosh: Okay, last time i checked we put bundled deps in the dev deps secton (things like react, and react-dom, next does this by default) :)

packages/slice-machine/package.json
MarcMcIntosh: moduleAliases has caused a lot of headaches, so it's nice to see it being phased out

package.json
MarcMcIntosh: Yes, but there used to be more packages there.... maybe it took to long 🤷‍♂️

package.json
MarcMcIntosh: sad

.github/workflows/ci.yml
MarcMcIntosh: Didn't always work though... that's why the config was added that way :), so if you see some weird 40x errors when publishing it'll be that :)

.github/actions/restore-cache/action.yml
MarcMcIntosh: noted

.github/workflows/ci.yml
bapmrl: Ok, thanks ^^ We'll try and if required, we'll fix it. I'll have that in mind.

packages/slice-machine/package.json
bapmrl: I've created this Linear issue to track this topic: https://linear.app/prismic/issue/DT-1409/consider-moving-bundled-dependencies-into.

PR #1005, opened 2023-06-12
Mm/dt 1333

MarcMcIntosh:
## Context

Editor config for unsplash and oembed does not work with a trailing slash

- [ ] The CI is successful.


packages/manager/src/lib/addTrailingSlash.ts
```diff
@@ -3,3 +3,7 @@ export const addTrailingSlash = (
 ): string | undefined => {
 	return url?.replace(/\/?$/, "/");
 };
+
+export const removeTrailingSlash = (url: string): string => {
```
bapmrl, line 7: ⚠️ The filename is misleading. We could rename it to `trailingSlash.ts`.

packages/manager/test/lib-addTrailingSlash.test.ts
```diff
@@ -1,12 +1,30 @@
-import { expect, it } from "vitest";
+import { expect, it, describe } from "vitest";
```
bapmrl: ⚠️ The filename is misleading. We could rename it to `lib-trailingSlash.test.ts`.

packages/manager/test/lib-addTrailingSlash.test.ts
```diff
@@ -1,12 +1,30 @@
...
+		expect(addTrailingSlash("https://prismic.io")).toBe("https://prismic.io/");
+		expect(addTrailingSlash("https://prismic.io/")).toBe("https://prismic.io/");
+	});
 });
 
-it("adds a single trailing slash to endpoints", async () => {
-	expect(addTrailingSlash("https://prismic.io")).toBe("https://prismic.io/");
-	expect(addTrailingSlash("https://prismic.io/")).toBe("https://prismic.io/");
+describe("removeTrailingSlash", () => {
+	it("will removed a trailing slash from a string", () => {
```
bapmrl: 💡 There's a typo (extra "d" to "removed"). We could change the message to "removes a single trailing slash to endpoints" for consistency or "will remove a trailing slash from a string" for your suggestion.

packages/manager/test/lib-addTrailingSlash.test.ts
```diff
@@ -1,12 +1,30 @@
...
+		expect(removeTrailingSlash("https://prismic.io/")).toBe(
+			"https://prismic.io",
+		);
+		expect(removeTrailingSlash("https://prismic.io")).toBe(
+			"https://prismic.io",
+		);
+
+		expect(removeTrailingSlash("foo/bar")).toBe("foo/bar");
```
bapmrl: ❓ What's the purpose of this expectation? Isn't it the same thing as `expect(removeTrailingSlash("https://prismic.io")).toBe("https://prismic.io")`?

bapmrl (changes requested):
⚠️ Why are we changing the `package-lock.json` file if this PR doesn't include changes to the `package.json` dependencies?

packages/manager/test/lib-addTrailingSlash.test.ts
MarcMcIntosh: no trailing slash... double checking i guess

bapmrl (approved):

PR #1008, opened 2023-06-13
changes up to date blank slate

hypervillain:
Update the Changes page empty state, to use new BlankSlate component.
I also updated the default SEO Tab of `createCustomType` to facilitate using it as website metadata


packages/slice-machine/components/ChangesEmptyPage/NoChangesPage.tsx
xrutayisire, line 1: Do you think we could start to move this component to the "src" folder that should be the new folder with clean new code?

packages/slice-machine/components/ChangesEmptyPage/NoChangesPage.tsx
```diff
@@ -1,33 +1,56 @@
...
+import { Button, Image, tokens } from "@prismicio/editor-ui";
+
+import { FiExternalLink } from "react-icons/fi";
+import { TextLink } from "@src/components/TextLink";
+import { useSelector } from "react-redux";
+import { SliceMachineStoreType } from "@src/redux/type";
+import { getDocumentsListEndpoint } from "@src/modules/environment";
 
 export const NoChangesPage = () => {
+  const { documentsListEndpoint } = useSelector(
```
xrutayisire: I think we need to continue our discussion for the "redux removal" task. Here it's so easy to use a new selector compare to thinking of a new way. 
But I'm afraid that the goal to remove redux is not a good one if we don't try, step by step, to do it.

At the end for your use case it's super simple what you did

packages/slice-machine/components/ChangesEmptyPage/NoChangesPage.tsx
```diff
@@ -1,33 +1,56 @@
...
-        <Text sx={{ fontSize: "13px", lineHeight: "24px" }}>
-          Your local project is in sync with your remote Prismic repository.
-        </Text>
-      </Flex>
-    </Flex>
+    <BlankSlate
+      style={{ width: "100%", alignSelf: "center", marginTop: tokens.size[72] }}
+    >
+      <BlankSlateImage>
+        <Image src={"/blank-slate-changes-uptodate.png"} sizing="cover" />
```
xrutayisire:
```suggestion
        <Image src="/blank-slate-changes-uptodate.png" sizing="cover" />
```

packages/slice-machine/components/ChangesEmptyPage/NoChangesPage.tsx
```diff
@@ -1,33 +1,56 @@
...
+      </BlankSlateImage>
+      <BlankSlateContent>
+        <BlankSlateTitle>Everything up-to-date</BlankSlateTitle>
+        <BlankSlateDescription>
+          You have no changes staged. Your changes appear here after you have
+          saved them, while they're waiting to be pushed to the Page Builder.
+          Ready to get going?
+        </BlankSlateDescription>
+        <BlankSlateActions>
+          <Button size="medium" onClick={() => void router.push("/")}>
```
xrutayisire: No end icon with the "+" button as the Figma?

packages/slice-machine/src/components/TextLink/TextLink.css.ts
```diff
@@ -0,0 +1,108 @@
...
+import { colors, selectors, sprinkles } from "@prismicio/editor-ui";
+
+export const root = style([
+  sprinkles({
+    all: "unset",
+    cursor: "pointer",
+    display: "flex",
+    alignItems: "center",
+    color: {
+      ...colors.indigo11,
```
xrutayisire: Should be purple11 here, no?

packages/slice-machine/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+type IconProps = Readonly<{
+  className?: string;
+}>;
+
+type Variant = keyof typeof textStyles.variant;
+
+export interface TextLinkProps {
+  children: string;
+  color?: keyof typeof styles.colorVariant;
```
xrutayisire, line 15: We are adding the color but here only support primary, right?

xrutayisire (changes requested):
Like this change 😊

⚠️ Don't forget, target branch should be: dev-next-release

packages/slice-machine/components/ChangesEmptyPage/NoChangesPage.tsx
hypervillain: In this case it's a bit hard if there is not a top-level query of environment values. I think moving selectors would be very easy once it's done. WDYT?

packages/slice-machine/components/ChangesEmptyPage/NoChangesPage.tsx
hypervillain: Nope!

packages/slice-machine/src/components/TextLink/TextLink.css.ts
hypervillain: Oh it should not be specified here actually. Good catch!

packages/slice-machine/src/components/TextLink/TextLink.tsx
hypervillain: Yes, it's just a way to extend it. Editor uses indigo instead of purple. Indigo would have a "secondary" key, but we need to discuss it with design team

packages/slice-machine/components/ChangesEmptyPage/NoChangesPage.tsx
xrutayisire: Yes, I agree that here we don't have top level env values. Maybe that's a case where we would need Context? :D

packages/slice-machine/src/components/ChangesEmptyPage/AuthErrorPage.tsx
xrutayisire, line 1: Sorry if my comments was not clear, the goal we discussed with Baptiste would be to move the files that are new / updated / clean to be in src. Meaning this one for example is a file with old theme ect so we would like to update it to the new style. It will help us to track remaining work, what do you think?

hypervillain: You would like to move only the component that I touched and leave the rest out of src?

xrutayisire: If I remember correctly, yes the goal was to only have clean code inside new src/components folder, so we know the gap to remove theme ui ect..

packages/slice-machine/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { Text } from "@prismicio/editor-ui";
+
+import * as styles from "./TextLink.css";
+import * as textStyles from "@prismicio/editor-ui/dist/components/Text/Text.css";
+
+type IconProps = Readonly<{
+  className?: string;
+}>;
+
+type Variant = keyof typeof textStyles.variant;
```
bapmrl: 💡 `Pick<keyof typeof textStyles.variant, "h1">`

packages/slice-machine/src/components/TextLink/TextLink.css.ts
```diff
@@ -0,0 +1,103 @@
...
+      color: {
+        ...colors.purple9,
+        ...selectors.hover(colors.purple8),
+        ...selectors.active(colors.purple9),
+      },
+    }),
+  ],
+});
+
+export const text = style([
```
bapmrl: 💡 You can write `sprinkles({ alignItems: "center", display: "flex" })`.

xrutayisire (approved):
LGTM, only @bapmrl comments left

packages/slice-machine/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,55 @@
+import type { FC, ReactNode } from "react";
+import { isValidElement, cloneElement } from "react";
+
+import clsx from "clsx";
+import { Text } from "@prismicio/editor-ui";
+
+import * as styles from "./TextLink.css";
+import * as textStyles from "@prismicio/editor-ui/dist/components/Text/Text.css";
```
bapmrl: ⚠️ `@prismicio/editor-ui/dist` shouldn't be used as it's not part of the package's public API (defined in the [exports field](https://github.com/prismicio/editor/blob/f56586f63162d57f35f9178a1779bbc81e1e51d8/packages/editor-ui/package.json#L64-L72)). The most simple solution is to simply change `textVariant?: keyof Pick ...` to `textVariant?: "normal" | "smallBold" | "inherit"`.

packages/slice-machine/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+type IconProps = Readonly<{
+  className?: string;
+}>;
+
+export interface TextLinkProps {
```
bapmrl: 💡 There's no need to export something that isn't used elsewhere. It artifically increases the surface area of this module's API.

packages/slice-machine/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export const TextLink: FC<TextLinkProps> = (props) => {
+  const {
+    children,
+    color = "primary",
+    endIcon,
+    href,
+    textVariant = "smallBold",
+  } = props;
```
bapmrl, line 29: 💡 Don't change it, but I would advise spreading `otherProps` to let consumers of UI components add extra `data-*` and `aria-*` props.

packages/slice-machine/src/components/TextLink/TextLink.stories.tsx
```diff
@@ -0,0 +1,27 @@
+import type { Meta, StoryObj } from "@storybook/react";
+
+import { TextLink } from "./TextLink";
+
+type Story = StoryObj<typeof meta>;
+
+const meta = {
+  component: TextLink,
```
bapmrl, line 9: ⚠️ You haven't specified `argTypes`, which means Storybook users (like Benjamin — when we'll continuously deploy Storybook) cannot easily select the values supported by the component. You should add something like this:
```
argTypes: {
    color: {
        control: { type: "select" },
        options: ["primary", "secondary"]
    },
    endIcon: { control: { disable: true } },
    textVariant: {
        control: { type: "select" },
        options: ["normal", "smallBold", "inherit"]
    }
}
```

packages/slice-machine/src/components/TextLink/TextLink.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+} satisfies Meta<typeof TextLink>;
+
+export default meta;
+
+export const Default = {
```
bapmrl, line 25: 💡 You're not showing users can add an icon with the `endIcon` prop.

packages/slice-machine/src/components/TextLink/TextLink.css.ts
```diff
@@ -0,0 +1,70 @@
...
+  }),
+  {
+    ":focus-visible": {
+      outline: "none",
+      textDecorationLine: "underline",
+    },
+  },
+]);
+
+export const colorVariant = styleVariants({
```
bapmrl: 💡 Nice, but can be simplified like this:
```
export const colorVariant = {
    primary: sprinkles({
        color: {
            ...colors.purple9,
            ...selectors.hover(colors.purple8),
            ...selectors.active(colors.purple9),
        },
    }),
    secondary: sprinkles({
        color: {
            ...colors.indigo11,
            ...selectors.hover(colors.indigo10),
            ...selectors.active(colors.indigo11),
        },
    })
}
```

packages/slice-machine/src/components/TextLink/TextLink.css.ts
```diff
@@ -0,0 +1,70 @@
...
+      color: {
+        ...colors.indigo11,
+        ...selectors.hover(colors.indigo10),
+        ...selectors.active(colors.indigo11),
+      },
+    }),
+  ],
+});
+
+export const text = sprinkles({
```
bapmrl: ⚠️ I don't think it's useful and it's weird to change the `display` property of a text element to `flex`.

packages/slice-machine/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
+    <a
+      href={href}
+      target="_blank"
+      className={clsx(styles.root, styles.colorVariant[color])}
+      rel="noreferrer"
+    >
+      <Text color="inherit" variant={textVariant} className={styles.text}>
```
bapmrl: ⚠️ You need to add `component="span"`, otherwise we will end up with a `p` element and that's not what we want in an `a` element: https://github.com/prismicio/editor/blob/main/packages/editor-ui/src/components/Text/Text.ts#L63.

packages/slice-machine/src/components/TextLink/TextLink.css.ts
```diff
@@ -0,0 +1,70 @@
...
+      fontSize: "1.2rem",
+    },
+  ],
+  smallBold: [
+    {
+      fontSize: "1rem",
+    },
+  ],
+  inherit: [
+    sprinkles({ fontFamily: "inherit" }),
```
bapmrl, line 58: ❓ There's no `fontSize: "inherit"` anymore in `sprinkles`? I remember I added it a long time ago.

packages/slice-machine/src/components/TextLink/TextLink.css.ts
```diff
@@ -0,0 +1,70 @@
...
+  ],
+});
+
+export const text = sprinkles({
+  display: "flex",
+  alignItems: "center",
+});
+
+/* TODO move paddingLeft to iconVariant style objects */
+export const endIcon = style([
```
bapmrl: 💡 Can be simplified to `export const endIcon = sprinkles({ paddingLeft: 4 })`.

packages/slice-machine/src/features/changes/BlankSlates/NoChangesPage.tsx
```diff
@@ -0,0 +1,59 @@
+import router from "next/router";
```
bapmrl: 💡 It's probably a bit premature to create a `BlankSlates` subdirectory in `features/changes`.

packages/slice-machine/src/features/changes/BlankSlates/NoChangesPage.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const { repoName } = useSelector((state: SliceMachineStoreType) => ({
+    repoName: getRepoName(state),
+  }));
+
+  const documentsListEndpoint =
+    createDocumentsListEndpointFromRepoName(repoName);
+
+  return (
+    <BlankSlate
+      style={{ width: "100%", alignSelf: "center", marginTop: tokens.size[72] }}
```
bapmrl: 💡 I'm not sure `width: "100%"` is necessary.

packages/slice-machine/src/features/changes/BlankSlates/NoChangesPage.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+import { FiExternalLink } from "react-icons/fi";
+import { TextLink } from "@src/components/TextLink";
+import { useSelector } from "react-redux";
+import { SliceMachineStoreType } from "@src/redux/type";
+import { getRepoName } from "@src/modules/environment";
+
+import { createDocumentsListEndpointFromRepoName } from "@lib/utils/repo";
+
+export const NoChangesPage = () => {
```
bapmrl: ⚠️ That's not a page, but just a "linked"/"connected" `BlankSlate`. Maybe call the file/component `NoChangesBlankSlate`?

packages/slice-machine/src/features/changes/BlankSlates/NoChangesPage.tsx
```diff
@@ -0,0 +1,59 @@
...
+      </BlankSlateImage>
+      <BlankSlateContent>
+        <BlankSlateTitle>Everything up-to-date</BlankSlateTitle>
+        <BlankSlateDescription>
+          You have no changes staged. Your changes appear here after you have
+          saved them, while they're waiting to be pushed to the Page Builder.
+          Ready to get going?
+        </BlankSlateDescription>
+        <BlankSlateActions>
+          <Button size="medium" onClick={() => void router.push("/")}>
```
bapmrl: ❓ Benjamin used `large` in his `BlankSlateActions` component. I'm surprised buttons are not always the same size, but maybe it's what he wants.

packages/slice-machine/src/features/changes/BlankSlates/NoChangesPage.tsx
```diff
@@ -0,0 +1,59 @@
...
+            Create a page type
+          </Button>
+          <TextLink
+            endIcon={<FiExternalLink />}
+            textVariant="normal"
```
bapmrl: 💡 Maybe it means `textVariant`'s default value could be `normal`?

bapmrl (changes requested):
I'm sorry that I'm adding 15 comments, but most of them are really small and I've tried to propose the updated code when I could. I've also put priorities using ⚠️, 💡 and ❓.

packages/slice-machine/src/features/changes/BlankSlates/NoChangesPage.tsx
hypervillain: It depends if we want to comply with the editor-ui default, which is smallBold. It would make merging them easier but is it the plan?

bapmrl: I see 🙂 Let's keep it that way.

packages/slice-machine/src/features/changes/BlankSlates/NoChangesPage.tsx
hypervillain: The "real" design includes 3 of them, not are not doable easily yet but will!

packages/slice-machine/src/features/changes/BlankSlates/NoChangesPage.tsx
hypervillain: It's medium everywhere now

bapmrl: Then, you need to change the `ButtonGroup`'s `size="large"` to `size="medium"` in `BlankSlateActions` 🙂

hypervillain: This requires to update multiple components that are not part of the review though

packages/slice-machine/src/components/TextLink/TextLink.css.ts
hypervillain: Done but can you explain me when `styleVariants` should be used? When you also have non-sprinkles values?

packages/slice-machine/src/components/TextLink/TextLink.css.ts
hypervillain: It seems not :/

packages/slice-machine/src/components/TextLink/TextLink.css.ts
bapmrl: Exactly :) When you need some properties not assignable through `sprinkles`.

hypervillain: I see! It was always unclear to me, thanks

bapmrl (approved):
Perfect 😊

PR #1012, opened 2023-06-15
Page type snippet

hypervillain:
https://linear.app/prismic/issue/DT-1403/aauser-i-can-see-the-header-of-the-custom-type-and-page-type-builder
https://linear.app/prismic/issue/DT-1404/aauser-in-the-header-of-a-page-type-i-can-click-on-a-code-button-that

Implemented a `MainContainer` and its Header, used in both Custom type and Page type builders. The Page type builder now contains a "Page snippet" button. It opens a Dialog that displays placeholder text.

Implemented a `PageTypeSnippet` component, it requests fake content via `useRequest` and renders it inside a Dialog.


packages/slice-machine/src/components/MainContainer/MainContainer.stories.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+import { MainContainer, MainContainerHeader } from "./";
+import { Button } from "@prismicio/editor-ui";
+
+type Story = StoryObj<typeof meta>;
+
+const meta = {
+  component: MainContainer,
+  parameters: {
+    docs: { story: { height: 256 * 2, inline: false } },
```
bapmrl: 💡 Try to remove it?

packages/slice-machine/src/components/MainContainer/Header/Header.tsx
```diff
@@ -0,0 +1,29 @@
+import { ReactElement } from "react";
+
+import { Button, ButtonGroup, Icon } from "@prismicio/editor-ui";
+
+import { Breadcrumb } from "@src/components/Breadcrumb";
+
+import * as styles from "./Header.css";
+
+type HeaderProps = {
+  Actions?: ReactElement[];
```
bapmrl: 💡 Personally I always use `ReactNode`, but don't know if it's better.

packages/slice-machine/src/components/MainContainer/Header/Header.css.ts
```diff
@@ -0,0 +1,24 @@
...
+    alignItems: "center",
+    marginBottom: 16,
+    height: 32,
+  }),
+  {
+    gridArea: "header",
+  },
+]);
+
+export const flex = style([
```
bapmrl: 💡 `spaceBetween`?

bapmrl:
Completes DT-1403 and DT-1404.

packages/slice-machine/components/AppLayout/index.tsx
```diff
@@ -34,6 +34,9 @@ const AppLayout: FC<Props> = ({ children }) => {
           px: [2, 4, null],
           overflow: "auto",
           display: "flex",
+          "& > :first-child": {
```
bapmrl, line 34: ❓ Have we checked it works on all pages (and there's only ever one child)?

packages/slice-machine/lib/builders/CustomTypeBuilder/index.tsx
```diff
@@ -1,16 +1,11 @@
 import { Box } from "theme-ui";
 
 import TabZone from "./TabZone";
-import { Header, Tabs } from "./Layout";
-
-import Container from "@components/Container";
```
bapmrl, line 6: ⚠️ The `@components/Container` is no longer used and should be deleted to avoid accumulating dead code.

packages/slice-machine/lib/builders/CustomTypeBuilder/index.tsx
```diff
@@ -1,16 +1,11 @@
...
+import { Tabs } from "./Layout";
 
 const CustomTypeBuilder = () => {
   return (
     <Box sx={{ flex: 1 }}>
```
bapmrl: ❓ Is this extra `Box` necessary?

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypesBuilderPage.tsx
```diff
@@ -72,11 +87,56 @@ type CustomTypesBuilderPageWithProviderProps = {
...
+  const router = useRouter();
+
+  const { initCustomTypeStore, saveCustomType } = useSliceMachineActions();
 
   useEffect(() => {
     initCustomTypeStore(customType, remoteCustomType);
   }, []);
 
-  return <CustomTypeBuilder />;
+  const { currentCustomType, hasPendingModifications, isSavingCustomType } =
```
bapmrl, line 98: ❓ What is the difference between `customType` and `currentCustomType`? If they are the same, we could probably avoid the extra `if (currentCustomType === null)` condition.

packages/slice-machine/src/components/MainContainer/Header/Header.tsx
```diff
@@ -0,0 +1,29 @@
+import { ReactElement } from "react";
+
+import { Button, ButtonGroup, Icon } from "@prismicio/editor-ui";
+
+import { Breadcrumb } from "@src/components/Breadcrumb";
+
+import * as styles from "./Header.css";
+
+type HeaderProps = {
+  actions?: ReactElement[];
```
bapmrl: 💡 It would probably be more "natural" to be able to pass a `ReactNode` instead. That way, we could pass just one element or a React `Fragment`.

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypesBuilderPage.tsx
```diff
@@ -72,11 +87,56 @@ type CustomTypesBuilderPageWithProviderProps = {
...
+      ? [<PageTypeSnippet key="trigger-snippet-view" />]
+      : []),
+    <Button
+      key="save-to-fs"
+      data-cy="builder-save-button"
+      onClick={saveCustomType}
+      loading={isSavingCustomType}
+      disabled={!hasPendingModifications || isSavingCustomType}
+    >
+      save
```
bapmrl: ❓ Shouldn't it be `Save`?

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypesBuilderPage.tsx
```diff
@@ -72,11 +87,56 @@ type CustomTypesBuilderPageWithProviderProps = {
...
+    </Button>,
+  ];
+
+  return (
+    <MainContainer>
+      <MainContainerHeader
+        backTo={() => void router.push(config.tablePagePathname)}
```
bapmrl, line 130: ❓ Would https://nextjs.org/docs/pages/api-reference/functions/use-router#routerback be better suited? Would it remove the need to pass the `backTo` prop altogether?

packages/slice-machine/src/features/customTypes/customTypesBuilder/PageTypeSnippet.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+function wait(milliseconds: number) {
+  return new Promise((resolve) => {
+    setTimeout(resolve, milliseconds);
+  });
+}
+
+async function getPageTypeSnippet(): Promise<string> {
+  await wait(3000);
+  return `Create a new page component in your  pages  directory with the filename  [uid].jsx . (If you already have a file with that name, you can create the file
```
bapmrl, line 23: ❓ Were the double spaces `  ` added on purpose?

packages/slice-machine/src/features/customTypes/customTypesBuilder/PageTypeSnippet.tsx
```diff
@@ -0,0 +1,84 @@
...
+  Dialog,
+  DialogContent,
+  DialogHeader,
+  ErrorBoundary,
+  Icon,
+  ProgressCircle,
+  vars,
+} from "@prismicio/editor-ui";
+
+function wait(milliseconds: number) {
```
bapmrl, line 15: 💡 You don't have to change it, but one little advice: most people would expect the `PageTypeSnippet` component to appear first as it's the main export of this file. Then, functions, constants, etcetera could appear in "order of reading" from top to bottom.

packages/slice-machine/src/features/customTypes/customTypesBuilder/PageTypeSnippet.tsx
```diff
@@ -0,0 +1,84 @@
...
+  });
+}
+
+async function getPageTypeSnippet(): Promise<string> {
+  await wait(3000);
+  return `Create a new page component in your  pages  directory with the filename  [uid].jsx . (If you already have a file with that name, you can create the file 
+        in a child directory like  pages/marketing/[uid].jsx .)`;
+}
+
+const PluginPageTypeSnippet = () => {
```
bapmrl, line 27: ⚠️ This component doesn't bring much value and the style could directly be applied in `PageTypeSnippet`.

packages/slice-machine/src/features/customTypes/customTypesBuilder/PageTypeSnippet.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const content = useRequest(getPageTypeSnippet, []);
```
bapmrl, line 28: ⚠️ It would be better to write a `usePageTypeSnippet` hook like: 
https://github.com/prismicio/slice-machine/blob/5cd4d26b4d37715173202c27ed3bf1dd6ac669ff/packages/slice-machine/src/features/customTypes/customTypesTable/useCustomTypes.ts#L18
Thay way, we could easily extract it in the future, if need be.

packages/slice-machine/src/features/customTypes/customTypesBuilder/PageTypeSnippet.tsx
```diff
@@ -0,0 +1,84 @@
...
+  return <div style={{ padding: vars.size[16] }}>{content}</div>;
+};
+
+export const PageTypeSnippet = () => {
+  return (
+    <Dialog
+      trigger={
+        <Button
+          key="open-snippet"
```
bapmrl: ⚠️ Don't add unnecessary `key`s.

packages/slice-machine/src/features/customTypes/customTypesBuilder/PageTypeSnippet.tsx
```diff
@@ -0,0 +1,84 @@
...
+  return `Create a new page component in your  pages  directory with the filename  [uid].jsx . (If you already have a file with that name, you can create the file 
+        in a child directory like  pages/marketing/[uid].jsx .)`;
+}
+
+const PluginPageTypeSnippet = () => {
+  const content = useRequest(getPageTypeSnippet, []);
+  return <div style={{ padding: vars.size[16] }}>{content}</div>;
+};
+
+export const PageTypeSnippet = () => {
```
bapmrl, line 32: 💡 Maybe `PageTypeSnippetButton` would better show that a button will be displayed when using this component?

packages/slice-machine/src/features/customTypes/customTypesBuilder/PageTypeSnippet.tsx
```diff
@@ -0,0 +1,84 @@
...
+        <Button
+          key="open-snippet"
+          variant="secondary"
+          startIcon={<Icon name="add" />}
+        >
+          Page snippet
+        </Button>
+      }
+    >
+      <DialogHeader icon="add" title="Page Snippet" />
```
bapmrl, line 44: ❓ "Page Snippet" or "Page snippet"?

packages/slice-machine/src/features/customTypes/customTypesBuilder/PageTypeSnippet.tsx
```diff
@@ -0,0 +1,84 @@
...
+          }}
+        >
+          <Suspense
+            fallback={
+              <div
+                style={{
+                  display: "flex",
+                  justifyContent: "center",
+                  alignItems: "center",
+                  marginBottom: vars.size[32],
```
bapmrl, line 69: ❓ What is the purpose of this `marginBottom`?

packages/slice-machine/src/components/MainContainer/Header/Header.tsx
```diff
@@ -0,0 +1,29 @@
...
+  backTo?: () => void;
+  breadcrumb: string;
+};
+
+export const Header = ({ actions, backTo, breadcrumb }: HeaderProps) => (
```
bapmrl, line 15: 💡 There's not a lot of code so `Header` probably doesn't warrant its own directory. Here, it's fine, but doing that too much can increase the likelihood of duplicating code and styles (like @xrutayisire experienced when he worked on the `Table` component).

packages/slice-machine/src/components/MainContainer/MainContainer.tsx
```diff
@@ -0,0 +1,22 @@
+import { clsx } from "clsx";
+import type { FC, PropsWithChildren } from "react";
+
+import * as styles from "./MainContainer.css";
+
+type MainContainerProps = PropsWithChildren<{
+  className?: string;
```
bapmrl: ❓ Why add unused props? We're not implementing a library of reusable components (reusable outside of Slice Machine).

packages/slice-machine/src/components/MainContainer/MainContainer.tsx
```diff
@@ -0,0 +1,22 @@
...
+}>;
+
+export const MainContainerContent: FC<PropsWithChildren> = ({ children }) => (
+  <div className={styles.content}>{children}</div>
+);
+
+export const MainContainer: FC<MainContainerProps> = ({
```
bapmrl: 💡 A shorter version could be:
```
export const MainContainer: FC<MainContainerProps> = (props) => <div {...props} />
```

packages/slice-machine/src/components/MainContainer/MainContainer.tsx
```diff
@@ -0,0 +1,22 @@
+import { clsx } from "clsx";
+import type { FC, PropsWithChildren } from "react";
+
+import * as styles from "./MainContainer.css";
+
+type MainContainerProps = PropsWithChildren<{
+  className?: string;
+}>;
+
+export const MainContainerContent: FC<PropsWithChildren> = ({ children }) => (
```
bapmrl, line 6: 💡 It's not "natural" to read `MainContainerContent` before `MainContainer`.

packages/slice-machine/src/components/MainContainer/Header/Header.css.ts
```diff
@@ -0,0 +1,24 @@
...
+  {
+    gridArea: "header",
+  },
+]);
+
+export const spaceBetweenFlex = style([
+  sprinkles({
+    display: "flex",
+    alignItems: "center",
+    justifyContent: "space-between",
```
bapmrl: ⚠️ I don't think you need `justifyContent: "space-between"` so `spaceBetweenFlex` wouldn't be a good name. Maybe `navigation` would be a good name?

packages/slice-machine/src/components/MainContainer/Header/Header.css.ts
```diff
@@ -0,0 +1,24 @@
...
+    alignItems: "center",
+    marginBottom: 16,
+    height: 32,
+  }),
+  {
+    gridArea: "header",
+  },
+]);
+
+export const spaceBetweenFlex = style([
```
bapmrl: 💡 No need to call `style` here.

packages/slice-machine/src/components/MainContainer/MainContainer.stories.tsx
```diff
@@ -0,0 +1,33 @@
+import type { Meta, StoryObj } from "@storybook/react";
+
+import { MainContainer, MainContainerHeader, MainContainerContent } from "./";
```
bapmrl, line 3: 💡 Using `./` is confusing. I would recommend using the real relative path.

packages/slice-machine/src/components/MainContainer/MainContainer.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export const Default = {
+  args: {
+    children: (
+      <>
+        <MainContainerHeader
+          backTo={() => console.log("back to")}
+          breadcrumb="folder/file"
+          actions={[
+            <Button key="action-1">Action 1</Button>,
```
bapmrl, line 25: 💡 I don't think `key`s are necessary here.

bapmrl (changes requested):
Nice work 😊

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypesBuilderPage.tsx
hypervillain: `currentCustomType` is the CT selected in Redux to be manipulated later in the CT builder. They could be merged once we move away from Redux or by passing down `customType` to children. But the calls to Redux are nested deep down so it's not an easy change

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypesBuilderPage.tsx
hypervillain: you can come from URL history on browser and therefore not have a previous place to go back to

packages/slice-machine/src/features/customTypes/customTypesBuilder/PageTypeSnippet.tsx
hypervillain: It's "fake" code!

packages/slice-machine/src/features/customTypes/customTypesBuilder/PageTypeSnippet.tsx
hypervillain: "Page Snippet" in the design!

packages/slice-machine/src/components/MainContainer/MainContainer.stories.tsx
hypervillain: I think the opposite: everything that you can and are allowed to access from a story is what's exported from index, with its definitive name also

packages/slice-machine/src/features/customTypes/customTypesBuilder/PageTypeSnippet.tsx
hypervillain: I would tend to to the exact opposite actually: caller/parent at the bottom, then everything used by it on top of it, with functions not being components on top of everything

packages/slice-machine/components/AppLayout/index.tsx
hypervillain: It does not seem to break anything. But this can probably be done differently? I could not find how but we could discuss it

bapmrl:
Continued in https://github.com/prismicio/slice-machine/pull/1019.

PR #1014, opened 2023-06-16
ci: fix end-to-end tests execution

mdeclercq:
## Context

- Fix end-to-end tests

image

## The Solution

- Fix a tgz file name that was not found. 
- Do not use create-next-app `src-dir` option. It was incompatible with the generated `slice-simulator.tsx` file

image

## Impact / Dependencies

None


bapmrl (approved):
Thank you 😍

PR #1016, opened 2023-06-20
chore(dev): Fix tracking sent to Prod instead of Dev when working as a contributor

xrutayisire:
## Context

- Completes DT-1407 - Running "slicemachine:dev" to dev send tracker events to [Prod] SliceMachine

## The Solution

We need to always have SM_ENV to development to send the tracking to the correct location. I created a script, since with SM_ENV development you have env variables to declare

## Impact / Dependencies

- No impact

- [x] The CI is successful.

### DIFF
--- a/e2e-projects/next/package.json
+++ b/e2e-projects/next/package.json
@@ -8,7 +8,7 @@
     "start": "next start",
     "lint": "next lint",
     "slicemachine": "start-slicemachine",
-    "slicemachine:dev": "NODE_ENV=development start-slicemachine"
+    "slicemachine:dev": "NODE_ENV=development SM_ENV=staging start-slicemachine"
   },
   "dependencies": {
     "@headlessui/react": "^1.5.0",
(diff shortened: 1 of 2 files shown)


bapmrl (commented):
Let's ask Lucie or Angelo if it would be a good idea to add fallbacks for all environment variables when `SM_ENV=development` in: https://github.com/prismicio/slice-machine/blob/e0b4c1630387312250982dda353baacc822d6311/packages/manager/src/constants/API_ENDPOINTS.ts#L25-L30

It would remove the need to specify in one additional file (the `slicemachine.sh` file you created) all environment variables.

bapmrl:
One additional comment: given the changes you made, the `package-lock.json` file shouldn't have changed.

xrutayisire:
> One additional comment: given the changes you made, the `package-lock.json` file shouldn't have changed.

@bapmrl  We still have a problem then I guess, every time I checkout the dev-next-release branch and npm i, it update the package-lock

xrutayisire:
Lot more simple now, thank you @bapmrl 😅

bapmrl:
Yes 😊

Have you seen this?
https://github.com/prismicio/slice-machine/blob/e0b4c1630387312250982dda353baacc822d6311/package.json#L26-L27

Should it be using `SM_ENV=development`?

xrutayisire:
SM_ENV=development means you want to define all the endpoints manually and it's the reason of the old scripts, here we just want Segment event sent to Dev and not Prod so Staging is made for that.

bapmrl (approved):

PR #1017, opened 2023-06-21
feat(adapter-next): support the App Router and `src` directory

angeloashmore:
## Types of changes

## Description

This PR adds the folloing features to `@slicemachine/adapter-next`:

- [x] Creates files in the `app` directory to support the App Router
- [x] Creates files in the `src` directory if the project uses it.
- [x] Creates `/api/preview` and `/api/exit-preview` endpoints to support Prismic previews.
- [x] Creates an `/api/revalidate` endpoint to support on-demand revalidation when the App Router is used.
- [x] Adds a default `localSliceSimulatorURL` value (`http://localhost:3000/slice-simulator`).
- [x] Adds the recommended `fetchOptions` parameter value to `prismicio.ts`'s `createClient()`.
- [x] Allow configuring the location of the generated TypeScript types file (`prismicio-types.d.ts`)

## Checklist:

- [x] My change requires an update to the official documentation.
- [x] All [TSDoc](https://tsdoc.org) comments are up-to-date and new ones have been added where necessary.


packages/adapter-next/src/hooks/project-init.ts
```diff
@@ -205,15 +284,270 @@ const createSliceSimulatorPage = async ({
...
+	// Add Slice Simulator URL.
+	project.config.localSliceSimulatorURL ||=
+		"http://localhost:3000/slice-simulator";
+
+	// Nest default Slice Library in `src` if used.
+	if (
+		JSON.stringify(project.config.libraries) === JSON.stringify(["./slices"]) &&
+		hasSrcDirectory
+	) {
+		project.config.libraries = ["./src/slices"];
```
bapmrl: ❓ What would happen if the user is running the `init`, but already has slices in the `slices` directory? Would it break? Should files be automatically moved?

packages/adapter-next/src/hooks/project-init.ts
```diff
@@ -40,31 +42,141 @@ const createPrismicIOFile = async ({
...
+		} else {
+			createClientContents = source`
+				/**
+				 * Creates a Prismic client for the project's repository. The client is used to
+				 * query content from the Prismic API.
+				 *
+				 * @param {prismicNext.CreateClientConfig} config - Configuration for the Prismic client.
+				 */
+				export const createClient = (config = {}) => {
+					const client = prismic.createClient(repositoryName, {
```
bapmrl, line 97: 💡 It seems only the comments and the declaration of the `createClient` function change depending on the `isTypeScriptProject` constant. Maybe the implementation of the `createClient` function could be moved into a `const` defined in the `if (hasAppRouter)` condition? Although, maybe it's not a good idea because it could make the code harder to follow.

packages/adapter-next/src/hooks/project-init.ts
```diff
@@ -205,15 +284,270 @@ const createSliceSimulatorPage = async ({
...
+}: SliceMachineContext<PluginOptions>) => {
+	const hasSrcDirectory = await checkHasSrcDirectory({ helpers });
+	const hasAppRouter = await checkHasAppRouter({ helpers });
+	const isTypeScriptProject = await checkIsTypeScriptProject({
+		helpers,
+		options,
+	});
+
+	const extension = isTypeScriptProject ? "ts" : "js";
+	const filePath = helpers.joinPathFromRoot(
```
bapmrl, line 300: 💡 I've seen this pattern used several times. Maybe it would warrant its own helper function?

bapmrl (approved):
Thanks Angelo!

I don't have enough context to comment on or test the feature, but the implementation seems good.

❓ However, should some of these changes be applied to `@slicemachine/adapter-nuxt` and `@slicemachine/adapter-nuxt2` as well? I'm thinking mostly of the changes made in the `modifySliceMachineConfig` function.

Also, note I had to skip reviewing the unit tests in `plugin-project-init.test.ts` because I have to leave and work on something else tomorrow morning.

→ I'm approving. Feel free to merge your PR into `master`. I'll handle conflicts in `dev-next-release` for you (as we're moving to Yarn, moving dependencies and deleting the `package-lock.json` file).

packages/adapter-next/src/hooks/project-init.ts
angeloashmore: Good point! I added a check to ensure the directory doesn't exist or, if it does, is empty: https://github.com/prismicio/slice-machine/pull/1017/commits/bc16ab45a1994defcb13b7c80b67d7d64b7b3b98

angeloashmore: I don't think we should move files if they exist, as that would modify a project's files. `init` should only add files as much as possible.

packages/adapter-next/src/hooks/project-init.ts
angeloashmore: There are two pieces that change, which is why we see four separate code blocks:

1. If the project uses TypeScript, we use TSDoc and inline the parameter's type. Otherwise, we use JSDoc for both.
2. If the project uses the App Router, we add an App Router-specific `fetchOptions` parameter to `@prismicio/client`.

We could potentially use multiple variables to construct the resulting string, but I felt it would be harder to follow. The current method is certainly longer and the differences may not be obvious between snippets, but I think it ultimately is more maintainable.

Happy to reconsider if you have a strong opinion!

packages/adapter-next/src/hooks/project-init.ts
angeloashmore: Done! Good idea. There's now a `getJSFileExtension()` helper, which replaces the more limited `getJSOrTSXFileExtension()` helper.

https://github.com/prismicio/slice-machine/pull/1017/commits/12ac14cf3f95504c15ae85a72bf83348e9bdf5ec

angeloashmore:
> @bapmrl: ❓ However, should some of these changes be applied to @slicemachine/adapter-nuxt and @slicemachine/adapter-nuxt2 as well? I'm thinking mostly of the changes made in the modifySliceMachineConfig function.

Yes, I think we should port the `localSliceSimulatorURL` logic to `adaper-nuxt` and `adapter-nuxt2`. The rest of the addition in this PR are specific to Next.js. We'll implement that in a separate PR if that works for you.

bapmrl:
Thanks Angelo for all your answers and the small changes 😊

>We'll implement that in a separate PR if that works for you.

Of course it works for us.

PR #1019, opened 2023-06-23
Page type snippet markdown

hypervillain:
Follows #1012, adds Markdown renderer and optional "documentation:read" hook to manager. Renders Page type snippets in a Dialog


packages/plugin-kit/src/index.ts
```diff
@@ -185,6 +185,13 @@ export type {
 	ProjectInitHookReturnType,
 } from "./hooks/project-init";
 
+// documentation
+export type {
+	DocumentationReadHook,
+	DocumentationReadHookData,
+	DocumentationReadHookReturnType,
+} from "./hooks/documentation-read";
```
lihbr: I'd add "Documentation" as an export also

package.json
```diff
@@ -42,11 +42,19 @@
...
     "cypress-wait-until": "^1.7.2",
-    "start-server-and-test": "^1.15.5"
+    "hast": "^1.0.0",
+    "hast-util-to-string": "^2.0.0",
+    "react-markdown": "^8.0.7",
+    "remark-directive": "^2.0.1",
+    "remark-gfm": "^3.0.1",
+    "start-server-and-test": "^1.15.5",
+    "unified": "^10.1.2",
+    "unist-util-visit": "^4.1.2"
```
lihbr: Don't install package-dependencies top level! (unless I'm missing the reason)

packages/manager/src/managers/documentation/DocumentationManager.ts
```diff
@@ -0,0 +1,57 @@
...
+import { HookError, DocumentationReadHookData } from "@slicemachine/plugin-kit";
+import { BaseManager } from "../BaseManager";
+import { DecodeError } from "../../lib/DecodeError";
+import { assertPluginsInitialized } from "../../lib/assertPluginsInitialized";
+import { decodeHookResult } from "../../lib/decodeHookResult";
+
+type Documentation = {
+	label?: string;
+	content: string;
+};
```
lihbr: Exactly why I'm suggesting exporting that from the plugin kit so you don't have to recreate it here :)

packages/manager/src/managers/documentation/DocumentationManager.ts
```diff
@@ -0,0 +1,57 @@
...
+
+type DocumentationManagerReadReturnType = {
+	documentation: Documentation | Documentation[] | undefined;
```
lihbr: For convenience, non array should be casted to array, so that return type can simply be `documentation: Documentation[]` (always defined), makes using the hook much more straightforward after that.

See the `SnippetManager` code for example: https://github.com/prismicio/slice-machine/blob/1ffb8cff5f38e82be9f534c9a2a249666af2b7a4/packages/manager/src/managers/snippets/SnippetsManager.ts#L20-L45

packages/manager/src/managers/documentation/DocumentationManager.ts
```diff
@@ -0,0 +1,57 @@
...
+				documentation: data,
+				errors,
+			};
+		} catch (e) {
+			return {
+				documentation: undefined,
+				errors: [],
+			};
+		}
+	}
```
lihbr: Does this ever throw? Not sure it does 🤔

packages/manager/src/managers/documentation/DocumentationManager.ts
```diff
@@ -0,0 +1,57 @@
...
+				args,
+			);
+
+			const { data, errors } = decodeHookResult(
+				documentationUnionCodec,
+				hookResult,
+			);
+
+			return {
+				documentation: data,
```
lihbr: I think data is an array of results, so you'll nee to flatten it like in the `SnippetManager`: https://github.com/prismicio/slice-machine/blob/1ffb8cff5f38e82be9f534c9a2a249666af2b7a4/packages/manager/src/managers/snippets/SnippetsManager.ts#L20-L45

package.json
hypervillain: wrong package.json 😬

packages/manager/src/managers/documentation/DocumentationManager.ts
hypervillain: If the hook is not implemented, right?

bapmrl:
Completes DT-1401, DT-1403, DT-1404 and DT-1405.

packages/plugin-kit/src/hooks/documentation-read.ts
```diff
@@ -0,0 +1,53 @@
...
+	ExtendSliceMachineHook,
+	SliceMachinePluginOptions,
+	SliceMachineHook,
+} from "../types";
+
+/**
+ * An object representing a documentation text.
+ */
+export type Documentation = {
+	label?: string;
```
bapmrl, line 13: ❓ Why is the `label` optional? Does it mean documentation could be used outside of tabs?

packages/plugin-kit/src/hooks/documentation-read.ts
```diff
@@ -0,0 +1,53 @@
...
+	kind: "PageSnippet";
+	data: {
+		model: CustomType;
+	};
+};
+
+/**
+ * Return value for `documentation:read` hook handlers.
+ */
+export type DocumentationReadHookReturnType = Documentation[] | undefined;
```
bapmrl: 💡 @lihbr and @hypervillain, why return `undefined` when there's no documentation? Why not just return an empty array?

packages/adapter-next/src/hooks/documentation-read.ts
```diff
@@ -0,0 +1,47 @@
...
+`;
+
+export const documentationRead: DocumentationReadHook<PluginOptions> = (
+	data,
+) => {
+	if (data.kind === "PageSnippet") {
+		return [
+			{
+				label: "Page router",
+				content: md,
```
bapmrl: 💡 Why not use `stripIndent` (or maybe `source`?) from `common-tags` to inline the `content`? It's a pattern that's been used in multiple other files and it would prevent having to declare as many `const md` as there are elements in the returned arrays.

packages/slice-machine/package.json
```diff
@@ -59,23 +61,26 @@
     "monocle-ts": "^2.3.13",
     "newtype-ts": "^0.3.5",
     "next": "^12.3.4",
-    "puppeteer": "^19.11.1",
```
bapmrl: ⚠️ `puppeteer` shouldn't be removed from the `dependencies`, otherwise the "Take a screenshot" feature won't work in production.

packages/slice-machine/package.json
```diff
@@ -50,6 +51,7 @@
     "fast-deep-equal": "^3.1.3",
     "formik": "^2.2.9",
     "fp-ts": "^2.13.1",
+    "hast-util-to-string": "^2.0.0",
```
bapmrl: ⚠️ This dependency doesn't appear to be used in `slice-machine-ui` and should be removed.

packages/slice-machine/package.json
```diff
@@ -59,23 +61,26 @@
...
     "react": "^18.2.0",
     "react-beautiful-dnd": "^13.1.1",
     "react-datepicker": "^4.10.0",
     "react-dom": "^18.2.0",
     "react-icons": "^4.8.0",
+    "react-markdown": "^8.0.7",
     "react-modal": "^3.16.1",
     "react-redux": "8.0.7",
     "react-select": "^5.7.0",
+    "react-syntax-highlighter": "15.5.0",
```
bapmrl: ⚠️ For the sake of consistency, this should use `^` as it's a dependency.

packages/slice-machine/package.json
```diff
@@ -59,23 +61,26 @@
...
     "react-tabs": "^5.2.0",
     "react-tiny-popover": "^7.2.4",
     "react-toastify": "^9.1.1",
     "react-tooltip": "^4.5.1",
     "redux": "^4.2.1",
     "redux-persist": "^6.0.0",
     "redux-saga": "^1.2.2",
+    "remark-directive": "^2.0.1",
```
bapmrl: ⚠️ This dependency doesn't appear to be used in `slice-machine-ui` and should be removed.

packages/manager/src/managers/documentation/DocumentationManager.ts
bapmrl: ⚠️ I don't think it throws (look at `CustomTypesManager` for instance). You can remove the `try`/`catch`.

packages/manager/test/SliceMachineManager-documentation-read.test.ts
```diff
@@ -0,0 +1,65 @@
+import { expect, it, vi } from "vitest";
+
+import { createTestPlugin } from "./__testutils__/createTestPlugin";
+import { createTestProject } from "./__testutils__/createTestProject";
+
+import { createSliceMachineManager } from "../src";
+
+const mockedDocumentation = { label: "My label", content: "" };
+
+it("calls plugins' `slice:create` hook", async (ctx) => {
```
bapmrl: ⚠️ `documentation:read`.

bapmrl (changes requested):

packages/plugin-kit/src/hooks/documentation-read.ts
hypervillain: Yes, if there is only one documentation page, no tabs are displayed, and the label is therefore not used

packages/plugin-kit/src/hooks/documentation-read.ts
hypervillain: Good catch, it's actually empty array

packages/adapter-next/src/hooks/documentation-read.ts
hypervillain: I agree but this is fake content, I'll let @angeloashmore and @lihbr decide how they want to take care of it

packages/slice-machine/package.json
xrutayisire: My bad on this one, the merge with conflicts took too much time (1 hour 🥵), so I planned to rely on you guys to check this 😬

packages/plugin-kit/src/hooks/documentation-read.ts
bapmrl: Okay, I understand. Thanks!

packages/adapter-next/src/hooks/documentation-read.ts
```diff
@@ -0,0 +1,47 @@
...
+			},
+			{
+				label: "App router",
+				content: "# App router",
+			},
+		];
+	}
+
+	return undefined;
```
bapmrl: ❓ Should it be `[]` as you removed `undefined` from the hook's return type?

packages/slice-machine/src/icons/FileIcons.tsx
```diff
@@ -0,0 +1,50 @@
+import { FC } from "react";
+
+export const Typescript: FC = () => (
```
bapmrl: ⚠️ I think it can be confusing to mix multiple icons into a single file. Indeed, it makes it harder to quickly know what icons are available just by checking the contents of the `icons` folder.

packages/slice-machine/src/icons/FileIcons.tsx
```diff
@@ -0,0 +1,50 @@
...
+      fill="#6F6E77"
+    />
+  </svg>
+);
+
+export const Javascript: FC = () => (
```
bapmrl: 💡 "TypeScript" and "JavaScript" are written with a capital "S".

packages/slice-machine/src/components/MainContainer/MainContainer.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+export const Default = {
+  args: {
+    children: (
+      <>
+        <MainContainerHeader
+          backTo={() => console.log("back to")}
+          breadcrumb="folder/file"
+          actions={
+            <>
+              <Button key="action-1">Action 1</Button>
```
bapmrl: ❓ I don't understand why we need to pass React `key`s here.

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypesBuilderPage.tsx
```diff
@@ -72,11 +87,60 @@ type CustomTypesBuilderPageWithProviderProps = {
...
+      disabled={!hasPendingModifications || isSavingCustomType}
+    >
+      Save
+    </Button>,
+  ];
+
+  return (
+    <MainContainer>
+      <MainContainerHeader
+        backTo={() => void router.push(config.tablePagePathname)}
```
bapmrl, line 142: 💡 We can fix that later, but there's a software design issue. `MainContainerHeader` has too many responsibilities. We could create a "connected" component that would handle navigation for us. That component could be located in a `core/navigation` folder and would use `Breadcrumb` and `Button`. It would implement the behavior and use the browser's `history` API to know whether we're allowed to call `router.back` or not. When not allowed (because the history is empty), the button could be `disabled`. But let's discuss this again at a later point so that we can merge this Pull Request 🙂

packages/slice-machine/src/features/documentation/useDocumentation.ts
```diff
@@ -0,0 +1,25 @@
+import { useRequest } from "@prismicio/editor-support/Suspense";
+
+import { managerClient } from "@src/managerClient";
+import {
+  Documentation,
+  DocumentationReadHookData,
+} from "@slicemachine/plugin-kit";
```
bapmrl: ⚠️ We should not import `@slicemachine/plugin-kit`. It's not part of `slice-machine-ui`'s `dependencies` and we probably don't want to add it. Adding it would increase the number of relationships between workspaces and blurrying the overall architecture of the project. I think the front end should not be aware of the plugin system (at least, it's my understanding of what DevX proposed).

packages/slice-machine/src/features/documentation/useDocumentation.ts
```diff
@@ -0,0 +1,25 @@
...
+
+async function getDocumentation(str: string): Promise<Documentation[]> {
+  const args = JSON.parse(str) as DocumentationReadHookData;
```
bapmrl: ⚠️ Why call `JSON.stringify` and then `JSON.parse`? It's so complicated and makes the code very hard to understand for other people. Let's be more explicit and pass a limited set of well typed arguments.

packages/slice-machine/src/features/documentation/useDocumentation.ts
```diff
@@ -0,0 +1,25 @@
...
+  const { errors, documentation } = await managerClient.documentation.read({
+    kind: args.kind,
+    data: args.data,
+  });
+  if (errors.length > 0 || documentation.length === 0) {
```
bapmrl: ❓ Why throw and potentially invalidate the cache if there's no documentation? What do we want to do UX-wise? Maybe when there's no documentation, we should simply render a message explaining that we can't provide snippets for the currently used adapter?

bapmrl (changes requested):
I'm sorry, but this Pull Request is too big and I'll have to continue commenting Monday morning. Here's a second set of comments.

packages/adapter-next/src/hooks/documentation-read.ts
hypervillain: It does not really matter as it is decoded by the DocumentationManager, but it's maybe more accurate yes

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypesBuilderPage.tsx
hypervillain: The button should not be disabled if no history is empty, it should still redirect to the table of customTypes. It does not mean we can move it to `core/navigation` of course, but right now the behaviour is correct. Maybe it's outside the scope of this PR

packages/slice-machine/src/features/documentation/useDocumentation.ts
hypervillain: We want to hide the button when there is no documentation for the moment. But I should be able to display nothing without throwing to not re-fetch the documentation each time users go on the page

packages/slice-machine/src/features/documentation/useDocumentation.ts
hypervillain: Ah you're right. My bad. I've removed typings to let TypeScript resolve the types correctly (with the JSON parsing of the argument, which is ugly indeed, as I don't see how to change that fast)

packages/slice-machine/src/features/documentation/useDocumentation.ts
hypervillain: From my understanding, useRequest is too limited to use objects and needs strings to behave properly

bapmrl (approved):

lihbr:
Some issues I encountered:

<details>
<summary>1/ <del>Vue syntax is not recognized</del></summary>
<br />

image

</details>

<details>
<summary>2/ <del>Inline code blocks are not styled as such</del></summary>
<br />

image

</details>

<details>
<summary>3/ <del>Line numbers are not aligned when the number of digits increases with Vue (I assume it's because it's not highlighted)</del></summary>
<br />

image

</details>

<details>
<summary>4/ Blockquotes are not styled (happy to postpone this one)</summary>
<br />

image

Markdown blockquote: https://www.markdownguide.org/basic-syntax/#blockquotes-1
</details>

hypervillain:
I think adding Vue syntax would simply require you to add `SyntaxHighlighter.registerLanguage("vue", vue);` next to the tsx one

PR #1023, opened 2023-06-29
Use EditDropdown in table and builder page

hypervillain:
Make Custom types EditDropdown reusable, use it in both tables and builder page

Editing a label from inside the builder does not update the rename inside the fileSystem but adds the change to the local state. Rename from CT tables is untouched.

This introduces a difference between how you treat _available_ and _selected_ custom types. Wherever possible, changes that need to be saved immediately are handled directly by the manager, and only the `.success` actions are preserved. It was not possible to do it with "selected CT" as it would require a change of the whole "selected" reducer


e2e-projects/next/customtypes/reusa/index.json
```diff
@@ -0,0 +1,41 @@
+{
```
xrutayisire: Do you want to add a new Page Type for others?

packages/slice-machine/components/DeleteCTModal/index.tsx
```diff
@@ -1,54 +1,51 @@
...
-        store,
-        LoadingKeysEnum.DELETE_CUSTOM_TYPE
-      ),
-    })
-  );
+> = ({ customType, format, onClose }) => {
   const customTypesMessages = CUSTOM_TYPES_MESSAGES[format];
-  const { closeModals, deleteCustomType } = useSliceMachineActions();
+
+  const [isDeleting, setIsdeleting] = useState(false);
```
xrutayisire:
```suggestion
  const [isDeleting, setIsDeleting] = useState(false);
```

packages/slice-machine/lib/builders/CustomTypeBuilder/TabZone/index.tsx
```diff
@@ -44,14 +51,12 @@ const TabZone: React.FC<TabZoneProps> = ({ tabId, fields, sliceZone }) => {
...
-      currentCustomType: selectCurrentCustomType(store),
-      poolOfFields: selectCurrentPoolOfFields(store),
-    })
-  );
+  const { poolOfFields } = useSelector((store: SliceMachineStoreType) => ({
+    poolOfFields: selectCurrentPoolOfFields(store),
+  }));
 
-  if (!currentCustomType || !poolOfFields) {
+  // TODO: fix this
```
xrutayisire: TODO here, just for a reminder

packages/slice-machine/src/features/customTypes/EditDropdown.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+export const EditDropdown: FC<EditDropdownProps> = ({
+  format,
+  customType,
+  isChangesLocal,
+}) => {
+  const router = useRouter();
+  const { saveCustomTypeSuccess } = useSliceMachineActions();
+
+  const [isDeleting, setIsDeleting] = useState(false);
```
xrutayisire: For the 2 variables, isDeleting, isRenaming I would not use this wording. I was really confused at first as this means the process isDeleting is in progress. 

On my work I used "isCreateSliceModalOpen" for example. So it's clear is just the boolean that handle the modal, not the actual action behind.

packages/slice-machine/src/features/customTypes/EditDropdown.tsx
```diff
@@ -0,0 +1,128 @@
...
+    setCustomTypeBeingConverted(true);
+    await convertCustomToPageType(customType, saveCustomTypeSuccess);
+    const customPagePathname =
+      CUSTOM_TYPES_CONFIG.custom.getBuilderPagePathname(customType.id);
+    if (router.asPath === customPagePathname) {
+      const pagePagePathname = CUSTOM_TYPES_CONFIG.page.getBuilderPagePathname(
+        customType.id
+      );
+      void router.replace(pagePagePathname);
+      setTimeout(() => {
```
xrutayisire: Do you really need that when you do these setTimeout in your code? I mean, since there is a redirection for me, it's not required, right?

packages/slice-machine/components/DeleteCTModal/index.tsx
```diff
@@ -1,54 +1,51 @@
...
+
+  const { deleteCustomTypeSuccess } = useSliceMachineActions();
+
+  const onConfirm = async () => {
+    setIsdeleting(true);
+    await deleteCustomType(customType, () =>
+      deleteCustomTypeSuccess(customType.id)
+    );
+    onClose(true);
```
xrutayisire: Having a props "onDeleteSuccess" on the component could be clearer, no? I feel here we are a bit mixing the close logic and the logic of the success case. But maybe it's too small details here.

packages/slice-machine/src/features/customTypes/actions/renameCustomType.ts
```diff
@@ -0,0 +1,46 @@
+import { CustomType } from "@prismicio/types-internal/lib/customtypes";
+import { managerClient } from "@src/managerClient";
+import { CUSTOM_TYPES_MESSAGES } from "../customTypesMessages";
+import { CustomTypeFormat } from "@slicemachine/manager/*";
```
xrutayisire:
```suggestion
import { CustomTypeFormat } from "@slicemachine/manager";
```

When important manager stuff it's by default like that, bad conf. Not necessary and strange.

packages/slice-machine/src/features/customTypes/customTypesBuilder/PageSnippetDialog.tsx
```diff
@@ -0,0 +1,79 @@
...
+};
+
+type PageSnippetDialogProps = { model: CustomType };
+
+export const PageSnippetDialog: FC<PageSnippetDialogProps> = ({ model }) => {
+  return (
+    <div>
+      <ErrorBoundary
+        title="Request failed"
+        description={`An error occurred while fetching page types snippets.`}
```
xrutayisire:
```suggestion
        description="An error occurred while fetching page types snippets."
```

packages/slice-machine/src/features/customTypes/customTypesBuilder/PageSnippetDialog.tsx
```diff
@@ -0,0 +1,79 @@
...
+        renderError={() => null}
+      >
+        <Suspense
+          fallback={
+            <Button variant="secondary" startIcon={<Icon name="code" />}>
```
xrutayisire: The fallback button should not be disabled?

xrutayisire (changes requested):
First review done, I will take a closer look when it's not DRAFT anymore 🙂
I so much love the code without redux-saga!

e2e-projects/next/customtypes/reusa/index.json
hypervillain: Accidental commit and push :/

packages/slice-machine/src/features/customTypes/EditDropdown.tsx
hypervillain: Ah yes I strongly agree

packages/slice-machine/src/features/customTypes/EditDropdown.tsx
hypervillain: The redirection takes some time to be applied. If you remove it you're left with a blank page, and possibly a "NOT FOUND" redirection. I agree the time is very arbitrary though

packages/slice-machine/lib/builders/CustomTypeBuilder/TabZone/index.tsx
hypervillain: I would like to leave it as is because it's not typed and I'm not sure what's expected here. It's just very probable that the condition is never true

packages/slice-machine/components/Forms/RenameCustomTypeModal/RenameCustomTypeModal.tsx
```diff
@@ -1,62 +1,73 @@
...
 
   return (
     <ModalFormCard
+      isOpen
       dataCy="rename-custom-type-modal"
-      isOpen={isRenameCustomTypeModalOpen}
       widthInPx="530px"
       formId={`rename-custom-type-modal-${customTypeId}`}
-      close={closeModals}
+      close={onClose}
```
xrutayisire:
```suggestion
      onSubmit={handleOnSubmit}
```

packages/slice-machine/lib/builders/CustomTypeBuilder/TabZone/index.tsx
xrutayisire: So for me we could remove the TODO, a TODO without an issue and without any real idea behing is not a todo. Or it could be a simple comment. But she should be careful with TODO and we should use then only when we want to do someting. no?

packages/slice-machine/lib/builders/CustomTypeBuilder/index.tsx
```diff
@@ -2,18 +2,22 @@ import { Box } from "theme-ui";
 
 import TabZone from "./TabZone";
 import { Tabs } from "./Layout";
+import { CustomTypeSM } from "@lib/models/common/CustomType";
 
-const CustomTypeBuilder = () => {
+type CustomTypeBuilderArgs = {
```
xrutayisire:
```suggestion
type CustomTypeBuilderProps = {
```

It's props here right?

packages/slice-machine/src/features/customTypes/actions/renameCustomType.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    onSuccess(renamedCustomType);
+
+    toast.success(
+      `${customTypesMessages.name({
+        start: true,
+        plural: false,
+      })} renamed`
+    );
+  } catch (e) {
```
xrutayisire, line 38: Could be good to add a console.error here


```suggestion
  } catch (e) {
    const errorMessage = `Internal Error: ${customTypesMessages.name({
        start: true,
        plural: false,
      })} could not be renamed`
    console.error(errorMessage, e)
    toast.error(errorMessage)
```

packages/slice-machine/src/features/customTypes/actions/deleteCustomType.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    onSuccess();
+
+    toast.success(
+      `${customTypesMessages.name({
+        start: true,
+        plural: false,
+      })} deleted`
+    );
+  } catch (e) {
```
xrutayisire, line 30: Could be good to add a console.error here


```suggestion
  } catch (e) {
    const errorMessage = `Internal Error: ${customTypesMessages.name({
        start: true,
        plural: false,
      })} could not be deleted`
    console.error(errorMessage, e)
    toast.error(errorMessage)
```

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypesBuilderPage.tsx
```diff
@@ -108,6 +116,12 @@ const CustomTypesBuilderPageWithProvider: React.FC<
   const messages = CUSTOM_TYPES_MESSAGES[currentCustomType.format];
 
   const actions = [
+    <EditDropdown
+      isChangesLocal
```
xrutayisire, line 120: Not sure I understand the naming of this props here, I get the idea behind but the naming make me question myself. Could you explain me a bit this props?

packages/slice-machine/src/features/customTypes/EditDropdown.tsx
xrutayisire: What I don't understand is, this feature is from before and I didn't need this hack. Why that for you?
Here it's just used for a loader not a modal so It's even simpler 🤔

packages/slice-machine/src/modules/availableCustomTypes/index.ts
```diff
@@ -166,13 +151,10 @@ export const availableCustomTypesReducer: Reducer<
       };
     }
 
-    case getType(renameCustomTypeCreator.success): {
+    case getType(renameAvailableCustomType): {
       const id = action.payload.renamedCustomType.id;
       const customType = state[id];
 
-      // Rename only applies for custom type with local data
```
xrutayisire, line 173: Don't sure to get this previous condition 😅

packages/slice-machine/src/modules/loading/types.ts
```diff
@@ -6,7 +6,6 @@ export enum LoadingKeysEnum {
   CHECK_SIMULATOR = "CHECK_SIMULATOR",
   CHECK_SIMULATOR_IFRAME = "CHECK_SIMULATOR_IFRAME",
   CREATE_CUSTOM_TYPE = "CREATE_CUSTOM_TYPE",
-  RENAME_CUSTOM_TYPE = "RENAME_CUSTOM_TYPE",
```
xrutayisire, line 9: Also no need:
  DELETE_CUSTOM_TYPE = "DELETE_CUSTOM_TYPE",

Correct?

packages/slice-machine/src/features/customTypes/actions/renameCustomType.ts
```diff
@@ -0,0 +1,46 @@
...
+import { toast } from "react-toastify";
+
+type RenameCustomTypeArgs = {
+  model: CustomType;
+  newName: string;
+  onSuccess: (renamedCustomType: CustomType) => void;
+};
+export async function renameCustomType({
+  model,
+  newName,
```
xrutayisire: You added "newLabel" in another place and newName here, short thing but I feel better to have one naming "newLabel", WDYT?

packages/slice-machine/src/modules/useSliceMachineActions.ts
```diff
@@ -196,12 +168,34 @@ const useSliceMachineActions = () => {
   const cleanupCustomTypeStore = () =>
     dispatch(cleanupCustomTypeStoreCreator());
   const saveCustomType = () => dispatch(saveCustomTypeCreator.request());
+
+  /** Success actions = sync store state from external actions.
```
xrutayisire, line 172: I was wondering, should we create another file and another hook like useSliceMachineSuccessActions, or else? It could help us to know the good redux stuff and the "bad" 🤔

packages/slice-machine/src/modules/availableCustomTypes/index.ts
hypervillain: Yeah.. I think it was due to typing rather than an actual case

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypesBuilderPage.tsx
hypervillain: I called changes to models (slice or custom type) that are done against the Redux store "local", because they are not propagated to the Filesystem. In that case the RenameModal will dispatch an action on Redux, in the other case it will commit changes to the Filesystem. I preferred to pass a boolean rather than a full function because it makes both the tables and the builder easier to read. The behaviour might also evolve and it will be easier to deprecate the property

e2e-projects/next/customtypes/reusa/index.json
xrutayisire: Still here 🙂

packages/slice-machine/src/features/customTypes/actions/deleteCustomType.ts
```diff
@@ -0,0 +1,44 @@
...
+    const errorMessage = `Internal Error: ${customTypesMessages.name({
+      start: true,
+      plural: false,
+    })} could not be deleted`;
+    console.error(errorMessage, e);
+
+    toast.error(
```
xrutayisire: Why not using the new errorMessage const

packages/slice-machine/src/features/customTypes/actions/renameCustomType.ts
```diff
@@ -0,0 +1,52 @@
...
+      })} renamed`
+    );
+  } catch (e) {
+    const errorMessage = `Internal Error: ${customTypesMessages.name({
+      start: true,
+      plural: false,
+    })} could not be renamed`;
+    console.error(errorMessage, e);
+
+    toast.error(
```
xrutayisire: Why not using the new errorMessage const

packages/slice-machine/src/modules/useSliceMachineActions.ts
xrutayisire: Ok for later if needed

xrutayisire (approved):
LGTM, Pair Review together done 🙂

PR #1030, opened 2023-07-05
Mm/academy link

MarcMcIntosh:
## Context
Add a link to the academy when using next

- [ ] The CI is successful.


xrutayisire:
I have some comments for you, but best would be to call after my meeting :)

xrutayisire (approved):
LGTM, I just let you test the Suspense that I showed you :)
And btw, I will create an issue to not have condition for Next in the front-end.

PR #1031, opened 2023-07-06
refactor(redux-saga): Remove create slice saga

xrutayisire:
## Context

- Completes DT-1443 - Convert createSliceSaga to a simple TypeScript function

## The Solution

- Remove createSliceSaga
- Add the New Slice and Update Slices button in the header all the time (DESIGN)
- Create a "createSlice" simple TS function that don't know anything about state management
- Handle the custom logic in CreateSliceModal
- Update Redux store in CreateSliceModal

## Impact / Dependencies

- Another issue has been created to handle the removal of the getState call in the CreateSliceModal: DT-1453

- [ ] The CI is successful.


packages/slice-machine/components/Forms/CreateSliceModal/CreateSliceModal.tsx
```diff
@@ -1,62 +1,124 @@
...
+          };
+          await managerClient.customTypes.updateCustomType({
+            model: CustomTypes.fromSM(newCustomType),
+          });
+          saveCustomTypeSuccess(CustomTypes.fromSM(newCustomType));
+        }
+
+        // Redirect to the slice page
+        const variationId = newSlice.variations[0].id;
+        const sliceLocation = `/${libraryName.replace(
```
hypervillain: I would prefer this to be an external function, a bit like you did for the CT tables

packages/slice-machine/src/features/slices/actions/createSlice.ts
```diff
@@ -0,0 +1,66 @@
...
+    id: snakelize(sliceName),
+    type: "SharedSlice",
+    name: sliceName,
+    description: sliceName,
+    variations: [
+      {
+        id: DEFAULT_VARIATION_ID,
+        name: pascalize(DEFAULT_VARIATION_ID),
+        // TODO: What should this be?
+        docURL: "...",
```
hypervillain, line 57: this property is unused

packages/slice-machine/src/features/slices/actions/createSlice.ts
```diff
@@ -0,0 +1,66 @@
...
+        // TODO: What should this be?
+        version: "initial",
```
hypervillain, line 59: this is set by the CT API, any value here is fine

packages/slice-machine/src/features/slices/actions/createSlice.ts
```diff
@@ -0,0 +1,66 @@
...
+        description: pascalize(DEFAULT_VARIATION_ID),
+        // TODO: What should this be? The type requires it.
+        imageUrl: "",
```
hypervillain, line 64: This is required in the model but stripped by SM to be compared to the remote slice. Empty string is fine I believe, we just need to make sure it's not saved in the JSON model

hypervillain: (although saving it to the model should not cause any issue)

hypervillain (changes requested):
Very cool

packages/slice-machine/components/Forms/CreateSliceModal/CreateSliceModal.tsx
xrutayisire: Done

packages/slice-machine/src/features/slices/actions/createSlice.ts
xrutayisire: Done

packages/slice-machine/src/features/slices/actions/createSlice.ts
xrutayisire: Done

packages/slice-machine/src/features/slices/actions/createSlice.ts
xrutayisire: Done

PR #1032, opened 2023-07-06
feat(slice-machine-ui): add FloatingBackButton

bapmrl:
Completes DT-1423.


packages/slice-machine/src/features/customTypes/customTypesBuilder/useCustomType.ts
```diff
@@ -0,0 +1,20 @@
...
+import type { CustomType } from "@prismicio/types-internal/lib/customtypes";
+
+import { managerClient } from "@src/managerClient";
+
+type UseCustomTypeReturnType = Readonly<{
+  customType: CustomType | undefined;
+}>;
+
+export function useCustomType(id: string): UseCustomTypeReturnType {
+  return { customType: useRequest(readCustomType, [id]) };
```
xrutayisire: Why returning an object how to directly "CustomType | undefined"

packages/slice-machine/src/features/customTypes/customTypesBuilder/useCustomType.ts
```diff
@@ -0,0 +1,20 @@
+import { useRequest } from "@prismicio/editor-support/Suspense";
+import type { CustomType } from "@prismicio/types-internal/lib/customtypes";
+
+import { managerClient } from "@src/managerClient";
+
+type UseCustomTypeReturnType = Readonly<{
+  customType: CustomType | undefined;
```
xrutayisire: When can it be undefined? It doesn't throw if it's undefined?

packages/slice-machine/src/features/customTypes/customTypesConfig.ts
```diff
@@ -23,3 +25,22 @@ export const CUSTOM_TYPES_CONFIG = {
     Icon: DatabaseIcon,
   },
 };
+
+export function matchesBuilderPagePathname(
```
xrutayisire, line 29: These 2 functions could be outside the config no? They are dependent on it but are not part of it

packages/slice-machine/src/features/customTypes/customTypesConfig.ts
```diff
@@ -23,3 +25,22 @@ export const CUSTOM_TYPES_CONFIG = {
...
+  pathname: string,
+  customTypeId: string
+): boolean {
+  return Object.values(CUSTOM_TYPES_CONFIG).some(({ getBuilderPagePathname }) =>
+    pathname.startsWith(getBuilderPagePathname(customTypeId))
+  );
+}
+
+export function readBuilderPageDynamicSegment(
```
xrutayisire, line 38: I feel what you want is just a "find" no?

```suggestion
export function readBuilderPageDynamicSegment(
  query: NextRouter["query"]
): string | undefined {
  const customTypesConfigValues = Object.values(CUSTOM_TYPES_CONFIG);

  const matchingConfigValue = customTypesConfigValues.find(
    ({ builderPageDynamicSegment }) =>
      typeof query[builderPageDynamicSegment] === "string"
  );

  return matchingConfigValue ? query[matchingConfigValue.builderPageDynamicSegment] : undefined;
}
```

I always favor dedicated array utils function over simple "for" to explain exactly what you're doing to the other devs.

packages/slice-machine/src/features/sliceBuilder/FloatingBackButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+} from "@src/features/customTypes/customTypesConfig";
+import { type Route, useRouteChange } from "@src/hooks/useRouteChange";
+import { CloseIcon } from "@src/icons/CloseIcon";
+import { UndoIcon } from "@src/icons/UndoIcon";
+
+export const FloatingBackButton: FC = () => {
+  const { source } = useRouteChange();
+  const sourceCustomTypeId = getSourceCustomTypeId(source);
+  return sourceCustomTypeId !== undefined ? (
+    // TODO: we shouldn't have to pass an empty description and title to the ErrorBoundary.
```
xrutayisire, line 19: We should create an issue, no? To prevent having never ending TODO

packages/slice-machine/src/features/sliceBuilder/FloatingBackButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+function getSourceCustomTypeId(source: Route): string | undefined {
+  const sourceCustomTypeId = readBuilderPageDynamicSegment(source.query);
+  return sourceCustomTypeId !== undefined &&
+    matchesBuilderPagePathname(source.asPath, sourceCustomTypeId)
+    ? sourceCustomTypeId
+    : undefined;
+}
+
+type BackButtonProps = Readonly<{ sourceCustomTypeId: string }>;
```
xrutayisire: We said no need of ReadOnly anymore right for Props?

```suggestion
type BackButtonProps = { sourceCustomTypeId: string }
```

packages/slice-machine/src/features/sliceBuilder/FloatingBackButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+const BackButton: FC<BackButtonProps> = ({ sourceCustomTypeId }) => {
+  const { customType: sourceCustomType } = useCustomType(sourceCustomTypeId);
+  const [visible, setVisible] = useState(true);
+  const router = useRouter();
+  if (sourceCustomType !== undefined && visible) {
+    const { format, id } = sourceCustomType;
+    // TODO: format should not be nullable.
```
xrutayisire, line 52: Same, should we need an issue? It's a typing thing here, in types internal in should be ok to not have a format since there is a fallback to support old one. If here we always have it, maybe we should create a specific type that override the one from internal. But an issue could be good then.

If we don't need issue, then we don't need TODO, maybe. That's my philosophy btw haha

packages/slice-machine/src/hooks/useRouteChange.tsx
```diff
@@ -0,0 +1,42 @@
...
+} from "react";
+
+type RouteChange = Readonly<{ source: Route; destination: Route }>;
+
+export type Route = Readonly<Pick<NextRouter, "asPath" | "query">>;
+
+const RouteChangeContext = createContext<RouteChange | undefined>(undefined);
+
+export const RouteChangeProvider: FC<PropsWithChildren> = (props) => (
+  <RouteChangeContext.Provider {...props} value={_useRouteChange()} />
```
xrutayisire, line 17: Is this a pattern we want to add in the code guideline? Starting a function by "_" when we want to keep it private?

packages/slice-machine/src/icons/CloseIcon.tsx
```diff
@@ -0,0 +1,19 @@
+import type { FC, SVGProps } from "react";
+
+export const CloseIcon: FC<SVGProps<SVGSVGElement>> = (props) => (
+  <svg
+    width="24"
+    height="24"
+    viewBox="0 0 24 24"
```
xrutayisire, line 7: First two icons with viewBox 24, here. It's the viewbox of the editor, on our side Benjamin has a viewbox of 32, is that normal?

xrutayisire (changes requested):
Ready for pair review to be quicker when you want 🙂

xrutayisire (approved):
LGTM after Pair Review

PR #1052, opened 2023-07-12
feat(Changelog): Add information about the adapters update in the cha…

xrutayisire:
## Context

- Completes DT-1465

## The Solution

- Create new function in the managers to be able to get the required information
- Update the condition to display the warning in the Side Nav to take into account the Adapter
- Update the color of the CURRENT tag when necessary
- Add a warning when only the adapter need an update
- The command for latest SM version now always display the way to update adapter at the same time

## Left To Do

- Clean code
- Check for improvements
- Check more the testing part
- Wait for Benjamin design confirmation

## Impact / Dependencies

- User should now be able to easily update both Slice Machine and the adapter
- It's not perfect and confusing because here it creates an artificial link between an adapter and SM that could not exist in the future

- [ ] The CI is successful.

## Preview

Screenshot 2023-07-13 at 00 29 09

Screenshot 2023-07-13 at 00 28 57

Screenshot 2023-07-13 at 00 30 41

### DIFF
--- a/packages/manager/src/managers/project/ProjectManager.ts
+++ b/packages/manager/src/managers/project/ProjectManager.ts
@@ -193,6 +193,27 @@ export class ProjectManager extends BaseManager {
 		return path.dirname(sliceMachinePackageJSONPath);
 	}
 
+	async getAdapterName(): Promise<string> {
+		const sliceMachineConfig = await this.getSliceMachineConfig();
+		const adapterName =
+			typeof sliceMachineConfig.adapter === "string"
+				? sliceMachineConfig.adapter
+				: sliceMachineConfig.adapter.resolve;
+
+		return adapterName;
+	}
+
+	async locateAdapterDir(): Promise<string> {
+		const projectRoot = await this.getRoot();
+		const adapterName = await this.getAdapterName();
+		const require = createRequire(path.join(projectRoot, "index.js"));
+		const adapterPackageJSONPath = require.resolve(
+			`${adapterName}/package.json`,
+		);
+
+		return path.dirname(adapterPackageJSONPath);
+	}
+
 	async initProject(args?: ProjectManagerInitProjectArgs): Promise<void> {
 		assertPluginsInitialized(this.sliceMachinePluginRunner);
 
--- a/packages/manager/src/managers/versions/VersionsManager.ts
+++ b/packages/manager/src/managers/versions/VersionsManager.ts
@@ -38,6 +38,41 @@ const detectVersionBumpKind = (
 	}
 };
 
+const filterStableVersions = (versions: string[]): string[] => {
+	return versions.filter((version) => {
+		// Exclude tagged versions (e.g. `1.0.0-alpha.0`).
+		// Exclude versions < 0.1.0 (e.g. `0.0.1`).
+		return (
+			/^[1-9]\d*\.\d+\.\d+$/.test(version) ||
+			/^\d+\.[1-9]\d*\.\d+$/.test(version)
+		);
+	});
+};
+
+const readPackageJSON = async (packageDir: string) => {
+	const packageJSONContents = await fs.readFile(
+		path.join(packageDir, "package.json"),
+		"utf8",
+	);
+
+	let packageJSON: unknown;
+	try {
+		packageJSON = JSON.parse(packageJSONContents);
+	} catch {
+		// noop
+	}
+
+	const { value, error } = decodePackageJSON(packageJSON);
+
+	if (error) {
+		throw new Error(
+			`Invalid ${packageDir} \`package.json\` file. ${error.errors.join(", ")}`,
+		);
+	}
+
+	return value;
+};
+
 type SliceMachineManagerGetReleaseNotesForVersionArgs = {
 	version: string;
 };
@@ -53,45 +88,16 @@ export class VersionsManager extends BaseManager {
 
 	async getRunningSliceMachineVersion(): Promise<string> {
 		const sliceMachineDir = await this.project.locateSliceMachineUIDir();
+		const packageJSON = await readPackageJSON(sliceMachineDir);
 
-		const sliceMachinePackageJSONContents = await fs.readFile(
-			path.join(sliceMachineDir, "package.json"),
-			"utf8",
-		);
-
-		let sliceMachinePackageJSON: unknown;
-		try {
-			sliceMachinePackageJSON = JSON.parse(sliceMachinePackageJSONContents);
-		} catch {
-			// noop
-		}
-
-		const { value, error } = decodePackageJSON(sliceMachinePackageJSON);
-
-		if (error) {
-			throw new Error(
-				`Invalid ${SLICE_MACHINE_NPM_PACKAGE_NAME} \`package.json\` file. ${error.errors.join(
-					", ",
-				)}`,
-			);
-		}
-
-		return value.version;
+		return packageJSON.version;
 	}
 
 	async getAllStableSliceMachineVersions(): Promise<string[]> {
 		const versions = await fetchNPMPackageVersions({
 			packageName: SLICE_MACHINE_NPM_PACKAGE_NAME,
 		});
-
-		const filteredVersions = versions.filter((version) => {
-			// Exclude tagged versions (e.g. `1.0.0-alpha.0`).
-			// Exclude versions < 0.1.0 (e.g. `0.0.1`).
-			return (
-				/^[1-9]\d*\.\d+\.\d+$/.test(version) ||
-				/^\d+\.[1-9]\d*\.\d+$/.test(version)
-			);
-		});
+		const filteredVersions = filterStableVersions(versions);
 
 		return semver.rsort(filteredVersions);
 	}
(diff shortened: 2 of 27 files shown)


packages/manager/src/managers/project/ProjectManager.ts
```diff
@@ -193,6 +193,27 @@ export class ProjectManager extends BaseManager {
...
+		const sliceMachineConfig = await this.getSliceMachineConfig();
+		const adapterName =
+			typeof sliceMachineConfig.adapter === "string"
+				? sliceMachineConfig.adapter
+				: sliceMachineConfig.adapter.resolve;
+
+		return adapterName;
+	}
+
+	async locateAdapterUIDir(): Promise<string> {
```
angeloashmore: This method might be better named `locateAdapterDir`. The similar method named `locateSliceMachineUIDir` is named such because it locates the `slice-machine-ui` package.

packages/manager/src/managers/versions/VersionsManager.ts
```diff
@@ -53,28 +88,7 @@ export class VersionsManager extends BaseManager {
...
-		const { value, error } = decodePackageJSON(sliceMachinePackageJSON);
-
-		if (error) {
-			throw new Error(
-				`Invalid ${SLICE_MACHINE_NPM_PACKAGE_NAME} \`package.json\` file. ${error.errors.join(
-					", ",
-				)}`,
-			);
-		}
+		const value = await readPackageJSON(sliceMachineDir);
```
angeloashmore: Small suggestion: rename `value` (a generic name) to `packageJSON` (a descriptive name).

packages/manager/test/SliceMachineManager-versions-checkIsAdapterUpdateAvailable.test.ts
```diff
@@ -0,0 +1,64 @@
...
+	const manager = createSliceMachineManager({
+		nativePlugins: { [adapter.meta.name]: adapter },
+		cwd,
+	});
+	const packageJSON = {
+		name: adapter.meta.name,
+		version: "1.0.0",
+	};
+
+	await mockAdapterUIDirectory({
```
angeloashmore: Similar to the previous comment, we can remove the "UI" part of the name.

packages/manager/test/SliceMachineManager-versions-checkIsSliceMachineUpdateAvailable.test.ts
```diff
@@ -24,7 +24,7 @@ it("returns true if an update is available", async (ctx) => {
 		versions: ["1.0.0", "2.0.0"],
 	});
 
-	const res = await manager.versions.checkIsUpdateAvailable();
```
angeloashmore, line 27: Should we still have a `checkIsUpdateAvailable()` method available? It could tell very generally if an update is available (checking for an SM UI and adapter update), which may be used in SM UI to display the sidebar update message.

packages/manager/test/SliceMachineManager-versions-getAllStableAdapterVersions.test.ts
```diff
@@ -0,0 +1,43 @@
+import { expect, it } from "vitest";
+
+import { createTestPlugin } from "./__testutils__/createTestPlugin";
+import { createTestProject } from "./__testutils__/createTestProject";
+import { mockNPMRegistryAPI } from "./__testutils__/mockNPMRegistryAPI";
+
+import { createSliceMachineManager } from "../src";
+
+it("returns stable Adapter versions", async (ctx) => {
```
angeloashmore: Super tiny nitpick: "adapter" should be lowercase

packages/slice-machine/components/AppLayout/Navigation/index.test.tsx
```diff
@@ -60,10 +60,17 @@ function renderApp({ canUpdate }: { canUpdate: boolean }): RenderReturnType {
...
-          updateAvailable: canUpdate,
-          latestNonBreakingVersion: null,
-          versions: [],
+          sliceMachine: {
+            currentVersion: "",
+            updateAvailable: canUpdate,
+            latestNonBreakingVersion: null,
+            versions: [],
+          },
+          adapter: {
```
angeloashmore, line 69: Should the `adapter` object contain the same properties as `sliceMachine`?

packages/slice-machine/components/Changelog/versionDetails/UpdateCommandBox.tsx
```diff
@@ -39,6 +49,18 @@ export const UpdateCommandBox: React.FC<UpdateCommandBoxProps> = ({
...
+      {isOnlyAdapterUpdate && isLatestSliceMachineVersion && (
+        <Text
+          sx={{
+            fontSize: "12px",
+            fontWeight: 500,
+            lineHeight: "16px",
+          }}
+        >
+          It looks like you are using an outdated version of the adapter. Run
+          this command to update your adapter:
```
angeloashmore, line 61: Could we get Sam Littlefair's (@samlfair) approval on this message if he hasn't already reviewed it?

packages/slice-machine/components/Changelog/versionDetails/UpdateCommandBox.tsx
```diff
@@ -63,18 +85,40 @@ export const UpdateCommandBox: React.FC<UpdateCommandBoxProps> = ({
...
     case "pnpm":
     case "pnpm@6":
-      return `pnpm add --save-dev ${packageSpec}`;
+      return `pnpm add --save-dev ${packagesSpecsStr}`;
     case "yarn":
     case "yarn@berry":
-      return `yarn add --dev ${packageSpec}`;
+      return `yarn add --dev ${packagesSpecsStr}`;
   }
 }
```
angeloashmore, line 104: We might want to use the existing logic from the manager to retrieve the installation command: https://github.com/prismicio/slice-machine/blob/1db109d2d74159ee5abbc3ebcf6c456b9430716a/packages/manager/src/lib/installDependencies.ts#L30-L59

Note that we would need to call an async method via `managerClient` if we made that change, which may introduce some unnecessary complexity. You can be the judge on whether or not it's worth it. 🙂

angeloashmore (commented):
Nice work! I left a few comments re: naming and some opportunities for reusing code. It looks very close to complete to me. 👍

packages/manager/test/SliceMachineManager-versions-getAllStableAdapterVersions.test.ts
xrutayisire: Done :)

packages/manager/test/SliceMachineManager-versions-checkIsAdapterUpdateAvailable.test.ts
xrutayisire: Done, thanks 🙂

packages/manager/src/managers/versions/VersionsManager.ts
xrutayisire: Done, thanks 🙂

packages/manager/src/managers/project/ProjectManager.ts
xrutayisire: Done, that's a mistake on my side, was too fast here, thank you 🙂

packages/manager/test/SliceMachineManager-versions-checkIsSliceMachineUpdateAvailable.test.ts
xrutayisire: I ask myself the same question, even if it could help to have only method, the two methods called within will not change at the end. So I choose two functions, mostly to prevent us to be force to call NPM for the adapters if we don't need the info. Today we only call both of them at the same time, but it's a performance issue if not one day, right?

Let me know if you think it's better, and I could merge the two for sure 🙂

packages/slice-machine/components/AppLayout/Navigation/index.test.tsx
xrutayisire: At the moment, we don't need the same properties. I decided to only save in the store what is required. I don't know how we want to handle the changelog page in the future, so I kept it short in purpose. It would mean to save unused value in store, and it could be hard to debug and understand later on, WDYT?

packages/slice-machine/components/Changelog/versionDetails/UpdateCommandBox.tsx
xrutayisire: Yep, this message is coming from him 🙂

packages/slice-machine/components/Changelog/versionDetails/UpdateCommandBox.tsx
xrutayisire: That could be a good idea for the front-end not having to know this management. My PR is already a bit too big, and it's a bit out of the score of this issue. When we will redesign this page, I could propose such a change and investigate if we would like this, ok for you?

packages/manager/test/SliceMachineManager-versions-checkIsSliceMachineUpdateAvailable.test.ts
angeloashmore: Sounds good to me! We can always introduce a convenience method that combines all update checkers if needed in the future. 🙂

packages/slice-machine/components/AppLayout/Navigation/index.test.tsx
angeloashmore: Makes sense! 🙂

packages/slice-machine/components/Changelog/versionDetails/UpdateCommandBox.tsx
angeloashmore: Yep, works for me. I definitely think we should consider it when the page is reworked, but it's not needed now.

PR #1057, opened 2023-07-17
track copy code snippet

hypervillain:
Completes https://linear.app/prismic/issue/DT-1466/aapm-i-can-track-the-usage-of-code-snippets


packages/slice-machine/src/features/documentation/MarkdownRenderer/index.tsx
```diff
@@ -29,10 +32,20 @@ const MarkdownCodeBlock = (props: CodeProps) => {
     return null;
   })();
 
+  const onCopy = () => {
+    if (adapter !== undefined) {
+      void telemetry.track({
+        event: "page-type:copy-snippet",
+        framework: adapter,
```
xrutayisire, line 41: Sorry, I just realized also, but the issue says adapter and framework. Here and in the previous event, we are mixing the two :/

xrutayisire (changes requested):
Not sure about how to handle the properties. We should ask Côme.

xrutayisire:
Also, you didn't create your branch from an up-to-date version of dev-next-release. Maybe best would be to be up-to-date.

packages/slice-machine/src/features/documentation/MarkdownRenderer/index.tsx
hypervillain: although we call them adapters in SM, we've called it framework in Amplitude. A bit weird, but I think it's easier to check for PMs

xrutayisire: Ok if you can confirm with Côme before the exact thing we will give since it's not what we have in the issue :/

xrutayisire:
Could you add the repository in the property so it's tracked later on? As discussed in the thread

xrutayisire (approved):
LGTM 👍

packages/manager/src/managers/telemetry/types.ts
```diff
@@ -114,6 +117,11 @@ type OpenPageSnippetSegmentEvent = SegmentEvent<
 	{ framework: string }
 >;
 
+type CopyPageSnippetSegmentEvent = SegmentEvent<
+	typeof SegmentEventType.copy_page_snippet,
+	{ framework: string; repository: string }
```
xrutayisire:
```suggestion
	{ framework: string; }
```

No need then finally

xrutayisire (changes requested):
Just saw the little thing

PR #1060, opened 2023-07-19
Init login if necessary

hypervillain:
Completes https://linear.app/prismic/issue/DT-1501/aauser-in-the-init-flow-i-dont-have-to-log-in-when-its-not-needed

### DIFF
--- a/packages/init/src/SliceMachineInitProcess.ts
+++ b/packages/init/src/SliceMachineInitProcess.ts
@@ -122,7 +122,6 @@ export class SliceMachineInitProcess {
 			);
 
 			await this.beginCoreDependenciesInstallation();
-			await this.loginAndFetchUserData();
 
 			if (this.options.repository) {
 				await this.useRepositoryFlag();
@@ -350,6 +349,20 @@ export class SliceMachineInitProcess {
 		]);
 	}
 
+	protected loginIfNecessary(): Promise<void> {
+		return listrRun([
+			{
+				title: "",
+				task: async (_) => {
+					const isLoggedIn = await this.manager.user.checkIsLoggedIn();
+					if (!isLoggedIn || this.context.userProfile === undefined) {
+						await this.loginAndFetchUserData();
+					}
+				},
+			},
+		]);
+	}
+
 	protected loginAndFetchUserData(): Promise<void> {
 		return listrRun([
 			{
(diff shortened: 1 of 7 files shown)


packages/init/src/SliceMachineInitProcess.ts
```diff
@@ -215,6 +223,31 @@ export class SliceMachineInitProcess {
 		}
 	}
 
+	protected async checkIsLoginRequired(): Promise<boolean> {
+		try {
+			assertExists(this.context.repository, "");
```
xrutayisire, line 228: With what I saw you just need to put again the same message as the other places:

```suggestion
			assertExists(
			   this.context.repository,
			   "Repository selection must be available through context to proceed",
		          );
```

packages/init/src/SliceMachineInitProcess.ts
```diff
@@ -135,12 +135,20 @@ export class SliceMachineInitProcess {
...
+
+			const isLoginRequired = await this.checkIsLoginRequired();
+			if (isLoginRequired) {
+				await this.loginAndFetchUserData();
+				if (this.context.repository.exists) {
+					this.validateWriteAccess();
+				}
+			}
 			if (!this.context.repository.exists) {
 				await this.createNewRepository();
```
xrutayisire, line 149: How long do you think this take?

hypervillain: the `validateWriteAccess` method?

packages/init/src/SliceMachineInitProcess.ts
hypervillain: It's not necessary here because the assertion is caught and is used to return a need for login!

packages/init/src/SliceMachineInitProcess.ts
xrutayisire: the await this.createNewRepository() method?

packages/init/src/SliceMachineInitProcess.ts
xrutayisire: I don't understand? For me it's just that all places I saw it was repeating the text even if not needed, no?

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
```diff
@@ -461,8 +466,13 @@ export class PrismicRepositoryManager extends BaseManager {
 			body: args.body ? JSON.stringify(args.body) : undefined,
 			headers: {
 				// Some endpoints rely on the authorization header...
-				Authorization: `Bearer ${cookies["prismic-auth"]}`,
-				Cookie: serializeCookies(cookies),
+
+				...(cookies !== undefined
```
xrutayisire, line 477: Could you explain me the change here?

packages/init/src/SliceMachineInitProcess.ts
```diff
@@ -428,39 +497,23 @@ export class SliceMachineInitProcess {
...
-							domain,
-							existsFn: (domain) =>
-								this.manager.prismicRepository.checkExists({ domain }),
-						});
+					if (validation.LessThan4 || validation.MoreThan30) {
 						const errorMessage = getErrorMessageForRepositoryDomainValidation({
-							validation,
+							validation: {
+								...validation,
+								AlreadyExists: false,
```
xrutayisire, line 515: Why A in uppercase?

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
xrutayisire: Ok basically, this is the place you remove the forced authent?

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
```diff
@@ -134,7 +134,12 @@ export class PrismicRepositoryManager extends BaseManager {
 			`./app/dashboard/repositories/${args.domain}/exists`,
 			API_ENDPOINTS.PrismicWroom,
 		);
-		const res = await this._fetch({ url });
+		const res = await fetch(url.toString(), {
```
xrutayisire: We should keep using this._fetch, no?

xrutayisire: Could be pass a parameter that prevent authent on this._fetch?
I'm afraid if we do something to all fetch we will forget this one in the future

xrutayisire: You could add the same condition, if no cookies just don't pass it, actually. Without a paramater so.

packages/init/src/SliceMachineInitProcess.ts
xrutayisire: Oh I understand!

xrutayisire (approved):
Let's ship that 🚀

PR #1061, opened 2023-07-19
refactor: video tooltip should not close on hover events

MarcMcIntosh:
## Context
Ticket: https://linear.app/prismic/issue/DT-1504/aauser-when-i-interact-out-side-of-the-video-tool-tip-it-should-not

Given I have a next project

[x] when have not seen the video tooltip or cleared local storage the video tooltip should open
[x] When I clicking out side of the tool-tip it should close
[x] when clicking the close button it should close
[x] when clicking the "got it" button it should close
[x] when hovering in and out should not close the tooltip
[x] Also ensure the close icon is properly aligned

- [ ] The CI is successful.

### DIFF
--- a/packages/slice-machine/src/components/VideoPopover/VideoPopover.css.ts
+++ b/packages/slice-machine/src/components/VideoPopover/VideoPopover.css.ts
@@ -93,7 +93,11 @@ export const videoButton = style([
 
 export const closeButton = style([
   block,
-  sprinkles({ color: colors.purple12 }),
+  sprinkles({
+    color: colors.purple12,
+    padding: 6,
+    display: "flex",
+  }),
   {
     background: "none",
     border: "none",
--- a/packages/slice-machine/src/components/VideoPopover/VideoPopover.stories.tsx
+++ b/packages/slice-machine/src/components/VideoPopover/VideoPopover.stories.tsx
@@ -50,7 +50,7 @@ export const CloseWhenClickingOutSide = {
           onPlay={noop}
           cloudName="dmtf1daqp"
           publicId="Tooltips/pa-course-overview_eaopsn"
-          delay={5000}
+          delay={1500}
         >
           <button>Popover</button>
         </VideoPopover>
--- a/packages/slice-machine/src/components/VideoPopover/VideoPopover.stories.tsx
+++ b/packages/slice-machine/src/components/VideoPopover/VideoPopover.stories.tsx
@@ -39,3 +39,23 @@ export const VideoPlayer = {
     />
   ),
 };
+
+export const CloseWhenClickingOutSide = {
+  render: () => {
+    return (
+      <div>
+        <VideoPopover
+          open={true}
+          onClose={noop}
+          onPlay={noop}
+          cloudName="dmtf1daqp"
+          publicId="Tooltips/pa-course-overview_eaopsn"
+          delay={5000}
+        >
+          <button>Popover</button>
+        </VideoPopover>
+        <button>Outside</button>
+      </div>
+    );
+  },
+};
--- a/packages/slice-machine/src/components/VideoPopover/VideoPopover.test.tsx
+++ b/packages/slice-machine/src/components/VideoPopover/VideoPopover.test.tsx
@@ -68,4 +68,74 @@ describe("VideoPopover", () => {
 
     expect(onClose).toHaveBeenCalled();
   });
+
+  test("give it is open, it should close when a user clicks else where on the screen", async () => {
+    const onClose = vi.fn();
+    const onPlay = vi.fn();
+
+    const App = () => {
+      return (
+        <div>
+          <VideoPopover
+            open={true}
+            onClose={onClose}
+            onPlay={onPlay}
+            delay={0}
+            cloudName="dmtf1daqp"
+            publicId="Tooltips/pa-course-overview_eaopsn"
+          >
+            <div>Thing</div>
+          </VideoPopover>
+
+          <button>other</button>
+        </div>
+      );
+    };
+
+    const { user } = render(<App />);
+
+    await waitFor(() => screen.getByText("Got it"));
+
+    const otherButton = screen.getByText("other");
+
+    await user.click(otherButton);
+
+    expect(screen.queryByText("Got it")).not.toBeInTheDocument();
+  });
+
+  test("hovering in and out should not close the popover", async () => {
+    const onClose = vi.fn();
+    const onPlay = vi.fn();
+
+    const App = () => {
+      return (
+        <div>
+          <VideoPopover
+            open={true}
+            onClose={onClose}
+            onPlay={onPlay}
+            delay={0}
+            cloudName="dmtf1daqp"
+            publicId="Tooltips/pa-course-overview_eaopsn"
+          >
+            <div>Thing</div>
+          </VideoPopover>
+
+          <button>other</button>
+        </div>
+      );
+    };
+
+    const { user } = render(<App />);
+
+    const closeButton = await waitFor(() => screen.getByText("Got it"));
+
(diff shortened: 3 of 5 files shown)


packages/slice-machine/src/components/VideoPopover/VideoPopover.stories.tsx
```diff
@@ -39,3 +39,23 @@ export const VideoPlayer = {
...
+  render: () => {
+    return (
+      <div>
+        <VideoPopover
+          open={true}
+          onClose={noop}
+          onPlay={noop}
+          cloudName="dmtf1daqp"
+          publicId="Tooltips/pa-course-overview_eaopsn"
+          delay={5000}
```
xrutayisire: For the delay in storybook, for all stories I would put something really fast like 1 or 2 seconds, so we don't wait as long haha, WDYT?

packages/slice-machine/src/components/VideoPopover/VideoPopover.test.tsx
```diff
@@ -68,4 +68,74 @@ describe("VideoPopover", () => {
...
+          </VideoPopover>
+
+          <button>other</button>
+        </div>
+      );
+    };
+
+    const { user } = render(<App />);
+
+    const closeButton = await waitFor(() => screen.getByText("Got it"));
```
xrutayisire: As discussed:

```suggestion
    const closeButton = await screen.findByText("Got it");
```

And thanks to this, you could not have to do it in two lines:

```suggestion
    await user.hover(await screen.findByText("Got it"));
```

packages/slice-machine/src/components/VideoPopover/VideoPopover.tsx
```diff
@@ -77,77 +159,24 @@ function useTimerWhen(condition: boolean, fn: () => void, delay: number) {
...
-  );
-
-  useTimerWhen(open, handleDefaultOpening, delay);
-
-  const handleChange = (value: boolean) => {
-    if (value === false) {
-      handleClose();
-    } else if (value === true && open === true) {
-      setOpen(value);
+function useOnClickOutside<T extends HTMLElement = HTMLElement>(
```
xrutayisire: Following your question, yes, I would put this in our hooks folder.

packages/slice-machine/src/components/VideoPopover/VideoPopover.tsx
```diff
@@ -77,77 +159,24 @@ function useTimerWhen(condition: boolean, fn: () => void, delay: number) {
...
+  ref: React.MutableRefObject<T | null>,
+  handler: () => void
+) {
+  useEffect(() => {
+    function checkWhereWasClicked(event: MouseEvent) {
```
xrutayisire: `checkWhereWasClicked` is somewhat vague and not really what the function does.

I can propose `handleClickOutside` to be clearer, as it explains exactly when the function will be called, do you like it?

packages/slice-machine/src/components/VideoPopover/VideoPopover.tsx
```diff
@@ -77,77 +159,24 @@ function useTimerWhen(condition: boolean, fn: () => void, delay: number) {
...
-        onClose={handleClose}
-        onPlay={onPlay}
-        cloudName={cloudName}
-        publicId={publicId}
-        thumbnail={thumbnail}
-      />
-    </HoverCard>
-  );
-};
+    document.addEventListener("mousedown", checkWhereWasClicked);
```
xrutayisire: Maybe we should add and remote the event listenner only when the ref is not null?

```suggestion
    if (ref.current !== null) {
      document.addEventListener("mousedown", handleClickOutside);
    }

    return () => {
      if (ref.current !== null) {
        document.removeEventListener("mousedown", handleClickOutside);
      }
    }
```

packages/slice-machine/src/components/VideoPopover/VideoPopover.tsx
```diff
@@ -77,77 +159,24 @@ function useTimerWhen(condition: boolean, fn: () => void, delay: number) {
...
-      handleClose();
-    } else if (value === true && open === true) {
-      setOpen(value);
+function useOnClickOutside<T extends HTMLElement = HTMLElement>(
+  ref: React.MutableRefObject<T | null>,
+  handler: () => void
+) {
+  useEffect(() => {
+    function checkWhereWasClicked(event: MouseEvent) {
+      if (
```
xrutayisire: Best is not to use case if possible. Also, we can use optional chaining to make it shorter.

You could try something like:

```suggestion
     if (
        event.target instanceof Node &&
        ref.current?.contains(event.target) === false
      ) {
        handler();
      }
```

packages/slice-machine/src/components/VideoPopover/VideoPopover.css.ts
```diff
@@ -93,7 +93,11 @@ export const videoButton = style([
 
 export const closeButton = style([
   block,
-  sprinkles({ color: colors.purple12 }),
+  sprinkles({
+    color: colors.purple12,
+    padding: 6,
```
xrutayisire, line 98: Not sure to understand, I see 4px as global padding and then 6px here so 10px before the icon, is that what we want?

xrutayisire (changes requested):
Looks nice, no really big changes but small details I saw 🙂

packages/slice-machine/src/components/VideoPopover/VideoPopover.css.ts
MarcMcIntosh: Depend which hover-card you look at: this one says 6 https://www.figma.com/file/wb7ujCJ4j7EiP7aDDGRYXB/SliceMachine%E2%84%A2-UI?node-id=4018%3A320396&mode=dev

but I can change it back to four and it shouldn't make much difference...

xrutayisire: Oh no that's perfect your link! Was not on that one, thank you!

packages/slice-machine/src/components/VideoPopover/VideoPopover.tsx
MarcMcIntosh: there's a similar hook in `usehooks-ts` ... not sure if it's worth inistalling the whole library for just a few hooks :/

packages/slice-machine/src/components/VideoPopover/VideoPopover.tsx
MarcMcIntosh: for removing it, i'd say it shouldn't matter if it's null or not (as the ref is mutablle)
for addding yes, that makes sense :)

packages/slice-machine/src/components/VideoPopover/VideoPopover.stories.tsx
MarcMcIntosh: 1.5

packages/slice-machine/src/components/VideoPopover/VideoPopover.tsx
MarcMcIntosh: can do

packages/slice-machine/src/components/VideoPopover/VideoPopover.tsx
xrutayisire: No don't think so, putting it in hooks folder is great already for me

packages/slice-machine/src/components/VideoPopover/VideoPopover.tsx
xrutayisire: Good one! 100% agree

xrutayisire (approved):
LGTM 🙂

PR #1062, opened 2023-07-20
feat(SideNav): Add info in Changes menu item when user is disconnected

xrutayisire:
## Context

- Completes DT-1503

## The Solution

- Create a new component, "ChangesRightElement" to better manage that information and the conditions
- Added tests for that
- Increase a bit the max-width so it fit 
  - The max-width is strictly for the ellipsis as you can see, so no problem to increase it
  - max-width depend on the text that still need to be validated

## Impact / Dependencies

- No impact

- [ ] The CI is successful.

## Preview

Screenshot 2023-07-20 at 17 25 50

### DIFF
--- a/packages/slice-machine/components/AppLayout/Navigation/ChangesRightElement.tsx
+++ b/packages/slice-machine/components/AppLayout/Navigation/ChangesRightElement.tsx
@@ -1,0 +1,33 @@
+import { FC } from "react";
+
+import { useUnSyncChanges } from "@src/hooks/useUnSyncChanges";
+import { RightElement } from "@src/components/SideNav/SideNav";
+import { AuthStatus } from "@src/modules/userContext/types";
+import { useNetwork } from "@src/hooks/useNetwork";
+
+export const ChangesRightElement: FC = () => {
+  const isOnline = useNetwork();
+  const { unSyncedSlices, unSyncedCustomTypes, authStatus } =
+    useUnSyncChanges();
+  const numberOfChanges = unSyncedSlices.length + unSyncedCustomTypes.length;
+
+  if (
+    !isOnline ||
+    authStatus === AuthStatus.UNAUTHORIZED ||
+    authStatus === AuthStatus.FORBIDDEN
+  ) {
+    return <RightElement>Login required</RightElement>;
+  }
+
+  if (numberOfChanges === 0) {
+    return null;
+  }
+
+  const formattedNumberOfChanges = numberOfChanges > 9 ? "+9" : numberOfChanges;
+
+  return (
+    <RightElement type="pill" data-cy="changes-number">
+      {formattedNumberOfChanges}
+    </RightElement>
+  );
+};
--- a/packages/slice-machine/components/AppLayout/Navigation/index.test.tsx
+++ b/packages/slice-machine/components/AppLayout/Navigation/index.test.tsx
@@ -1,29 +1,40 @@
 // @vitest-environment jsdom
 import { describe, test, expect, vi } from "vitest";
+import Router from "next/router";
+import { act } from "react-dom/test-utils";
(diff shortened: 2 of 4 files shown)


packages/slice-machine/components/AppLayout/Navigation/ChangesRightElement.tsx
```diff
@@ -0,0 +1,33 @@
+import { FC } from "react";
+
+import { useUnSyncChanges } from "@src/hooks/useUnSyncChanges";
+import { RightElement } from "@src/components/SideNav/SideNav";
+import { AuthStatus } from "@src/modules/userContext/types";
+import { useNetwork } from "@src/hooks/useNetwork";
+
+export const ChangesRightElement: FC = () => {
+  const isOnline = useNetwork();
```
MarcMcIntosh, line 9: `useUnsyncedChanges` also returns `isOnline`

packages/slice-machine/components/AppLayout/Navigation/index.test.tsx
```diff
@@ -133,6 +156,95 @@ describe("Side Navigation", () => {
...
+    expect(await screen.findByText("Changes")).toBeVisible();
+    const changesItem = screen
+      .getByText("Changes")
+      .closest("a") as HTMLAnchorElement;
+
+    expect(within(changesItem).getByText("Login required")).toBeVisible();
+  });
+
+  test("should display the information that user is disconnected on the 'Changes' item when offline", async () => {
+    vi.spyOn(navigator, "onLine", "get").mockReturnValueOnce(false);
```
MarcMcIntosh, line 215: hmm, that's an odd edge case, as they could have a valid login token but there's no why to check it

packages/slice-machine/src/components/SideNav/SideNav.css.ts
```diff
@@ -316,7 +316,7 @@ export const rightElementText = style([
     alignSelf: "baseline",
     fontSize: "12px",
     lineHeight: "16px",
-    maxWidth: "68px",
+    maxWidth: "82px",
```
MarcMcIntosh, line 319: That's pretty big right?

xrutayisire: It's for the text "Login required" so the ellipsis don't trigger.
I tested it for the version ad it works well with our alpha one

packages/slice-machine/components/AppLayout/Navigation/index.test.tsx
xrutayisire: True, but I did the same stuff as we are doing in the Changes page, so if it's offline you cannot see the page.

packages/slice-machine/components/AppLayout/Navigation/ChangesRightElement.tsx
xrutayisire: What is the best for you, I used it from there at the beginning but since it's coming from useNetwork, I kept it like this.
When we will rework useUnsyncedChanges it could simplify the hook. Each hook should have separate concern and isOnline don't look too much related. WDYT?

MarcMcIntosh: Is there rework to be done on useUnsyncedChanges?.... that may change things

MarcMcIntosh (approved):

packages/slice-machine/components/AppLayout/Navigation/ChangesRightElement.tsx
xrutayisire: Yes for me, we should try to not mix everything, there is no direct link between useUnsyncedChanges and useNetwork. It create a bit of a all in one hook.

xrutayisire: But we will see later on :)

PR #1068, opened 2023-07-25
feat(HoverCard): Convert VideoPopover to HoverCard and support both Video or Image

xrutayisire:
## Context

- Complete DT-1529

**PR is big only because of renaming that git was not able to get, creating deleted files and new files for nothing. PR is about 200 line changes tops** 

## The Solution

- Rename "HoverCard" to "BaseHoverCard" as it cannot be used directly but needs to be used within another component
- Rename "VideoPopover" to "HoverCard" that will support both Video and Image media type
- Use HTML img component for the Image because editor-ui one don't match our needs

## Impact / Dependencies

- Visual of existing VideoPopover is changed according to our needs
- Current Figma is not yet up-to-date, Benjamin will take care of that, and so he will do a design review to ensure its pixel perfect

- [x] The CI is successful.

## Preview

Screenshot 2023-07-25 at 15 21 52


packages/slice-machine/src/components/HoverCard/HoverCard.test.tsx
```diff
@@ -1,24 +1,129 @@
 // @vitest-environment jsdom
+import React from "react";
 import { describe, test, expect, vi } from "vitest";
-import { render, screen } from "../../../test/__testutils__";
-import { HoverCard } from "./HoverCard";
 
-// TBD: it's odd that this needs to be included in this test
+import { screen, render, waitFor } from "../../../test/__testutils__"; // skip usinng this
```
MarcMcIntosh: Which one?

MarcMcIntosh (approved):
Approved, left a question about one of the comments though.

packages/slice-machine/src/components/HoverCard/HoverCard.test.tsx
xrutayisire: Which one what?  🤔

It's your comment 😛

I will remove it, for me, we should not add comment like this we never know why it's for after :/

Screenshot 2023-07-25 at 16 20 30

xrutayisire: Removed 🙂

MarcMcIntosh: oh, I remember now... it's because the component doesn't need to have the initial state setup :)

xrutayisire: I see, but it's not a big problem, no?

PR #1079, opened 2023-07-28
fix(HoverCard): Fix HoverCard display to prevent having more than one on screen

xrutayisire:
## Context

- For DT-1493 we want two changes:
  - Fix tooltips closing when clicking outside
  - Fix tooltips being visible all at the same time

## The Solution

We define an order for the tooltip
- 1st that you should see is the Tutorial tooltip when you open SM the first time
- 2nd is the Simulator tooltip when you open a Slice
- 3rd is the Changes tooltip when you save a Custom type or a Slice.

With that order, I can prevent a Tooltip to be displayed if the previous one was not displayed yet
- Simulator tooltip is only visible if the user already saw Tutorial one. After closing the Tutorial tooltip and if you're still on Slice page, you will see the Stimulator tooltip after 5s.
- Changes tooltip is only visible if the user already saw the Tutorial one AND the Simulator one. After closing the Tutorial tooltip and also Simulator tooltip, you will see the Changes tooltip after 1.5s.

## Impact / Dependencies

- You have to open a Slice at least one time to be able to see the Changes tooltip

- [ ] The CI is successful.

## Preview

https://github.com/prismicio/slice-machine/assets/19946868/518ecf33-d4a6-46b6-b8a0-e75352f3f51d


packages/slice-machine/components/AppLayout/Navigation/ChangesListItem.tsx
```diff
@@ -36,7 +40,7 @@ export const ChangesListItem = forwardRef<HTMLLIElement, ChangesListItemProps>(
     return (
       <HoverCard
         open={open}
-        openDelay={1500}
+        openDelay={3000}
```
MarcMcIntosh, line 43: This is intentional right?

packages/slice-machine/components/AppLayout/Navigation/ChangesListItem.tsx
```diff
@@ -67,12 +71,19 @@ export const ChangesListItem = forwardRef<HTMLLIElement, ChangesListItemProps>(
 );
 
 const useOpenChangesHoverCard = () => {
```
MarcMcIntosh, line 73: This is good, it does what it needs to do.

What I'm wondering though is should we make this generic enough to be reused by the other or future hover cards by returning all the boolean properties in an object?

packages/slice-machine/src/components/HoverCard/HoverCard.stories.tsx
```diff
@@ -63,33 +63,3 @@ export const Video = {
     ),
   },
 } satisfies Story;
-
-export const CloseWhenClickingOutSide = {
```
MarcMcIntosh, line 67: Is this no longer a requirement?

MarcMcIntosh (commented):
Good work,, I have a few questions though :)

packages/slice-machine/src/components/HoverCard/HoverCard.stories.tsx
xrutayisire: Yep finally :/

https://prismic-team.slack.com/archives/C02L3FN3AJK/p1690529552348829

packages/slice-machine/components/AppLayout/Navigation/ChangesListItem.tsx
xrutayisire: Yes: https://prismic-team.slack.com/archives/C02L3FN3AJK/p1690534832385109?thread_ts=1690529552.348829&cid=C02L3FN3AJK

packages/slice-machine/components/AppLayout/Navigation/ChangesListItem.tsx
xrutayisire: Hum maybe, but for now this is the first and only tooltip that wait for something to be displayed. Other one are just if there is false in LS, display it.

So I guess it would be hard to make it generic now. WDYT?

MarcMcIntosh: I think if we went that way it would mostly look like a wrapper or util for getting the `hasSeen...` from redux. Not need for now so I'll resolve this one :)

MarcMcIntosh (approved):

PR #1087, opened 2023-08-01
fix(adapter-nuxt2): Nuxt 2 doesn't support subpath exports correctly

lihbr:
This PR fixes subpath exports for Nuxt 2 which doesn't handle them (legacy framework).

See:
- Slack thread: https://prismic-team.slack.com/archives/C014VAACCQL/p1690800041933869
- Forum thread: https://community.prismic.io/t/missing-dependencies-nuxt-2-slicemachine-adapter-nuxt2/13536

## The Solution

We change the import of `@prismicio/slice-simulator/kit` to its CommonJS equivalent. This is not pretty but I think it's good enough since it only impacts Nuxt 2.

An alternative would be to inline `@prismicio/slice-simulator` inside `@slicemachine/adapter-nuxt2`. But I think this makes less sense.

- [x] The CI is successful.

### DIFF
--- a/packages/adapter-nuxt2/src/simulator/SliceSimulator.ts
+++ b/packages/adapter-nuxt2/src/simulator/SliceSimulator.ts
@@ -2,6 +2,12 @@ import Vue, { PropType, VNodeChildren } from "vue";
 import { CreateElement, ExtendedVue } from "vue/types/vue";
 
 import {
+	SliceSimulatorState,
+	SliceSimulatorOptions,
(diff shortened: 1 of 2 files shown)


angeloashmore:
LGTM! Since this is a Nuxt 2-specific issue, let's keep the solution simple.

As an alternative solution, we could add a root-level `kit.js` file to `@prismicio/simulator` that re-exports files from `dist/kit`. However, `exports` is the future, and there isn't a strong reason to support older bundlers like Webpack 4.

PR #1088, opened 2023-08-03
feat(plugin-kit): add file system plugin helpers

angeloashmore:
## Types of changes

## Description

This PR adds a new entry to `@slicemachine/plugin-kit`: `@slicemachine/plugin-kit/fs`.

`@slicemachine/plugin-kit/fs` exports helpers for file system-based plugins (and, by extension, adapters). They are suited for plugins like `@slicemachine/adapter-next` and `@slicemachine/adapter-nuxt`.

Existing and new file system-based adapters should be written with the new helpers to promote code reuse and minimize bugs. All helpers are fully tested.

### Included helpers

The following helpers are included:

- `buildCustomTypeDirectoryPath`
- `buildCustomTypeFilePath`
- `buildCustomTypeLibraryDirectoryPath`
- `buildSliceDirectoryPath`
- `buildSliceFilePath`
- `buildSliceLibraryDirectoryPath`
- `checkHasProjectFile`
- `checkIsTypeScriptProject`
- `deleteAllCustomTypeFiles`
- `deleteAllSliceFiles`
- `deleteCustomTypeFile`
- `deleteProjectFile`
- `deleteSliceFile`
- `readCustomTypeFile`
- `readCustomTypeLibrary`
- `readCustomTypeModel`
- `readProjectFile`
- `readSliceFile`
- `readSliceLibrary`
- `readSliceModel`
- `renameSlice`
- `upsertGlobalTypeScriptTypes`
- `writeCustomTypeFile`
- `writeCustomTypeModel`
- `writeProjectFile`
- `writeSliceFile`
- `writeSliceModel`

## Checklist:

- [x] All [TSDoc](https://tsdoc.org) comments are up-to-date and new ones have been added where necessary.

🐤


packages/plugin-kit/src/createSliceMachineHelpers.ts
```diff
@@ -99,6 +99,6 @@ export class SliceMachineHelpers {
 	};
 
 	joinPathFromRoot = (...paths: string[]): string => {
-		return path.join(this._project.root, ...paths);
+		return path.resolve(this._project.root, ...paths);
```
lihbr: Just curious about the context behind this change? (because now the method name is counterintuitive as it's now resolving path from root)

packages/plugin-kit/src/fs/buildSliceDirectoryPath.ts
```diff
@@ -0,0 +1,26 @@
...
+	BuildSliceLibraryDirectoryPathArgs,
+} from "./buildSliceLibraryDirectoryPath";
+
+export type BuildSliceDirectoryPathArgs = BuildSliceLibraryDirectoryPathArgs &
+	ResolveSliceModelArgs;
+
+export async function buildSliceDirectoryPath(
+	args: BuildSliceDirectoryPathArgs,
+): Promise<string> {
+	const model = await resolveSliceModel(args);
```
lihbr, line 20: Is there really a need to abstract the `async` case? (model is not provided and has to be resolved)

It's fine-ish, I just find it counterintuitive for this method to be async. I get the DX benefit here, but could perhaps lead to performance issue or infinite recursion (e.g. `slice:read` relies on this method without providing the whole model)

packages/plugin-kit/src/fs/checkIsTypeScriptProject.ts
```diff
@@ -0,0 +1,21 @@
...
+	args: CheckIsTypeScriptProjectArgs,
+): Promise<boolean> => {
+	try {
+		await fs.access(args.helpers.joinPathFromRoot(TSCONFIG_FILENAME));
+
+		return true;
+	} catch {
+		return false;
+	}
+};
```
lihbr: Can't this reuse `checkHasProjectFile()` ?

packages/plugin-kit/src/fs/buildSliceDirectoryPath.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import {
+	buildSliceLibraryDirectoryPath,
+	BuildSliceLibraryDirectoryPathArgs,
+} from "./buildSliceLibraryDirectoryPath";
+
+export type BuildSliceDirectoryPathArgs = BuildSliceLibraryDirectoryPathArgs &
+	ResolveSliceModelArgs;
+
+export async function buildSliceDirectoryPath(
```
lihbr: Any reason this is not an arrow function?

packages/plugin-kit/src/fs/checkHasProjectFile.ts
```diff
@@ -0,0 +1,22 @@
...
+): Promise<boolean> {
+	const filePath = args.helpers.joinPathFromRoot(args.filename);
+
+	try {
+		await fs.access(filePath);
+
+		return true;
+	} catch {
+		return false;
+	}
```
lihbr: Should we use `./lib/checkPathExists` here?

packages/plugin-kit/src/fs/readProjectFile.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export async function readProjectFile(
+	args: ReadProjectFileArgs,
+): Promise<Buffer>;
+export async function readProjectFile(
+	args: ReadProjectFileArgs & { encoding: BufferEncoding },
+): Promise<string>;
+export async function readProjectFile(
+	args: ReadProjectFileArgs & { encoding?: BufferEncoding },
+): Promise<Buffer | string> {
```
lihbr, line 20: Any thoughts on using the arrow function overload syntax instead for those helpers?

```typescript
export const readProfileFile: {
	(args: ReadProjectFileArgs): Promise<Buffer>;
	(args: ReadProjectFileArgs & { encoding: BufferEncoding }): Promise<string>;
} = async(
	args: ReadProjectFileArgs & { encoding?: BufferEncoding },
): Promise<Buffer | string> {}
```

We did so on client helpers: https://github.com/prismicio/prismic-client/blob/76fa2ca4e207d70fa3a19f131d9a416a9dfb180a/src/helpers/asText.ts

packages/plugin-kit/src/fs/readCustomTypeLibrary.ts
```diff
@@ -0,0 +1,74 @@
...
+					} catch {
+						// noop
+					}
+				}
+			}
+		}),
+	);
+
+	return {
+		ids: ids.sort(),
```
lihbr, line 85: Should this be `customTypeIDs` to align with `readSliceLibrary()`?

packages/plugin-kit/src/fs/writeProjectFile.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+export async function writeProjectFile(
+	args: WriteProjectFileArgs,
+): Promise<string> {
+	const filePath = args.helpers.joinPathFromRoot(args.filename);
+
+	let contents = args.contents;
+
+	if (args.format) {
+		contents = await args.helpers.format(contents.toString(), filePath);
```
lihbr: What if we need to provide additional options to Prettier? For example I had(? not sure if I had to) to provide the Vue parser to format Vue files: https://github.com/prismicio/slice-machine/blob/c3a5117c0d35e86020df6b5d16dd600e4231fda3/packages/adapter-nuxt/src/hooks/project-init.ts#L144-L146

Could just be a `prettierOptions` optional argument 🤔

packages/plugin-kit/src/fs/deleteAllCustomTypeFiles.ts
```diff
@@ -0,0 +1,24 @@
+import {
+	buildCustomTypeDirectoryPath,
+	BuildCustomTypeDirectoryPathArgs,
+} from "./buildCustomTypeDirectoryPath";
+import { deleteProjectFile } from "./deleteProjectFile";
+
+export type DeleteAllCustomTypeFilesArgs = BuildCustomTypeDirectoryPathArgs;
+
+export async function deleteAllCustomTypeFiles(
```
lihbr: Since it also deletes the directory, maybe `deleteCustomTypeDirectory` is clearer? Naming confused me at first.

packages/plugin-kit/src/fs/deleteAllSliceFiles.ts
```diff
@@ -0,0 +1,24 @@
+import {
+	buildSliceDirectoryPath,
+	BuildSliceDirectoryPathArgs,
+} from "./buildSliceDirectoryPath";
+import { deleteProjectFile } from "./deleteProjectFile";
+
+export type DeleteAllSliceFilesArgs = BuildSliceDirectoryPathArgs;
+
+export async function deleteAllSliceFiles(
```
lihbr: Same as `deleteAllCustomTypeFiles`, maybe `deleteSliceDirectory` is clearer?

packages/plugin-kit/src/fs/readCustomTypeLibrary.ts
```diff
@@ -0,0 +1,74 @@
...
+
+					try {
+						const modelContents = await readJSONFile(modelPath);
+
+						if (isCustomTypeModel(modelContents)) {
+							ids.push(modelContents.id);
+						}
+					} catch {
+						// noop
+					}
```
lihbr: Should it warn? The same way `readSliceModel()` warns?

packages/plugin-kit/src/fs/readSliceLibrary.ts
```diff
@@ -0,0 +1,66 @@
...
+
+				try {
+					const modelContents = await readJSONFile(modelPath);
+
+					if (isSharedSliceModel(modelContents)) {
+						sliceIDs.push(modelContents.id);
+					}
+				} catch {
+					// noop
+				}
```
lihbr: Should it warn? The same way `readSliceModel()` warns? Maybe not as this prohibits arbitrary directory 🤔

packages/plugin-kit/test/fs/writeCustomTypeModel.test.ts
```diff
@@ -0,0 +1,68 @@
...
+		format: true,
+		helpers: ctx.pluginRunner.rawHelpers,
+	});
+
+	const contents = await fs.readFile(
+		path.join(ctx.project.root, "customtypes", model.id, "index.json"),
+		"utf8",
+	);
+
+	expect(contents).not.toBe(JSON.stringify(model));
```
lihbr: I think this is always true since the model is still "lightly formatted" before being written: https://github.com/prismicio/slice-machine/blob/aa/plugin-kit-fs-helpers/packages/plugin-kit/src/fs/writeCustomTypeModel.ts#L29C1-L29C1, I guess we're looking at `expect(contents).not.toBe(JSON.stringify(model, null, 2);`

Might need to test with a dummy prettier config (e.g. tab instead of spaces)

lihbr (approved):
Really great job! I like it. Excellent simplification of existing tests also!

Left a few notes, but nothing major. Here are general ones:
- Looks like we started to mix functions/arrow functions, is there a reason for that? Some helpers are functions, some are arrow functions 🤔
- We have a `renameSlice` helper but not a `renameCustomType` helper, we don't have a need for it, right?

packages/plugin-kit/src/createSliceMachineHelpers.ts
angeloashmore: Oops, that was accidentally left in while debugging some path issues. It should still be `join`.

packages/plugin-kit/src/fs/buildSliceDirectoryPath.ts
angeloashmore: In most cases, we only provide a Slice ID to hooks, which means we need to load the model and derive a path using the Slice's name. We need to do this with Slices because their directory names are based on the name, not the ID.

packages/plugin-kit/src/fs/checkIsTypeScriptProject.ts
angeloashmore: Yep, it can. Good catch; this helper was written before `checkHasProjectFile()` was written.

packages/plugin-kit/src/fs/checkHasProjectFile.ts
angeloashmore: Yes, we should!

packages/plugin-kit/src/fs/readProjectFile.ts
angeloashmore: I gave it a try and ran into type issues with the return type. It doesn't seem to be an issue with `asText()` since the return types are the same across each overload.

packages/plugin-kit/src/fs/readCustomTypeLibrary.ts
angeloashmore: `ids` was selected to align with `custom-type-library:read`'s return type. We used `ids` because there is no ambiguity in what `ids` refers to.

In `slice-library:read`'s return type, `sliceIDs` is namespaced because `id` (the Slice library's ID) is also included.

packages/plugin-kit/src/fs/buildSliceDirectoryPath.ts
angeloashmore: No reason - everything's an arrow function now! (Except the function overloads mentioned in other comments.)

packages/plugin-kit/src/fs/writeProjectFile.ts
angeloashmore: Done! Added an optional `formatOptions` argument that accepts `format`'s third argument.

packages/plugin-kit/src/fs/deleteAllCustomTypeFiles.ts
angeloashmore: Works for me!

packages/plugin-kit/test/fs/writeCustomTypeModel.test.ts
angeloashmore: I'll finish this fix tomorrow!

packages/plugin-kit/src/fs/readProjectFile.ts
lihbr: Alright, no worries, works for me :)

packages/plugin-kit/src/fs/readCustomTypeLibrary.ts
lihbr: OK, works for me

packages/plugin-kit/test/fs/writeCustomTypeModel.test.ts
angeloashmore: Fixed in https://github.com/prismicio/slice-machine/pull/1088/commits/d2e0f7ebb04e1692aeaf01787e03402fc4bbdaad. Thanks for the suggestion!

angeloashmore:
> @lihbr: We have a renameSlice helper but not a renameCustomType helper, we don't have a need for it, right?

We technically don't need it since it would only re-write a custom type's model, which can be accomplished with `writeCustomTypeModel()`. However, it would be good to include for consistency with the Slice helpers, and in the future, we may want to adjust how and where those files are saved.

I added `renameCustomType()` as effectively an alias here: https://github.com/prismicio/slice-machine/pull/1088/commits/a5dacd829895280749d010ac624afe31878d774c

packages/plugin-kit/src/fs/readCustomTypeLibrary.ts
angeloashmore: Good idea. The same mechanism from `readSliceModel()` was transferred into `readCustomTypeLibrary()` and `readSliceLibrary()`.

angeloashmore:
@lihbr The PR should be ready to merge now. Could you give it a final look to make sure I applied your feedback correctly?

lihbr (approved):
All good :)

PR #1089, opened 2023-08-03
feat(plugin-kit): `updateSliceMachineConfig()` helper

lihbr:
## The Solution

This PR introduces a standard `updateProjectConfig` helper on the Plugin Kit, allowing adapters to update Slice Machine configuration upon init.

This enables adapters to insert their Slice Simulator URL path, not requiring users to do it manually anymore, therefore reducing the Time To First Slice (cf. OKRs).

Additional context: https://www.notion.so/prismic/Adapter-Slice-Simulator-Config-e594f57a83274afd800684bdaf4df4a0?pvs=4

## Impact / Dependencies

Adapters have been refactored to use the standardized helper.

- [x] The CI is successful.

HennahaneThurstableGIF

### DIFF
--- a/packages/plugin-kit/src/createSliceMachineHelpers.ts
+++ b/packages/plugin-kit/src/createSliceMachineHelpers.ts
@@ -6,7 +6,7 @@ import { stripIndent } from "common-tags";
 
 import { decodeSliceMachineConfig } from "./lib/decodeSliceMachineConfig";
 
-import { SliceMachineProject } from "./types";
+import { SliceMachineConfig, SliceMachineProject } from "./types";
 
 type FormatOptions = {
 	prettier?: prettier.Options;
@@ -72,6 +72,28 @@ export class SliceMachineHelpers {
 			...this._project,
 			config: sliceMachineConfig,
 		};
+	};
+
+	updateProjectConfig = async (
+		config: SliceMachineConfig,
+		format?: boolean,
+	): Promise<void> => {
+		const { value: sliceMachineConfig, error } =
+			decodeSliceMachineConfig(config);
+
+		if (error) {
+			// TODO: Write a more friendly and useful message.
+			throw new Error(`Invalid config provided. ${error.errors.join(", ")}`);
+		}
+
+		const configFilePath = this.joinPathFromRoot("slicemachine.config.json");
+		let content = JSON.stringify(sliceMachineConfig);
+
+		if (format) {
+			content = await this.format(content, configFilePath);
+		}
+
+		await fs.writeFile(configFilePath, content);
 	};
 
 	format = async (
--- a/packages/plugin-kit/src/createSliceMachineHelpers.ts
+++ b/packages/plugin-kit/src/createSliceMachineHelpers.ts
@@ -8,6 +8,10 @@ import { decodeSliceMachineConfig } from "./lib/decodeSliceMachineConfig";
 
 import { SliceMachineConfig, SliceMachineProject } from "./types";
 
+type UpdateSliceMachineConfigOptions = {
+	format?: boolean;
+};
+
 type FormatOptions = {
 	prettier?: prettier.Options;
 	/**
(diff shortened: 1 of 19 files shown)


packages/plugin-kit/src/createSliceMachineHelpers.ts
```diff
@@ -74,6 +74,28 @@ export class SliceMachineHelpers {
...
+		const { value: sliceMachineConfig, error } =
+			decodeSliceMachineConfig(config);
+
+		if (error) {
+			// TODO: Write a more friendly and useful message.
+			throw new Error(`Invalid config provided. ${error.errors.join(", ")}`);
+		}
+
+		const configFilePath = this.joinPathFromRoot("slicemachine.config.json");
+		let content = JSON.stringify(sliceMachineConfig);
```
angeloashmore: Let's use something that looks nicer, like `JSON.stringify(sliceMachineConfig, null, 2)`.

packages/plugin-kit/src/createSliceMachineHelpers.ts
```diff
@@ -74,6 +74,28 @@ export class SliceMachineHelpers {
 		};
 	};
 
+	updateProjectConfig = async (
+		config: SliceMachineConfig,
```
angeloashmore: What do you think of accepting a partial object, which would let adapters specify patches to the config rather than replace the whole thing?

For example:

```ts
context.actions.updateProjectConfig({
  localSliceSimulatorURL: "http://localhost:3000/slice-simulator",
});
```

Rather than:

```ts
const project = context.actions.getProject();

project.config.localSliceSimulatorURL = "http://localhost:3000/slice-simulator";

context.actions.updateProjectConfig(project.config)
```

If we go this route, I think we should avoid `defu` since it tries to do too much and adds complexity. We could write something that simply merges nested object properties, completely overriding things like functions and arrays. WDYT?

packages/plugin-kit/test/SliceMachineHelpers-getProject.test.ts
```diff
@@ -35,3 +35,43 @@ it("returns Slice Machine project metadata", async () => {
 
 	await fs.rm(project.root, { recursive: true });
 });
+
+it("throws on not found config", async () => {
```
angeloashmore: Small nit pick: "throws when a config cannot be found"

angeloashmore (commented):
It looks good to me! I suggested some changes, but the suggestions are up for discussion and don't necessarily need to be accepted.

`updateProjectConfig()` in its current form works for me.

Edit: forgot to mention that `updateSliceMachineConfig()` may be more accurate, as that is how we refer to it in the manager and types.

packages/plugin-kit/src/createSliceMachineHelpers.ts
```diff
@@ -74,6 +74,28 @@ export class SliceMachineHelpers {
...
+		format?: boolean,
```
angeloashmore: Rather than using a positional argument, which doesn't scale to more arguments, we could use named arguments as the second argument:

```ts
type UpdateProjectConfigConfig = {
  format?: boolean
}

// ...

updateProjectConfig(
  sliceMachineConfig: SliceMachineConfig,
  config: UpdateProjectConfigConfig = {}
) {
  // ...
}
```

packages/plugin-kit/src/createSliceMachineHelpers.ts
lihbr: But we don't have any object properties apart from a potential adapter config? So everything should be overridden? Gave it a simple go, let me know.

Similarly, the above wouldn't work as simply because we have to check if there's no `localSliceSimulatorURL` before adding a default.

lihbr: Eventually decided against that for now:
- There's no clear/convenient way of merging the config correctly as you might want to remove a value, add one in an array, filter them.
- Because of the above, you still have to read the project on the plugin's end to decide on what patches you want to make.

lihbr:
Should be good, thanks for the review!

packages/plugin-kit/src/createSliceMachineHelpers.ts
angeloashmore: Makes sense to me! Using the current implementation still lets someone use a lib like defu internally (though that would be unusual).

angeloashmore:
LGTM! 🐎

PR #1090, opened 2023-08-03
Refactor Screenshot DropZone behaviour + paste images

hypervillain:
Completes https://linear.app/prismic/issue/DT-1514/aauser-i-can-paste-from-clipboard-a-screenshot-taken-manually-with-my


packages/slice-machine/components/ChangesItems/ChangesItems.tsx
```diff
@@ -33,6 +33,10 @@ export const ChangesItems: React.FC<ChangesItemsProps> = ({
 
   const { sliceFilterFn, defaultVariationSelector } = modalPayload;
 
+  const screenshotChangesSlices = unSyncedSlices.filter(
```
hypervillain, line 36: this is a fix for an error I encountered during coding: users were able to access screenshot modal of a deleted slice, and therefore could not add a screenshot. This removes deleted slices from the list of slices to display

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
```diff
@@ -7,17 +7,39 @@ import { AiOutlinePicture } from "react-icons/ai";
...
+      fontSize: "12px",
+      lineHeight: "24px",
+      border: "1px solid #E4E2E4",
+      boxShadow: "0px 1px 0px 0px rgba(0, 0, 0, 0.04)",
+      width: "100%",
+      height: "40px",
+    }}
+  >
+    <FigmaIcon />
+    Use cmd+shift+c to copy any frame as .png, then just paste it here
```
xrutayisire: This is only for Mac, we need to adapt depending on Windows / Linux / OSX

packages/slice-machine/components/ScreenshotChangesModal/VariationDropZone.tsx
```diff
@@ -0,0 +1,251 @@
...
+  variationID,
+  slice,
+  imageTypes = acceptedImagesTypes,
+}) => {
+  const maybeScreenshot = slice.screenshots[variationID];
+
+  const { openToaster } = useSliceMachineActions();
+  const [isDragActive, setIsDragActive] = useState(false);
+  const [isHover, setIsHover] = useState(false);
+  useHotkeys("meta+v", () => void handlePaste(), []);
```
xrutayisire: Did you check on Windows / Linux?

Is the meta key this: https://www.w3schools.com/jsref/event_key_metakey.asp#:~:text=On%20Windows%20the%20META%20key,the%20Cmd%20(%E2%8C%98)%20key.

If yes, then on Windows it will be the Window key and it's not how users paste something :/

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
```diff
@@ -143,7 +165,11 @@ const ScreenshotChangesModal = ({
   useEffect(() => {
     setVariationSelector(variationSetter(defaultVariationSelector, slices));
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [defaultVariationSelector, isOpen]);
+  }, [
+    defaultVariationSelector?.sliceID,
```
xrutayisire, line 192: You still need that at the end?

packages/slice-machine/package.json
```diff
@@ -32,6 +32,7 @@
   },
   "dependencies": {
     "@slicemachine/manager": "workspace:^",
+    "react-hotkeys-hook": "4.4.1",
```
xrutayisire: We should not add it to dependencies, I guess. It's something that is internal for us to manage. Then it's like any other dependencies we have in devDependencies. If it's here, it could be mixed with our user's deps.

Maybe @bapmrl you can confirm on that since you did the changes on that part.

bapmrl: Yes, it should be in `devDependencies` like 99% of the other ones ^^

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
hypervillain: yes, for whatever reason deep state reloads happen some time to time, resetting the object passed and therefore resetting forcefully the tab

xrutayisire: Ok when refactoring the whole modal I guess it will not be useful anymore or at least not like this

packages/slice-machine/components/ScreenshotChangesModal/VariationDropZone.tsx
```diff
@@ -0,0 +1,251 @@
...
+  variationID,
+  slice,
+  imageTypes = acceptedImagesTypes,
+}) => {
+  const maybeScreenshot = slice.screenshots[variationID];
+
+  const { openToaster } = useSliceMachineActions();
+  const [isDragActive, setIsDragActive] = useState(false);
+  const [isHover, setIsHover] = useState(false);
+  useHotkeys(["meta+v", "ctrl+v"], () => void handlePaste(), []);
```
xrutayisire, line 88: Problem is now it's also listening to control+v on OSX and that's not what we want, no?

xrutayisire: Same for Windows, listening to window+v

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
```diff
@@ -7,17 +7,39 @@ import { AiOutlinePicture } from "react-icons/ai";
...
 import { ComponentUI } from "@lib/models/common/ComponentUI";
 import useSliceMachineActions from "@src/modules/useSliceMachineActions";
 
+import VariationDropZone from "./VariationDropZone";
+
+import { FigmaIcon } from "@src/icons/FigmaIcon";
+
 export type SliceVariationSelector = { sliceID: string; variationID: string };
 
+const FigmaTip = () => (
```
xrutayisire: So it's not a link anymore, right? Maybe design needs to be updated?

packages/slice-machine/components/ScreenshotChangesModal/VariationDropZone.tsx
hypervillain: I prefer to not add a conditional for this, I think it's perfectly fine to listen to ctrl+v too on OSX and vice versa

packages/slice-machine/src/components/Kbd/Kbd.tsx
```diff
@@ -0,0 +1,7 @@
+import { FC, ReactNode } from "react";
+
+import * as styles from "./Kbd.css";
+
+export const Kbd: FC<{ children: ReactNode }> = ({ children }) => (
```
xrutayisire:
```suggestion
export const Kbd: FC<PropsWithChildren> = ({ children }) => (
```

packages/slice-machine/src/components/Kbd/Kbd.css.ts
```diff
@@ -0,0 +1,14 @@
+import { style } from "@vanilla-extract/css";
+import { colors, sprinkles } from "@prismicio/editor-ui";
+
+export const root = style([
+  sprinkles({
+    paddingInline: 4,
+    color: colors.grey11,
+  }),
+  {
+    border: "1px solid #E4E2E4",
```
xrutayisire: Ask @benjammartin if we can have a color from Design System for both color here.

packages/slice-machine/components/ScreenshotChangesModal/index.tsx
```diff
@@ -7,17 +7,62 @@ import { AiOutlinePicture } from "react-icons/ai";
...
+        lineHeight: "24px",
+        border: "1px solid #E4E2E4",
+        boxShadow: "0px 1px 0px 0px rgba(0, 0, 0, 0.04)",
+        width: "100%",
+        minHeight: "40px",
+      }}
+    >
+      <FigmaIcon />
+      <div>
+        Use&nbsp;{keys} to copy any frame as .png, then just paste it here
```
xrutayisire, line 60: The space after works but not before?

xrutayisire (approved):
Let's go!

PR #1092, opened 2023-08-04
refactor(black-slate): add spacious variation

MarcMcIntosh:
## Context
[Ticket](https://linear.app/prismic/issue/DT-1524/aadev-i-need-to-add-a-spacious-option-to-the-blankslate-component)

Adds a spacious variation to the blank slate component, this sets the width and height to 100%
completes DT-1524

Screenshot 2023-08-04 at 13 29 50

- [ ] The CI is successful.

### DIFF
--- a/packages/slice-machine/src/components/BlankSlate/BlankSlate.css.ts
+++ b/packages/slice-machine/src/components/BlankSlate/BlankSlate.css.ts
@@ -23,14 +23,33 @@ export const root = style([
   { maxWidth: 498, minHeight: 400 },
 ]);
 
-export const withBackground = style([
+export const large = style([
   sprinkles({
     textAlign: "center",
     width: "100%",
   }),
   {
-    backgroundSize: "cover",
     height: "50vh",
+    maxWidth: "100%",
+  },
+]);
+
+export const withBackground = style([
+  sprinkles({
+    textAlign: "center",
+  }),
+  {
+    backgroundSize: "cover",
+  },
+]);
+
+export const spacious = style([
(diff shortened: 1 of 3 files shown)


xrutayisire:
I'm not sure to get this issue and maybe you can help me.

When I read:
> This already has been implemented by @xavier.rutayisire, but I wonder if we could simplify the API by simply having to pass a variant or boolean to control how the BlankSlateImage is displayed — instead of having a redundant backgroundImage prop.

I feel like the goal is to remove the backgroundImage prop on the main component and instead keep using BlankSlateImage but by setting a variant spacious it would put the image in full width like visually it is today with my code.

@bapmrl maybe you could confirm or not 🙂

MarcMcIntosh:
@xrutayisire we discussed the options for removing the background prop during the daily and we decided that it would be simpler to keep it rather than iterating over the children to maybe find an image child, and take the image prop from that :)

xrutayisire:
Ok thanks @MarcMcIntosh.

I don't really understand the code:
- large is a variation?
- what's the difference visually between the two stories "Background" and "Spacious"
- what is the use case the change is for?

I feel like the code is a bit harder to understand now, with hasBackground, spacious, large, variation. Maybe I'm missing something.
If needed, we can call to sync 🙂

MarcMcIntosh:
Sure a call would be best :)

bapmrl:
Continued in https://github.com/prismicio/slice-machine/pull/1105. Thanks a lot Marc 😊

PR #1094, opened 2023-08-07
refactoring(editor): Update editor packages

xrutayisire:
## Context

- Completes DT-1461

## The Solution

- Update editor packages
- Update usage of ErrorBoundary
- Wrap SharedSliceEditor in a Suspense and ErrorBoundary
- Add status in editorConfig properties

## Impact / Dependencies

- Fix the upload button problem from DT-1527

- [ ] The CI is successful.

### DIFF
--- a/packages/slice-machine/components/Simulator/index.tsx
+++ b/packages/slice-machine/components/Simulator/index.tsx
@@ -1,5 +1,13 @@
-import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import {
+  Suspense,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
 import { EditorConfig, SharedSliceEditor } from "@prismicio/editor-fields";
+import { DefaultErrorMessage, ErrorBoundary } from "@prismicio/editor-ui";
 
 import { defaultSharedSliceContent } from "@src/utils/editor";
 
@@ -72,6 +80,8 @@ const Simulator: ComponentWithSliceProps = ({ slice, variation }) => {
       unsplash: {
         url: endpoints.PrismicUnsplash,
       },
+      // TODO(DT-1543): Remove status property when optional
+      status: "published",
     };
   }, [endpoints.PrismicOembed, endpoints.PrismicUnsplash]);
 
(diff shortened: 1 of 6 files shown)


packages/slice-machine/components/Simulator/index.tsx
```diff
@@ -301,21 +311,37 @@ const Simulator: ComponentWithSliceProps = ({ slice, variation }) => {
...
-                }}
-                sharedSlice={sharedSlice}
-              />
+              {/* 
+                The editor warn us it's recommended to wrap  SharedSliceEditor in a Suspense and an ErrorBoundary.
+                Warning: It is not recommended to use editor-support->Suspense without a editor-ui->ErrorBoundary above it 
+              */}
+              <ErrorBoundary
+                renderError={() => (
+                  <DefaultErrorMessage
```
bapmrl, line 320: ❓ Have we tested the `DefaultErrorMessage` fits nicely in the parent container/pane?

packages/slice-machine/src/features/customTypes/customTypesBuilder/PageSnippetDialog/PageSnippetDialog.tsx
```diff
@@ -78,9 +79,12 @@ export const PageSnippetDialog: FC<PageSnippetDialogProps> = ({ model }) => {
   return (
     <div>
       <ErrorBoundary
-        title="Request failed"
-        description="An error occurred while fetching page types snippets."
-        renderError={() => null}
+        renderError={() => (
+          <DefaultErrorMessage
```
bapmrl: ❓ Does it fit in the dialog? Before, we were apparently rendering `null`.

xrutayisire: Oh good one, I copy-paste the title and description, but it was returning null. I will just set an empty ErrorBoundary like you back button then 👍

packages/slice-machine/components/Simulator/index.tsx
xrutayisire: I'm not even sure what are the use cases for us to catch that here... Do you think we should just wrap without message, btw?

bapmrl (approved):

PR #1105, opened 2023-08-16
feat(slice-machine-ui): partially redesign CustomTypeBuilder

bapmrl:
Completes DT-1440, DT-1520, DT-1528 and DT-1543.


packages/slice-machine/components/Simulator/index.tsx
```diff
@@ -71,19 +71,14 @@ const Simulator: ComponentWithSliceProps = ({ slice, variation }) => {
...
-      },
+  const editorConfig: EditorConfig = useMemo(
+    () => ({
+      embedsUrl: endpoints.PrismicOembed,
       env: "prod",
-      unsplash: {
-        url: endpoints.PrismicUnsplash,
-      },
-      // TODO(DT-1543): Remove status property when optional
-      status: "published",
```
xrutayisire, line 84: 🙏 Thank you for this one

packages/slice-machine/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
```diff
@@ -155,69 +156,62 @@ const SliceZone: React.FC<SliceZoneProps> = ({
...
-          authStatus={authStatus}
-          isOnline={isOnline}
-          format={customType.format}
-        />
-      )}
+              ) : undefined}
+            </>
+          }
+          toggle={
+            customType.format !== "page" || customType.tabs[0].key !== tabId ? (
```
xrutayisire, line 179: ❓You changed the condition here, right? Before it was always possible to remove the slice zone, even the first tab. I think what you did is good but prefer to ask

packages/slice-machine/lib/builders/CustomTypeBuilder/TabModal/update.tsx
```diff
@@ -86,77 +44,44 @@ const CreateCustomtypeForm = ({
         }
       }}
       content={{
-        title,
+        title: "Update Tab",
```
xrutayisire: ⚠️ I know it's not yet the new modal but Since the label is Rename should be put Rename Tab, before Edit / Update was for both action, now it's only Renaming.

packages/slice-machine/lib/builders/CustomTypeBuilder/index.tsx
```diff
@@ -1,28 +1,143 @@
...
+                          tabKey: tab.key,
+                        });
+                      }}
+                      startIcon={<Icon name="edit" />}
+                    >
+                      Rename
+                    </DropdownMenuItem>
+                    <DropdownMenuItem
+                      color="tomato"
+                      disabled={customType.tabs.length < 2}
```
xrutayisire: ℹ️ Just so you know, personally I prefer these conditions:

- `customType.tabs.length <= 1`
- `customType.tabs.length === 1`

Just because when I read 2, I'm always like what what, why 2 haha

packages/slice-machine/lib/builders/CustomTypeBuilder/index.tsx
```diff
@@ -1,28 +1,143 @@
...
+                customType={customType}
+                fields={tab.value}
+                sliceZone={tab.sliceZone}
+                tabId={tab.key}
+              />
+            </WindowTabsContent>
+          ))}
+        </WindowTabs>
+      </Window>
+      {dialog?.type === "CREATE_CUSTOM_TYPE" ? (
```
xrutayisire, line 96: ❓ You prefer to do `myVar ? JSX : undefined` vs `myVar && JSX` right?

packages/slice-machine/src/components/Window/Window.css.ts
```diff
@@ -0,0 +1,156 @@
+import { colors, sprinkles, vars } from "@prismicio/editor-ui";
```
xrutayisire, line 1: ℹ️ I will not really get into the specific of this file 😅 If you feel I should for the review let me know.

packages/slice-machine/src/components/Window/Window.tsx
```diff
@@ -0,0 +1,105 @@
+import {
```
xrutayisire, line 1: 👏 Great job for all this Window component, amazing work from all!

xrutayisire (commented):
Really great work! 👏

packages/slice-machine/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
bapmrl: Yes 🙂

It's what Côme asked in https://linear.app/prismic/issue/DT-1528/aauser-i-expect-the-window-and-list-components-to-be-used-on-the:
> Ensure the switch for toggling the Slice Zone is always present, except in the main tab of Page Types.

packages/slice-machine/lib/builders/CustomTypeBuilder/index.tsx
bapmrl: I do ^^ I don't really like "too smart" expressions, especially when `myVar` is a `string` for instance. Also, I prefer to explicitly state what value should be returned when `myVar` is `false`.

packages/slice-machine/src/components/Window/Window.css.ts
bapmrl: I think it's okay. Let's put more emphasis on testing it in multiple browsers 😊

packages/slice-machine/src/components/Window/Window.tsx
bapmrl: Thank you Xavier 😊

packages/slice-machine/lib/builders/CustomTypeBuilder/index.tsx
xrutayisire: Good to know, maybe it could be in the code guidelines? It's simple and for example I would like to do like you. I don't have a strong preference, and it would keep the code coherent. WDYT?

packages/slice-machine/lib/builders/CustomTypeBuilder/TabModal/update.tsx
bapmrl: I've made the change 🙂

packages/slice-machine/lib/builders/CustomTypeBuilder/index.tsx
bapmrl: I hesitated and agree it's better to use `<=`. I've made the change.

packages/slice-machine/lib/builders/CustomTypeBuilder/index.tsx
bapmrl: What do you think about this @AlexGalays?

PR #1106, opened 2023-08-16
build(adapter-nuxt2): change bundling strategy to fully support webpack 4

lihbr:
## Context

This is a follow-up PR to #1087. It appears changing the import is not enough to fully fix the issue with Nuxt 2 in all scenario, especially with Nuxt v2.17 (latest version to date of Nuxt 2).

## The Solution

This PR changes the fix strategy to inlining `@prismicio/simulator`, which is not ideal but good enough for Nuxt 2 which will reach its EOL at the end of the year.

<details>
<summary>With the previous PR, we're stuck in that kind of back & forth loop with Nuxt v2.17</summary>

image
image

</details>

- [x] The CI is successful.


angeloashmore:
Looks good to me! Inlining `@prismicio/simulator` shouldn't have a significant negative effect since it would be included in the bundle anyway.

PR #1120, opened 2023-08-29
feat(SlicesTemplates): Update slice:create hook to accept a custom co…

xrutayisire:
## Context

- Completes DT-1573

## The Solution

The goal is to let the "slice:create" hook accept a new argument "contents" so it will be created with that instead of the default string there is.

It's the responsibility of the caller to give the correct JS or TS version

## Impact / Dependencies

- Not a breaking change

- [ ] The CI is successful.

### DIFF
--- a/packages/adapter-next/src/hooks/slice-create.ts
+++ b/packages/adapter-next/src/hooks/slice-create.ts
@@ -32,15 +32,17 @@ const createComponentFile = async ({
 	const filename = `index.${extension}`;
 	const pascalName = pascalCase(data.model.name);
 
-	let contents: string;
+	let componentContents: string;
 
 	const isTypeScriptProject = await checkIsTypeScriptProject({
 		helpers,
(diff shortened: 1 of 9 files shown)


angeloashmore:
@xrutayisire What do you think of renaming `contents` to `componentContents`? "contents" is ambiguous and could refer to any type of content: the model's content, mock data content, etc.

xrutayisire:
> @xrutayisire What do you think of renaming contents to componentContents? "contents" is ambiguous and could refer to any type of content: the model's content, mock data content, etc.

@angeloashmore, good idea 👍

Btw, At the end with Hugo after lonnnnng discussions we came up with a good way to handle slices templates we think. Taking care of screenshots, mocks also at creation time.
Hugo should open a PR tomorrow, so you will be able to see the changes next work day for you, don't hesitate to comment 🙂

PR #1125, opened 2023-09-01
feat(adapter-nuxt): move or delete default `app.vue`

lihbr:
## Context

When creating a brand new Nuxt 3 project, the template includes a default `app.vue` that disables Nuxt's pages directory, not allowing users to use the simulator without additional setup. This PRs solves that by moving or removing the default `app.vue`

## The Solution

- If there's a default `app.vue` ...
   - ... and no index page ...
      - ... then move `app.vue` to `pages/index.vue`
   - ... and an index page ...
      - ... then delete `app.vue`

## Impact / Dependencies

N/A

## QA

1. Create a brand new Nuxt app `npx nuxi@latest init && cd nuxt-app`
2. Init Slice Machine `npx @slicemachine/init@latest`
3. Check that it's working without additional setup
   - [ ] Root `app.vue` shouldn't be there anymore
   - [ ] Simulator should be working
   - [ ] Navigating to `http://localhost:3000` should show you Nuxt's default index page
   - [ ] Creating a new page type in Slice Machine should give you a snippet that works to create the page if you follow the process

- [x] The CI is successful.

### DIFF
--- a/packages/adapter-nuxt/src/hooks/project-init.ts
+++ b/packages/adapter-nuxt/src/hooks/project-init.ts
@@ -6,6 +6,8 @@ import type {
 } from "@slicemachine/plugin-kit";
 import {
 	checkHasProjectFile,
+	deleteProjectFile,
+	readProjectFile,
 	writeProjectFile,
 } from "@slicemachine/plugin-kit/fs";
 import { stripIndent } from "common-tags";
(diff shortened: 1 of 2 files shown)


xrutayisire (approved):
LGTM, we just need to be careful if Next change the "<NuxtWelcome" ref that we are using.

PR #1126, opened 2023-09-01
Slice template feature

hypervillain:
Slice template feature


packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
```diff
@@ -0,0 +1,111 @@
...
+				}),
+			),
+			hookReadResult,
+		);
+
+		if (errors) {
+			return {
+				errors,
+			};
+		}
```
lihbr: Why don't you reuse the method from above?

packages/adapter-next/src/hooks/slice-template-library-read.ts
```diff
@@ -0,0 +1,32 @@
...
+export const sliceTemplateLibraryRead: SliceTemplateLibraryReadHook<
+	PluginOptions
+> = async (args: SliceTemplateLibraryReadHookData, { helpers }) => {
+	const isTypeScriptProject = await checkIsTypeScriptProject({
+		helpers,
+	});
+
+	const templates =
+		args.templateIds && args.templateIds.length
+			? initialTemplates.filter((t) => args.templateIds?.includes(t.model.id))
```
lihbr: What's the use case for filtering templates here? Shouldn't this just happen on the front end if it needs to?

This adds complexity to the hook which would be better to offload elsewhere to keep the hook dumb (because here it's forcing it to behave quite differently based on options).

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
```diff
@@ -0,0 +1,111 @@
...
+					}),
+				),
+			}),
+			hookResult,
+		);
+
+		return {
+			templates: data
+				.flat()[0]
+				.templates.map((t) => ({ model: t.model, screenshots: t.screenshots })),
```
lihbr: Maybe this should just return everything? A bit weird to filter the "read" result in a "read" method. Are there performance concerns with returning more data?

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
```diff
@@ -0,0 +1,111 @@
...
+						screenshots: t.any,
+					}),
+				),
+			}),
+			hookResult,
+		);
+
+		return {
+			templates: data
+				.flat()[0]
```
lihbr: Why do we take only the first template here? This reads only templates from one adapter/plugin

For context, we do that with adapter only hooks because they are the only one able to return, but plugins could also return templates no?

packages/plugin-kit/src/hooks/slice-create.ts
```diff
@@ -12,6 +12,7 @@ import type {
 export type SliceCreateHookData = {
 	libraryID: string;
 	model: SharedSlice;
+	componentContents?: string;
```
lihbr, line 15: Maybe just `component`? (to align with `model`) Not sure

lihbr (commented):
Looking nice so far, maybe I'd just name the manager "TemplateManager" in case we want to introduce other kind of templates later on (thinking page templates, etc.)

packages/adapter-next/src/hooks/slice-template-library-read.ts
lihbr: Basically, it swings the method between:
- If no template IDs return all
- If template IDs return some, but likely just return one (I assume that's your usecase here, only reading one)

Somewhere making it a bit not "pure" 🤔

packages/plugin-kit/src/hooks/slice-template-library-read.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+/**
+ * Return value for `slice-template-library:read` hook handlers.
+ */
+export type SliceTemplateLibraryReadHookReturnType = {
+	templates: {
+		model: SharedSlice;
+		mocks: string;
+		componentContents: string;
```
lihbr: same as above, maybe "component", but I'm not sure as it could make sense to be obvious that it's the content and not a file path whatsoever 🤔 (but since it's also the content for model/mocks, maybe it's fine/clear also)

packages/adapter-next/src/sliceTemplates/CallToAction/index.ts
```diff
@@ -0,0 +1,271 @@
...
+					},
+				},
+			},
+			items: {},
+		},
+	],
+};
+
+export const screenshotPaths = {
+	default: ["CallToAction", "screenshot-default.png"].join("/"),
```
xrutayisire: ❓ Cannot do that:

```suggestion
	default: 'CallToAction/screenshot-default.png',
```

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
```diff
@@ -0,0 +1,179 @@
...
+			assertPluginsInitialized(this.sliceMachinePluginRunner);
+
+			return this.slices.updateSliceMocks({
+				libraryID: libraries[0].libraryID,
+				sliceID: slice.model.id,
+				mocks: slice.mocks,
+			});
+		});
+
+		const mocksResults = await Promise.all(mocksPromises);
```
xrutayisire, line 162: We could wait both mocks and screenshots at the same time, right?

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
```diff
@@ -0,0 +1,179 @@
...
+				libraryID: libraries[0].libraryID,
+				model: slice.model,
+				componentContents: slice.componentContents,
+			});
+		});
+
+		// Wait for all slices to be created
+		const creationResults = await Promise.all(creationPromises);
+
+		const mocksPromises = slicesToCreate.map((slice) => {
```
xrutayisire, line 154: I guess we don't want to continue if there are errors on the creation, we should opt out here with the creationErrors

packages/adapter-next/src/hooks/slice-template-library-read.ts
hypervillain: We did it like this because we re-fetch the slice templates list once user has selected template IDs (they will be able to select more than one day one). But I agree that we could fetch everything once again at creation time and then filter

packages/adapter-next/src/hooks/slice-create.ts
```diff
@@ -32,15 +32,17 @@ const createComponentFile = async ({
 	const filename = `index.${extension}`;
 	const pascalName = pascalCase(data.model.name);
 
-	let contents: string;
+	let componentContents: string;
```
angeloashmore: `contents` can remain `contents` in this file since the function's name implies it is for the component (in contrast to `data.componentContents` where `contents` is ambiguous).

packages/adapter-next/src/plugin.ts
```diff
@@ -30,6 +30,7 @@ import { projectInit } from "./hooks/project-init";
 import { sliceCreate } from "./hooks/slice-create";
 import { sliceSimulatorSetupRead } from "./hooks/sliceSimulator-setup-read";
 import { snippetRead } from "./hooks/snippet-read";
+import { sliceTemplateLibraryRead } from "./hooks/slice-template-library-read";
```
angeloashmore: Small nit pick: could we sort this group of imports by alphabetical order? (Or even better: automate it with [`prettier-plugin-sort-imports`](https://github.com/trivago/prettier-plugin-sort-imports)).

```ts
import { documentationRead } from "./hooks/documentation-read";
import { projectInit } from "./hooks/project-init";
import { sliceCreate } from "./hooks/slice-create";
import { sliceSimulatorSetupRead } from "./hooks/sliceSimulator-setup-read";
import { sliceTemplateLibraryRead } from "./hooks/slice-template-library-read";
import { snippetRead } from "./hooks/snippet-read";
```

packages/adapter-next/src/sliceTemplates/CallToAction/index.ts
```diff
@@ -0,0 +1,271 @@
+import type { SharedSlice } from "@prismicio/types-internal/lib/customtypes";
+import { SharedSliceContent } from "@prismicio/types-internal/lib/content";
+
+export const mocks: SharedSliceContent[] = [
```
angeloashmore, line 8: I know this is an MVP for this feature, but thinking long-term: it is unwieldy for adapters to write and manage their own mocks in this format. Is there a plan for making this easier?

packages/adapter-nuxt2/src/hooks/slice-create.ts
```diff
@@ -25,7 +25,9 @@ const createComponentFile = async ({
 	actions,
 	options,
 }: CreateComponentFileArgs) => {
-	const contents = stripIndent`
+	const contents =
+		data.componentContents ??
+		stripIndent`
```
angeloashmore, line 30: Small nit pick: the contents of `stripIndent` should be indented on level more than the `stripIndent` label. For example:

```ts
stripIndent`
  This is indented content
`
```

packages/manager/src/managers/telemetry/types.ts
```diff
@@ -164,7 +164,7 @@ type CustomTypeSavedSegmentEvent = SegmentEvent<
 
 type SliceCreatedSegmentEvent = SegmentEvent<
 	typeof SegmentEventType.slice_created,
-	{ id: string; name: string; library: string }
+	{ id: string; name: string; library: string; sliceTemplate?: string }
```
angeloashmore, line 175: "sliceTemplate" sounds redundant to me since we already know this is a slice-based event. What do you think of changing it to just "template"?

packages/plugin-kit/src/createSliceMachinePluginRunner.ts
```diff
@@ -141,8 +141,9 @@ export class SliceMachinePluginRunner {
 					path.resolve(this._project.root, "noop.js"),
 				)(resolve);
 				plugin = raw.default || raw;
-			} catch {
-				// noop
+			} catch (e) {
+				// uncomment if you need to debug adapter registration
```
angeloashmore: This practice is prone to accidentally leaving it in the production version. Maybe we could put a safeguard in?

```ts
catch (error) {
  if (import.meta.env.DEV) {
    console.error(error)
  }
}
```

(Also, could we use a descriptive variable name like `error` rather than `e`?)

packages/plugin-kit/src/hooks/slice-create.ts
angeloashmore: `componentContents` is clearer, in my opinion, but either option is good.

(The property was originally named `contents`, which was not clear enough.)

packages/plugin-kit/src/hooks/slice-template-library-read.ts
angeloashmore: Copying my above reply:

> componentContents is clearer, in my opinion, but either option is good.
>
> (The property was originally named contents, which was not clear enough.)

packages/plugin-kit/src/hooks/slice-template-library-read.ts
```diff
@@ -0,0 +1,47 @@
...
+	ExtendSliceMachineHook,
+	SliceMachinePluginOptions,
+	SliceMachineHook,
+} from "../types";
+
+/**
+ * Data provided to `slice-template-library:read` hook handlers.
+ */
+export type SliceTemplateLibraryReadHookData = {
+	isTypeScriptProject: boolean;
```
angeloashmore: What happens when we support a framework that isn't JavaScript? Is there a way we can abstract this data completely to the plugin rather than returning a `isTypeScriptProject` boolean?

For example, the component contents returned by the plugin would already automatically switch between JavaScript and TypeScript as needed internally.

If this is for showing icons in the UI or something, maybe we could let plugins return a list of tags (e.g. `["typescript", "tailwindcss"]`).

angeloashmore: I just realized this is the data sent to hooks from the manager. In that case, the manager really shouldn't be aware if the project is TypeScript-based or not. It's the plugin's responsibility to determine that.

(Think: How would this work in a Ruby on Rails project? We won't be supporting RoR, but thinking that way will help keep SM framework-agnostic and allow us to be more flexible going forward.)

packages/plugin-kit/src/hooks/slice-template-library-read.ts
```diff
@@ -0,0 +1,47 @@
...
+	templateIds?: string[];
```
angeloashmore: I believe we capitalize "ID" in all other hooks: `templateIDs`.

packages/adapter-next/src/hooks/slice-template-library-read.ts
```diff
@@ -0,0 +1,46 @@
+import fs from "node:fs";
```
angeloashmore: This file should be named `sliceTemplateLibrary-read` to match the other hook file names.

`slice-template-library:read` => `sliceTemplateLibrary-read`
`custom-type:read` => `customType-read`
`slice:read` => `slice-read`

packages/adapter-next/src/hooks/slice-template-library-read.ts
```diff
@@ -0,0 +1,46 @@
...
+			const { mocks, model, createComponentContents, screenshotPaths } = t;
+
+			return {
+				mocks,
+				model,
+				componentContents: createComponentContents(isTypeScriptProject),
+				screenshots: Object.entries(screenshotPaths).reduce(
+					(acc, curr) => ({
+						...acc,
+						[curr[0]]: fs.readFileSync(
```
angeloashmore: The manager and all adapters use `node:fs/promise` exclusively. In this case, that means changing from `readFileSync` to `readFile`. I understand that adds a bit more code since we need to await the `Promise`s, but it keeps the code async.

packages/adapter-next/src/hooks/slice-template-library-read.ts
```diff
@@ -0,0 +1,46 @@
...
+							path.join(globalThis.__dirname, curr[1]),
```
angeloashmore: Consider using `import.meta.url` rather than `__dirname`; `__dirname` does not exist in ES modules.

packages/adapter-next/src/index.ts
```diff
@@ -1,3 +1,8 @@
+import path from "path";
+
+const rootFile = import.meta.url.slice("file://".length);
+globalThis.__dirname = path.dirname(rootFile);
```
angeloashmore: Node.js has a standard function for this:

```ts
import { fileURLToPath } from "node:url";

fileURLToPath(import.meta.url)
```

Example:

```ts
import { fileURLToPath } from "node:url";

fileURLToPath(new URL("./relative/path/to/file", import.meta.url))
```

If possible, could we avoid adding variables to the global scope? It shouldn't be necessary in this case.

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
```diff
@@ -0,0 +1,179 @@
...
+			model: SharedSlice,
+			componentContents: t.string,
+			mocks: t.array(SharedSliceContent),
+			screenshots: t.record(t.string, t.any),
+		}),
+	),
+});
+
+export class SliceTemplateLibraryManager extends BaseManager {
+	async read(
```
angeloashmore: To keep consistent with the other managers, could these methods be explicit about what it is reading? For example, `readLibrary()` or `readSliceTemplateLibrary()`.

Being explicit helps us add other read-related methods, like `readSliceLibraryTemplateMetadata` (if we needed something like that).

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
```diff
@@ -0,0 +1,179 @@
...
+		args: SliceTemplateLibraryReadData,
+	): Promise<SliceTemplateLibraryManagerReadReturnType> {
+		assertPluginsInitialized(this.sliceMachinePluginRunner);
+
+		const isTypeScriptProject = await this.project.checkIsTypeScript();
```
angeloashmore: Stated in another comment already, but plugins should determine if a project is TypeScript themselves.

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
```diff
@@ -0,0 +1,179 @@
...
+		const hookResult = await this.sliceMachinePluginRunner.callHook(
+			"slice-template-library:read",
+			{ ...args, isTypeScriptProject },
+		);
+
+		const { data, errors } = decodeHookResult(readHookCodec, hookResult);
+
+		return {
+			templates: data.flat()[0].templates,
```
angeloashmore: Does this mean only adapters can return templates? What if plugins want to provide templates?

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
```diff
@@ -0,0 +1,179 @@
...
+			errors,
+		};
+	}
+	async create(
+		args: SliceTemplateLibraryCreateData,
+	): Promise<SliceTemplateLibraryManagerCreateReturnType> {
+		assertPluginsInitialized(this.sliceMachinePluginRunner);
+
+		// Reading all available slice templates
+		const isTypeScriptProject = await this.project.checkIsTypeScript();
```
angeloashmore: Stated in another comment already, but plugins should determine if a project is TypeScript themselves.

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
```diff
@@ -0,0 +1,179 @@
...
+		);
+
+		const { data, errors } = decodeHookResult(readHookCodec, hookResult);
+
+		return {
+			templates: data.flat()[0].templates,
+			errors,
+		};
+	}
+	async create(
```
angeloashmore: The "create" name implies that it is creating a slice template library, but that is not the case here.

Should this logic be part of SliceManager instead?

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
```diff
@@ -0,0 +1,179 @@
...
+			readHookCodec,
+			hookReadResult,
+		);
+		if (readErrors.length > 0) {
+			return {
+				errors: readErrors,
+			};
+		}
+
+		// Extract all existing slice IDs into an array
```
angeloashmore, line 86: You can use `this.slices.readAllSlices().map(slice => slice.model.id)` to get an array of all slice IDs.

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
```diff
@@ -0,0 +1,179 @@
...
+					id,
+					name:
+						counter > 1 ? template.model.name + counter : template.model.name,
+				},
+			};
+		});
+
+		// Initiate the slice creation process for all slices
+		const creationPromises = slicesToCreate.map((slice) => {
+			assertPluginsInitialized(this.sliceMachinePluginRunner);
```
angeloashmore: `this.slices.createSlice()` calls `assertPluginsInitialized` so it is not needed here.

`assertPluginsInitialized` is primarily used to narrow the type of `this.sliceMachinePluginRunner` when you need access it directly, which we don't do here.

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
```diff
@@ -0,0 +1,179 @@
...
+				model: slice.model,
+				componentContents: slice.componentContents,
+			});
+		});
+
+		// Wait for all slices to be created
+		const creationResults = await Promise.all(creationPromises);
+
+		const mocksPromises = slicesToCreate.map((slice) => {
+			assertPluginsInitialized(this.sliceMachinePluginRunner);
```
angeloashmore: `this.slices.updateSliceMocks()` calls `assertPluginsInitialized` so it is not needed here.

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
```diff
@@ -0,0 +1,179 @@
...
+
+		// Check for any errors in the creation results
+		const creationErrors = creationResults.flatMap((result) => result.errors);
+		const mocksErrors = mocksResults.flatMap((result) => result.errors);
+		const screenshotErrors = sliceScreenshotsResults.flatMap((result) =>
+			result.flat().flatMap((r) => r.errors),
+		);
+
+		// Extract the slice IDs from the creation results (assuming each result has an ID)
+		const sliceIds = slicesToCreate.map((slice) => slice.model.id);
```
angeloashmore: Style: `sliceIDs`

angeloashmore (changes requested):

packages/plugin-kit/src/hooks/slice-create.ts
lihbr: Works for me!

packages/adapter-next/src/sliceTemplates/CallToAction/index.ts
lihbr: IMO mocks should be optionally provided by the hook, library authors can edit fancy mocks if they want to, but it should be optional and default to auto mocking(?)

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
lihbr: Just to add on that, but whether a project is TypeScript or not can also be configured explicitly on the adapter level, so hardcoding it here makes it quite inconsistent

packages/adapter-next/src/sliceTemplates/CallToAction/index.ts
hypervillain: I don't know, if screenshots are mandatory, I would like to end up with the same slice in the simulator. I think it will be easy enough to make them optional in next iteration

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
hypervillain: done

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
hypervillain: we could wait everything at the same time, including the first contents file. In a sense, that's what the `create` function does :) But screenshots is an array of array of promises and I think it's clearer to keep it separate

packages/adapter-nuxt2/src/hooks/slice-create.ts
hypervillain: done

packages/plugin-kit/src/hooks/slice-template-library-read.ts
hypervillain: done

packages/adapter-next/src/index.ts
hypervillain: I'm unable to use URL in adapters, for whatever reason (with an error like "self is not referenced" when initialising the plugin)

packages/adapter-next/src/hooks/slice-create.ts
xrutayisire: Good one, done

packages/adapter-next/src/plugin.ts
xrutayisire: Done

packages/manager/src/managers/telemetry/types.ts
xrutayisire: I agree, but It's what product (Côme) ask me to handle it clearly in Amplitude (maybe for search, things like this, not sure and he just left). 

WDYT @hypervillain, should I still change it?

hypervillain: No I think it's part of the product requirements

packages/plugin-kit/src/createSliceMachinePluginRunner.ts
xrutayisire: I really like this idea of always throwing an error in DEV mode on this noop catch

xrutayisire: Done

packages/plugin-kit/src/hooks/slice-template-library-read.ts
xrutayisire: Done

packages/adapter-next/src/hooks/slice-template-library-read.ts
xrutayisire: Done

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
xrutayisire: Done

xrutayisire: - readLibrary
- createSlices

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
xrutayisire: Done

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
xrutayisire: Done

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
xrutayisire: Personally I feel like keeping all things relative to slices templates in SliceTemplateLibrary manager, easy to maintain, but maybe that's not the idea?

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
xrutayisire: Nice one, done in multiple lines anyway for await + error handling

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
xrutayisire: done

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
xrutayisire: done

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
xrutayisire: done

xrutayisire:
Ok, I'm resolving threads that we did directly from your propositions @angeloashmore, @lihbr. It will help us see clearer this PR. Don't hesitate to check resolve threads if you want still.

We will try to release today hopefully, so that's why we could come back just after to changes that we may want to discuss together.

packages/adapter-next/src/hooks/slice-template-library-read.ts
xrutayisire: Done

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
xrutayisire: You're right, fixed, I return all templates now

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
xrutayisire: Fixed, I returned all templates

packages/adapter-next/src/sliceTemplates/CallToAction/index.ts
xrutayisire: Done

lihbr:
Cool, just highlighting this for consideration (from first review message)

> Looking nice so far, maybe I'd just name the manager "TemplateManager" in case we want to introduce other kind of templates later on (thinking page templates, etc.)

packages/adapter-next/src/hooks/slice-template-library-read.ts
xrutayisire: Done you're right

packages/adapter-next/src/index.ts
xrutayisire: Done, at the end we succeed to do it

xrutayisire:
> Looking nice so far, maybe I'd just name the manager "TemplateManager" in case we want to introduce other kind of templates later on (thinking page templates, etc.)

It's another small change, but we lack of time for today release. We will consider it later if necessary, for sure. I agree if we have other types of template, we would need to rename

packages/adapter-next/src/hooks/slice-template-library-read.ts
xrutayisire: TODO: Discuss that for a change in next iteration

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
xrutayisire: TODO: Discuss that for a change in next iteration

packages/adapter-next/src/sliceTemplates/CallToAction/index.ts
xrutayisire: TODO: Discuss that for a change in next iteration

packages/adapter-next/src/index.ts
angeloashmore: Strange error! If you know the reason, I'm curious to know what it was.

packages/manager/src/managers/sliceTemplateLibrary/SliceTemplateLibraryManager.ts
angeloashmore: Works for me, especially with the method rename. I just wanted to make sure we considered it.

PR #1128, opened 2023-09-07
dev-next-release - v1.13.0

xrutayisire:
Pull Request for next release


packages/slice-machine/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
```diff
@@ -223,7 +223,7 @@ const SliceZone: React.FC<SliceZoneProps> = ({
                         onSelect={openSlicesTemplatesModal}
                         startIcon={<Icon name="contentCopy" />}
                       >
-                        Slice templates
+                        From templates
```
gijsbotje, line 226: Shouldn't this also be singular?
```suggestion
                        From template
```

xrutayisire: Thank you for your proposal 🙂

But for this one, we want to focus on the fact you can add from multiple templates at once. 
On the other hand, we choose "From library" because most of our users only have one library and when you have multiple libraries of slices you will understand. It's to not confuse new people. Anyway, we are working on a better wording for "library" 🤞

PR #1129, opened 2023-09-07
Slice template next AlternateGrid

hypervillain:
closes https://linear.app/prismic/issue/DT-1563/aauser-i-have-a-new-slice-template-available-alternate-grid


packages/adapter-next/src/sliceTemplates/AlternateGrid/index.ts
```diff
@@ -0,0 +1,937 @@
...
+				edit: {
+					zoom: 1,
+					crop: {
+						x: 0,
+						y: 0,
+					},
+					background: "transparent",
+				},
+				credits: null,
+				alt: null,
```
xrutayisire, line 67: alt values need to be set for all image to prevent a console error in simulator

packages/adapter-next/src/sliceTemplates/AlternateGrid/index.ts
```diff
@@ -0,0 +1,937 @@
...
+								}
+								}
+								
+								@media (min-width: 1200px) {
+								.es-alternate-grid__content--with-image {
+									grid-template-columns: repeat(2, 1fr);
+								}
+								}
+								
+								.es-alternate-grid__image {
```
xrutayisire, line 573: If we don't have a width height = auto, then the component don't look goods when you don't have tailwinds default that does that for you

PR #1132, opened 2023-09-08
Next slice templates - AlternateGrid

hypervillain:
closes https://linear.app/prismic/issue/DT-1563/aauser-i-have-a-new-slice-template-available-alternate-grid


packages/slice-machine/lib/builders/CustomTypeBuilder/SliceZone/sliceTemplatesComingSoon.ts
```diff
@@ -1,39 +1,6 @@
...
-          docURL: "",
-          id: "Other",
-          name: "Other",
-          version: "sktwi1xtmkfgx8626",
-        },
-      ],
-    },
-    screenshots: {
-      default: {
-        url: "/alternateGrid.png",
```
xrutayisire, line 33: Don't forget to remove the image also in public

xrutayisire:
Was not possible to have mobile support as in the Storybook?

Screenshot 2023-09-08 at 16 32 33
Screenshot 2023-09-08 at 16 32 39

packages/adapter-next/src/sliceTemplates/AlternateGrid/index.ts
```diff
@@ -0,0 +1,939 @@
...
+										: "es-alternate-grid__image--left"
+									}
+								\`}
+							/>
+						)}
+						<div className="es-alternate-grid__primary-content">
+							<div className="es-alternate-grid__primary-content__intro">
+								{isFilled.keyText(slice.primary.eyebrowHeadline) && (
+									<p className="es-alternate-grid__primary-content__intro__eyebrow">
+										{slice.primary.eyebrowHeadline}
```
xrutayisire, line 492: No need to use simple PrismicText

xrutayisire:
Can you check the indentation to be like the CallToAction one:

For the inline style, you will see that it's a bit tricky to make it perfect after creation but it's working for CallToAction 🙂

Screenshot 2023-09-08 at 16 34 13
Screenshot 2023-09-08 at 16 35 20
Screenshot 2023-09-08 at 16 34 08

packages/adapter-next/src/sliceTemplates/AlternateGrid/index.ts
hypervillain: I don't understand

xrutayisire: Sorry I wanted to add a question mark haha

Should we use PrismicText for these kind of things?

hypervillain: It depends if the designer considers that something else than a p tag should be used. In this case, I reused the first implementation from Sarah and Angelo

xrutayisire (approved):
LGTM

PR #1133, opened 2023-09-08
feat(Page Type): Specific blank slate right after Page Type creation

xrutayisire:
## Context

Completes DT-1578, DT-1566

## The Solution

- Add a new URL query "newPageType" at true when we are doing the redirection after the Page Type creation
- Handle a full page blank slate when "newPageType" URL query is at true

## Impact / Dependencies

- None

- [ ] The CI is successful.

## Preview

https://github.com/prismicio/slice-machine/assets/19946868/2e685886-1aef-4f41-a7e1-25880a1cffb8


hypervillain (approved):
I'm not a fan of of adding that much "product logic" in the code, as these requirements change and require a careful update every time. But it _is_ the requirement at the minute so it's fine

PR #1144, opened 2023-09-22
feat(slice-machine-ui): add Card component

bapmrl:
Completes DT-1604.


packages/slice-machine/.storybook/preview.tsx
```diff
@@ -3,11 +3,6 @@ import "@prismicio/editor-ui/style.css";
 import type { Preview } from "@storybook/react";
 
 const preview: Preview = {
-  argTypes: {
```
xrutayisire, line 6: ❓ What is the idea behind this change?
I see the diff on the stories but would like to understand the whole picture

packages/slice-machine/src/components/Card/Card.css.ts
```diff
@@ -0,0 +1,142 @@
...
+  solid: [sprinkles({ backgroundColor: colors.grey3, overflowX: "hidden" })],
+  outlined: [sprinkles({ backgroundColor: colors.grey2 })],
+});
+
+export const interactive = style({
+  selectors: {
+    "&[data-disabled]": { opacity: vars.opacity["50%"] },
+    "&:not([data-disabled])": { cursor: vars.cursor.pointer },
+    "&:focus-visible": {
+      boxShadow: "0px 0px 0px 4px hsla(251, 63%, 63%, 0.1)",
```
xrutayisire: ⚠️ No box shadow from the system design, I guess? Is that normal that we don't have it?

packages/slice-machine/src/components/Card/Card.css.ts
```diff
@@ -0,0 +1,142 @@
...
+    },
+  }),
+};
+
+export const media = style([
+  sprinkles({ boxSizing: "border-box" }),
+  {
+    order: 0,
+    selectors: {
+      [`${size.small} > &`]: { height: "180px" },
```
xrutayisire: ❓ When we will be using small or medium?

packages/slice-machine/src/components/Card/Card.css.ts
```diff
@@ -0,0 +1,142 @@
...
+      [`${size.medium} > &`]: { height: "260px" },
+      [`${variant.solid} > &`]: {
+        padding: calc.subtract(vars.space[32], vars.borderWidth[1]),
```
xrutayisire, line 69: ⚠️ For me, we should avoid doing this, by talking to Benjamin it's just that he don't have the correct setup in Figma, and he is ok to do 32px here for example most of the time. Did you see that with him? Having to subtract the border feel strange for me 😕

packages/slice-machine/src/components/Card/Card.css.ts
```diff
@@ -0,0 +1,142 @@
...
+        marginTop: calc.subtract(vars.space[16], vars.borderWidth[1]),
+        padding: vars.space[16],
+      },
+    },
+  },
+]);
+
+export const mediaComponent = styleVariants({
+  div: [sprinkles({ all: "unset", display: "grid" })],
+  img: [
```
xrutayisire, line 85: ❓ You don't want to all: unset for img?

packages/slice-machine/src/components/Card/Card.css.ts
```diff
@@ -0,0 +1,142 @@
...
+    sprinkles({ objectFit: "contain" }),
+    { selectors: { "&:not([src])": { visibility: "hidden" } } },
```
xrutayisire, line 87: ⚠️ When this can happen? Should it not be a JavaScript condition for an error or something here?

packages/slice-machine/src/components/Card/Card.css.ts
```diff
@@ -0,0 +1,142 @@
...
+        paddingInline: calc.subtract(vars.space[16], vars.borderWidth[1]),
+      },
+    },
+  },
+]);
+
+export const footerTexts = style([
+  column,
+  sprinkles({ flexGrow: 1, minWidth: 0 }),
+  { marginBottom: `calc((${24 / 16}rem - ${16 / 16}rem) / 2)` },
```
xrutayisire, line 169: ❓ Where does this come from?

packages/slice-machine/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,108 @@
...
+        checked?: never;
+        disabled?: never;
+        onClick?: never;
+      }
+  )
+>;
+
+export const Card: FC<CardProps> = (props) => {
+  const {
+    checked: _checked,
```
xrutayisire: ⚠️ Why extracting it with "_" to remove it from otherProps, while you would extract it normally and using it instead of doing `props.checked`. No?

packages/slice-machine/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,108 @@
...
+    disabled: _disabled,
+    interactive: _interactive,
+    onClick: _onClick,
+    size = "medium",
+    variant = "solid",
```
xrutayisire: ❓I see that "solid" and "medium" are the default, how did you define it?

packages/slice-machine/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,108 @@
...
+  action?: ReactNode;
+  subtitle?: ReactNode;
+  title?: ReactNode;
+};
+
+export const CardFooter: FC<CardFooterProps> = ({
+  action,
+  subtitle,
+  title,
+  ...otherProps
```
xrutayisire, line 141: ❓ I'm always wondering, here you are typing the props with strictly 3 props. But you spread the rest. It should have no rest, at least it's not visible it's allowed. Should we extend the type to explain we accept some other HTML default props?

xrutayisire (changes requested):
First code only review, always nice work with you! 🙂

packages/slice-machine/.storybook/preview.tsx
bapmrl: Before that change, all stories had (disabled) controls for `className` and `style`, even if the corresponding components didn't use these props. Now, `className` and `style` are hidden in all stories.

packages/slice-machine/src/components/Card/Card.css.ts
bapmrl: I've checked and in [Prismic - Design System](https://www.figma.com/file/V5pub1BwPMJGY52YHmhTmX/Prismic---Design-System?mode=dev) I couldn't find an outline that is `4px` wide. I've only found `2px` (which exists in `@prismicio/editor-ui`) and `3px` (which doesn't exist in `@prismicio/editor-ui`). I'm not sure what to do. Maybe it's fine for now to keep it that way, as we only have one such outline.

packages/slice-machine/src/components/Card/Card.css.ts
bapmrl: `small` and `outlined` will be used to select slices in the dialog to update the slice zone. `medium` and `solid` will be used everywhere else.

packages/slice-machine/src/components/Card/Card.css.ts
bapmrl: For me, it's either:
1. @benjammartin configures Figma so that borders sit outside (instead of inside) components. It would then remove the need to make subtractions in the styles to match Figma.
2. @benjammartin keeps it the current way (so borders sit inside components in Figma), but we have to make subtractions.

Am I missing something?

packages/slice-machine/src/components/Card/Card.css.ts
xrutayisire: Ok, for me, we should just keep track with @benjammartin of this kind of things, so we know when to add it to the system design if needed.

packages/slice-machine/src/components/Card/Card.css.ts
xrutayisire: Good summary, If it's possible to change the configuration, I see only something simple. Much more to using sprinkles and all.

benjammartin: Make subtractions. That's ok like this.

packages/slice-machine/src/components/Card/Card.css.ts
bapmrl: Good question. It's not done by the CSS reset I inspired from when I decided to use `all: unset` (see https://github.com/elad2412/the-new-css-reset/blob/main/css/reset.css#L11)). It's probably not necessary for `img`, `video`... and may even break some browsers?

packages/slice-machine/src/components/Card/Card.css.ts
bapmrl: The default story doesn't specify a `src` and I didn't like how Chrome rendered the `img` in that case. I wanted to illustrate the component in it's "default"/"empty" state to highlight slots (where actual components can be passed).

packages/slice-machine/src/components/Card/Card.css.ts
xrutayisire: I personally don't like it, it means we cannot rely on the data from Figma directly, that we always have to think about it and that we cannot use sprinkles directly. Is it complicated to update the configuration?
For me, we will lose time over this during all SM life.

packages/slice-machine/src/components/Card/Card.css.ts
xrutayisire: Oh good to know! 
Interesting.

You should definitely create a Notion page with all your references, like System Design, Components libraries, this inspiration. You have so many good things to share in your head 😛

packages/slice-machine/src/components/Card/Card.css.ts
bapmrl: It comes from Figma, but I agree it's a bit obscure. Benjamin slightly offset the texts so that they appear to be vertically centered despite the `title` having a higher `lineHeight` than the `subtitle`.

Should I add a comment?

I would have preferred to have access to the `fontSize` and `lineHeight` in the theme, but they have been removed and I don't really have time to add them back.

packages/slice-machine/src/components/Card/Card.tsx
bapmrl: 1. I need access to the unspread `props` so that TypeScript can narrow down the type when `props.interactive` is `true`.
2. I don't want `checked`, `disabled`, `interactive` and `onClick` to be passed as is to the underlying `div`. These props help me derive other props, which are passed to the underlying `div` only when the `Card` is `interactive`.
3. I had to prefix `checked`, `disabled`, `interactive` and `onClick` with `_` otherwise ESLint would complain.

packages/slice-machine/src/components/Card/Card.tsx
bapmrl: It's what will be most frequently used in the app.

packages/slice-machine/src/components/Card/Card.css.ts
xrutayisire: Yeah no problem with the editor problem. A small comment could be nice for maintenance yes 🙌

packages/slice-machine/src/components/Card/Card.tsx
xrutayisire: Ok good one sorry, thanks for the explanation 🙂

packages/slice-machine/src/components/Card/Card.tsx
xrutayisire: The other comment made me understand thanks

packages/slice-machine/src/components/Card/Card.css.ts
benjammartin: A quick explanation. Figma doesn't handle borders like a browser does.

Having an outside or inside border has no impact on the dimensions of a component in Figma. This is not the case when it's rendered in a browser.

This means that if we want to create an implementation that matches the Figma design exactly, we'll have to cheat in some way.

packages/slice-machine/src/components/Card/Card.tsx
bapmrl: I'm spreading the rest because consumers may pass `data-` attributes. Also, if the component where to be used as a Radix trigger, Radix would pass props which we wouldn't be aware of.

packages/slice-machine/src/components/Card/Card.css.ts
bapmrl: That's true. The problem is always time ^^

packages/slice-machine/src/components/Card/Card.css.ts
benjammartin: Edit : Except by turning the component in Autolayout.

xrutayisire: Ok Baptiste, I let you decide what you prefer in the meantime. We just need to react quickly after so we don't have too many things to change like the editor. 🙂

xrutayisire (approved):
LGTM, great work! 🙂

Last question, if we have in a slice zone, some Card with CardActions + CardFooter, and some only with CardFooter, is it ok that in terms of Design size (height) will be different?

bapmrl:
> LGTM, great work! 🙂
> 
> Last question, if we have in a slice zone, some Card with CardActions + CardFooter, and some only with CardFooter, is it ok that in terms of Design size (height) will be different?

I've fixed this issue ^^ It was an error on my side.

packages/slice-machine/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,157 @@
...
+          if (findFocusableAncestor(target) === event.currentTarget) {
+            onClick(event);
+          }
+        }}
+        tabIndex={disabled === true ? undefined : 0}
+      />
+    );
+  } else if (props.interactive === true) {
+    // prettier-ignore
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
xrutayisire: Why these prettier-ignore and eslint-disable-next-line?

bapmrl: I'm removing those as discussed together ^^

PR #1146, opened 2023-09-22
fix(adapter-next): don't include route resolver configuration by default

angeloashmore:
## Context

This PR updates `@slicemachine/adapter-next` by removing the generated `prismicio.ts`'s default route resolver entries.

The entries were provided as an example for developers to update in their project, but it meant projects were not immediately runnable.

## The Solution

Remove the default entries. The existing entries were kept, but commented out, to continue to provide an example configuration.

Note that **API responses will always default to `null` without route resolver entries**. The adapter's page code snippet describes that users should add an entry to the project's route resolver.

## Impact / Dependencies

Developers will now be able to run a project immediately after running `@slicemachine/init`, even if it means they are still missing necessary route resolver configuration.

- [x] The CI is successful.


xrutayisire (approved):
LGTM 👍

PR #1161, opened 2023-10-10
fix(slice-machine-ui): inactive slices menu item

bapmrl:
Completes DT-1603.

I'm also changing the way `packages/slice-machine/components/Navigation/index.tsx` works. Links in the menu now use `next/link` instead of custom logic that prevented the browser's default behaviour.


packages/slice-machine/components/Navigation/index.test.tsx
```diff
@@ -128,34 +125,6 @@ describe("Side Navigation", () => {
     expect(element).toBeNull();
   });
 
-  test.each([
```
xrutayisire, line 131: As I said in Slack, I tried with https://www.npmjs.com/package/next-router-mock#example-nextlink-with-react-testing-library and it was still not working.

But I would prefer to skip the test instead of removing it so we can unskip and update it with vite. WDYT?

xrutayisire (approved):
LGTM, super nice this fix 🙌

packages/slice-machine/components/Navigation/index.test.tsx
bapmrl: Done 😊

PR #1162, opened 2023-10-10
fix(init): Missing reason in init errors

xrutayisire:
## Context

- Part of DT-1636

## The Solution

- `cause: res` was not the right thing we would like to have as a cause since cause should be a readable string

## Impact / Dependencies

- Let us know exactly why some calls fails

- [ ] The CI is successful.

### DIFF
--- a/packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
+++ b/packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
@@ -174,10 +174,10 @@ export class PrismicRepositoryManager extends BaseManager {
 		const text = await res.text();
 
 		// Endpoint returns repository name on success, which must be more than 4 characters and less than 30
-		// if (!res.ok) {
-		if (!res.ok || text.length < 4 || text.length > 30) {
+		// Even when there is an error, we get a 200 success and so we have to check the name thanks to that
+		if (!res.ok || text.length < 4 || text.length > 63) {
 			throw new Error(`Failed to create repository \`${args.domain}\``, {
-				cause: res,
+				cause: text,
 			});
 		}
 	}
@@ -244,7 +244,7 @@ export class PrismicRepositoryManager extends BaseManager {
 				throw new Error(
 					`Failed to push documents to repository \`${args.domain}\`, repository is not empty`,
 					{
-						cause: res,
+						cause: reason,
 					},
 				);
 			}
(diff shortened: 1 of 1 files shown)


packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
```diff
@@ -174,10 +174,9 @@ export class PrismicRepositoryManager extends BaseManager {
 		const text = await res.text();
 
 		// Endpoint returns repository name on success, which must be more than 4 characters and less than 30
-		// if (!res.ok) {
-		if (!res.ok || text.length < 4 || text.length > 30) {
```
xrutayisire, line 178: @lihbr on this PR I removed a strange condition you added.

When the back-end return that the creation was ok, the name should be well formatted and even if it's not a good one, it's a success anyway. Strange to fail when back-end return that it's ok, no?

If you can review this PR at the same time, I would like to have the detailled error message with your release

lihbr: I added that back then because the backend returns 200 OK when it actually encounters 500 (IIRC it happens when the auth state is expired/invalid)

The only "real" 200 OK for this call is when these conditions are all met:
- Response status is 200
- Response text is the created repository ID (which can sometime differ from the one send since the backend also slugifies the ID to some extends on its end)

All other responses are errors. So I guess the condition could be updated to check that `text.length <= 63` (subdomain length limit)?

lihbr (approved):
LGTM but we need to update the condition/resolve the opened comment :)

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
lihbr: You actually can easily test that by sending a POST like this, this returns a 200 that is actually a 500/401:
```
POST /authentication/newrepository?app=slicemachine HTTP/2
Host: prismic.io
user-agent: insomnia/2023.4.0
accept: */*
content-length: 108

{
	"domain": "test",
	"framework": "nuxt",
	"plan": "personal",
	"isAnnual": "false",
	"role": "developer"
}
```

xrutayisire: Thank you for the explanation!

PR #1172, opened 2023-10-16
fix(page type): Convert to page type should create slice zone for main tab if necessary

xrutayisire:
## Context

- Completes DT-1687

## The Solution

- Create shared functions that should allow us to easily manipulate custom type without CustomTypeSM
- Ensure slice zone is created for main tab if necessary when converting a custom type to page type

## Impact / Dependencies

- New cool functions to use now 🙂

- [ ] The CI is successful.

## Preview

https://github.com/prismicio/slice-machine/assets/19946868/6f952300-4af1-4c88-902c-f2f8cc328acc

### DIFF
--- a/packages/slice-machine/src/domain/__tests__/CustomTypeModel.test.ts
+++ b/packages/slice-machine/src/domain/__tests__/CustomTypeModel.test.ts
@@ -1,0 +1,247 @@
+import { describe, expect } from "vitest";
+
+import {
+  CustomType,
+  DynamicSection,
+} from "@prismicio/types-internal/lib/customtypes";
+
+import * as CustomTypeModel from "../customType";
+
+describe("CustomTypeModel test suite", () => {
+  const mainSection: DynamicSection = {
+    uid: {
+      config: {
+        label: "MainSectionField",
+      },
+      type: "UID",
+    },
+    slices: {
+      type: "Slices",
+      fieldset: "Slice Zone",
+      config: {
+        choices: {
+          hero_banner: {
+            type: "SharedSlice",
+          },
+          promo_section_image_tiles: {
+            type: "SharedSlice",
+          },
+        },
+      },
+    },
+  };
+  const anotherSection: DynamicSection = {
+    uid: {
+      config: {
+        label: "AnotherSectionField",
+      },
+      type: "UID",
+    },
+  };
+  const mockCustomType: CustomType = {
+    format: "custom",
+    id: "id",
+    json: {
+      mainSection,
+      anotherSection,
+    },
+    label: "lama",
+    repeatable: true,
+    status: true,
+  };
+
+  it("getSectionEntries should return the sections entries", () => {
+    expect(CustomTypeModel.getSectionEntries(mockCustomType)).toEqual([
+      ["mainSection", mainSection],
+      ["anotherSection", anotherSection],
+    ]);
+  });
+
+  it("getSectionEntries should return an empty array if there are no sections", () => {
+    expect(
+      CustomTypeModel.getSectionEntries({
+        ...mockCustomType,
+        json: {},
+      })
+    ).toEqual([]);
+  });
+
+  it("getMainSectionEntry should return the first section even if not named Main", () => {
+    expect(CustomTypeModel.getMainSectionEntry(mockCustomType)).toEqual([
+      "mainSection",
+      mainSection,
+    ]);
+  });
+
+  it("getMainSectionEntry should return undefined if there is are sections", () => {
+    expect(
+      CustomTypeModel.getMainSectionEntry({
+        ...mockCustomType,
+        json: {},
+      })
+    ).toEqual(undefined);
+  });
+
+  it("getSection should return the section matching the key", () => {
+    expect(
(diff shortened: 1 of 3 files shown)


packages/slice-machine/src/domain/CustomTypeModel.ts
```diff
@@ -0,0 +1,93 @@
+import {
+  CustomType,
+  DynamicSection,
+  DynamicSlicesConfig,
+} from "@prismicio/types-internal/lib/customtypes";
+
+export const CustomTypeModel = {
```
xrutayisire: Hello @hypervillain, @lihbr, @angeloashmore

I discussed a bit with Baptiste on two things:

- Having shared functions in `src/domains` like Baptiste did in `src/domains/slice.ts` and I did on this file. We know there are existing shared functions, but we feel either some are missing or too complicated. We would also like to push for not relying on specific SM types, but only the ones shared with the managers.
- We also discussed how to share this kind of shared domain functions:
  - a file like `src/domains/slice.ts` where functions are directly exported
  - a file like `src/domains/CustomTypeModel.ts` where an object that contains all the functions is exported
  
It's mostly for the second part that we would like to have your preferences. We discussed together a bit the two, and we would like to see what you prefer before giving you our arguments 🙂 
No hard blocker for one or the other, just good to agree on a way to do things.

cc @bapmrl

packages/slice-machine/src/domain/__tests__/CustomTypeModel.test.ts
```diff
@@ -0,0 +1,190 @@
...
+        json: {
+          MainSection: {},
+          AnotherSection,
+        },
+      })
+    ).toEqual({
+      ...mockCustomType,
+      json: {
+        MainSection: {
+          slices: {
```
xrutayisire, line 237: @hypervillain 

Here on the old code, the key was created with a unique name depending on all slice zone of all tabs.

Do we really want that? I don't get it, since each slice zone is specific to a tab.

Old code for the key: `slices${i !== 0 ? i.toString() : ""}`

packages/slice-machine/src/domain/CustomTypeModel.ts
```diff
@@ -0,0 +1,93 @@
...
+  getSectionsEntries(customType: CustomType): [string, DynamicSection][] {
+    return Object.entries(customType.json);
+  },
+
+  getMainSectionEntry(
```
xrutayisire: @hypervillain,

Can you confirm that it's the only way to get the main tab?

packages/slice-machine/src/domain/CustomTypeModel.ts
lihbr: I prefer option 1 imo `src/domains/slices.ts`, option 2 doesn't feel like a pattern I often encountered (I assume because it makes it harder to treeshake the object which is not a concern for us here, but uncommon afaik)

xrutayisire: @bapmrl also prefers option 1, let's go for option 1, thanks 🙂

hypervillain: I have a small preference for objects, I don't like the micro frustrations of forgetting the name of an import, or an import not being completed properly. We don't have large perf problems so let's use this fact to be simpler!

xrutayisire: Baptiste also proposed we can do that with option one:

`import * as CustomTypeModel from "src/features/custom-type.ts"`

If you don't know the function name exactly, for example.

packages/slice-machine/src/domain/CustomTypeModel.ts
```diff
@@ -0,0 +1,93 @@
+import {
+  CustomType,
+  DynamicSection,
+  DynamicSlicesConfig,
+} from "@prismicio/types-internal/lib/customtypes";
+
+export const CustomTypeModel = {
+  getSectionsEntries(customType: CustomType): [string, DynamicSection][] {
```
bapmrl: ❓ `getSectionEntries`?

bapmrl (approved):

packages/slice-machine/src/domain/__tests__/CustomTypeModel.test.ts
xrutayisire: Done, it's required

packages/slice-machine/src/domain/CustomTypeModel.ts
xrutayisire: Done, the `import * as CustomTypeModel from "src/features/custom-type.ts"` is done as an example in the test file. Less imports so.

packages/slice-machine/src/domain/CustomTypeModel.ts
xrutayisire: Done

packages/slice-machine/src/domain/CustomTypeModel.ts
xrutayisire: Done, yes it's the only way

bapmrl (approved):

PR #1173, opened 2023-10-18
feat(adapter-sveltekit): support previews (DT-1678)

angeloashmore:
## Context

[DT-1678](https://linear.app/prismic/issue/DT-1678/aauser-i-can-support-prismic-previews-in-my-sveltekit-website)

## The Solution

`@slicemachine/adapter-sveltekit` now performs the following in the `project:init` hook:

- Creates a `src/routes/[[preview=preview]]` directory with a `README.md` file.
- Creates a `src/params/preview.js` file to define the `preview` route matcher.
- Creates a `src/routes/+layout.server.js` file to set `prerender = "auto"` for the whole app.
- Creates a `src/routes/api/preview/+server.js` file that uses `@prismicio/svelte/kit`'s `redirectToPreviewURL()`.

`@slicemachine/adapter-sveltekit` also has a new page snippet that supports previews.

Finally, the `prismicio.js` file generated by `@slicemachine/adapter-sveltekit` now uses `@prismicio/svelte/kit`'s `enableAutoPreviews()`.

## Impact / Dependencies

SvelteKit developers will be able to implement previews in their app.

- [x] The CI is successful.

### DIFF
--- a/packages/adapter-sveltekit/src/hooks/documentation-read.ts
+++ b/packages/adapter-sveltekit/src/hooks/documentation-read.ts
@@ -32,8 +32,6 @@ export const documentationRead: DocumentationReadHook<PluginOptions> = async (
 			dataFileContent = source`
 				import { createClient } from "$lib/prismicio";
 
-				export const prerender = true;
-
 				export async function load({ params, fetch }) {
 					const client = createClient({ fetch });
 
@@ -58,8 +56,6 @@ export const documentationRead: DocumentationReadHook<PluginOptions> = async (
 			dataFileContent = source`
 				import { createClient } from "$lib/prismicio";
 
-				export const prerender = true;
-
 				export async function load({ params, fetch }) {
 					const client = createClient({ fetch });
 
@@ -68,6 +64,10 @@ export const documentationRead: DocumentationReadHook<PluginOptions> = async (
 					return {
 						page,
 					};
+				}
+
+				export async function entries() {
+					return [{}]
 				}
 			`;
 		}
--- a/packages/adapter-sveltekit/src/hooks/project-init.ts
+++ b/packages/adapter-sveltekit/src/hooks/project-init.ts
@@ -243,6 +243,8 @@ const createPreviewRouteDirectory = async ({
 
 		- \`/example-route\` (prerendered)
 		- \`/preview/example-route\` (server-rendered)
+
+		See <https://prismic.io/docs/svelte-preview> for more information.
 	`;
 
 	await writeProjectFile({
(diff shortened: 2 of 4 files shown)


packages/adapter-sveltekit/src/hooks/documentation-read.ts
```diff
@@ -69,6 +65,10 @@ export const documentationRead: DocumentationReadHook<PluginOptions> = async (
 						page,
 					};
 				}
+
+				export async function entries() {
+					return [{}]
```
samlfair, line 70: @angeloashmore This looks like it will always return the root, `/`? Should it be something like the repeatable one?
```
export async function entries() {
	const client = createClient();

	const page = await client.getSingle("${model.id}");

        return { uid: page.uid };
}
```

packages/adapter-sveltekit/src/hooks/project-init.ts
```diff
@@ -200,6 +196,86 @@ const createSliceSimulatorPage = async ({
...
+		"[[preview=preview]]",
+		"README.md",
+	);
+
+	if (await checkHasProjectFile({ filename, helpers })) {
+		return;
+	}
+
+	const contents = source`
+		This directory adds support for optional \`/preview\` routes. Do not remove this directory.
```
samlfair, line 240: @angeloashmore If I publish the preview doc tomorrow, can we link to it here?

https://prismic.io/docs/svelte-preview

angeloashmore: Yes, good idea. I added this line:

```md
See <https://prismic.io/docs/svelte-preview> for more information.
```

packages/adapter-sveltekit/src/hooks/documentation-read.ts
angeloashmore: From what I can tell, `{}` doesn't mean root. Instead, it means there are no route parameters to define for the current route.

In this case, because the document is a singleton, it likely does not contain a UID and the route will likely not use a dynamic parameter. Thus, `{}` is used for a route like `src/routes/[[preview=preview]]/foo/bar/+page.server.js`.

Like the repeatable page snippet, we don't include `preview: "preview"` because we want it to be server rendered.

angeloashmore: I tested it locally and it works as it should: `/foo/bar` is prerendered and `/preview/foo/bar` is server-rendered.

samlfair: @angeloashmore Ah, I understand! Sounds good.

lihbr (approved):
LGTM, I'll merge!

PR #1174, opened 2023-10-18
feat(slice-machine-ui): update List component

bapmrl:
Completes DT-1679 and DT-1680.

### DIFF
--- a/packages/slice-machine/src/components/Window/Window.css.ts
+++ b/packages/slice-machine/src/components/Window/Window.css.ts
@@ -18,6 +18,9 @@ export const root = style([
     borderWidth: 1,
     overflowX: "hidden",
   }),
+  // The following `{}` is necessary for the `root` class to work in selectors.
+  // It helps us remove the `List` component's borders when used in a `Window`.
+  {},
 ]);
 
 export const frame = style([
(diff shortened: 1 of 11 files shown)


packages/slice-machine/src/components/Window/Window.css.ts
```diff
@@ -18,6 +18,7 @@ export const root = style([
     borderWidth: 1,
     overflowX: "hidden",
   }),
+  {},
```
xrutayisire, line 23: ⚠️ We should add a comment here, since it's hard to understand that first this is made to let the possibility to use root as a selector. And second used in other style files, but not in this one.
I don't really like having to do this, actually it creates a hard to maintain dependency but I don't have a better solution 👀.
I would recommend having a comment here to explain where it's used or at least it's used elsewhere ^^

xrutayisire (changes requested):
Super work! Really nice to see a simplification of the overall codebase 🤩
One small comment that I think we should add.

packages/slice-machine/src/components/Window/Window.css.ts
bapmrl: I've added a comment as discussed together 😊

xrutayisire (approved):
All good! ✅

PR #1175, opened 2023-10-19
fix(repository name): Repository name length can be up to 63 and not only 30 characters

xrutayisire:
## Context

- Completes DT-1688

## The Solution

- Created constant in order to save the correct min max repository name length

## Impact / Dependencies

- Prevent user to have an error after creating a valid repository name in wroom

- [ ] The CI is successful.

### DIFF
--- a/packages/init/src/SliceMachineInitProcess.ts
+++ b/packages/init/src/SliceMachineInitProcess.ts
@@ -17,6 +17,7 @@ import {
 	SliceMachineManager,
 	PackageManager,
 	StarterId,
+	REPOSITORY_NAME_VALIDATION,
 } from "@slicemachine/manager";
 
 import pkg from "../package.json";
@@ -683,7 +684,7 @@ export class SliceMachineInitProcess {
 							this.manager.prismicRepository.checkExists({ domain }),
 					});
 
-					if (validation.LessThan4 || validation.MoreThan30) {
+					if (validation.LessThanMin || validation.MoreThanMax) {
 						const errorMessage = getErrorMessageForRepositoryDomainValidation({
 							validation: {
 								...validation,
@@ -846,17 +847,25 @@ export class SliceMachineInitProcess {
 				const domain = formatRepositoryDomain(rawDomain);
 				const validation = validateRepositoryDomain({ domain });
 
-				const minRule = validation.LessThan4
+				const minRule = validation.LessThanMin
 					? chalk.red(
-							`1. Name must be ${chalk.bold("4 characters long or more")}`,
+							`1. Name must be ${chalk.bold(
+								REPOSITORY_NAME_VALIDATION.Min + " characters long or more",
+							)}`,
 					  )
-					: `1. Name must be ${chalk.cyan("4 characters long or more")}`;
+					: `1. Name must be ${chalk.cyan(
+							REPOSITORY_NAME_VALIDATION.Min + " characters long or more",
+					  )}`;
 
-				const maxRule = validation.MoreThan30
+				const maxRule = validation.MoreThanMax
 					? chalk.red(
-							`1. Name must be ${chalk.bold("30 characters long or less")}`,
+							`1. Name must be ${chalk.bold(
+								REPOSITORY_NAME_VALIDATION.Max + " characters long or less",
+							)}`,
 					  )
-					: `1. Name must be ${chalk.cyan("30 characters long or less")}`;
+					: `1. Name must be ${chalk.cyan(
+							REPOSITORY_NAME_VALIDATION.Max + " characters long or less",
+					  )}`;
 
 				this.msg = chalk.reset(
 					`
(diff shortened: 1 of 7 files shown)


packages/init/src/constants.ts
```diff
@@ -7,3 +7,7 @@ export const SLICE_MACHINE_INIT_USER_AGENT = "slice-machine-init";
 export const SENTRY_EXPRESS_DSN =
 	import.meta.env.VITE_SENTRY_EXPRESS_DSN ||
 	"https://d3e69d5189266d8d7f60f71ccc9353fc@o146123.ingest.sentry.io/4505917981982720";
+export const REPOSITORY_NAME_VALIDATION = {
```
bapmrl: ❓ Why not import it from the manager instead of defining it twice? It would prevent issues where we update one package, but not the other.

packages/manager/src/constants/REPOSITORY.ts
```diff
@@ -0,0 +1,4 @@
+export const REPOSITORY_NAME_VALIDATION = {
```
bapmrl: 💡 I would name the file `REPOSITORY_NAME_VALIDATION` as other constants seem to be named very "precisely".

packages/manager/src/constants/REPOSITORY.ts
```diff
@@ -0,0 +1,4 @@
...
+	min: 4,
```
bapmrl: 💡 It seems other constants which expose values in an object, have these values either `UpperCamelCased` or `UPPER_SNAKE_CASED`.

packages/init/src/SliceMachineInitProcess.ts
```diff
@@ -848,15 +849,23 @@ export class SliceMachineInitProcess {
 
 				const minRule = validation.LessThan4
```
bapmrl: ⚠️ There's something confusing here. Why is `LessThan4` explicitly stating `4` while we use the word `min` in `REPOSITORY_NAME_VALIDATION`? Same thing for `MoreThan30`.

bapmrl (changes requested):

packages/init/src/constants.ts
xrutayisire: That's a good question I asked myself, can I just reuse a constant from the manager. It has not been done yet, @lihbr maybe, WDYT?

packages/manager/src/constants/REPOSITORY.ts
xrutayisire: I will check to be more aligned, you're right

packages/init/src/SliceMachineInitProcess.ts
xrutayisire: Sorry, I forgot to update these naming here, I don't want to have this stored directly in the code but only in the constant now 🙂

lihbr: Yes, let's find a more generic name

packages/init/src/constants.ts
lihbr: Indeed, this whole "repository name validation/generation" module quite deserve a unique place to exists, maybe it makes more send to move it as methods of the repository manager(?)

xrutayisire: I feel like it's a bit annoying having to call the manager and passing it every place needed just to get the constant, no? Or maybe that's only what we want to do?

lihbr: No for constants I agree, imports are enough and I'm fine having them coming from the manager package.

I was thinking about all the helpers to validate repository names & such.

xrutayisire: Done 🙂

packages/manager/src/constants/REPOSITORY.ts
xrutayisire: Done 🙂

packages/manager/src/constants/REPOSITORY.ts
xrutayisire: Done 🙂

packages/init/src/SliceMachineInitProcess.ts
xrutayisire: Done 🙂

bapmrl: I still see `LessThan4` and `MoreThan30` ^^

xrutayisire: I forgot to push my whole fix commit haha thanks

packages/init/src/lib/repositoryDomain.ts
```diff
@@ -114,8 +116,8 @@ const assertFormattedRepositoryDomain = (
 };
 
 export const ValidationErrors = {
-	LessThan4: "LessThan4",
-	MoreThan30: "MoreThan30",
+	Min: "Min",
```
bapmrl: 💡 `LessThanMin` and `MoreThanMax` would be clearer ^^

bapmrl (approved):

packages/init/src/lib/repositoryDomain.ts
xrutayisire: Done thanks

PR #1177, opened 2023-10-19
feat(slice-machine-ui): update Slice Builder layout

bapmrl:
Completes DT-1682:
- Implement new `Sidebar` (with the new `Card`) and use `List` (and its `ListHeader` subcomponent) everywhere.
- Make `NewField` wrap so that users can more easily input their field name/id with a small viewport.
- Move code closer to usage (for example, `VariationsPopover`).
- Update Cypress tests.
- Remove dead code.


angeloashmore (commented):
### General feedback

- ❓ When creating a new slice variation, should the new variation be selected? It currently stays on the variation selected when clicking the "Add a new variation" button.

- ❓ The `NewField` component looks awkward with the new wrapping. The name and ID fields don't align, for example. Do we have a proper design?

   As an aside, could we change "field ID" to "Field ID"?

   (I understand that this may be an intermediate design until we implement the full redesign.)

   image

packages/slice-machine/lib/builders/SliceBuilder/Sidebar/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+    <>
+      <Box flexDirection="column" gap={16}>
+        {slice.model.variations.map((v) => (
+          <SharedSliceCard
+            action={{ type: "menu" }}
+            key={v.id}
+            mode="navigation"
+            onUpdateScreenshot={() => {
+              onOpenModal({
+                sliceFilterFn: (s) => s,
```
angeloashmore, line 39: I'm not exactly sure what `sliceFilterFn` does, but since this is an identity function, do we need to pass it?

packages/slice-machine/lib/builders/SliceBuilder/Sidebar/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+  return (
+    <>
+      <Box flexDirection="column" gap={16}>
+        {slice.model.variations.map((v) => (
+          <SharedSliceCard
+            action={{ type: "menu" }}
+            key={v.id}
+            mode="navigation"
+            onUpdateScreenshot={() => {
+              onOpenModal({
```
angeloashmore: Style: `onOpenModal` doesn't convey what modal it is connected to. If we trace the code, we can see it comes from `useScreenshotChangesModal()`. Using something like the following may be easier for others to more quickly understand:

```ts
const screenshotChangesModal = useScreenshotChangesModal();

// ...

screenshotChangesModal.onOpenModal({ ... })
```

Totally your call though!

packages/slice-machine/lib/builders/SliceBuilder/Sidebar/index.tsx
bapmrl: `sliceFilterFn` is required. I could try to refactor the `onOpenModal` function (which is more complicated than necessary 😅), but I prefer to limit risks when handling older code I don't know. When we'll implement redesigned dialogs, we'll use the Editor's `Dialog` component (based on Radix UI's `Dialog` primitive) and it will work in a completely different way ^^

packages/slice-machine/lib/builders/SliceBuilder/Sidebar/index.tsx
bapmrl: You're right! I've made the change 😊

bapmrl:
> * ❓ When creating a new slice variation, should the new variation be selected? It currently stays on the variation selected when clicking the "Add a new variation" button.

It was the case before and I've just fixed it. Thank you for noticing this!

> * ❓ The `NewField` component looks awkward with the new wrapping. The name and ID fields don't align, for example. Do we have a proper design?

The change on the `NewField` component isn't great, but I didn't want to spend too much time on it knowing we'll eventually reimplement it. It's more like a quick/temporary fix because I know a lot of people are complaining because inputs are ridiculously small when creating new fields 😅

But we do have a [new design](https://www.figma.com/file/wb7ujCJ4j7EiP7aDDGRYXB/SliceMachine%E2%84%A2-UI?type=design&node-id=5698-692481&mode=dev) (I don't know when we'll implement it). Benjamin has also designed responsive pages and I'll ask Côme when we can work on that. Feel free to comment on Figma ^^ I'm sure Benjamin will be more than happy to discuss his work with you.

>   As an aside, could we change "field ID" to "Field ID"?

I've made the change.

PR #1180, opened 2023-10-19
fix(plugin-kit): ignore empty slice folders (DT-1589)

lihbr:
## Context

See: DT-1589

## The Solution

Refactored the shared `readSliceLibrary` helpers from the plugin kit to adopt the same strategy as `readCustomTypeLibrary` to deal with empty folders.

## How to QA

1. Get a project up and running with the latest Slice Machine alpha
2. Create an empty folder inside the `slices` folder
3. Restarting Slice Machine should yield no error

## Impact / Dependencies

This change will be propagated to adapters as the plugin kit is used by all of them for that kind of operations.

- [x] The CI is successful.

### DIFF
--- a/packages/plugin-kit/src/createSliceMachinePluginRunner.ts
+++ b/packages/plugin-kit/src/createSliceMachinePluginRunner.ts
@@ -142,7 +142,8 @@ export class SliceMachinePluginRunner {
 				)(resolve);
 				plugin = raw.default || raw;
 			} catch (error) {
-				if (import.meta.env.DEV) {
+				// Only log in development and when plugin is unavailable through native plugins library (to prevent noise)
+				if (import.meta.env.DEV && !(resolve in this._nativePlugins)) {
 					console.error(error);
 				}
 			}
--- a/packages/plugin-kit/src/createSliceMachinePluginRunner.ts
+++ b/packages/plugin-kit/src/createSliceMachinePluginRunner.ts
@@ -142,8 +142,11 @@ export class SliceMachinePluginRunner {
 				)(resolve);
 				plugin = raw.default || raw;
 			} catch (error) {
-				// Only log in development and when plugin is unavailable through native plugins library (to prevent noise)
-				if (import.meta.env.DEV && !(resolve in this._nativePlugins)) {
+				// Only log in development, but not during tests when a native plugin matches.
+				if (
+					import.meta.env.DEV &&
+					!(import.meta.env.TEST && resolve in this._nativePlugins)
+				) {
 					console.error(error);
 				}
 			}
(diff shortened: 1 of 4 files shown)


packages/plugin-kit/src/createSliceMachinePluginRunner.ts
```diff
@@ -142,7 +142,8 @@ export class SliceMachinePluginRunner {
 				)(resolve);
 				plugin = raw.default || raw;
 			} catch (error) {
-				if (import.meta.env.DEV) {
+				// Only log in development and when plugin is unavailable through native plugins library (to prevent noise)
+				if (import.meta.env.DEV && !(resolve in this._nativePlugins)) {
```
xrutayisire: ❓ Why kind of noise we are talking here. I tend to like having errors always display, so you don't miss a problem.

xrutayisire (commented):
Super nice 🙂 
I have one small question for the console.error

packages/plugin-kit/src/createSliceMachinePluginRunner.ts
lihbr: This kind of noise: https://github.com/prismicio/slice-machine/actions/runs/6575577367/job/17863112184?pr=1178, the unit test logs are unreadable because all tests are relying on native plugins.

From my understanding it only happens on this package's test because when running Vitest the code is in dev mode, but maybe there's a better way? (I considered using `!import.meta.env.TEST` at first, but thought native plugins were cleaner)

To my understanding, we added this line to make clearer while developing what prevented a plugin from loading which is unrelated to plugin testing with native plugins(?)

xrutayisire: I agree with your last question. But are we sure it will be the only error that can be thrown here?
If yes, your change is perfect. 
If not, maybe we could add a condition to not display it specifically in test (process.env.VITEST for example)?

lihbr: We can do both yes, something like that if that makes more sense:

```ts
// dev but not tests
import.meta.env.DEV && !import.meta.env.TEST

// dev but not tests when matching a native plugin
import.meta.env.DEV && !(import.meta.env.TEST && resolve in this._nativePlugins)
```

xrutayisire: I'm ok for either of the two personally 👍

lihbr: Done 👌

xrutayisire (approved):
LGTM, nice work 🙂

PR #1160, opened 2023-10-10
feat: fetch list of repository environments (DT-1646)

angeloashmore:
## Context

[DT-1646](https://linear.app/prismic/issue/DT-1646/aadev-i-want-to-consolelog-the-list-of-environments-in-repositoryname)

## The Solution

- Add a `manager.prismicRepository.fetchEnvironments()` method that returns a list of the project's environments.
- Add an `<Environments>` component to Slice Machine UI's sidebar that logs the environments.

## Impact / Dependencies

This PR will grow to support a dropdown allowing developers to switch between their repository's environments. After switching, the project will be adapted to the selected environment.

## Checklist before requesting a review

- [ ] I hereby declare my code ready for review.
- [x] If it is a critical feature, I have added tests.
- [x] The CI is successful.
- [x] If there could backward compatibility issues, it has been discussed and planned.

PR #1182, opened 2023-10-20
feat: support environments in `@slicemachine/plugin-kit`, `@slicemachine/manager`, and adapters (DT-1650, DT-1721)

angeloashmore:
> [!IMPORTANT]  
> This PR is based on #1160.

## Context

[DT-1650](https://linear.app/prismic/issue/DT-1650/[manager]-aauser-i-want-to-update-my-framework-specific-env-file-with), [DT-1721](https://linear.app/prismic/issue/DT-1721/aadev-i-want-to-use-correct-types-for-environments-list)

## The Solution

- [x] Add two new hooks:
    - [x] `project:environment:read`
    - [x] `project:environment:update`
- [x] Add two new manager methods:
    - [x] `manager.project.readEnvironment()`
    - [x] `manager.project.updateEnvironment()`
- [x] Add two new `@slicemachine/plugin-kit/fs` helpers:
    - [x] `readProjectEnvironment()`
    - [x] `writeProjectEnvironment()`
- [x] Implement the new hooks in the adapters:
    - [x] `@slicemachine/adapter-next`
    - [x] `@slicemachine/adapter-nuxt`
    - [x] `@slicemachine/adapter-nuxt2`
    - [x] `@slicemachine/adapter-sveltekit`

## Impact / Dependencies

This work will coincide with work on the Slice Machine API.

- [x] The CI is successful.


packages/plugin-kit/src/createSliceMachinePluginRunner.ts
```diff
@@ -48,6 +48,9 @@ export const REQUIRED_ADAPTER_HOOKS: SliceMachineHookTypes[] = [
 	"custom-type:asset:read",
 	"custom-type:asset:delete",
 	"custom-type-library:read",
+	"project:init",
+	"project:environment:read",
+	"project:environment:update",
```
lihbr: Have we considered possibly making those hooks optional in order not to raise the bar to implement new minimal adapters?

lihbr (commented):
Just a pre-review, but it looks good to me so far :)

For Nuxt it'll be able to work similarly, the environment variable will be: `NUXT_PUBLIC_PRISMIC_ENDPOINT`

Happy to do it, but if you want to give it a shot, sure thing!

As you said, it'll be nice to abstract the env-reading utils to `plugin-kit/fs`

angeloashmore:
@lihbr Thanks for the pre-review! I implemented support in all adapters except replacing `repositoryName` with the environment variable in the Nuxt and Nuxt 2 adapters.

Could you help implement it those adapters' `project:init` hook?

packages/plugin-kit/src/createSliceMachinePluginRunner.ts
angeloashmore: They are optional now. 🙂 However, with the possible introduction of dev-only environments, we may want to promote it to a required hook in the future. Dev environments may be a feature we want to push developers to use.

packages/plugin-kit/src/fs/writeProjectEnvironment.ts
```diff
@@ -0,0 +1,62 @@
...
+		? await readProjectFile({
+				filename: args.filename,
+				helpers: args.helpers,
+				encoding: "utf8",
+		  })
+		: "";
+
+	const hasExistingVariable = new RegExp(`^${args.variableName}=.*$`, "m").test(
+		contents,
+	);
```
lihbr: Would it be "safer" to use `dotenv.parse` instead here?

Not sure since there's no real workaround for after 🤔 Too bad `dotenv.stringify` does not exists.

packages/adapter-nuxt/test/plugin-project-environment-update.test.ts
```diff
@@ -0,0 +1,101 @@
+import { expect, test } from "vitest";
+import { createSliceMachinePluginRunner } from "@slicemachine/plugin-kit";
+import * as fs from "node:fs/promises";
+import * as path from "node:path";
+
+import adapter from "../src";
+
+test("writes the environment to default the env file", async (ctx) => {
```
hypervillain: I think it should read "to the default env file"

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
```diff
@@ -423,6 +424,29 @@ export class PrismicRepositoryManager extends BaseManager {
 		}
 	}
 
+	async fetchEnvironments(): Promise<Environment[]> {
+		const repositoryName = await this.project.getRepositoryName();
+
+		const url = new URL("./environments", API_ENDPOINTS.PrismicWroom);
```
hypervillain: `PrismicCoreAPI` rather than `PrismicWroom`

lihbr (approved):
LGTM!

There's nothing to update for Nuxt init hooks here since the client is initialized within the module, I'll update them and link to this PR.

packages/manager/src/managers/prismicRepository/types.ts
```diff
@@ -155,3 +155,25 @@ export type StarterId =
 	| "nuxt_multi_page"
 	| "nuxt_blog"
 	| "nuxt_multi_lang";
+
+export const Environment = t.type({
```
hypervillain: I'm curious if the Core API might be typed somewhere already and if it would make sense to use that type here or before (like at query time). Also some data here might be restricted for us, eg. `pendingUsers`.
Let's ask around

lihbr:
Nuxt 3 support: https://github.com/nuxt-modules/prismic/pull/202 (v3.1.0)
Nuxt 2 support: https://github.com/nuxt-modules/prismic/pull/203 (v1.5.0)

packages/adapter-nuxt/test/plugin-project-environment-update.test.ts
angeloashmore: Fixed!

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
angeloashmore: We don't have a `PrismicCoreAPI` value currently. What URLs should be used for production, staging, and development?

For reference, this is where we define the URLs: https://github.com/prismicio/slice-machine/blob/7c995fe301c4c7ae41faa433a339ce6cb3340ceb/packages/manager/src/constants/API_ENDPOINTS.ts#L14-L77

packages/plugin-kit/src/fs/writeProjectEnvironment.ts
angeloashmore: I figured we could get away with a RegExp, but you're right. `dotenv` would be safer, even if the likelihood of the RegExp failing is low.

I switched the RegExp for `dotenv.parse`.

We wouldn't want to do a `dotenv.parse` then `dotenv.print` (if it existed) because we would probably lose whitespace and comments. It's the same problem with parsing and printing a TypeScript file with `tsc`.

packages/manager/src/managers/prismicRepository/types.ts
angeloashmore: Ideally, I think we'd only want to type and check for the properties we need within Slice Machine. For example, if we aren't going to use `pendingUsers`, we could omit it from the type.

I think this would decouple Slice Machine file external libraries (like the Core API), making it easier to maintain with greater flexibility. It also would ensure we could continue using `io-ts` (like all other data validation in SM) without caring with the Core API uses.

What do you think?

angeloashmore:
@lihbr In the Nuxt modules, how does the environment get passed into the `environment` option? Wouldn't we need to reference the `NUXT_PUBLIC_PRISMIC_ENVIRONMENT` environment variable somewhere?

Should that be something we add as a default option or as part of `project:init`'s templates?

packages/plugin-kit/src/fs/writeProjectEnvironment.ts
lihbr: Oh, right, great then!

lihbr:
It gets automatically loaded by Nuxt through runtime config: https://nuxt.com/docs/api/composables/use-runtime-config#environment-variables ☺️

`NUXT_PUBLIC_PRISMIC_ENVIRONMENT` maps to `useRuntimeConfig().public.prismic.environment`

lihbr (approved):

angeloashmore:
> @lihbr: It gets automatically loaded by Nuxt through runtime config: https://nuxt.com/docs/api/composables/use-runtime-config#environment-variables ☺️
> 
> `NUXT_PUBLIC_PRISMIC_ENVIRONMENT` maps to `useRuntimeConfig().public.prismic.environment`

Oh clever! I see it now. It's nice that that `PRISMIC_` maps to the existing config.

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
hypervillain: I'm still working on the lambda so I don't have production URL yet :/
I just thought PrismicCoreAPI is a more explicit name

packages/manager/src/managers/prismicRepository/types.ts
hypervillain: Perfect, I agree

lihbr (approved):
LGTM, really nice work!

angeloashmore:
Will be merged into a separate PR.

PR #1183, opened 2023-10-20
feat(deps): upgrade to prettier 3 (#1055,DT-1614)

lihbr:
## Context

Prettier 2 is used to format code within the manager and adapters. It is not ESM-fully-capable, leading to issues for users using more recent, ESM-only, prettier plugins.

A reminder that CommonJS sunset with Node 16 EOL last September. The whole Node.js community is on its last mile getting out of CJS/ESM hell.

This PR fixes: #1055, DT-1614

## The Solution

Prettier 3 now support ESM properly: https://prettier.io/blog/2023/07/05/3.0.0

This PRs updates Prettier to version 3 in `@slicemachine/plugin-kit` and `@slicemachine/manager` to format user code.

Because we also use Prettier within Slice Machine monorepo, we also take advantage of this update to update our Prettier version to version 3 on the whole repository.

Prettier 3 comes with some changes regarding its default formatting preferences. To reduce the amount of impacted changes on our end I reverted [the new `trailingComma` behavior in Prettier version 3](https://prettier.io/blog/2023/07/05/3.0.0#javascript) to the one of version 2, bringing down the amount of impacted files from 300-ish to 20-ish.

I recommend looking at the commits to better understand the changes:
- https://github.com/prismicio/slice-machine/commit/6163561ce021a7b36a51c478ce4707354919526f - All the `package.json` updates to use Prettier 3 on the monorepo
- https://github.com/prismicio/slice-machine/commit/4d5b250f602315fa8e544d369671fd4808d34e20 - All the programmatic changes required in code by Prettier 3, tl;dr; `prettier.format` now returns a promise that needs to be awaited.
- https://github.com/prismicio/slice-machine/commit/ecead9971dab87a379eeaffd668c4fd7bd502399 - All the style changes coming with Prettier 3, you can see that I added a root `.prettierrc` to bring back the old behavior of trailing comma. Eventually, we'll want to get rid of it.
- https://github.com/prismicio/slice-machine/pull/1183/commits/e9ba134e77ea05dd6eac232447413682ad8b35ef - Updates tests to accommodate Prettier 3 ESM bundle.

## How to QA DT-1614

In `e2e-projects/next-upgrade`:
1. Install `prettier` and `prettier-plugin-tailwindcss` v0.4 or v0.5
   ```bash
   yarn workspace next-upgrade add -D prettier prettier-plugin-tailwindcss
   ```
2. Create or update Prettier config to use `prettier-plugin-tailwindcss`
   ```js
   module.exports = {
      plugins: ['prettier-plugin-tailwindcss'],
   }
   ```
3. Update `packages/adapter-next/src/hooks/slice-create.ts` to create slices with some shuffled TailwindCSS classes (line 62)
   ```diff
   + className="px-4 text-white py-2"
   ```
4. Launch dev script and dev server
   ```bash
   # root
   yarn dev

   # e2e-projects/next-upgrade
   npx cross-env NODE_ENV=development start-slicemachine
   ```
5. Go to Slice Machine UI and create a slice
   ✅ Slice should be created without errors and formatted correctly (`px-4` and `py-2` should have been grouped together)

## Impact / Dependencies

Change should be transparent to users. However: **Do we want to catch further issues with formatting applied by Slice Machine to just warn about an error instead of failing hard?**

- [x] The CI is successful.


.prettierrc
```diff
@@ -0,0 +1,3 @@
+{
+  "trailingComma": "es5"
```
xrutayisire: ❓ Why not keeping default prettier config also for that?

packages/start-slicemachine/.prettierrc
xrutayisire, line 1: ℹ️ Just informative here, but personally I always prefer to keep 100% prettier default conf. It's here for that, to not care and discuss what each other prefer. I feel like it's always simpler not to have a `.prettierrc` file.

xrutayisire (approved):
LGTM, super cool for our users! Following a tracking I did, it happened to quite a lot of people. Thanks a lot 👍

Just asked a small question about "es5" setting.

.prettierrc
lihbr: From the above:
> Prettier 3 comes with some changes regarding its default formatting preferences. To reduce the amount of impacted changes on our end I reverted [the new `trailingComma` behavior in Prettier version 3](https://prettier.io/blog/2023/07/05/3.0.0#javascript) to the one of version 2, bringing down the amount of impacted files from 300-ish to 20-ish.

Happy to turn it on however if we want to merge that quickly (if not it will pile up conflicts)

xrutayisire: I see! You right. I'm ok to keep it like this for now. I guess it can come with the change on not having override on prettier one day 😇
But anyway I agree, you decide

PR #1186, opened 2023-10-23
fix(init): Rework init end message

xrutayisire:
## Context

- Complete DT-1698

## The Solution

- One path for all, only propose to run Slice Machine at the end

## Impact / Dependencies

- Users will always start slice machine directly and stop going sometimes to the repository

- [ ] The CI is successful.

## Preview

Screenshot 2023-10-23 at 14 12 21

### DIFF
--- a/packages/init/src/SliceMachineInitProcess.ts
+++ b/packages/init/src/SliceMachineInitProcess.ts
@@ -198,106 +198,38 @@ export class SliceMachineInitProcess {
 		console.log(
 			`\n${chalk.bgGreen(` ${chalk.bold.white("Slice Machine")} `)} ${chalk.dim(
 				"→",
-			)} Initialization successful!`,
+			)} Initialization successful!
+			
+Continue with next steps in Slice Machine.
+`,
 		);
 
 		try {
-			const runSmCommand = this.context.projectInitialization?.patchedScript
-				? await getRunScriptCommand({
-						agent: this.context.packageManager || "npm",
-						script: "slicemachine",
-				  })
-				: await getExecuteCommand({
-						agent: this.context.packageManager || "npm",
-						script: "start-slicemachine",
-				  });
+			if (this.options.startSlicemachine) {
+				const runSmCommand = this.context.projectInitialization?.patchedScript
+					? await getRunScriptCommand({
+							agent: this.context.packageManager || "npm",
+							script: "slicemachine",
+					  })
+					: await getExecuteCommand({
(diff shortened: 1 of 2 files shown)


packages/init/src/SliceMachineInitProcess.ts
```diff
@@ -198,106 +198,38 @@ export class SliceMachineInitProcess {
...
 					"startSlicemachine"
 				>({
 					type: "confirm",
 					name: "startSlicemachine",
-					message: finalSmCommand.message,
+					message: `Run Slice Machine (${chalk.cyan(runSmCommand)})?`,
 					initial: true,
 				});
 
 				if (startSlicemachine) {
```
lihbr, line 229: If the answer is no, should we tell users how to start Slice Machine still?

packages/init/src/SliceMachineInitProcess.ts
```diff
@@ -198,106 +198,38 @@ export class SliceMachineInitProcess {
...
-	`);
-
-			if (this.options.starter && this.options.repository) {
-				const { openDashboard } = await prompt<boolean, "openDashboard">({
-					type: "confirm",
-					name: "openDashboard",
-					message: "Would you like to open your repository?",
-					initial: true,
-				});
+			if (this.options.startSlicemachine) {
```
lihbr, line 208: If the flag is set to false, shouldn't we tell users how to start Slice Machine?

lihbr (approved):
LGTM, just some small suggestions in case user don't want to start right away 🤔

packages/init/src/SliceMachineInitProcess.ts
xrutayisire: Hey, on that one I don't think it's necessary since it's displayed in the question, no?

packages/init/src/SliceMachineInitProcess.ts
xrutayisire: I think we added that to be consistent but I don't think anyone is using that, WDYT?
For me if it's used it could be something related to CI, and so no need to display the command, no?

lihbr: OK, fair enough yes! That's what I assumed.

packages/init/src/SliceMachineInitProcess.ts
lihbr: Oh, right! My bad, all good then ☺️

PR #1196, opened 2023-10-27
feat: add environment selector (DT-1649)

angeloashmore:
> [!IMPORTANT]
> This PR is based on #1160 and #1182.

## Context

DT-1649

## The Solution

Add a dropdown menu to change the active environment.

## Impact / Dependencies

This PR builds on the following PRs:

- https://github.com/prismicio/slice-machine/pull/1160
- https://github.com/prismicio/slice-machine/pull/1182

- [ ] The CI is successful.


packages/slice-machine/src/components/SideNav/SideNavEnvironmentSelector.tsx
```diff
@@ -0,0 +1,142 @@
...
+  );
+
+  return (
+    <DropdownMenu modal>
+      <DropdownMenuTrigger>
+        <Button startIcon="unfoldMore" variant="secondary" />
+      </DropdownMenuTrigger>
+      <DropdownMenuGroupContent align="end">
+        <DropdownMenuGroupLabel>Regular Environments</DropdownMenuGroupLabel>
+        {productionEnvironment ? (
```
bapmrl: 💡 I think there's probably a way to replace these 3 `map`s with a single one and make the code simpler. What about something like the following?
```
environments
  .sort(/* place production first, then rest and finally dev */)
  .map((environment) => (
    <>
      {environment.type === "dev" ? (
        <DropdownMenuGroupLabel>Personal Environments</DropdownMenuGroupLabel>
        <EnvironmentDropdownMenuItem
            key={environment.domain}
            {...environment}
          />
      ) : undefined}
    </>
  ));
```

packages/slice-machine/src/components/SideNav/index.ts
```diff
@@ -1 +1,2 @@
 export * from "./SideNav";
```
bapmrl, line 1: 💡 I'm not found of the `SideNav` component. I think it's too tied to Slice Machine's business concerns and would have preferred something similar to what I did with the `Card` component. If you take a look at it, you'll notice I never mention slices in the `.tsx` and `.css.ts` files. It's only in the `.stories.tsx` file, by composing the "general"  props and subcomponents of `Card`, that I build cards for slices. This approach is better, in my opinion, because we could easily imagine using the `Card` component in other contexts or moving it to the Editor's repository. It's more flexible and better separates responsibilities. A `Card` should only be concerned with laying out texts, images, data... whatever data it may be. Styles and React (sub)components are already super difficult to write correctly and I think we don't have to solve all problems (meaning, domain concerns) in the same components. So, I would recommend new components in `src/components` follow the approach I've taken with the `Card` (focusing on styles and pure UI concerns) 😊 Of course, it can be discussed within the team.

→ All that said, the `SideNav` component wasn't developed this way and what you did gives something consistent... you can't rewrite the whole `SideNav` so the path you've taken is the best one in this case 😊

packages/slice-machine/src/components/SideNav/SideNavEnvironmentSelector.tsx
```diff
@@ -0,0 +1,142 @@
...
+  name: string;
+  type: "prod" | "stage" | "dev";
+  domain: string;
+};
+
+type SideNavEnvironmentSelectorProps = {
+  environments: Environment[];
+};
+
+export const SideNavEnvironmentSelector: FC<SideNavEnvironmentSelectorProps> = (
```
bapmrl: 💡 I think it's far more natural to have code naturally flow from top to bottom, rather than bottom to top. The first things I would expect to read in this file are `SideNavEnvironmentSelectorProps`, then `SideNavEnvironmentSelector`, then the components that `SideNavEnvironmentSelector` uses, etcetera.

packages/slice-machine/src/components/SideNav/SideNavEnvironmentSelector.css.ts
```diff
@@ -0,0 +1,54 @@
...
+    boxSizing: "border-box",
+    fontFamily: "body",
+    all: "unset",
+  }),
+]);
+
+const blockWithDisplayRevert = style([
+  block,
+  sprinkles({
+    display: "revert",
```
bapmrl: 💡 I personally tend to only use `display: "flex"`, `display: "grid"` (when necessary) and `display: "revert"` for `img`, `svg`... I'm basically following the same reset strategy as https://github.com/elad2412/the-new-css-reset/blob/main/css/reset.css#L11.

packages/slice-machine/src/components/DropdownMenuGroup/DropdownMenuGroup.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+);
+
+const Template: StoryFn<typeof DropdownMenu> = () => (
+  <div style={{ display: "flex", gap: 16 }}>{DefaultDropdownMenu}</div>
+);
+
+export const Default = {
+  render: Template,
+  name: "DropdownMenuGroup",
+  args: { children: null },
```
bapmrl: ❓ Why do you have to pass `null` to `args.children`?

packages/slice-machine/src/components/DropdownMenuGroup/DropdownMenuGroup.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+    </DropdownMenuGroupContent>
+  </DropdownMenu>
+);
+
+const Template: StoryFn<typeof DropdownMenu> = () => (
+  <div style={{ display: "flex", gap: 16 }}>{DefaultDropdownMenu}</div>
+);
+
+export const Default = {
+  render: Template,
```
bapmrl: ❓ Is there a way to avoid using the `render` property? It's the first story where it's being used. I think you did that do be able to pass `style={{ display: "flex", gap: 16 }}`? Shouldn't this be handled by the component's implementation?

packages/slice-machine/src/components/DropdownMenuGroup/DropdownMenuGroup.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+type DropdownMenuGroupItemProps = ComponentProps<typeof DropdownMenuItem>;
+
+export const DropdownMenuGroupItem: FC<DropdownMenuGroupItemProps> = (
+  props,
+) => {
+  return (
+    <DropdownMenuItem
+      {...props}
+      // TODO: How do I fix this? - Angelo
```
bapmrl: ⚠️ I'm very sorry Angelo that you had to start without any input/help from the team. This implementation raises a lot of questions in fact... and you couldn't know what to do.

As a general rule, we don't want to fork components that already have an implementation in the `@prismicio/editor-ui` package. It means, in this case, either Benjamin needs to use the existing version (without insets and with dividers) or we need to update the component in the Editor's repository. I've called Anthony and he told me he'll discuss this specific dropdown issue with Benjamin... in the meantime, I'm not sure what you can do 😥 Depending on the result of this discussion, you'll either have to delete this `DropdownMenuGroup` component or add new props to the Editor's `DropdownMenu` component (which will take you some time, as the code isn't necessarily easy and it's being used, for instance, by the slash commands — so you'd need to be cautious to avoid regressions). If you end up having to add props, I would recommend inspiring from [GitHub Primer's implementation](https://primer.style/components/action-list/react) and adding a `variant?: "inset" | "full"` prop and a `showDividers?: boolean` prop.

To summarise:
1. When you have to implement a component, if it already exists in the `@prismicio/editor-ui` package, don't fork it. Either discuss with Benjamin so that he uses the existing component in Slice Machine's Figma (if it makes sense), or update the Editor's implementation to match Benjamin's new design. When the change seems a bit involved, it's good to come with a proposal and present it to the Editor team (I usually do this with Alex).
2. When the component you have to implement doesn't already exist in either `@primicio/editor-ui` or Slice Machine's `src/components`, discuss with Benjamin and/or Anthony (and maybe Alex if it's more technical than design) to assess whether it could be useful to the Editor. If it can be and doesn't take much more time, we should implement it in the Editor as it's currently acting as the shared package (design system). If it isn't useful to the Editor (because it's specific to Slice Machine, like the `SideNav`), implement it in Slice Machine's `src/components`.

All these discussions should happen while the corresponding Linear issues are in *Specify → Active*. I know you couldn't wait, but it should be a team priority to unblock you (and anybody else) and reduce risks/unknowns before implementing anything.

bapmrl (changes requested):
Hello Angelo,

That's excellent work, but you'll see in my comments that we have to discuss with Benjamin and Anthony to know how to proceed. I'm not sure how to help you right know 😕 Monday I'll ask them if they came up with a decision and I'll tell you.

Also, if you think it can be useful and you need it, I can take some time on Monday evening (after the retro) to discuss with you about the overall strategy around implementing UI components (and maybe even the current state of the Slice Machine's front end). Would it be helpful to you?

Thank you,

bapmrl:
⚠️ Benjamin has just updated the Figma and it should be feasible with the current implementation of the Editor (or by opening a very small PR on their repository). Don't hesitate if you face any other issue!

packages/slice-machine/src/features/environments/useEnvironments.ts
```diff
@@ -1,23 +1,102 @@
...
+  await managerClient.project.updateEnvironment({
+    environment: environment.domain,
+  });
+  revalidateData(getActiveEnvironmentDomain, []);
+}
+
 export function useEnvironments() {
   const environments = useRequest(getEnvironments, []);
+  const activeEnvironmentDomain = useRequest(getActiveEnvironmentDomain, []);
+  const repositoryName = useRequest(getRepositoryName, []);
```
lihbr: Cannot this make use of `useRepositoryName()` instead?

angeloashmore: `useRepositoryName()` was created temporarily. It has been removed (as well as the commented-on lines).

PR #1197, opened 2023-10-27
chore: fix cypress-setup.sh

bapmrl:
Our E2E tests can no longer run because:
1. Next.js 14 has just been released.
2. Our `cypress-setup.sh` script uses a version of `create-next-app` (version `13.4.6`) that always installs the latest version of Next.js.
3. Our `@slicemachine/adapter-next` package isn't compatible with Next.js 14.

Changing the version of `create-next-app` to `13.5.4` solves the issue, as it no longer installs the latest version of Next.js, but the version listed in its `package.json` file.

### DIFF
--- a/cypress-setup.sh
+++ b/cypress-setup.sh
@@ -26,7 +26,7 @@ else
 fi
 
 rm -rf e2e-projects/cypress-next-app \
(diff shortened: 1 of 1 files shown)


xrutayisire (approved):
LGTM ✅

PR #1198, opened 2023-10-27
feat(tutorial): Update tutorial tooltip condition

xrutayisire:
## Context

- Completes DT-1699

## The Solution

- Copy update
- Change condition to display tooltip

## Impact / Dependencies

- None

- [x] The CI is successful.


packages/slice-machine/components/Navigation/VideoItem.tsx
```diff
@@ -44,7 +48,7 @@ export const VideoItem = forwardRef<HTMLLIElement, VideoItemProps>(
       >
         <SideNavListItem ref={ref}>
           <SideNavLink
-            title="Tutorial"
+            title={isNext ? "Academy" : "Tutorial"}
```
lihbr: That kind of pattern is _really_ something we should avoid. It's fine as a shortcut here but ideally we'd be looking at using some kind of `documentation:read` hook instead to get that data from adapters.

Can we add a comment warning about that here (or line 40) so that future developers aren't tempted to reuse this pattern?

lihbr (approved):
LGTM to me, happy to QA it in an alpha

Maybe someone more experienced with React should also have a look?

packages/slice-machine/components/Navigation/VideoItem.tsx
xrutayisire: That's a super interesting topic because with @bapmrl we agree with you 😛

But @hypervillain, I think you told us that for this video tooltip we should not rely on a hook and that's the responsibility of Slice Machine UI. Correct?

hypervillain: Yes. Of course if it gets unmanageable I would reconsider it, but I don't think Marketing links should be part of adapters, for plenty of reasons

lihbr: I can see a world where this makes some sense for marketing copy OK.

Maybe scoping it under some kind of general explicit hook/helper would at least help make it clearer and scale better.

(example hook)
```typescript
// src/hooks/useMarketingCopy.ts
import { useAdapterName } from "./useAdapterName";

// Not sure if we have one of these somewhere already
type Framework = "next" | "nuxt" | "nuxt2" | "sveltekit";

type UseMarketingCopyArgs<T> = [specificCopy: Record<Framework, T>, fallbackCopy: T];

/**
 * This hook is only meant to be used for marketing copy/marketing specific tasks.
 * Rely on adapter's `documentation:read` hooks or other specific hooks to display
 * other kind of context-aware information to users.
 */
export function useMarketingCopy<T>(...args: UseMarketingCopyArgs<T>): T {
	const [specificCopy, fallbackCopy] = args;

	const adapterName = useAdapterName();
	const framework = adapterName.replace("@slicemachine/adapter-", "") as Framework;

	return framework in specificCopy ? specificCopy[framework] : fallbackCopy;
}
```

(example usage)
```ts
// packages/slice-machine/components/Navigation/VideoItem.tsx
const copy = useMarketingCopy(
	{
		next: {
			title: "Academy",
			video: "https://example.com/next",
			url: "https://prismic.io/academy/prismic-and-nextjs"
		}
	},
	{
		title: "Tutorial",
		video: "https://example.com/prismic",
		url: "https://prismic.io/docs"
	},
);

// Alternative
const title = useMarketingCopy({ next: "Academy" }, "Tutorial");
```

xrutayisire: I really like it as a middle ground. I will use your proposal 🙂

lihbr: (might need some tweaking as I just wrote that there, but cool!)

xrutayisire: Done 🙂
It should be easy with my proposal to drop and replace by an adapter hook one day if we feel it's too complicated to maintain in Slice Machine UI.

lihbr (approved):
LGTM

PR #1205, opened 2023-10-31
feat(logs): Add more logs for Sentry

xrutayisire:
## Context

- Completes DT-1718

## The Solution

- Increase the max value of Sentry logs as it was clearly too low for some errors
- Add `cause` on most of the endpoints that can benefit more information in Sentry

## Impact / Dependencies

- Checking limitation of Sentry logs

- [x] The CI is successful.


lihbr (approved):
LGTM! Thanks for taking that :)

PR #1209, opened 2023-11-07
fix(init): limit the amount of custom types created concurrently (DT-1758)

lihbr:
## Context

Since https://github.com/prismicio/slice-machine/releases/tag/slice-machine-ui%401.17.0, Slice Machine init command downloads remote custom types during the init process so that users can work on them locally.

When users have many custom types (60+), the init command can fail to allocate memory as it was creating all custom types concurrently which is a costly operation.

Resolves: DT-1758

## The Solution

This PR limits the rate at which custom types are created by the init command so that no more than 8 custom types are being created at the same time.

## How to QA

1. Create a repository with many custom types (60+) or ask me for a clone of such a repository
2. Try running the current init command inside a project, it **should** fail
```bash
npx create-next-app@latest
npx @slicemachine/init@latest
```
3. Try running the alpha init command, it **should** succeed
```bash
npx create-next-app@latest
npx @slicemachine/init@alpha
```

## Impact / Dependencies

/

- [ ] The CI is successful.


lihbr:
Side notes:

We've been fixing multiple load issues recently using `p-limit` and other strategies. I'm considering making it more global by updating the [`HookSystem`](https://github.com/prismicio/slice-machine/blob/54a538b74763294f74647bd8c4155f52897bdeac/packages/plugin-kit/src/lib/HookSystem.ts) class to support an optional map of options for hooks. This would allow us to say something like this:

```typescript
import { HookSystem } from "./lib/HookSystem";
import type { SliceMachineHooks } from "./types";

/**
 * @internal
 */
export const createSliceMachineHookSystem =
	(): HookSystem<SliceMachineHooks> => {
		return new HookSystem<SliceMachineHooks>({
			"slice:create": { concurrencyLimit: 8 },
		});
	};
```

What do you think? @angeloashmore, you might have an opinion on this 🤔

package.json
```diff
@@ -75,4 +75,4 @@
     "e2e-projects/next-upgrade",
     "packages/*"
   ]
-}
+}
```
bapmrl: ❓ Why this change? Running `yarn install` after this pull request is merged might generate an unnecessary diff, no?

packages/init/src/SliceMachineInitProcess.ts
```diff
@@ -1286,9 +1287,17 @@ ${chalk.cyan("?")} Your Prismic repository name`.replace("\n", ""),
...
 									await this.manager.customTypes.fetchRemoteCustomTypes();
 
+								const createCustomTypeLimit = pLimit(8);
+
+								let pulled = 0;
+								task.title = `Pulling existing types... (0/${remoteTypes.length})`;
 								await Promise.all(
 									remoteTypes.map(async (model) => {
-										await this.manager.customTypes.createCustomType({ model });
+										await createCustomTypeLimit(() =>
```
bapmrl, line 1296: 💡 Apparently, it could also be written:
```
await limit(this.manager.customTypes.createCustomType, { model })
```

It's apparently used to avoid creating unnecessary closures, but we don't need that here ^^

bapmrl (approved):
Excellent! I just have 2 questions.

bapmrl:
> Side notes:
> 
> We've been fixing multiple load issues recently using `p-limit` and other strategies. I'm considering making it more global by updating the [`HookSystem`](https://github.com/prismicio/slice-machine/blob/54a538b74763294f74647bd8c4155f52897bdeac/packages/plugin-kit/src/lib/HookSystem.ts) class to support an optional map of options for hooks. This would allow us to say something like this:
> 
> ```ts
> import { HookSystem } from "./lib/HookSystem";
> import type { SliceMachineHooks } from "./types";
> 
> /**
>  * @internal
>  */
> export const createSliceMachineHookSystem =
> 	(): HookSystem<SliceMachineHooks> => {
> 		return new HookSystem<SliceMachineHooks>({
> 			"slice:create": { concurrencyLimit: 8 },
> 		});
> 	};
> ```
> 
> What do you think? @angeloashmore, you might have an opinion on this 🤔

I'm not super aware of this topic, but searching for `p-limit`, it seems our use of it isn't necessarily tied to calling hooks and can happen anywhere we await a promise?

package.json
lihbr: Yeah, I'll revert it, it's ESLint adding one, Yarn removing it, driving me crazy

packages/init/src/SliceMachineInitProcess.ts
lihbr: I've read about it too! Looking at the README, this way feels like the default one though unless we have performance concerns here beyond that 🤔

lihbr:
Agreed yes, just seems like recent issues were mostly revolving around hooks

bapmrl:
LGTM.

PR #1215, opened 2023-11-16
feat(slice-machine-ui): disable SSR

bapmrl:
Slice Machine UI uses React components from the Editor that aren't tested on the Node.js runtime. This is because they don't use Next.js and we can't ask them to support Server-Side Rendering just for us.

Moreover, Slice Machine UI doesn't take any benefit from SSR as we currently only use Next.js's file system based router and don't need SEO.

Consequently, to prevent further bugs and reduce work for the Editor team, I'm opening this pull request to disable SSR on all pages.

### DIFF
--- a/packages/slice-machine/pages/_app.tsx
+++ b/packages/slice-machine/pages/_app.tsx
@@ -23,7 +23,8 @@ import {
 } from "@prismicio/editor-ui";
 import { ConnectedRouter } from "connected-next-router";
 import type { NextPage } from "next";
-import App, { type AppContext, type AppInitialProps } from "next/app";
+import type { AppContext, AppInitialProps } from "next/app";
+import dynamic from "next/dynamic";
 import Head from "next/head";
 import Router from "next/router";
 import { type FC, type ReactNode, useEffect, useState, Suspense } from "react";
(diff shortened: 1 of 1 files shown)


xrutayisire:
This LGTM, I'm ok with the change. I let other team member comment to ensure we are all aligned on this work.

packages/slice-machine/pages/_app.tsx
```diff
@@ -171,8 +172,4 @@ function MyApp({
   );
 }
 
-MyApp.getInitialProps = async (appContext: AppContext) => {
-  return await App.getInitialProps(appContext);
-};
-
-export default MyApp;
+export default dynamic(() => Promise.resolve(App), { ssr: false });
```
angeloashmore, line 175: Just double checking: is `Promise.resolve()` necessary?

angeloashmore:
LGTM as well. This will prepare us to move off Next.js in the future.

packages/slice-machine/pages/_app.tsx
bapmrl: `dynamic(App)` doesn't compile, but `dynamic(Promise.resolve(App))` and `dynamic(() => Promise.resolve(App))` do. I think the latest is the cleanest, no? Otherwise, with the second version, the runtime would create the promise a bit early (which isn't a real issue).

Here's the documentation page: https://nextjs.org/docs/pages/building-your-application/optimizing/lazy-loading#with-no-ssr.

hypervillain: Yes, the second one looks good

PR #1219, opened 2023-11-23
feat: track screen size

bapmrl:
Completes DT-1723.

### DIFF
--- a/packages/slice-machine/src/hooks/useSMTracker.ts
+++ b/packages/slice-machine/src/hooks/useSMTracker.ts
@@ -71,5 +71,9 @@ async function trackPageView(): ReturnType<typeof telemetry.track> {
     title: document.title,
     referrer: document.referrer,
     adapter,
+    outerWidth: window.outerWidth,
+    outerHeight: window.outerHeight,
+    screenWidth: window.screen.width,
+    screenHeight: window.screen.height,
   });
 }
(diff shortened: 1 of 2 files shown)


packages/slice-machine/src/hooks/useSMTracker.ts
```diff
@@ -71,5 +71,9 @@ async function trackPageView(): ReturnType<typeof telemetry.track> {
     title: document.title,
     referrer: document.referrer,
     adapter,
+    outerWidth: window.outerWidth,
```
xrutayisire, line 74: So for me:

- `window.screen.width`: global screen size of your real screen, will not change on browser resize
- `window.outerWidth`: window size of the browser, will change on browser resize

But the most important would be:

- `window.innerWidth` no? It would give you the actual browser window size of the opened page, if you open dev tools it will give you the space specific for SM. Why not using it?

bapmrl: `window.innerWidth` is expressed in CSS pixels and changes based on the viewport size, indeed, but also based on the screen density and the browser's zoom level. We can't easily get the browser's zoom level, so it would be very hard interpreting the data.

xrutayisire (approved):
LGTM, just small comments on why these in the code could be good.

PR #1220, opened 2023-11-23
fix(sentry): Upgrade Segment package and fix Axios vulnerability

xrutayisire:
## Context

- Completes DT-1780

## The Solution

- Remove deprecated package https://github.com/segmentio/analytics-node
- Use new package https://github.com/segmentio/analytics-next/tree/master/packages/node#readme
- Follow guide https://segment.com/docs/connections/sources/catalog/libraries/server/node/migration/

## Impact / Dependencies

- New Analytics class required this usage

Screenshot 2023-11-23 at 09 52 41

- [x] The CI is successful.

### DIFF
--- a/packages/manager/src/managers/telemetry/TelemetryManager.ts
+++ b/packages/manager/src/managers/telemetry/TelemetryManager.ts
@@ -1,6 +1,6 @@
 import { randomUUID } from "node:crypto";
 
-import SegmentClient from "analytics-node";
+import { Analytics, GroupParams, TrackParams } from "@segment/analytics-node";
 
 import { readPrismicrc } from "../../lib/prismicrc";
 
@@ -41,9 +41,9 @@ type TelemetryManagerContext = {
 };
 
 function assertTelemetryInitialized(
-	segmentClient: SegmentClient | undefined,
+	segmentClient: (() => Analytics) | undefined,
 ): asserts segmentClient is NonNullable<typeof segmentClient> {
-	if (segmentClient == undefined) {
+	if (segmentClient === undefined) {
 		throw new Error(
 			"Telemetry has not been initialized. Run `SliceMachineManager.telemetry.prototype.initTelemetry()` before re-calling this method.",
 		);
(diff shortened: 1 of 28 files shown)


packages/manager/src/managers/telemetry/TelemetryManager.ts
```diff
@@ -41,39 +41,47 @@ type TelemetryManagerContext = {
 };
 
 function assertTelemetryInitialized(
-	segmentClient: SegmentClient | undefined,
+	segmentClient: (() => Analytics) | undefined,
```
bapmrl, line 44: 💡 Would it be better to change all references to segment (and in this case, `segmentClient`) to `analytics`?

packages/manager/src/managers/telemetry/TelemetryManager.ts
```diff
@@ -41,39 +41,47 @@ type TelemetryManagerContext = {
...
-			errorHandler: () => {
+			analytics.on("error", (error) => {
 				// noop - We don't care if the tracking event
 				// failed. Some users or networks intentionally
 				// block Segment, so we can't block the app if
 				// a tracking event is unsuccessful.
-			},
-		});
+				console.error(
+					`An error occurred during telemetry initialization`,
```
bapmrl: ⚠️ It's a callback and errors might be thrown after the initialization.

packages/manager/src/managers/telemetry/TelemetryManager.ts
```diff
@@ -41,39 +41,47 @@ type TelemetryManagerContext = {
...
-		}
+		const isTelemetryEnabled = await this.checkIsTelemetryEnabled();
+
+		this._segmentClient = () => {
+			const analytics = new Analytics({
+				writeKey: API_TOKENS.SegmentKey,
+				// Since it's a local app, we do not benefit from event batching the way a server would normally do, all tracking event will be awaited.
+				maxEventsInBatch: 1,
+				// TODO: Verify that this actually does not send data to Segment when false.
+				disable: isTelemetryEnabled === false,
```
bapmrl: 💡 `!isTelemetryEnabled`.

bapmrl (changes requested):

packages/manager/src/managers/telemetry/TelemetryManager.ts
xrutayisire: I thought exactly the same but before in the old package doc it was already `Analytics`, but they decided to name it `SegmentClient`, I guess for more consistency in our codebase. Before it was am unmanned export vs now a named one but I guess how we are writing it does not depend on that.

So for me, we can keep this name as it doesn't really depend on the package version or type. WDYT?

See: https://github.com/segmentio/analytics-node#usage

packages/manager/src/managers/telemetry/TelemetryManager.ts
xrutayisire: In Slice Machine UI we even have a rule to prevent that usage, right?

I started since I'm at Prismic to do explicit comparison, so I guess we should fix how we want to do, no?

packages/manager/src/managers/telemetry/TelemetryManager.ts
xrutayisire: That's how I want it 😛 If I understood your comment.

What I want is on any Segment error that can happen in the life of the client, I want to console.error. It's documented here: https://segment.com/docs/connections/sources/catalog/libraries/server/node/#error-handling

PR #1224, opened 2023-11-29
fix(slice-machine-ui): environment-related UI bugs (DT-1808)

angeloashmore:
## Context

DT-1808

### `<Divider>` bug

The CSS `color-mix` function is used to make `currentColor` transparent, which is necessary to have a correct transparent gradient. Vanilla Extract (or whatever minifies our CSS) converts `0%` to `0` in production, which is invalid syntax within `color-mix`.

Interestingly, `100%` is not converted to `1`, so we can use `transparent 100%` instead of `currentColor 0%`.

### Top border bug

Default to the `purple` style. This regression accidentally came from a previous refactor.

## Impact / Dependencies

N/A

- [x] The CI is successful.

### DIFF
--- a/packages/slice-machine/src/components/PageLayout/PageLayout.tsx
+++ b/packages/slice-machine/src/components/PageLayout/PageLayout.tsx
@@ -1,4 +1,5 @@
 import type { FC, PropsWithChildren } from "react";
+import clsx from "clsx";
 
 import * as styles from "./PageLayout.css";
 
@@ -13,7 +14,9 @@ export const PageLayout: FC<PageLayoutProps> = ({
 }) => (
   <div {...otherProps} className={styles.root}>
     {children}
-    <div className={styles.borderTopColor[borderTopColor]} />
+    <div
+      className={clsx(styles.borderTop, styles.borderTopColor[borderTopColor])}
+    />
   </div>
 );
 
--- a/packages/slice-machine/src/components/PageLayout/PageLayout.tsx
+++ b/packages/slice-machine/src/components/PageLayout/PageLayout.tsx
@@ -7,13 +7,13 @@ type PageLayoutProps = PropsWithChildren<{
 }>;
 
 export const PageLayout: FC<PageLayoutProps> = ({
-  borderTopColor,
+  borderTopColor = "purple",
   children,
   ...otherProps
 }) => (
   <div {...otherProps} className={styles.root}>
     {children}
-    <div className={styles.borderTopColor[borderTopColor ?? "purple"]} />
+    <div className={styles.borderTopColor[borderTopColor]} />
   </div>
 );
 
--- a/packages/slice-machine/src/components/PageLayout/PageLayout.tsx
+++ b/packages/slice-machine/src/components/PageLayout/PageLayout.tsx
@@ -13,11 +13,7 @@ export const PageLayout: FC<PageLayoutProps> = ({
 }) => (
   <div {...otherProps} className={styles.root}>
     {children}
-    <div
-      className={
-        borderTopColor ? styles.borderTopColor[borderTopColor] : undefined
-      }
-    />
+    <div className={styles.borderTopColor[borderTopColor ?? "purple"]} />
   </div>
 );
 
--- a/packages/slice-machine/src/components/Divider/Divider.css.ts
+++ b/packages/slice-machine/src/components/Divider/Divider.css.ts
@@ -1,5 +1,5 @@
 import { colors, sprinkles } from "@prismicio/editor-ui";
-import { style } from "@vanilla-extract/css";
+import { style, styleVariants } from "@vanilla-extract/css";
 
 const base = style([
   sprinkles({
@@ -9,7 +9,7 @@ const base = style([
   }),
 ]);
 
-export const variants = {
+export const variants = styleVariants({
   dashed: [
     base,
     sprinkles({
@@ -28,4 +28,4 @@ export const variants = {
         "linear-gradient(to right, color-mix(in srgb, currentColor, transparent 100%), currentColor, color-mix(in srgb, currentColor, transparent 100%))",
     },
   ],
-};
+});
--- a/packages/slice-machine/src/components/Divider/Divider.css.ts
+++ b/packages/slice-machine/src/components/Divider/Divider.css.ts
@@ -23,5 +23,9 @@ export const variants = styleVariants({
     sprinkles({
       height: 1,
     }),
+    {
+      backgroundImage:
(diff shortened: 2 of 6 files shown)


packages/slice-machine/src/components/PageLayout/PageLayout.tsx
```diff
@@ -7,17 +7,13 @@ type PageLayoutProps = PropsWithChildren<{
...
   ...otherProps
 }) => (
   <div {...otherProps} className={styles.root}>
     {children}
-    <div
-      className={
-        borderTopColor ? styles.borderTopColor[borderTopColor] : undefined
-      }
-    />
+    <div className={styles.borderTopColor[borderTopColor]} />
```
bapmrl: 💡 That's better, but I think `styles.borderTopColor` has too many responsibilities (as it mixes positioning with the `borderTopBase` class and colour handling). We could handle the color in a separate CSS class.

In `packages/slice-machine/src/components/PageLayout/PageLayout.tsx`:
```
<div className={clsx(styles.borderTop, styles.borderTopColor[borderTopColor])} />
```

In `packages/slice-machine/src/components/PageLayout/PageLayout.css.ts`:
```

export const borderTop = style([
  sprinkles({
    all: "unset",
    display: "revert",
    height: 2,
    position: "fixed",
    top: 0,
  }),
  { width: "inherit" },

export const borderTopColor = {
  amber: sprinkles({ backgroundColor: colors.amber10 }),
  indigo: sprinkles({ backgroundColor: colors.indigo10 }),
  purple: sprinkles({ backgroundColor: colors.purple9 })
}
```

packages/slice-machine/src/components/PageLayout/PageLayout.tsx
```diff
@@ -7,17 +7,13 @@ type PageLayoutProps = PropsWithChildren<{
 }>;
 
 export const PageLayout: FC<PageLayoutProps> = ({
-  borderTopColor,
+  borderTopColor = "purple",
```
bapmrl, line 11: 💡 The `argTypes` definition can be simplified in `PageLayout`'s Storybook file:
```
import * as styles from "./PageLayout.css"

// ...

const meta = {
  argTypes: {
    borderTopColor: { control: "select", options: Object.keys(styles.borderTopColor) }
  },
  // ...
}
```

Also, when I write Storybook stories I always specify the default props in each story's `args`. I think it's nice to "document" what's the default state of the component. So, if you want, you could add `borderTopColor: "purple"` in `Default.args`.

Here's an example Storybook file I'm quite happy with: https://github.com/prismicio/slice-machine/blob/master/packages/slice-machine/src/components/Card/Card.stories.tsx.

packages/slice-machine/src/components/PageLayout/PageLayout.tsx
bapmrl, line 1: 👌 This UI component is in a better state now. I saw that environments had been introduced to it, but I believe UI components shouldn't include any domain concerns. It will allow us to move them to the Editor's repository if need be (as it acts as our central place for the design system) and get simpler code overall.

bapmrl (approved):
I'm approving as it solves the bugs, but I've added small comments to help you simplify the code.

packages/slice-machine/src/components/PageLayout/PageLayout.tsx
angeloashmore: Done! I also removed `all: unset; display: "revert"` and changed `width: "inherit"` to `width: "100%"`.

Are those styles actually necessary?

packages/slice-machine/src/components/PageLayout/PageLayout.tsx
angeloashmore: > The argTypes definition can be simplified in PageLayout's Storybook file:

With that version, how would we test an `undefined` value, which falls back to the default `purple` value?

`[...Object.keys(borderTopColor), undefined]` is used so we can include `undefined`.

> I think it's nice to "document" what's the default state of the component.

Done! 🙂

angeloashmore:
@bapmrl Thanks for the review! I updated the PR to address your comments.

packages/slice-machine/src/components/PageLayout/PageLayout.tsx
bapmrl: >With that version, how would we test an undefined value, which falls back to the default purple value?

True ^^ Let's keep the spread with `undefined`.

packages/slice-machine/src/components/PageLayout/PageLayout.tsx
bapmrl: It's because we don't have any global CSS reset. So, I've been doing that on each DOM element to reset their styling. I've chosen this strategy because it was pretty hard to find a solution that would work well with `theme-ui`'s `BaseStyles`... that overrides quite a lot of styles in our components.

Are you sure for `width: "100%"`? I think I tried that at the time I wrote the `PageLayout`, but it didn't work. Maybe it was because of the `unset`?

angeloashmore: > It's because we don't have any global CSS reset. So, I've been doing that on each DOM element to reset their styling. I've chosen this strategy because it was pretty hard to find a solution that would work well with theme-ui's BaseStyles... that overrides quite a lot of styles in our components.

Since this CSS is styling a bare `<div>`, do you think it's necessary? It seems to work fine without it, but I don't have the background info on this topic.

Maybe we should find a CSS reset that works for us at some point so we don't need to reset individual elements. 😄

Let me know if you think it's better to keep the reset in.

> Are you sure for width: "100%"? I think I tried that at the time I wrote the PageLayout, but it didn't work. Maybe it was because of the unset?

`width: "100%"` seems to work with or without `all: "unset"`. `100%` is more intuitive to me than `inherit`.

bapmrl: >Let me know if you think it's better to keep the reset in.

I would keep the resets just for the sake of consistency, as it's something that's been documented in the Notion style guide, but I agree that a global reset would be much simpler. I'll discuss this with @AlexGalays (or you can write a small message on Page Builder's Slack channel if you want).

>width: "100%" seems to work with or without all: "unset". 100%" is more intuitive to me than inherit`.

If it works on the latest versions of Firefox and Safari, okay ^^

bapmrl (approved):

packages/slice-machine/src/components/PageLayout/PageLayout.tsx
angeloashmore: We decided to put `all: unset; display: revert;` back in for consistency with other components. We'll remove it when we have a proper CSS reset.

PR #1226, opened 2023-11-30
fix(authentication): Not logged in management

xrutayisire:
## Context

- Completes DT-1700

## The Solution

- Fix a problem in the manager not being specific enough for 403 or 401 error.
- Handle the differentiation in the front-end for the changes page and the Environments switch.

## Impact / Dependencies

- It's here only a fix issue and not the redesign of the page. The redesign is DEVTD-7 and needs time in UI / UI / Product

- [ ] If it is a critical feature, I have added tests. ⇒ Didn't add tests since I didn't yet handle login / logout tests in playwright 
- [x] The CI is successful.

## Preview

Screenshot 2023-11-30 at 14 51 21

Screenshot 2023-11-30 at 16 26 34

### DIFF
--- a/packages/slice-machine/components/ChangesEmptyState/AuthErrorPage.tsx
+++ b/packages/slice-machine/components/ChangesEmptyState/AuthErrorPage.tsx
@@ -1,36 +1,52 @@
-import useSliceMachineActions from "@src/modules/useSliceMachineActions";
-import React from "react";
-import { Button, Flex, Text } from "theme-ui";
+import React, { FC } from "react";
+import { Box, Button, Text, TextLink } from "@prismicio/editor-ui";
 
-export const AuthErrorPage = () => {
+import useSliceMachineActions from "@src/modules/useSliceMachineActions";
+import { AuthStatus } from "@src/modules/userContext/types";
+
+export type AuthErrorPageProps = {
+  authStatus: AuthStatus.UNAUTHORIZED | AuthStatus.FORBIDDEN;
+};
+
+export const AuthErrorPage: FC<AuthErrorPageProps> = (props) => {
+  const { authStatus } = props;
   const { openLoginModal } = useSliceMachineActions();
+
   return (
-    <Flex sx={{ justifyContent: "center", alignItems: "center", flex: 1 }}>
-      <Flex
-        sx={{
-          flexDirection: "column",
-          alignItems: "center",
-          transform: "translateY(-50%)",
-        }}
-      >
-        <Text
-          sx={{
-            fontWeight: 600,
-            fontSize: "24px",
-            lineHeight: "32px",
-            mb: "8px",
-            mt: "2px",
-          }}
-        >
-          Could not access your repository
-        </Text>
-        <Text sx={{ fontSize: "13px", lineHeight: "24px" }}>
-          You need to log in to Prismic in order to push your changes.
-        </Text>
-        <Button sx={{ mt: "24px" }} onClick={() => openLoginModal()}>
-          Log in to Prismic
-        </Button>
-      </Flex>
-    </Flex>
+    <Box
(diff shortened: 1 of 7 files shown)


packages/slice-machine/components/ChangesEmptyState/AuthErrorPage.tsx
```diff
@@ -1,36 +1,46 @@
...
+      <Text variant="h3">
+        {authStatus === AuthStatus.FORBIDDEN
+          ? "It seems like you are logged out"
+          : "It seems like you don't have access to this repository"}
+      </Text>
+      <Text>
+        {authStatus === AuthStatus.FORBIDDEN
+          ? "Log in to connect to your repository."
+          : "Check that the repository name is correct, then contact your repository administrator."}
+      </Text>
```
angeloashmore: What do you think of this version instead? It pairs the heading with the content, which makes the copy easier to follow in my opinion.

```ts
{
  authStatus === AuthStatus.FORBIDDEN ? (
    <>
      <Text variant="h3">It seems like you are logged out</Text>
      <Text>Log in to connect to your repository.</Text>
    </>
  ) : (
    <>
      <Text variant="h3">
        It seems like you don't have access to this repository
      </Text>
      <Text>
        Check that the repository name is correct, then contact your repository
        administrator.
      </Text>
    </>
  );
}
```

Totally subjective, so your call. 🙂

packages/slice-machine/src/components/SideNav/SideNavEnvironmentSelector.tsx
```diff
@@ -18,7 +18,11 @@ import LogoIcon from "@src/icons/LogoIcon";
 import * as styles from "./SideNavEnvironmentSelector.css";
 
 type SideNavEnvironmentSelectorProps = {
-  variant?: "default" | "unauthorized";
+  variant?:
+    | "default"
+    | "offline"
+    | "UnauthorizedError"
+    | "UnauthenticatedError";
```
angeloashmore: Using the error names as variants seems a bit strange to me. It couples the component to `@prismicio/custom-types-client`, which are normally two separate entities.

Do you think we could use `"unauthorized" | "unauthenticated"` instead?

angeloashmore (commented):
Nice! Left a few comments.

angeloashmore (approved):
(It looks good to me overall, so I'm approving. The comments above are up to your discretion to implement.)

packages/slice-machine/components/ChangesEmptyState/AuthErrorPage.tsx
xrutayisire: I hesitated, I agree the JSX duplication is super low. I did your way, I prefer it!

packages/slice-machine/src/components/SideNav/SideNavEnvironmentSelector.tsx
xrutayisire: You are right. I updated the code.

Only problem is that it should not even be like that directly. In order to release we did a quick fix with @hypervillain.

In summary, if you check what has changed since your code, you will see that we don't throw from the manager anymore but return an object with the error and also that we don't use the manager function to check what kind of error it is. Both usage were not working. But for me there is something to improve here. The way we are checking the error is bad, and it's tech debt already.

The problem was only happening in production with an alpha, there is no problem in dev.

I'm talking these two commits:
- https://github.com/prismicio/slice-machine/pull/1217/commits/fa4b88cf1eb890625159955fa3a1ce2fd5a10144
- https://github.com/prismicio/slice-machine/pull/1217/commits/dcfd59c8c9754f6c0d64dc67a4b74dbafc304974

If you have an idea of the problem and how we could improve that part, it would be great 🙂

PR #1227, opened 2023-11-30
feat(manager): update PrismicUser endpoint

bapmrl:
Completes DT-1811.


xrutayisire (approved):
LGTM, tested login in both wroom.io and prismic.io to ensure it's working well.

PR #1228, opened 2023-11-30
feat(slice-machine): add Gradient in Sidebar and use medium-sized DialogActions

bapmrl:
Completes DT-1657, DT-1681 and DT-1803.

### DIFF
--- a/packages/slice-machine/lib/builders/SliceBuilder/Sidebar/index.tsx
+++ b/packages/slice-machine/lib/builders/SliceBuilder/Sidebar/index.tsx
@@ -77,10 +77,13 @@ export const Sidebar: FC<SidebarProps> = (props) => {
           backgroundColor="grey2"
           bottom={0}
           flexDirection="column"
-          padding={{ inline: 24 }}
+          padding={{ bottom: 40, inline: 24 }}
           position="sticky"
-          // @ts-expect-error - TODO: add missing props to the Editor's `Box` component.
-          style={{ marginBottom: "-16px", paddingBottom: "40px" }}
+          // As `PageLayoutContent` has a `16px` bottom padding, we need to
+          // apply an equal negative margin to the `Box` if we want it to sit
+          // flush with the bottom of the page.
+          // @ts-expect-error TODO(PBD-1080): write the new `ScrollAreaEndGradient` component instead of using a `Box`.
+          style={{ marginBottom: "-16px" }}
         >
           <Gradient sx={{ left: 0, position: "absolute", right: 0 }} />
           <Button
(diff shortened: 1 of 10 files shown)


packages/slice-machine/lib/builders/SliceBuilder/Sidebar/index.tsx
```diff
@@ -73,16 +73,32 @@ export const Sidebar: FC<SidebarProps> = (props) => {
...
-        </Button>
+          <Gradient sx={{ left: 0, position: "absolute", right: 0 }} />
+          <Button
+            onClick={() => {
+              setDialog({ type: "ADD_VARIATION" });
+            }}
+            startIcon="add"
+            // Set `position` to `relative` to position `Button` on top of
+            // `Gradient`.
+            sx={{ position: "relative" }}
```
xrutayisire, line 96: ❓ We could also use z-index here, no?

packages/slice-machine/lib/builders/SliceBuilder/Sidebar/index.tsx
```diff
@@ -73,16 +73,32 @@ export const Sidebar: FC<SidebarProps> = (props) => {
...
+          backgroundColor="grey2"
+          bottom={0}
+          flexDirection="column"
+          padding={{ bottom: 40, inline: 24 }}
+          position="sticky"
+          // As `PageLayoutContent` has a `16px` bottom padding, we need to
+          // apply an equal negative margin to the `Box` if we want it to sit
+          // flush with the bottom of the page.
+          // @ts-expect-error TODO(PBD-1080): write the new `ScrollAreaEndGradient` component instead of using a `Box`.
+          style={{ marginBottom: "-16px" }}
```
xrutayisire, line 86: ❓Are we sure editor-ui will never remove additional props? It could break our work without us knowing.

xrutayisire (approved):
LGTM! Great work. I see it should be improved with another component.

Tested it and it's super nice now!

packages/slice-machine/lib/builders/SliceBuilder/Sidebar/index.tsx
bapmrl: We could, but I prefer not specifying `zIndex` when I don't need to :) As the `Box` is positioned absolutely and the `Button` appears next in the DOM and is positioned relatively, the `Button` is displayed on top of the `Box`.

packages/slice-machine/lib/builders/SliceBuilder/Sidebar/index.tsx
bapmrl: We aren't. @AlexGalays proposed I use a simple `div`, but I find it even worse as I would have to use all the design tokens manually (by using the `tokens` import).

PR #1229, opened 2023-12-01
fix(slice-machine-ui): use the correct log in icon

angeloashmore:
## Context

DT-1836

## The Solution

Use `<IconButton>`'s new `icon={<MyIcon />}` functionality to render the correct custom icon.

## Impact / Dependencies

None

- [x] The CI is successful.


packages/slice-machine/src/icons/LogInIcon.tsx
```diff
@@ -0,0 +1,13 @@
+import type { FC, SVGProps } from "react";
+
+export const LogInIcon: FC<SVGProps<SVGSVGElement>> = (props) => (
```
bapmrl: 💡 I personally try to use the same names as Material Symbols for consistency (instead of relying on my imperfect English level). In this case, it would be [LoginIcon](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:login:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=login).

Don't forget to rename the CSS class if you decide to apply this suggestion 🙂

packages/slice-machine/src/components/SideNav/SideNavEnvironmentSelector.css.ts
```diff
@@ -67,3 +67,12 @@ export const actionRequiredLabel = style([
...
   },
 ]);
+
+export const logInIcon = style([
+  sprinkles({
+    position: "relative",
+    width: 24,
+    height: 24,
+  }),
+  { left: "-1px" },
```
bapmrl: ❓ Why do you have to move it by `1px`?

bapmrl (approved):

packages/slice-machine/src/icons/LogInIcon.tsx
xrutayisire: We need to agree on the naming btw since remember I use @benjammartin naming. Anyway for me our naming need to be the same as Figma.

angeloashmore: Good call. 🙂 I changed it to "log in" since it is a verb, but "login" is also valid here and is what Material icons uses.

@xrutayisire Following Material is a good convention, in my opinion. I assume their team puts thought into the naming structure, or at least has their own conventions, which means we can offload our thinking to them.

packages/slice-machine/src/components/SideNav/SideNavEnvironmentSelector.css.ts
angeloashmore: The icon was visually off-centered by 1px when put into a square button.

It would always be off-centered regardless where it is used, so I refactored the icon's SVG to always be shifted 1px to the left and removed the `loginIcon` class.

With the 1px offset:
image

Without the 1px offset:
image

bapmrl (approved):

PR #1231, opened 2023-12-04
test(playwright): Add logged in and onboarded setup for tests in Playwright

xrutayisire:
## Context

- Completes DT-1815, DT-1823

## The Solution

- Handle login management to be able to test logged in test or not logged in tests
- Handle onboarding management to be able to test onboarded test or not onboarded tests
- Each of these two configurations are test based only! Super cool to not care about this at a higher level
- Handle two env support wroom or prismic with an automatic switch in local and an manual setup in CI to speed up tests
- Prevent having tests report in CI in order to save time when it's a success (upload can take up to 8 min)
- Added support of mocking manager procedures
- Added an issue to take care of full scenarios without mocking at all, only for CI: DT-1840
- Added a mocks folder in order to share common mocks in all tests
- Improve README
- Added in-guide tests to check my work on onboarding config
- Added some changes page tests to check my work on log in and mock management 

## Impact / Dependencies

- We can now test almost anything, and that PR will be the last before my announcement to ask all devs to write some tests. 💪
- Playwright already have more tests than Cypress (not skipped) 🎉

- [x] The CI is successful.


playwright/pages/components/InAppGuideDialog.ts
```diff
@@ -0,0 +1,38 @@
+import { Locator, Page } from "@playwright/test";
+
+import { Dialog } from "./Dialog";
+
+export class InAppGuideDialog extends Dialog {
+  override readonly closeButton: Locator;
```
xrutayisire, line 6: I'm not yet 100% sure how to correctly target editor-ui close button. We may need a small update on the editor part.

playwright/README.md
```diff
@@ -63,6 +74,23 @@ To open a downloaded CI test report from anywhere in your computer:
...
+## Creating tests
+
+Use `test.run()` to create a test. `run` function take an optional object parameter `options` that let you configure how you want to run the test.
+You can configure if you want a logged in test and also an onboarded test.
+Default is not logged in and onboarded.
+
+Example for a logged in user not onboarded:
+
+```ts
+test.run({ loggedIn: true, onboarded: false })("I can ...",
```
angeloashmore, line 88: Do you see a benefit to preferring functions rather than configuration? For example:

```ts
test("I can...", async () => {
  await logIn()
  // await logIn({ username: "username", password: "password" })

  await onboard()
})
```

With functions, we can script when login/logout occurs, as well as any other actions we may want to orchestrate. In addition, we can configure the actions as needed, such as different usernames and passwords.

It also means we can use the vanilla Playwright API rather than a wrapper.

With a configuration based approach, we could end up in configuration overload with things like: `loggedIn`, `wroomUsername`, `wroomPassword`, `onboardBeforeLogin`, etc.

Just a suggestion since I'm not as familiar with the finer details of how this all works as you are. 🙂

playwright/fixtures/index.ts
```diff
@@ -35,113 +44,242 @@ export type Fixtures = {
...
+                  }),
+                },
+                {
+                  name: "slice-machine_isInAppGuideOpen",
+                  value: "false",
+                },
+              ],
+            },
+          ],
+        },
```
angeloashmore, line 155: Should we be initializing pages with default state? In an ideal test (which we don't always need to achieve, to be clear), we would simulate what a user would do to get to that state. For example, if we want to test a page when a user has already onboarded, we would open a fresh page and close the onboarding dialog (or whatever the action is).

Doing it that way tests the full onboarding flow and expected states. It also means we can change implementation details in SM UI, like the property names used in localStorage, without modifying our tests.

playwright/fixtures/index.ts
```diff
@@ -35,113 +44,242 @@ export type Fixtures = {
...
+
+      // Create new page object with new context
+      const page = await context.newPage();
+
+      // Logout user by default
+      try {
+        await fs.rm(path.join(os.homedir(), ".prismic"));
+      } catch (error) {
+        // Ignore since it means the user is already logged out
+      }
```
angeloashmore, line 181: Just checking: will this affect the user's actual `.prismic` file? Or is it mocked/virtualized?

playwright/fixtures/index.ts
```diff
@@ -35,113 +44,242 @@ export type Fixtures = {
...
+
+      // Login user if needed
+      if (loggedIn) {
+        // In CI we define a PRISMIC_URL env variable to fasten the tests
+        let prismicUrl: string | undefined = process.env["PRISMIC_URL"];
```
angeloashmore: Is `string | undefined` necessary? I'm not sure what the built-in type of `process.env["PRISMIC_URL"]` is, but I generally try to avoid declaring types.

playwright/fixtures/index.ts
```diff
@@ -35,113 +44,242 @@ export type Fixtures = {
...
+
+        // In local we get the Prismic URL from the browser, it helps to avoid
+        // switching manually between Wroom and Prismic
+        if (!prismicUrl) {
+          await page.route(
+            "*/**/_manager",
```
angeloashmore, line 192: `@slicemachine/manager/test` provides with a `createSliceMachineManagerMSWHandler()` test helper to mock manager RPC requests.

You can see it being used in `slice-machine-ui` here: https://github.com/prismicio/slice-machine/blob/d061d339995c8d849ccf8d1f7f84a22e7dffadd1/packages/slice-machine/test/pages/slices.test.tsx#L56-L61

Is there a way we can make a Playwright version of that helper? The helper is specific to MSW, but I think we could adapt it for Playwright's network APIs.

playwright/fixtures/index.ts
```diff
@@ -35,113 +44,242 @@ export type Fixtures = {
...
+            {
+              times: 1,
+            },
+          );
 
-    // Use the fixture value in the test.
-    await use(page);
+          // Visit the page to trigger the getState call and wait for it
+          await page.goto("/");
+          await page.waitForResponse("*/**/_manager");
```
angeloashmore, line 226: Is there a way to check for behavior rather than implementation details? For example, checking that a user is logged in using an element on the page rather than the internal `_manager` endpoint.

playwright/fixtures/index.ts
```diff
@@ -35,113 +44,242 @@ export type Fixtures = {
...
+        }
 
-    // Gracefully close up everything
-    await page.close();
-    await context.close();
-  },
-});
+        const activeEnv = prismicUrl?.includes("wroom") ? "WROOM" : "PRISMIC";
```
angeloashmore, line 229: Determining environments based on URL seems inflexible to me. What do you think of using a dedicated option for setting the environment instead? I'm not sure where that option would go (does Playwright have a way to configure that? Or would it need to be an environment variable?).

playwright/fixtures/index.ts
```diff
@@ -35,113 +44,242 @@ export type Fixtures = {
...
+        if (!prismicUrl) {
+          console.warn("Could not find Prismic URL.");
+        } else if (!email || !password) {
+          console.warn(
+            `Missing EMAIL or PASSWORD environment variables for ${activeEnv} environment.`,
+          );
+        } else {
+          // Do the authentication call
+          const res = await fetch(
+            new URL("./authentication/signin", prismicUrl).toString(),
```
angeloashmore, line 242: Do we call the actual API? Is there a way to mock this, or do you prefer calling the real API?

To be honest, I don't have much experiencing writing E2E tests (I usually stick to integration where we would normally mock everything with like-real responses), so I might have the wrong perspective.

playwright/fixtures/index.ts
```diff
@@ -35,113 +44,242 @@ export type Fixtures = {
...
+            // If the authentication fails, log the error
+            const reason = await res.text();
+            console.error(
+              "Could not authenticate to prismic. Please check the credentials.",
+              reason,
+            );
+          } else {
+            // If the authentication succeeded, save the cookies to persist it
+            await fs.writeFile(
+              path.join(os.homedir(), ".prismic"),
```
angeloashmore, line 265: Same question as above: are we writing to the actual `.prismic` file on the test runner's machine? It would be great if we could mock this somehow so we aren't overwriting our own logins.

playwright/pages/ChangesPage.ts
```diff
@@ -20,23 +25,57 @@ export class ChangesPage extends SliceMachinePage {
...
+    this.loginButton = this.body.getByText("Log in to Prismic");
+    this.pushChangesButton = page.getByText("Push Changes");
+    this.pushedMessaged = page.getByText(
+      "All slices and types have been pushed",
+    );
+    this.notLoggedInTitle = page.getByText("It seems like you are logged out");
+    this.notAuthorizedTitle = page.getByText(
+      "It seems like you don't have access to this repository",
+    );
+    this.blankSlateTitle = page.getByText("Everything up-to-date");
```
angeloashmore, line 37: If we use these queries elsewhere, should we abstract them to a set of helpers? (If we don't use it elsewhere, then we can skip the abstraction, of course).

angeloashmore: Should these be evaluated immediately? I'm wondering if we'll run into a race condition in tests, such as this:

```ts
  async goto() {
    await this.page.goto("/changes");
    await expect(this.breadcrumbLabel).toBeVisible();
    // What happens if breadcrumbLabel didn't exist until we loaded `/changes`?
  }
```

I would expect something like this instead:

```ts
  async goto() {
    await this.page.goto("/changes");
    await expect(this.getBreadcrumbLabel()).toBeVisible();
  }
```

playwright/pages/components/InAppGuideDialog.ts
angeloashmore: Indeed, I had to add a custom data property to the close buttons for the Page Builder demo (https://prismic.io/try). Doing more complex selectors seems fragile (e.g. `div > div > header > button`).

playwright/pages/components/InAppGuideDialog.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  constructor(page: Page) {
+    super(page, {
+      title: "Build a page in 5 minutes",
```
angeloashmore, line 10: Should we be hardcoding content like this?

playwright/tests/changes/changes.spec.ts
```diff
@@ -1,5 +1,134 @@
...
+
+test.describe("Changes", () => {
+  test.run({ loggedIn: true })(
+    "I cannot see the login screen when logged in",
+    async ({ changesPage }) => {
+      await changesPage.goto();
+      await expect(changesPage.loginButton).not.toBeVisible();
+      await expect(changesPage.notLoggedInTitle).not.toBeVisible();
+      await expect(changesPage.notAuthorizedTitle).not.toBeVisible();
+    },
```
angeloashmore, line 15: This is very concise and readable! 👍

playwright/tests/changes/changes.spec.ts
```diff
@@ -1,5 +1,134 @@
...
+  );
+
+  test.run()(
+    "I can see the login screen when logged out",
+    async ({ changesPage }) => {
+      await changesPage.goto();
+      await expect(changesPage.loginButton).toBeVisible();
+      await expect(changesPage.notLoggedInTitle).toBeVisible();
+    },
+  );
```
angeloashmore, line 25: The fact that this tests runs when logged out is not obvious. Should we make it obvious with something like `test.run({ loggedIn: false })`?

(I suggested making `loggedIn` into a helper function (`await logIn()`) in a previous comment. In this case, we would probably also make a `await logOut()` helper to make it obvious.)

playwright/tests/changes/changes.spec.ts
```diff
@@ -1,5 +1,134 @@
...
+            name: "getState",
+            data: (data) => ({
+              ...data,
+              clientError: {
+                status: 403,
+              },
+            }),
+          },
+        ],
+      });
```
angeloashmore: I'm not sure we should be mocking the manager's responses. It removes some of the test's thoroughness since we want to test that the manager is working properly. Is it possible to mock the external APIs instead? I think we would want to mock APIs as far away from Slice Machine's code as possible.

playwright/tests/common/inAppGuide.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { test } from "../../fixtures";
+
+test.describe("In-app guide", () => {
+  test.run({ onboarded: false })(
+    "I can see the in-app guide open by default",
+    async ({ sliceMachinePage }) => {
+      await sliceMachinePage.gotoDefaultPage();
+
+      await expect(sliceMachinePage.inAppGuideDialog.title).toBeVisible();
```
angeloashmore, line 11: Follow-up from a previous comment about evaluating element queries lazily: this test would look something like this when changing from an instance property to a getter:

```ts
    async ({ sliceMachinePage }) => {
      await sliceMachinePage.gotoDefaultPage();

      await expect(sliceMachinePage.inAppGuideDialog.getTitle()).toBeVisible();
    },
  );
```

playwright/utils/mockManagerProcedures.ts
```diff
@@ -0,0 +1,83 @@
...
+     * Whether to execute the procedure or not. Defaults to true.
+     */
+    execute?: boolean;
+  }[];
+};
+
+/**
+ * Mocks manager procedures from SliceMachineManagerClient
+ */
+export async function mockManagerProcedures(args: MockManagerProcedureArgs) {
```
angeloashmore: Copying my comment from above:

> `@slicemachine/manager/test` provides with a `createSliceMachineManagerMSWHandler()` test helper to mock manager RPC requests.
> 
> You can see it being used in `slice-machine-ui` here: https://github.com/prismicio/slice-machine/blob/d061d339995c8d849ccf8d1f7f84a22e7dffadd1/packages/slice-machine/test/pages/slices.test.tsx#L56-L61
> 
> Is there a way we can make a Playwright version of that helper? The helper is specific to MSW, but I think we could adapt it for Playwright's network APIs.

angeloashmore (commented):
This is awesome! The test API is clean and readable and should be nice to work with. 🙂

I left a bunch of comments on ways we could refine the API and their implementations. They are all just suggestions, and you'll know best if they are good or bad. Let me know if anything is unclear!

playwright/pages/components/InAppGuideDialog.ts
xrutayisire: Yes, We don't want to do complex selectors like this. Better to add a test id or an aria something in the editor package.

playwright/README.md
xrutayisire: I wanted to rely on Fixture functionality from Playwright:  https://playwright.dev/docs/test-fixtures

It gave us an isolated environment to work on. While doing your way is technically possible (I would need to try because you cannot do just `login()`, you need to pass down the browser. Also, I'm not sure because Fixture as I dit let you override the default page object that is used everywhere. 

So the recommendation by playwright is to do it with a Fixture, but then when you want to opt out of the default configuration, you handle it with `use` for example.
Problem with `use` is that it's for a whole file. And it's a bad DX for our usage, having it locally for a specific test is really better and so that why I'm only relying on Fixture with a function to do it.

Last point with fixtures is that you want to prevent mixing your setup inside the test. That's what fixture are for, for example if you need a custom type, you can usage the customType fixture and it will create a customType for you. 
Each test should only handle the test content and we try to prevent having a mix with the config.

At the moment only mocking is kind of allowed because it's also something we try to prevent doing. I will explain in another comment more suited for the mocking part.

I can try using a function inside test but I'm not sure of the outcome, much more with the comment about having an explicit logout for example. It's not the responsibility of the test to call logout if you just want to test an empty state for example.

It's a good point to discuss over a call maybe I would like to continue that point.

playwright/fixtures/index.ts
xrutayisire: Should we be initializing pages with default state? 
Short answer yes. And no 😅

So the goal here is separate testing concern. What you are describing is a test that need to be written, As a user I will close the popup etc, etc. But you should not do that on all tests.

Onboarding should be tested in isolation, you should verify that everything is working with that but then for any other tests nothing should bother you testing normally. If it can bother you then a specific test for onboarding should be written. 

e2e tests can take long time to execute, it's not a good practice to repeat too many actions on all the tests if you can isolate some scenarios. Here on our cases for example, I quickly tested the In-App guide. being bale to open it, close it, etc. It should no bother me for other tests.

We should never tests implementation details, but here it's the setup, it's not an actual test. And setup can rely on implementation details, and it should not be a big problem since it's shared. It prevents you from actually having an impact on tests that should not have. 

Without these kinds of setup, you would have more flakiness in your tests for something not related to your work.

Let me know if it was clear, maybe also quickly continuing this over a talk.

playwright/fixtures/index.ts
xrutayisire: Yes and it should for me, I saw another comment about this. e2e tests means really e2e as much as possible. Cutting the authentication part is possible but we want to login fully, we will want to fully push, get the envs, changes all. 
We want to be able to see problems that can happen on wroom. If the wroom part change, it should break in our side so we know we have something to do. 

We will differentiate two kind of tests. The standalone one, and the scenario one. At the moment it's only standalone, but later we will add scenarios that will really do everything.

playwright/fixtures/index.ts
xrutayisire: Oh good one, removing it!

playwright/fixtures/index.ts
xrutayisire: I will take a look, but the interception should be made on the browser side, so I'm not sure how it will work. I will take a look and see what we can do, good idea!

playwright/fixtures/index.ts
xrutayisire: Here it's the setup and it's also a tricky part. The goal is to get the env URL in order to improve the DX. Being able to test on stage or prod without changing an env variable. For that I want to wait the getState call but not an element in the UI. It's more faster. As I wrote, setup can rely on implementation details because it's shared and only in a unique place.

playwright/fixtures/index.ts
xrutayisire: Ah ^^ The goal was to be flexible. Maybe it's not clear enough but you can already used an environement variable. `PRISMIC_URL`, so all good. 
Why that then? As I just explain above in another comment, it's to improve DX. I did that for some use cases I saw while testing.

I often switch between `yarn run slicemachine:dev` and in another project `npm run slicemachine` so between wroom and prismic. Do I want to update an env variable in Playwright everytime, not really. I just don't wanna care. I wrote the email and password for both env and I want an automatic swtich. 

If you wan to hard set an env you can with `PRISMIC_URL` anyway. That's waht the CI do because you will save 3s per test that need login then and that's cool for CI

playwright/fixtures/index.ts
xrutayisire: I prefer the real API. As I said fully e2e, it's important to mock as less as possible. Still being able to mock is cool for standalone test.

But for scenario, as you said with the onboarding we will not setup anything. I agree that integration tests we would mock everything.

playwright/fixtures/index.ts
xrutayisire: Yes it will override it. For me it should work like this. I don't see a better option. Mocking the login kind of create an integration test on that part and can be tricky later on to handle for e2e tests. You think it's a problem to override your .prismic while you doing test?

playwright/pages/ChangesPage.ts
xrutayisire: > If we use these queries elsewhere

Hum, not sure to understand. It's already an abstraction here with POM model. ChangesPage is the abstraction here. these locators are specific to ChangesPage. `    this.blankSlateTitle = page.getByText("Everything up-to-date");
` is only for here.

xrutayisire: > Should these be evaluated immediately? I'm wondering if we'll run into a race condition in tests, such as this:
Oh no need with Playwright to do this. 

`await expect(this.breadcrumbLabel).toBeVisible();` is only checked at the moment the line is processed. It's how you're writing test in Playwright, no need to wrap everything in function. Locator are already function in a sense. 

see: https://playwright.dev/docs/locators

let me know If I didn't really answer your question

playwright/pages/components/InAppGuideDialog.ts
xrutayisire: What do you mean? It's POM model.

Meaning here you share an instance of the InAppGuideDialog and you define the expected title. If someone change it in the code, test will break because as a user it changed. So it's "normal" but maybe I didn't get your question?

playwright/tests/changes/changes.spec.ts
xrutayisire: For this one I agree that we could call it like this:

```
test.run({ loggedIn: false })(
    "I can see the login screen when logged out",
    async ({ changesPage }) => {
      await changesPage.goto();
      await expect(changesPage.loginButton).toBeVisible();
      await expect(changesPage.notLoggedInTitle).toBeVisible();
    },
  );
  ```
  
  But for the other one not related to login I'm not sure to force dev to write down every test `{ loggedIn: false }`. Because if it's not necessary to be logged in, better to test without a login. It's quicker and also ensure it's not mandatory. It can be too much to always have to setup the test while 85% of the tests need the same default, no? WDYT?

File: playwright/tests/changes/changes.spec.ts (hunk 15 above)
Comment (xrutayisire) [reply playwright/tests/changes/changes.spec.ts]: It's 100% sure that mocking should be used a little as possible since it's e2e. But for standalone tests, opposite as full scenario you want to be able to test on any running Slice Machine some default use case. 

It will break the e2e management but ensure consistency in your front-end. I 100% agree that it would be better to mock the API behind, but I have no clue how to do and if it's possible.  Do you think we can mock the API since Plawright mock what the browser can see?

File: playwright/tests/common/inAppGuide.spec.ts (hunk 16 above)
Comment (xrutayisire) [reply playwright/tests/common/inAppGuide.spec.ts:11]: No need, Locator are kind of already a function. Wrapping it in a function yourself will not make any difference. It's evaluated only when you await them.

File: playwright/README.md (hunk 2 above)
Comment (xrutayisire) [reply playwright/README.md:88]: Discussed together 🙂

File: playwright/fixtures/index.ts (hunk 3 above)
Comment (xrutayisire) [reply playwright/fixtures/index.ts:155]: Discussed together 🙂

File: playwright/fixtures/index.ts (hunk 5 above)
Comment (xrutayisire) [reply playwright/fixtures/index.ts]: Done

File: playwright/fixtures/index.ts (hunk 6 above)
Comment (xrutayisire) [reply playwright/fixtures/index.ts:192]: Discussed together 🙂

File: playwright/fixtures/index.ts (hunk 7 above)
Comment (xrutayisire) [reply playwright/fixtures/index.ts:226]: Discussed together 🙂

File: playwright/fixtures/index.ts (hunk 8 above)
Comment (xrutayisire) [reply playwright/fixtures/index.ts:229]: Discussed together 🙂

File: playwright/utils/mockManagerProcedures.ts [hunk 18]
```diff
@@ -0,0 +1,102 @@
...
+  page: Page;
+
+  /**
+   * Array of procedures to mock
+   */
+  procedures: {
+    /**
+     * Procedure path in the format of `getState` or `slices.readSlice`
+     */
+    path: ProcedurePath;
```
xrutayisire: @angeloashmore I was able to type the procedure path but couldn't succeed in typing the data depending on the procedure path.

Here we need one function for all mock since Playwright mock the url that is  `*/**/_manager`. 

It means I cannot use generics to guess the typing because the mock can have multiple mocks for different procedure. 

I spend some time trying without success. If you are ok, I would like to merge if the rest is ok and maybe someone could take a look at some point, WDYT?

PR #1233, opened 2023-12-05
refactor: use `@prismicio/custom-types-client`'s `fetchOptions` to send a User Agent

angeloashmore:
## Context

DT-1764

## The Solution

Update `@prismicio/custom-types-client` to `v1.2.0`, which includes a new `fetchOptions` option (see prismicio/prismic-custom-types-client#9). `fetchOptions` allows for setting arbitrary `fetch()` options, including headers.

With the new option, we can replace the existing alpha `@prismicio/custom-types-client` API usage with the production version.

Test have been added/updated to check for the `User-Agent` header.

## Impact / Dependencies

None

- [x] The CI is successful.

### DIFF
--- a/packages/manager/src/managers/SliceMachineManager.ts
+++ b/packages/manager/src/managers/SliceMachineManager.ts
@@ -3,10 +3,7 @@ import {
 	CustomType,
 } from "@prismicio/types-internal/lib/customtypes";
 import { SharedSliceContent } from "@prismicio/types-internal/lib/content";
-import {
-	ForbiddenError,
-	UnauthorizedError,
-} from "@prismicio/custom-types-client";
+import * as prismicCustomTypesClient from "@prismicio/custom-types-client";
 import {
 	SliceMachinePlugin,
 	SliceMachinePluginRunner,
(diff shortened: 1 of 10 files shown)


packages/manager/src/managers/SliceMachineManager.ts
```diff
@@ -3,10 +3,7 @@ import {
 	CustomType,
 } from "@prismicio/types-internal/lib/customtypes";
 import { SharedSliceContent } from "@prismicio/types-internal/lib/content";
-import {
-	ForbiddenError,
-	UnauthorizedError,
-} from "@prismicio/custom-types-client";
+import * as prismicCustomTypesClient from "@prismicio/custom-types-client";
```
xrutayisire, line 6: ❓ You cannot import like before anymore? Why this change?

xrutayisire (approved):
LGTM! Super great like this and allow us to add more options later if needed 🙂

Just a small question, not a blocker for a merge. Just wanna know.

packages/manager/src/managers/SliceMachineManager.ts
angeloashmore: We can import like before, but I changed it to a namespace import because:

1. It is consistent with other imports of `@prismicio/custom-types-client` in `@slicemachine/manager`.
2. It makes it clear that `ForbiddenError` and `UnauthorizedError` come from `@prismicio/custom-types-client` (e.g. `prismicCustomTypesClient.ForbiddenError`).

   I felt this was important because we have a Slice Machine-specific `UnauthorizedError` in `src/errors.ts`, which could cause confusion with `instanceof` checks.

xrutayisire: Super clear! thanks

PR #1235, opened 2023-12-05
chore: update release process

bapmrl:
Completes DT-1769, DT-1770, DT-1771, DT-1772, DT-1779 and DT-1810.


scripts/publish.ts
```diff
@@ -0,0 +1,104 @@
+import { $ } from "execa";
+import mri from "mri";
+import semverPrerelease from "semver/functions/prerelease";
+import semverValid from "semver/functions/valid";
+
+publish();
+
+type Options = { defaultBranchName?: string };
+
+async function publish(options?: Options): Promise<void> {
```
angeloashmore, line 26: Will `options` be passed anywhere? Should it be converted to a CLI argument?

scripts/publish.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+  const specifiers = args._.slice(1);
+  if (mode === "stable" && specifiers.length === 0) {
+    console.error(
+      "You must specify at least one package version (using the `<name>@<version>` format) to create a stable release.",
+    );
+    process.exit(1);
+  } else if (mode === "unstable" && specifiers.length > 0) {
+    console.error(
+      "You can't specify any package version when creating an unstable release.",
```
angeloashmore: Small phrasing change suggestion: "Versions cannot be specified when creating an unstable release."

scripts/publish.ts
```diff
@@ -0,0 +1,104 @@
...
+    if (currentBranchName !== defaultBranchName) {
+      console.error(
+        `You must be on the \`${defaultBranchName}\` branch to create a stable release.`,
+      );
+      process.exit(1);
+    }
+  }
+}
+
+async function getWorkspaceNames(): Promise<Set<string>> {
```
angeloashmore: Any ideas on how this script could be made package manager agnostic? The script currently relies on Yarn and workspaces, but this release workflow would be useful in our other packages, like `@prismicio/client`, that don't use those.

We don't need and answer now or support that any time soon, but it could be good to consider. 🙂 I'm pretty sure you already mentioned this idea.

scripts/publish.ts
```diff
@@ -0,0 +1,104 @@
...
+  return new Set(
+    stdout
+      .split("\n")
+      .map((line) => (JSON.parse(line) as { name: string }).name),
+  );
+}
+
+function parsePackageNameSpecifier(
+  specifier: string,
+): [name: string, version: string] | undefined {
```
angeloashmore, line 351: Why do you choose an array return value rather than an object? I don't see any problem with it, I'm just curious if there's a benefit we are missing out on elsewhere.

scripts/publish.ts
```diff
@@ -0,0 +1,104 @@
...
+        }
+        const validPackageVersion = semverValid(packageVersion);
+        if (validPackageVersion === null) {
+          console.error(
+            `Invalid package specifier: \`${specifier}\`. Version \`${packageVersion}\` must be a valid SemVer string.`,
+          );
+          process.exit(1);
+        } else if (semverPrerelease(validPackageVersion) !== null) {
+          console.error(
+            `Invalid package specifier: \`${specifier}\`. Version \`${validPackageVersion}\` mustn't include any prerelease component.`,
```
angeloashmore: "mustn't" is an uncommon contraction. "must not" is clearer. 🙂

scripts/publish.ts
```diff
@@ -0,0 +1,104 @@
...
+          );
+          process.exit(1);
+        }
+        return [packageName, validPackageVersion];
```
angeloashmore: We may want to verify that the provided version is higher than the currently published version.

scripts/publish.ts
```diff
@@ -0,0 +1,104 @@
...
+      }),
+    );
+    console.log(packageNameToSpecifiedVersion); // TODO: remove this line.
+  }
+
+  const { stdout: status } = await $`git status --porcelain`;
+  if (status.length > 0) {
+    console.error("This command must be executed on a clean repository.");
+    process.exit(1);
+  } else if (mode === "stable") {
```
angeloashmore: Style suggestion: a switch statement may be easier to maintain and read over time as we add new modes.

```ts
switch (mode) {
  case "stable": {
    // ...
  }

  default: {
    console.error("Invalid mode. Must be one of: stable, unstable.")
    process.exit(1)
  }
}
```

(That switch statement might be better suited elsewhere.)

angeloashmore (commented):
This is looking very good! I just had a handful of style suggestions and questions, but overall, it is definitely going in the right direction. 🙂

scripts/publish.ts
bapmrl: In an ideal world, I would like this function to live in a central place (/monorepo) and consumers would use it to create their own script, passing the `options` argument if needed. Here's an example consuming script, written in another Prismic repository (for instance, `@prismicio/client`):
```
import { publish } from "@prismicio/monorepo/scripts"

// Passing `master` as we haven't renamed the default branch to `main` yet.
publish({ defaultBranchName: "master" })
```

In my opinion, these `options` should be transparent to users in the CLI. Indeed, as a user, I don't want to think of the default branch name. I want it to be inferred automatically for me.

Note that this idea comes from Vite's release scripts: https://github.com/vitejs/vite/blob/main/scripts/release.ts.

scripts/publish.ts
bapmrl: Thanks!

I've changed:
- `You must specify at least one package version (using the '<name>@<version>' format) to create a stable release.` to `At least one package version must be specified (using the '<name>@<version>' format) to create a stable release.`.
- `You can't specify any package version when creating an unstable release.` to `Package versions cannot be specified when creating an unstable release.`.

Is `Package versions` correct in english or should I just say `Versions` like you did?

scripts/publish.ts
bapmrl: I don't know ^^ It's a very good question, but let's solve that after shipping a first version of the new release process.

scripts/publish.ts
bapmrl: I found it elegant because I need the left (`name`) and right (`version`) parts of something that looks like a 2-tuple (`<name>@<version>`). However, it could very well be done with an object instead ^^

scripts/publish.ts
bapmrl: Done! Should I avoid using the `'t` contraction in log messages?

scripts/publish.ts
bapmrl: Indeed, it's planned ^^

scripts/publish.ts
bapmrl: Hmmm, I'll come back to this comment when I have written more code. It'll help me think about it.

package.json
```diff
@@ -47,16 +44,22 @@
...
     "husky": "8.0.3",
-    "lerna": "7.4.1",
     "lint-staged": "13.2.0",
+    "mri": "1.2.0",
     "prettier": "3.0.3",
-    "start-server-and-test": "1.15.5"
+    "rimraf": "5.0.5",
+    "semver": "7.3.8",
+    "start-server-and-test": "1.15.5",
+    "tsx": "4.6.2"
```
angeloashmore, line 62: In the future, it would be nice to remove the need to transpile TypeScript at runtime. Are you considering making this CLI into its own non-Slice Machine package in the future?

packages/adapter-next/package.json
```diff
@@ -49,7 +49,7 @@
 		"dev": "vite build --watch",
 		"format": "prettier --write .",
 		"lint": "eslint --max-warnings 0 --ext .js,.ts .",
-		"prepack": "npm run build",
+		"prepack": "$npm_execpath run build",
```
angeloashmore, line 52: Is there a reason we need to use `$npm_execpath`? Why do we use `npm` instead of `yarn` like the other scripts?

scripts/utils/commandUtils.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export function exec(
+  file: string,
+  args?: string[],
+  options?: ExecaOptions & { dryRun?: false },
+): ExecaChildProcess;
+export async function exec(
+  file: string,
+  args?: string[],
+  options?: ExecaOptions & { dryRun: boolean },
```
angeloashmore, line 33: Should this line be `{ dryRun: true }`? I assume the current line works since the above signature narrows `boolean `to `true`, but it might be more explicit to use `true`. This is subjective, so it's up to you. 🙂

scripts/utils/packageManagerUtils.ts
```diff
@@ -0,0 +1,40 @@
+import { exec } from "./commandUtils";
+
+export async function fetchPackageVersionHistory(
+  name: string,
+): Promise<Set<string>> {
+  try {
+    const { stdout } = await exec("yarn", [
+      ...["npm", "info", name, "--fields", "versions", "--json"],
```
angeloashmore, line 8: Do we have to worry about caching by using `yarn`/`npm` to read version info?

angeloashmore, line 8: Is `...` needed? It looks like the arguments array is spread into an empty array.

scripts/utils/packageManagerUtils.ts
```diff
@@ -0,0 +1,40 @@
...
+    ]);
+    return new Set((JSON.parse(stdout) as { versions: string[] }).versions);
+  } catch (error) {
+    if (
+      error instanceof Error &&
+      "stdout" in error &&
+      typeof error.stdout === "string" &&
+      error.stdout.includes("Response Code: 404 (Not Found)")
```
angeloashmore: When checking for text content, I usually use a regex with case insensitivity. It avoids errors if the message ever changes to "Response **c**ode" rather than "Response **C**ode", for example.

```ts
error.stdout.includes(/response code: 404/i)
```

scripts/publish.ts
angeloashmore: I find that CLIs with a programmatic API have the same options exposed to both interfaces.

For example, the `--dry-run` CLI flag would also be available to `publish()` users. Likewise, the `defaultBranchName` programmatic option would be available to CLI users.

Without that level of interoperability, users may be forced to use one interface over the other, even if it isn't the best fit.

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+      default: { "dry-run": false, "tolerate-republish": false },
+    },
+  );
+  const dryRun = args["dry-run"];
+
+  // Validate the release mode and number of package specifiers.
+  const mode = args._[0];
+  if (mode !== "stable" && mode !== "unstable") {
+    throw new CommandError(
+      "The first argument must be either `stable` or `unstable`.",
```
angeloashmore, line 46: I find CLIs that tell me what I provided in addition to the valid options to be helpful. Knowing what the CLI received helps me know if I incorrectly passed options:

For example, the error would be obvious if I accidentally ran `yarn publish publish stable`:

```
The first argument must be either `stable` or `unstable`. Provided: `publish`
```

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+        if (parsedSpecifier !== undefined) {
+          const [packageName, packageReleaseType] = parsedSpecifier;
+          validatePackageName(packageName, specifier);
+          if (!isReleaseType(packageReleaseType)) {
+            throw new CommandError(
+              `Invalid package specifier: \`${specifier}\`. Increment \`${packageReleaseType}\` must be one of \`major\`, \`minor\` or \`patch\`.`,
+            );
+          }
+          return [packageName, packageReleaseType];
+        }
```
angeloashmore, line 76: It might just be me, but I find the lack of whitespace between lines hard to read. I tend to add empty lines around `if`, `throw`, `switch`, `return`, and groups of `const`/`let`.

Totally stylistic, so what works best for you should be the way to go. 🙂

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+      if (!nonPrivateWorkspaceNames.has(packageName)) {
+        throw new CommandError(
+          `Invalid package specifier: \`${specifier}\`. Package \`${packageName}\` must be a non-private workspace.`,
+        );
+      }
+    }
+  }
+
+  // Ensure the repository is clean and validate the current branch name.
+  const { stdout: gitStatus } = await exec("git", ["status", "--porcelain"]);
```
angeloashmore, line 108: Suggestion: abstract all CLI calls to functions (e.g. `checkIsGitClean()` or `getGitStatus()`). This keeps the publish script unaware of `stdout` and low-level CLI command management.

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+} from "./utils/packageManagerUtils";
+
+publish();
+
+type Options = { defaultBranchName?: string };
+
+async function publish(options?: Options): Promise<void> {
+  process.on("uncaughtException", handleUncaughtException);
+
+  const defaultBranchName = options?.defaultBranchName ?? "main";
```
angeloashmore, line 29: We should be able to detect the default branch name using the `origin` remote naming convention. This will make the script automatically run in both `main` and `master` repositories (as well as any other popular names, like `dev` or `stable`).

```sh
git rev-parse --abbrev-ref origin/HEAD

# => origin/master
```

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+  const packageNames = [
+    ...(packageNameToVersionOrReleaseType !== undefined
+      ? packageNameToVersionOrReleaseType.keys()
+      : nonPrivateWorkspaceNames),
+  ];
+  const packageNameToVersionHistory = new Map(
+    await Promise.all(
+      packageNames.map(
+        async (packageName) =>
+          [packageName, await fetchPackageVersionHistory(packageName)] as const,
```
angeloashmore, line 144: Do we have to worry about npm rate limits? I'm guessing no, but I figured it should be considered.

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+  }
+  function formatInvalidPackageVersionError(
+    packageName: string,
+    packageVersion: string,
+  ): string {
+    return `Package \`${packageName}\` has an invalid version \`${packageVersion}\` in its manifest.`;
+  }
+
+  // Ensure the new versions aren't already in use.
+  if (!args["tolerate-republish"]) {
```
angeloashmore, line 240: Could you explain the purpose of `tolerate-republish`?

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+  console.log(`✏️  Updating manifest${s}...`);
+  for (const nonPrivateWorkspaceName of nonPrivateWorkspaceNames) {
+    const nonPrivateWorkspaceReleaseVersion = packageNameToReleaseVersion.get(
+      nonPrivateWorkspaceName,
+    );
+    await exec(
+      "yarn",
+      [
+        ...["workspace", nonPrivateWorkspaceName],
+        ...["version", nonPrivateWorkspaceReleaseVersion ?? "decline"],
```
angeloashmore, line 264: Could you explain the `decline` value?

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+        "--deferred",
+      ],
+      { dryRun },
+    );
+  }
+  await exec("yarn", ["version", "apply", "--all"], { dryRun });
```
angeloashmore, line 270: Is there a benefit to using `--deferred` rather than applying the versions immediately>

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+
+  // If running in stable mode, both commit and tag the changes. Otherwise, only
+  // tag them.
+  console.log(`🏷️  ${mode === "stable" ? "Committing" : "Tagging"} changes...`);
+  const { stdout: gitDiff } = await exec("git", ["diff", "--name-only"]);
+  if (gitDiff.length === 0 && !dryRun) {
+    throw new CommandError(
+      "There are no changed package versions to commit or tag.",
+    );
```
angeloashmore, line 279: At this point, we have made changes to the repository's state. What happens when an error is thrown? Do we have any safeguards to revert what the script did?

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+    await exec("git", ["commit", "--all", "--message", commitMessage], {
+      dryRun,
+    });
+  }
+  for (const [
+    packageName,
+    packageReleaseVersion,
+  ] of packageNameToReleaseVersion) {
+    const gitTagName = `${packageName}@${packageReleaseVersion}`;
+    await exec(
```
angeloashmore, line 291: Suggestion for a future version: it could be nice to have a `--verbose`/`-v` mode where `exec` logs the command.

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+  // Reset the index and working tree.
+  console.log("🧹 Cleaning...");
+  await exec("git", ["reset", "--hard"], { dryRun });
+
+  // Everything went well.
+  if (dryRun) {
+    console.log("This was a dry run.");
+  } else {
+    const successMessage = `Successfully published ${packageNameToReleaseVersion.size} new package${s}.`;
+    console.log(chalk.green(successMessage));
```
angeloashmore, line 330: Do we log the published package versions?

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+      }
+      // `publishedPrereleaseNumber` is the last element of the
+      // `parsedPublishedVersion.prerelease` array.
+      const [publishedPrereleaseNumber] =
+        parsedPublishedVersion.prerelease.slice(-1);
+      return typeof publishedPrereleaseNumber === "number" &&
+        publishedPrereleaseNumber >=
+          (prevHighestPublishedPrereleaseNumber ?? Number.MIN_SAFE_INTEGER)
+        ? publishedPrereleaseNumber
+        : prevHighestPublishedPrereleaseNumber;
```
angeloashmore, line 387: Have you tried using `semver`'s range functions? There may be one that does what this function does, like `semverSort` or `semverMaxSatisfying` (assuming semver's sorting matches what we need).

angeloashmore (commented):
This is amazing! What is the plan to integrate this into our CI?

Some things I would like to see (which I'm sure you already have in mind):

- Publish alpha versions for all PRs.
- Publish alpha versions quickly so we can validate before tests finish.
- Add a comment to PRs with instructions on running the alpha.

---

### Possible bug

I ran the following command and ran into an issue where the package version did not correctly increment:

```sh
yarn publish --dry-run stable slice-machine-ui:patch
```

This was in the output:

```
git tag --annotate --force --message slice-machine-ui "slice-machine-ui@0.0.0"
```

I'm not sure if only the tag is incorrect or the published version would be as well, but it seems to be a bug. Any idea what could cause that?

scripts/publish.ts
bapmrl: In the end, it was helpful to be able to easily create `Map`s.

scripts/publish.ts
bapmrl: I see 😊 I think I can do that quickly.

However, what should win if both flags and options are supplied? I guess the CLI flags should always win over the API options?

Also, should we be able to pass the `mode` and `specifiers` as options? It would let consumers call the `publish` function without having to pass any CLI args.

scripts/publish.ts
bapmrl: I'm resolving this comment. I've made the change and it will help me better see what remains to be done 👍

scripts/publish.ts
bapmrl: I'm resolving this comment. It will help me better see what remains to be done 👍

scripts/publish.ts
bapmrl: I'm resolving this conversation. I've made the change and it will help me better see what remains to be done 👍

scripts/publish.ts
bapmrl: I'm checking whether the package version has already been published. It's not exactly the same as "higher than the currently published version", but it enables backporting fixes.

bapmrl: I'm resolving this conversation. It will help me better see what remains to be done 👍

scripts/publish.ts
bapmrl: I think a `switch` statement would make the code more verbose than it currently is. However, if more `mode`s are supported in the future, it would be a better fit.

package.json
bapmrl: I would like to, but I can't allocate time for it before Christmas. Let's re-discuss this later.

packages/adapter-next/package.json
bapmrl: If I'm not mistaken,`$npm_execpath` will be set to an absolute path to the packager manager calling that script. It's not super useful in the context of Slice Machine, but if someone was to link one of our packages locally, using for example `npm` or `pnpm`, it wouldn't call `yarn` but their package manager instead.

See https://yarnpkg.com/advanced/lifecycle-scripts#environment-variables.

bapmrl: I'm resolving this conversation. It will help me better see what remains to be done 👍

scripts/utils/commandUtils.ts
bapmrl: It needs to be a `boolean`, otherwise TypeScript won't find the overload:
Screenshot 2023-12-19 at 09 50 29

bapmrl: I'm resolving this conversation. It will help me better see what remains to be done 👍

scripts/utils/packageManagerUtils.ts
bapmrl: We do have to worry about caching. I think the bug you encountered isn't one. You probably forgot to run `yarn` (or `yarn install`) before running the `publish` script. I've noticed the `yarn` CLI doesn't return what's written in the `package.json` files, but some cached versions instead.

bapmrl: I'm resolving this conversation. It will help me better see what remains to be done 👍

bapmrl: It's a technique I've found to avoid getting one line of code for each argument, as they don't all fit on a single line.

bapmrl: I'm resolving this conversation. It will help me better see what remains to be done 👍

scripts/utils/packageManagerUtils.ts
bapmrl: Excellent suggestion! I've changed this.

scripts/publish.ts
bapmrl: I don't really like how this looks like and most errors include what the user provided:
```
bapmrl@baptistes-mbp slice-machine % yarn publish 
The first argument must be either `stable` or `unstable`. Provided: `undefined`.
bapmrl@baptistes-mbp slice-machine % yarn publish publish 
The first argument must be either `stable` or `unstable`. Provided: `publish`.
```

scripts/publish.ts
bapmrl: I've just tried, but I don't really like the look of it. I think I'm too used to how it's currently looking and I also think empty lines are better used to highlight the different big parts of the function.

scripts/publish.ts
bapmrl: I absolutely agree, but considering the time management constraints we have, I prefer not to do that right now 😅

bapmrl: I'm resolving this conversation. It will help me better see what remains to be done 👍

scripts/publish.ts
bapmrl: I actually don't know. Do you think I should use `p-limit`? We already have it in the repository.

scripts/publish.ts
bapmrl: If the script fails after committing/tagging (so during packing/publishing), we might have to re-run it. If `--tolerate-republish` isn't provided, the script will potentially fail, stating that some versions have already been published. Passing `--tolerate-republish` will disable this check and will forward the flag to Yarn's `npm publish` command.

bapmrl: I'm resolving this conversation. It will help me better see what remains to be done 👍

scripts/publish.ts
bapmrl: It tells Yarn to not bump the given workspace during the next release cycle. See: https://yarnpkg.com/cli/version.

bapmrl: I'm resolving this conversation. It will help me better see what remains to be done 👍

scripts/publish.ts
bapmrl: Applying the versions immediately didn't seem to work. After running `yarn workspace "@slicemachine/manager" version 0.17.3-alpha.dev-new-release-process.6` (without the `--deferred` flag), Yarn left a `.yarn/versions/2919f977.yml` file containing:
```
undecided:
  - "@slicemachine/adapter-next"
  - "@slicemachine/adapter-nuxt"
  - "@slicemachine/adapter-nuxt2"
  - "@slicemachine/adapter-sveltekit"
  - "@slicemachine/init"
  - "@slicemachine/plugin-kit"
  - slice-machine-ui
  - start-slicemachine
```

I think it's an expected behaviour, as we don't want to release just one package. In truth, what we want to do is create a bag of packages to release all at the same time. Passing `--deferred` and then running `version apply --all` is just a way to make Yarn apply all the changes at once.

bapmrl: I'm resolving this conversation. It will help me better see what remains to be done 👍

scripts/publish.ts
bapmrl: The updates are only local at this point, so the user can reset their repository's state by running `git reset --hard`. If they try to re-run without doing so, the script will fail earlier with the `This command must be executed on a clean repository` error.

bapmrl: I'm resolving this conversation. It will help me better see what remains to be done 👍

scripts/publish.ts
bapmrl: Excellent suggestion 👍

scripts/publish.ts
bapmrl: No, but the user can see them by running the script in dry mode, prior to creating a release. They can also see what was done by printing the tags with `git tag --list --points-at`.

bapmrl: I'm resolving this conversation. It will help me better see what remains to be done 👍

scripts/publish.ts
bapmrl: I've tried finding a `semver` function that could help, but we have a very specific use case. We don't want to find "any" max satisfying version, but a version that looks the same (meaning, with the same main part, the same prerelease identifier — with all subcomponents matching our maturity level and sanitised branch name — and with a number as its last subcomponent) and whose prerelease number is the highest.

bapmrl: I'm resolving this conversation. It will help me better see what remains to be done 👍

package.json
bapmrl: I'm resolving this conversation. It will help me better see what remains to be done 👍

scripts/publish.ts
bapmrl: I'm resolving this conversation. It will help me better see what remains to be done 👍

bapmrl:
Thanks @angeloashmore for the great help and feedback!

Here are the suggestions you've made that could be implemented in the future:
1. Make the script package manager agnostic.
2. Make the CLI into its own non-Slice Machine package.
3. Turn the CLI into a wrapper for the `publish` function to have the ability to call it programmatically (without having to parse command-line arguments).
4. Abstract all CLI calls to functions to keep the script unaware of low-level CLI command management.
5. Have a `--verbose` mode where exec logs the command.
6. Log the published package versions.
7. Add a comment to PRs with instructions on running the alpha.
8. Move each group of code to individual functions.

Tell me if I've forgotten or misunderstood anything.

I'm also adding @xrutayisire's suggestions:
1. Don't hang indefinitely if the `npm publish` sub-command prompts for an OTP.
2. Consider automatically setting up the upstream branch for the user.
3. Add more whitespace to the function.

After we're done with this pull request, I'll create a follow up Linear issue.

package.json
```diff
@@ -5,6 +5,7 @@
   },
   "packageManager": "yarn@3.6.0",
   "scripts": {
+    "clean": "rimraf --glob '{e2e-projects/*/{.next,.svelte-kit},packages/*/{.next,.svelte-kit,build,coverage,dist,out,storybook-static},playwright/{playwright-report,test-results}}'",
```
xrutayisire: ⚠️ Should we also clean *.tgz in this command and node_modules?

package.json
```diff
@@ -26,11 +27,7 @@
     "cypress-setup": "./cypress-setup.sh",
     "cypress": "yarn cypress-setup && start-server-and-test 'npm run dev --prefix e2e-projects/cypress-next-app' http://localhost:3000 'SM_ENV=development VITE_ENABLE_SENTRY=false npm run slicemachine --prefix e2e-projects/cypress-next-app' http://localhost:9999 'cypress run'",
     "cypress:dev": "yarn cypress-setup && start-server-and-test 'npm run dev --prefix e2e-projects/cypress-next-app' http://localhost:3000 'SM_ENV=staging VITE_ENABLE_SENTRY=false npm run slicemachine --prefix e2e-projects/cypress-next-app' http://localhost:9999 'cypress open'",
-    "bump:interactive": "lerna version prerelease --preid alpha --no-push --exact",
-    "bump:alpha": "lerna version prerelease --preid $npm_config_preid --no-changelog --exact --yes",
-    "bump:production": "lerna version --conventional-graduate --exact",
-    "publish:alpha": "lerna publish from-package --dist-tag alpha --yes",
-    "publish:production": "PUBLIC_SM_UI_SEGMENT_KEY=cGjidifKefYb6EPaGaqpt8rQXkv5TD6P PUBLIC_SM_INIT_SEGMENT_KEY=ED3O1FVsa8DteNyRmAuDyMcPoftJ6VdP lerna publish from-package"
+    "publish": "tsx scripts/publish.ts"
```
xrutayisire, line 30: ❓ Sentry will be a manual env variable as before, correct? Since we don't want to publicly let our Sentry key, I guess?

xrutayisire: ❓ Any reason why choosing `tsx` over classic old way `ts-node`?

package.json
```diff
@@ -47,16 +44,22 @@
...
   },
   "devDependencies": {
+    "@types/semver": "7.3.13",
+    "chalk": "4.1.2",
     "concurrently": "7.6.0",
     "cypress": "12.8.0",
     "cypress-localstorage-commands": "2.2.2",
     "cypress-real-events": "1.7.6",
     "cypress-wait-until": "1.7.2",
+    "execa": "7.1.1",
```
xrutayisire, line 54: ⚠️ `child_process` was not enough for your needs?

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+async function publish(options?: Options): Promise<void> {
+  process.on("uncaughtException", handleUncaughtException);
+
+  const defaultBranchName = options?.defaultBranchName ?? "main";
+
+  // Parse the command-line arguments.
+  const args = mri<{ "dry-run": boolean; "tolerate-republish": boolean }>(
+    process.argv.slice(2),
+    {
+      alias: { n: "dry-run" },
```
xrutayisire, line 35: ⚠️ Why an alias for dry-run and not tolerate-republish? Is `n` classic for dry-run?

scripts/publish.ts
xrutayisire: I also found the code too compact for me. I'm used to seeing more whitespace between lines. 

For example I would do:

```ts
  const nonPrivateWorkspaceNames = await readNonPrivateWorkspaceNames();
  let packageNameToVersionOrReleaseType: Map<string, string> | undefined;
  
  if (mode === "stable") {
    packageNameToVersionOrReleaseType = new Map(
      specifiers.map((specifier) => {
        let parsedSpecifier = parsePackageReleaseTypeSpecifier(specifier);
        
        if (parsedSpecifier !== undefined) {
          const [packageName, packageReleaseType] = parsedSpecifier;
          validatePackageName(packageName, specifier);
          
          if (!isReleaseType(packageReleaseType)) {
            throw new CommandError(
              `Invalid package specifier: \`${specifier}\`. Increment \`${packageReleaseType}\` must be one of \`major\`, \`minor\` or \`patch\`.`,
            );
          }
          
          return [packageName, packageReleaseType];
        }
```

At the moment it's hard for me to review the whole script compacted ^^ So I guess it would be the same to maintain.

bapmrl:
> This is amazing! What is the plan to integrate this into our CI?

We should have a workflow for stable releases early in January. Xavier is currently testing a workflow for unstable releases in parallel to his own issues.

> Some things I would like to see (which I'm sure you already have in mind):
> 
> * Publish alpha versions for all PRs.
> * Publish alpha versions quickly so we can validate before tests finish.
> * Add a comment to PRs with instructions on running the alpha.

I had the two first elements in mind and noted the last one in my previous comment summing-up future improvements.

> ### Possible bug
> I ran the following command and ran into an issue where the package version did not correctly increment:
> 
> ```shell
> yarn publish --dry-run stable slice-machine-ui:patch
> ```
> 
> This was in the output:
> 
> ```
> git tag --annotate --force --message slice-machine-ui "slice-machine-ui@0.0.0"
> ```
> 
> I'm not sure if only the tag is incorrect or the published version would be as well, but it seems to be a bug. Any idea what could cause that?

I think you didn't run `yarn` (or `yarn install`) before the `publish` script. It uses Yarn's CLI, which seems to return some cached versions instead of what the `package.json` files contain. Running `yarn` seems to update the cache.

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+  console.log(`🔍 Calculating new version${s}...`);
+  const maturityLevel = isOnDefaultBranch ? "beta" : "alpha";
+  const sanitizedCurrentBranchName = currentBranchName.replace(/[._/]/g, "-");
+  let packageNameToReleaseVersion: Map<string, string>;
+  if (packageNameToVersionOrReleaseType !== undefined) {
+    // We're creating a stable release, so we only need to use the specified
+    // versions or increments.
+    packageNameToReleaseVersion = new Map(
+      await Promise.all(
+        [...packageNameToVersionOrReleaseType].map(
```
xrutayisire, line 165: ⚠️ Why spreading here?

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+    );
+  } else {
+    // We're creating an unstable release, so for each package, we need to:
+    // 1. Browse the version history to find the highest published prerelease
+    //    number that matches the current branch / prerelease identifier.
+    // 2. Then:
+    //    - If we find a matching prerelease number, increment it and keep the
+    //      main part of the SemVer string (i.e., the `<major>.<minor>.<patch>`
+    //      part) unchanged.
+    //    - If we don't find any, initialize the prerelease number to `1` and
```
xrutayisire: ❓ Why not starting at 0 like it was before? Since version "0.0.0" also exist in semver

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+    return `Package \`${packageName}\` has an invalid version \`${packageVersion}\` in its manifest.`;
+  }
+
+  // Ensure the new versions aren't already in use.
+  if (!args["tolerate-republish"]) {
+    packageNameToReleaseVersion.forEach(
+      (packageReleaseVersion, packageName) => {
+        const packageVersionHistory =
+          packageNameToVersionHistory.get(packageName);
+        if (packageVersionHistory?.has(packageReleaseVersion)) {
```
xrutayisire, line 245: ⚠️ Why not just relying on npm to crash if it already exists? What the gain to prevent check here?

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+        "--deferred",
+      ],
+      { dryRun },
+    );
+  }
+  await exec("yarn", ["version", "apply", "--all"], { dryRun });
+
+  // If running in stable mode, both commit and tag the changes. Otherwise, only
+  // tag them.
+  console.log(`🏷️  ${mode === "stable" ? "Committing" : "Tagging"} changes...`);
```
xrutayisire, line 274: ⚠️ It can help to know that it commit and tag exactly and not only commit that could be possible also:

```suggestion
  console.log(`🏷️  ${mode === "stable" ? "Committing and tagging" : "Tagging"} changes...`);
```

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+    await exec(
+      "git",
+      ["tag", "--annotate", "--force", "--message", packageName, gitTagName],
+      { dryRun },
+    );
+  }
+
+  // Push the changes to GitHub.
+  console.log("🐱 Pushing to GitHub...");
+  await exec("git", ["push", "--atomic", "--follow-tags"], { dryRun });
```
xrutayisire, line 300: ❓`--atomic` is usefull because you're pushing multiple tags + commit possibly?

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+  // Publish the packages that are part of the release, both in parallel and in
+  // topological order.
+  console.log("📦 Packing and publishing to npm...");
+  let npmDistributionTag: string;
+  if (mode === "stable") {
+    npmDistributionTag = "latest";
+  } else {
+    npmDistributionTag = isOnDefaultBranch
+      ? maturityLevel
+      : `${maturityLevel}+${sanitizedCurrentBranchName}`;
```
xrutayisire: ❓ When we will want to init with an specific alpha we will be able to do:

`npx @slicemachine/init@alpha+my-branch.2`, that's the end goal?

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+  }
+  await exec(
+    "yarn",
+    [
+      ...["workspaces", "foreach", "--all"],
```
xrutayisire, line 310: ⚠️ Why spreading here, could be directly in the array, no? Is it for eadability?

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+      "--include",
+      `{${[...packageNameToReleaseVersion.keys()].join(",")}}`,
+      ...["--no-private", "--parallel", "--topological-dev"],
```
xrutayisire, line 313: ⚠️ I don't think we want parallel and at the same time topological-dev, are you sure about the parallel. Btw, I'm saying that because the CI don't seem to like it 😅

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+      ...["npm", "publish"],
+      ...["--tag", npmDistributionTag],
+      ...(args["tolerate-republish"] ? ["--tolerate-republish"] : []),
+    ],
+    { dryRun },
+  );
+
+  // Reset the index and working tree.
+  console.log("🧹 Cleaning...");
+  await exec("git", ["reset", "--hard"], { dryRun });
```
xrutayisire, line 323: ❓ Not sure to understand why this is necessary? In what use cases?

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+    const successMessage = `Successfully published ${packageNameToReleaseVersion.size} new package${s}.`;
+    console.log(chalk.green(successMessage));
+  }
+
+  process.off("uncaughtException", handleUncaughtException);
+}
+
+function parsePackageReleaseTypeSpecifier(
+  specifier: string,
+): [name: string, releaseType: string] | undefined {
```
xrutayisire, line 338: ⚠️ Maybe you told me, but I don't remember why not returning an object that will enforce the correct naming? 

```
{ name: string, releaseType: string }
```

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+  const groups = specifier.match(/^(?<name>\S+):(?<releaseType>\S+)$/)?.groups;
+  return groups?.name !== undefined && groups.releaseType !== undefined
+    ? [groups.name, groups.releaseType]
+    : undefined;
+}
+
+function isReleaseType(value: unknown): value is "major" | "minor" | "patch" {
+  return value === "major" || value === "minor" || value === "patch";
```
xrutayisire, line 346: ℹ️ Btw when doing more than 2 times the same kind of comparaison I like to do that:

```suggestion
  return ["major", "minor", "patch"].includes(value)
```

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+        // Either the main parts or the prerelease identifiers are different.
+        return prevHighestPublishedPrereleaseNumber;
+      }
+      // `publishedPrereleaseNumber` is the last element of the
+      // `parsedPublishedVersion.prerelease` array.
+      const [publishedPrereleaseNumber] =
+        parsedPublishedVersion.prerelease.slice(-1);
+      return typeof publishedPrereleaseNumber === "number" &&
+        publishedPrereleaseNumber >=
+          (prevHighestPublishedPrereleaseNumber ?? Number.MIN_SAFE_INTEGER)
```
xrutayisire, line 385: ❓ I don't understand how and what would be the result if the fallback `Number.MIN_SAFE_INTEGER` is triggered?

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+  ] of packageNameToReleaseVersion) {
+    const gitTagName = `${packageName}@${packageReleaseVersion}`;
+    await exec(
+      "git",
+      ["tag", "--annotate", "--force", "--message", packageName, gitTagName],
+      { dryRun },
+    );
+  }
+
+  // Push the changes to GitHub.
```
xrutayisire, line 298: ⚠️ When creating a branch then pushing the branch then executing yarn and then your script it still fails because of:

Screenshot 2023-12-18 at 19 19 41

Can we make the push command do that for us to prevent an unnecessary fail? To confirm my branch was pushed so it's annoying :/

scripts/publish.ts
```diff
@@ -0,0 +1,408 @@
...
+      : `${maturityLevel}+${sanitizedCurrentBranchName}`;
+  }
+  await exec(
+    "yarn",
+    [
+      ...["workspaces", "foreach", "--all"],
+      "--include",
+      `{${[...packageNameToReleaseVersion.keys()].join(",")}}`,
+      ...["--no-private", "--parallel", "--topological-dev"],
+      ...["npm", "publish"],
```
xrutayisire, line 314: ⚠️ You don't have OTP setup, right? (you should 😛) It's not working with it, the terminal hang indefinitely: 

I could see the error after cancelling the script executing with CTRL+C

Screenshot 2023-12-19 at 13 34 15

xrutayisire (changes requested):
Great work! This will help us in the future! 🚀

I agree with Angelo mentioning the alpha process.

For me, we should ensure to wait for the PR that will introduce alpha in CI to be ready to merge this PR. If not it means the process would be kind of harder for us than today. Because the alpha thing will still happen since it's another issue, and we will not have automatic publish.

Or maybe we should separate the two concern. Let the old deployment in CI and remove it ONLY when we have the new one. You could have done this script without touching the current deployment and let the ticket that touch the CI remove the old CI. WDYT?

package.json
bapmrl: I've added `package.tgz`, but I'm not comfortable cleaning any `.tgz` file and I think we should keep `node_modules`. Cleaning them would prevent us from running other commands after the `clean` script and `yarn install` should override most of them anyway.

package.json
bapmrl: > ❓ Sentry will be a manual env variable as before, correct? Since we don't want to publicly let our Sentry key, I guess?

I won't change how we handle environment variables. It will be the same as before.

> ❓ Any reason why choosing tsx over classic old way ts-node?

I think `tsx` was suggested by @angeloashmore and it seems to be a more modern alternative to `ts-node`. See https://www.npmjs.com/package/tsx#how-is-tsx-different-from-ts-node.

package.json
bapmrl: `child_process` is too hard to use and we're already using `execa` in our repository.

scripts/publish.ts
bapmrl: The `-n` shortcut is used by both Cargo (the Rust package manager) and Yarn for the `--dry-run` flag. Yarn has the `--tolerate-republish` flag, but no shortcut for it. I just mimicked what is done on other, much more widespread tools.

scripts/publish.ts
bapmrl: `packageNameToVersionOrReleaseType` is a `Map`, but `Map`s can't be mapped over so I need to turn it into an array.

scripts/publish.ts
bapmrl: I don't think version `0.0.0` is a commonly found first stable version for packages. I think `1` generally better indicates the notion of "first iteration" and is clearer to us, human beings ^^ We count from 1, not from 0.

xrutayisire: Ok, just good thing to know then 🙂

scripts/publish.ts
bapmrl: What's the point of publishing only a subset of packages we intend to release, just because the user made an error while specifying the desired versions, or the commit has been wrongly tagged? We have the history, so I'm using it to ensure the release makes sense. The `--tolerate-republish` flag is used by both us and Yarn to let users republish if something failed. It's a more explicit process.

scripts/publish.ts
bapmrl: Hmmm, I don't know. We can't document every detail in the log titles. I wanted them to be very short.

scripts/publish.ts
bapmrl: Yes, I want all the annotated tags (which are Git objects) and the commit itself to be either accepted or not. See https://github.blog/2015-04-30-git-2-4-atomic-pushes-push-to-deploy-and-more/#atomic-pushes.

scripts/publish.ts
bapmrl: Not exactly. The prerelease number doesn't make sense for distribution tags. When we validate/QA a branch, we only intend to test the latest version of it. So, we'll just write `npx @slicemachine/init@alpha+my-branch` (without the prerelease number).

scripts/publish.ts
bapmrl: Yes, it's to avoid having each parameter on their own line because of Prettier.

scripts/publish.ts
bapmrl: Yarn is orchestrating everything and correctly handles wait points / dependencies between workspaces, even while running in parallel mode.

scripts/publish.ts
bapmrl: It's necessary after an unstable release, as we don't commit the changed manifests (`package.json` files).

scripts/publish.ts
bapmrl: It's far more practical to create `Map`s from specifiers.

scripts/publish.ts
xrutayisire: Hum it's more because yarn already do it that for us, it's not required in our script. It's good now that we have it, but I feel it's an edge case since yarn handle it for us properly.

scripts/publish.ts
bapmrl: Screenshot 2023-12-19 at 16 18 45
It doesn't seem to be a good refinement for TypeScript.

scripts/publish.ts
bapmrl: `prevHighestPublishedPrereleaseNumber` can be `undefined` and any number, even negative, will be higher than `Number.MIN_SAFE_INTEGER`.

scripts/publish.ts
xrutayisire: Yep you're right sorry but did you check it's possible to do that?
I'm checking with one of your `alpha+dev-new-release-process` but npm don't like it:

```
npm ERR! code EINVALIDTAGNAME
npm ERR! Invalid tag name "alpha+dev-new-release-process" of package "start-slicemachine@alpha+dev-new-release-process": Tags may not have any characters that encodeURIComponent encodes.
```

scripts/publish.ts
bapmrl: I'm not sure it should be the responsibility of the script to configure upstreams for users. I've taken a look at both Yarn's and Vite's release scripts and they don't seem to do it either. I think it's okay to expect users have set the upstream before (why try to publish something that has no corresponding pull request?) and the script can easily recover from that state. The user will simply have to re-run it after having set the remote.

scripts/publish.ts
bapmrl: I have OTP configured, but I'm also logged in (using `yarn npm login`). I'm not sure how I can handle the OTP case because I don't know how prompts can be interacted with when running commands through `execa`. I've noticed Yarn has an `--otp` flag for the `npm publish` subcommand. I'll investigate further ;)

bapmrl: Maybe you know @angeloashmore?

xrutayisire: It could be totally part of the process actually, running yarn `npm login first`, WDYT?

scripts/publish.ts
xrutayisire: Arf, forgot the unknown value, good for string yes

scripts/publish.ts
bapmrl: I agree 🙂 I'm still curious if @angeloashmore knows whether it's easy to handle or not.

bapmrl:
> Great work! This will help us in the future! 🚀
> 
> I agree with Angelo mentioning the alpha process.
> 
> For me, we should ensure to wait for the PR that will introduce alpha in CI to be ready to merge this PR. If not it means the process would be kind of harder for us than today. Because the alpha thing will still happen since it's another issue, and we will not have automatic publish.
> 
> Or maybe we should separate the two concern. Let the old deployment in CI and remove it ONLY when we have the new one. You could have done this script without touching the current deployment and let the ticket that touch the CI remove the old CI. WDYT?

Hello @angeloashmore,

The workflow to publish unstable releases continuously is working as intended now 🙂 Xavier will open a pull request after we have merged this one.

Have a good day,

scripts/publish.ts
angeloashmore: You could export a function that performs all of the work. The CLI would be a wrapper for that function.

Showing it in code may be easier to describe 😅:

```ts
// src/myFunction.ts

export function myFunction(args: { foo: boolean }) {
  console.log("I do things.")
}
```

```ts
// src/cli.ts

import mri from "mri";
import { myFunction } from "./myFunction";

type Args = {
  help: boolean
  version: boolean
  foo: boolean
}

function cli() {
  const args = mri<Args>(process.argv.slice(2));

  if (args.help) {
    // ...
  }

  if (args.version) {
    // ...
  }

  myFunction({
    foo: args.foo,
  });
}
```

```ts
// src/index.ts

export { myFunction } from "./myFunction";
```

With this architecture, the CLI and the function are separate entries to the program, so there is no precedence to determine.

scripts/publish.ts
angeloashmore: As an alternative to adding whitespace to the large top-level function, each group of code could be moved to individual functions. Then the top-level function would call the smaller functions, removing the need to keep groups of code compact.

scripts/publish.ts
angeloashmore: Let's wait and see if we run into rate limiting. If we do, then sure, we can use `p-limit`. 🙂

scripts/publish.ts
angeloashmore: As a future improvement, maybe we could inform the user to run `git reset --hard` to reset state. Keeping the errored state around could help debug problems.

scripts/publish.ts
angeloashmore: `-n` is a convention for `--dry-run` in other CLIs as well. I have no idea why "n" was selected. 😄

scripts/publish.ts
angeloashmore: Oh, so I guess we can't use `+`? What about another `.`?

```
alpha.dev-new-release-process.1
```

scripts/publish.ts
angeloashmore: @bapmrl Out of curiosity, is there a benefit to using `Number.MIN_SAFE_INTEGER` over `Number.NEGATIVE_INFINITY`?

scripts/publish.ts
angeloashmore: @bapmrl Unfortunately, I don't know. Execa has options for handling inputs, but it looks like it only handles input on initialization (e.g. passing a file to a CLI immediately).

The `input` option allows an async iterator, so it's possible you could watch stdout and check for the OTP prompt, which could prompt you to listen to stdin, but it's not easy.

angeloashmore: If the publishing process is quick, you could ask for an OTP code upfront, store it in memory, and pass it to the publish command. That only works if the publish process quickly uses the OPT code.

scripts/publish.ts
bapmrl: Okay, thank you!

scripts/publish.ts
bapmrl: I'll take a look.

scripts/publish.ts
bapmrl: Okay, let's document we need to `yarn npm login` before launching the script and list OTP handling as a future improvement. Thank you!

scripts/publish.ts
bapmrl: I've added it to the list of future improvements. Thanks again!

scripts/publish.ts
bapmrl: I've implemented a `readDefaultBranchName` function using the `git rev-parse --quiet --symbolic-full-name --verify refs/remotes/origin/HEAD` command. It will fail gracefully if the repository has no default branch defined for the `origin` remote. Thank you!

scripts/publish.ts
bapmrl: I've added that to the list of future improvements. Thanks!

scripts/publish.ts
bapmrl: I won't fix that for this release. Let's reconsider this later.

bapmrl:
I've just created the [follow-up Linear issue](https://linear.app/prismic/issue/DT-1878/angelos-and-xaviers-suggestions-to-improve-the-nodejs-script-to). Thanks for all your suggestions to improve the script!

packages/init/src/lib/setupSentry.ts
```diff
@@ -3,22 +3,38 @@ import semver from "semver";
 
 import { SENTRY_EXPRESS_DSN } from "../constants";
 import { checkIsSentryEnabled } from "./checkIsSentryEnabled";
-import { version } from "../../package.json";
+import { name as pkgName, version as pkgVersion } from "../../package.json";
```
angeloashmore: It might be cleaner to use `import * as pkg` and `pkg.name`/`pkg.version`. I believe Vite/Rollup will convert it to `import { name, version }` automatically.

packages/init/src/lib/setupSentry.ts
```diff
@@ -3,22 +3,38 @@ import semver from "semver";
...
+	let environment;
+	if (parsedPkgVersion.prerelease.length === 0) {
+		environment = import.meta.env.MODE || "production";
+	} else if (
+		parsedPkgVersion.prerelease[0] === "alpha" ||
+		parsedPkgVersion.prerelease[0] === "beta"
+	) {
+		environment = parsedPkgVersion.prerelease[0];
+	} else {
+		throw new Error(
```
angeloashmore: Throwing errors in this function may halt `@slicemachine/init`. Since this pertains only to error tracking, should we be more tolerant and allow it to fail, or do you prefer the strict requirement?

(I'm not sure if we catch the error elsewhere.)

packages/slice-machine/next.config.mjs
```diff
@@ -1,30 +1,35 @@
...
+} else if (
+  parsedPkgVersion.prerelease[0] === "alpha" ||
+  parsedPkgVersion.prerelease[0] === "beta"
+) {
+  sentryEnvironment = parsedPkgVersion.prerelease[0];
+} else {
+  throw new Error(
+    `Invalid package version: \`${pkg.name}@${parsedPkgVersion.version}\`. The first prerelease component \`${parsedPkgVersion.prerelease[0]}\` must be either \`alpha\` or \`beta\`.`,
+  );
+}
```
angeloashmore, line 26: Since this code is duplicated in `@slicemachine/init`, `slice-machine-ui`, and `start-slicemachine`, should we abstract it to a common location?

Maybe it could be in `@slicemachine/manager` as a dedicated export:

```ts
import { getStaticEnvironmentInfo } from "@slicemachine/manager";
import * as pkg from "../package.json";

getStaticEnvironmentInfo(pkg);
//  => { version: string; environment: string }
```

(Or individual helpers, whichever feels better.)

packages/slice-machine/next.config.mjs
```diff
@@ -50,7 +55,7 @@ if (process.env.NODE_ENV !== "development") {
       // The Sentry webpack plugin always ignores some files when uploading sourcemaps
       // We actually need them (because of the static export?) to get the complete trace in Sentry
       ignore: [],
-      release: RELEASE_NUMBER,
+      release: parsedPkgVersion.version,
```
angeloashmore, line 58: Do we need to provide `environment` here as well, like `@slicemachine/init`?

packages/init/src/lib/framework.ts
```diff
@@ -58,11 +58,22 @@ export type Framework = {
...
+if (parsedPkgVersion.prerelease.length > 0 && process.env.NODE_ENV !== "test") {
+	const prereleaseIdentifier = parsedPkgVersion.prerelease.slice(-1).join(".");
+	npmDistributionTag = prereleaseIdentifier;
+} else {
+	npmDistributionTag = "latest";
+}
 
 const DEFAULT_DEV_DEPENDENCIES: Record<string, string> = {
-	"slice-machine-ui": isPrerelease ? "alpha" : "latest",
+	"slice-machine-ui": npmDistributionTag,
```
angeloashmore, line 75: Just want to confirm: will this install dependencies using pinned versions (which is what we want)?

I.e. `npm install slice-machine-ui@0.1.2-alpha.my-branch.1` rather than `npm install slice-machine-ui@^0.1.2-alpha.my-branch.1`

packages/init/src/lib/framework.ts
```diff
@@ -58,11 +58,24 @@ export type Framework = {
 	devDependencies: Record<string, string>;
 };
 
-const isPrerelease =
-	semver.prerelease(pkgVersion) !== null && process.env.NODE_ENV !== "test";
+const parsedPkgVersion = semver.parse(pkgVersion);
+if (parsedPkgVersion === null) {
+	throw new Error(
```
bapmrl: ❓ Is it fine to throw here?

packages/manager/src/lib/getSentryEnvironmentInfo.ts
angeloashmore, line 1: In general `lib` is reserved for code used internally by the package. In this case, since `getSentryEnvironmentInfo` is a top-level export, `src/getSentryEnvironmentInfo.ts` may be a better fit.

This isn't a strict rule, but I find it easier to find source code more easily when the export list matches the `src` directory structure.

packages/manager/src/lib/getSentryEnvironmentInfo.ts
```diff
@@ -0,0 +1,32 @@
+import semver from "semver";
+
+type PackageManifest = { name: string; version: string };
+
+type SentryEnvironmentInfo = { environment: string; release: string };
+
+export const getSentryEnvironmentInfo = (
```
angeloashmore: Is this information specific to Sentry? What do you think of making this more generic, like `getEnvironmentInfo()`?

If you think this function will evolve to eventually be Sentry specific, we can keep this name. 🙂

packages/start-slicemachine/src/lib/setupSentry.ts
```diff
@@ -24,16 +27,15 @@ export const setupSentry = async (
 		// User is not logged in, it doesn't matter
 	}
 
-	const isStableVersion =
-		semver.prerelease(sliceMachineVersion) === null &&
-		semver.lte("0.1.0", sliceMachineVersion);
+	const { environment, release } = getSentryEnvironmentInfo({
+		name: "slice-machine-ui",
```
angeloashmore, line 31: Should this be `slice-machine-ui` or `start-slicemachine`?

packages/init/src/SliceMachineInitProcess.ts
```diff
@@ -137,7 +137,11 @@ export class SliceMachineInitProcess {
 				appName: pkg.name,
 				appVersion: pkg.version,
 			});
-			await setupSentry();
+			try {
+				await setupSentry();
+			} catch (error) {
+				// Noop, we probably couldn't determine the Sentry environment
```
angeloashmore: Could we add a quick explicit message that we are intentionally not providing feedback or halting the process?

```js
// noop - We don't want to stop the user from using Slice Machine because of
// failed tracking set up. We probably couldn't determine the Sentry
// environment.
```

bapmrl: Done!

packages/start-slicemachine/src/lib/setupSentry.ts
bapmrl: I think it should be `slice-machine-ui` as we're calling `getRunningSliceMachineVersion` just above.

packages/manager/src/lib/getSentryEnvironmentInfo.ts
bapmrl: Done!

packages/manager/src/lib/getSentryEnvironmentInfo.ts
bapmrl: Done!

packages/start-slicemachine/src/lib/setupSentry.ts
angeloashmore: You're right 👍

packages/init/src/lib/framework.ts
angeloashmore: I'd avoid top-level throwing since it would likely crash any consumer immediately. Although it is very unlikely an error would be thrown, I think it's best we prevent total crashes.

We could convert `FRAMEWORKS` and `UNIVERSAL` to `getFrameworksMetadata()` and `getUniversalMetadata()` to functions that perform the version detection logic.

packages/slice-machine/next.config.mjs
bapmrl: Done, except for the Next.js configuration file.

packages/init/src/lib/framework.ts
bapmrl: To keep changes minimal and simple, I've reworked the code to not throw and to fallback to `latest`. Thanks!

packages/init/src/SliceMachineInitProcess.ts
```diff
@@ -137,7 +137,13 @@ export class SliceMachineInitProcess {
 				appName: pkg.name,
 				appVersion: pkg.version,
 			});
-			await setupSentry();
+			try {
+				await setupSentry();
+			} catch (error) {
```
xrutayisire, line 142: I'm just not sure what would happen if the setup fails, and we still call the trackError that will call Sentry.

Can we just to protect add a console.log like `console.log("Failed to setup Sentry")`, it would help us to know if someone copy-paste an init error in an issue on day.

packages/start-slicemachine/src/StartSliceMachineProcess.ts
```diff
@@ -82,7 +82,13 @@ export class StartSliceMachineProcess {
 		const isTelemetryEnabled =
 			await this._sliceMachineManager.telemetry.checkIsTelemetryEnabled();
 		if (isTelemetryEnabled) {
-			setupSentry(this._sliceMachineManager);
+			try {
+				await setupSentry(this._sliceMachineManager);
+			} catch (error) {
+				// noop - We don't want to stop the user from using Slice Machine
```
xrutayisire, line 88: Same here for a console.log?

xrutayisire (approved):
Amazing work! Good for me! 🚀

PR #1236, opened 2023-12-07
chore(manager): send environment kind with tracking events

angeloashmore:
## Context

DT-1851

## The Solution

`@slicemachine/manager`'s `telemetry.track()` method can now send an `environmentKind` property containing the active environment's kind (one of `prod`, `stage`, or `dev`).

The property is populated only when the `_includeEnvironmentKind` argument is `true`:

```ts
manager.telemetry.track({
  event: "...",
  _includeEnvironmentKind: true,
})
```

The property is opt-in at this time because it requires at least one network request per session to fetch the list of the Prismic repository's environments. The request is cached and only re-fetched if the active environment domain is not present in the cache.

There are cases when we do not want to include the `environmentKind` property, such as during the `@slicemachine/init` process.

In the future, we may switch the property from being opt-in to opt-out.

### Handled edge cases

- When the project's Slice Machine adapter does not support environments, `environmentKind` will be set to `prod` when `_includeEnvironmentKind` is enabled.
- When the user is not logged in, the user is not not authorized, or a network failure occurs when fetching the environments, `environmentKind` will be set to a special `_unknown` value when `_includeEnvironmentKind` is enabled. We cannot fetch metadata about the environment without access to the repository's environments.

### Initial use case

As part of this PR, the `environmentKind` property will be included in the `changes:pushed` event.

## Impact / Dependencies

All tracking events will include this new non-breaking property regardless of where the tracking event was fired. That means events from Slice Machine UI, `@slicemachine/init`, and `start-slicemachine` will now report `environmentKind`.

- [x] The CI is successful.

### DIFF
--- a/packages/manager/src/managers/project/ProjectManager.ts
+++ b/packages/manager/src/managers/project/ProjectManager.ts
@@ -15,6 +15,7 @@ import { DecodeError } from "../../lib/DecodeError";
 import { assertPluginsInitialized } from "../../lib/assertPluginsInitialized";
 import { decodeHookResult } from "../../lib/decodeHookResult";
 import { decodeSliceMachineConfig } from "../../lib/decodeSliceMachineConfig";
+import { findEnvironment } from "../../lib/findEnvironment";
 import { format } from "../../lib/format";
 import { installDependencies } from "../../lib/installDependencies";
 import { locateFileUpward } from "../../lib/locateFileUpward";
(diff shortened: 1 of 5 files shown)


packages/manager/src/managers/project/ProjectManager.ts
```diff
@@ -427,6 +439,49 @@ export class ProjectManager extends BaseManager {
...
+			this._cachedEnvironments = environments;
+		}
+
+		const activeEnvironment = findEnvironment(
+			activeEnvironmentDomain,
+			this._cachedEnvironments || [],
+		);
+
+		if (!activeEnvironment) {
+			throw new UnexpectedDataError(
```
hypervillain, line 475: Is there an action that user can do here? eg. check/delete the .env property?

hypervillain (commented):
Looks good! Just waiting for your answer on the error message to see if we can make sure users don't feel stuck

PR #1237, opened 2023-12-08
test(playwright): environment selector

angeloashmore:
## Context

DT-1825

## The Solution

- Add a `EnvironmentSelector` component to the general `SliceMachinePage`.
- Mock the manager as needed to support environments regardless of which repository is used.

## Impact / Dependencies

N/A

- [ ] The CI is successful.


packages/slice-machine/src/components/PageLayout/PageLayout.tsx
```diff
@@ -16,6 +16,7 @@ export const PageLayout: FC<PageLayoutProps> = ({
     {children}
     <div
       className={clsx(styles.borderTop, styles.borderTopColor[borderTopColor])}
+      data-cy="top-border"
```
xrutayisire: ⚠️ Maybe a bit more details?

Something like `app-top-border` or `window-top-border`?

packages/slice-machine/src/components/SideNav/SideNavEnvironmentSelector.tsx
```diff
@@ -44,9 +45,10 @@ export const SideNavEnvironmentSelector: FC<SideNavEnvironmentSelectorProps> = (
     <Box alignItems="center" gap={16}>
       <Box position="relative">
         <LogoIcon className={styles.logo} />
-        {activeEnvironment !== undefined && environments.length > 1 && (
+        {environments.length > 1 && (
```
xrutayisire, line 48: ✅ Nice catch thanks to the tests I guess!

packages/slice-machine/src/components/SideNav/SideNavEnvironmentSelector.tsx
```diff
@@ -165,13 +172,33 @@ const EnvironmentDropdownMenuItem: FC<EnvironmentDropdownMenuItemProps> = (
...
+  const humanReadableKind = humanReadableKindMap[kind];
 
-  return <div className={clsx(styles.environmentDot[kind], className)} />;
+  return (
+    <div
+      className={clsx(styles.environmentDot[kind], className)}
+      role={asStatus ? "status" : undefined}
+    >
+      {asStatus ? (
+        <VisuallyHidden.Root>
```
xrutayisire, line 200: ✅ Super cool that tests made us think about accessibility!

playwright/pages/SliceMachinePage.ts
```diff
@@ -2,11 +2,16 @@ import { Locator, Page } from "@playwright/test";
...
 import { InAppGuideDialog } from "./components/InAppGuideDialog";
+import { EnvironmentSelector } from "./components/EnvironmentSelector";
+import { LoginDialog } from "./components/LoginDialog";
 
 export class SliceMachinePage {
   readonly page: Page;
+  readonly topBorder: Locator;
   readonly menu: Menu;
   readonly inAppGuideDialog: InAppGuideDialog;
+  readonly environmentSelector: EnvironmentSelector;
```
xrutayisire: ⚠️ Environment selector should be in Menu for me, it's today specific of this area. WDYT?

playwright/pages/components/EnvironmentSelector.ts
```diff
@@ -0,0 +1,61 @@
...
+    this.page = page;
+
+    /**
+     * Static locators
+     */
+    this.loginTextButton = page.getByRole("button", {
+      name: "Login required",
+      exact: true,
+    });
+    // TODO: Replace with `getByRole` once `<IconButton>` supports labels.
```
xrutayisire: ⚠️ Maybe you could create a ticket for that and put ths ticket number like:
`// TODO(DT-1234): Replace with getByRole once <IconButton> supports labels.`

Same for others. WDYT?

playwright/pages/components/EnvironmentSelector.ts
```diff
@@ -0,0 +1,61 @@
...
+    return this.page.getByRole("menuitem", { name, exact: true });
+  }
+
+  /**
+   * Actions
+   */
+  async selectEnvironment(name: string) {
+    await this.dropdownTrigger.click();
+    await this.getEnvironmentMenuItem(name).click();
+    await expect(this.environmentName).toHaveText(name);
```
xrutayisire: ⚠️ I think it's better to always check visible elements with:

```
await expect(this.environmentName.getByText(name, { extact: true })).toBeVisible()
```

It ensures it's not hidden some way. WDYT adding something in best practices also?

playwright/pages/components/EnvironmentSelector.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  /**
+   * Assertions
+   */
+  // TODO: Type `kind` with `@slicemachine/manager`'s `Environment["kind"]`.
```
xrutayisire: ⚠️ Also add issue number, It already exist for this one:
`// TODO(DT-1854): Type kind with @slicemachine/manager's Environment["kind"].`

playwright/pages/components/EnvironmentSelector.ts
```diff
@@ -0,0 +1,61 @@
...
+  async checkSelectedEnvironmentKind(kind: "prod" | "stage" | "dev") {
+    const humanReadableKindMap = {
+      prod: "Production",
+      stage: "Staging",
+      dev: "Development",
+    } as const;
+
+    await expect(this.page.getByRole("status")).toHaveText(
```
xrutayisire: ⚠️ Dangerous since tomorrow with two "status" it would break, right? We could use something before instead of `page` to target the right element. 

Also, good one here to use `toHaveText` since it's not visible

playwright/pages/components/EnvironmentSelector.ts
```diff
@@ -0,0 +1,61 @@
...
+    await this.dropdownTrigger.click();
+    await this.getEnvironmentMenuItem(name).click();
+    await expect(this.environmentName).toHaveText(name);
+  }
+
+  /**
+   * Assertions
+   */
+  // TODO: Type `kind` with `@slicemachine/manager`'s `Environment["kind"]`.
+  async checkSelectedEnvironmentKind(kind: "prod" | "stage" | "dev") {
```
xrutayisire: ❓ I was a big confused with the naming, maybe we could be specific to what you're testing, the dot, no?

playwright/tests/common/environment.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+      await sliceMachinePage.gotoDefaultPage();
+      await sliceMachinePage.environmentSelector.loginIconButton.click();
+      await expect(sliceMachinePage.loginDialog.submitButton).toBeVisible();
+    },
+  );
+
+  test.run()(
+    'I can see "Production" environment when no environments exist',
+    async ({ sliceMachinePage }) => {
+      await mockManagerProcedures({
```
xrutayisire, line 29: ⚠️ Are you sure about your mocking here? I feel like "no env exist" means that either you have an empty array or no active env. Here you return "production", but what about the mock to active env? It could happen that a Slice Machine tested version have one.

xrutayisire: I guess can you could mock `fetchEnvironments` and `readEnvironment` to return nothing and so it should be "Production" as a fallback.

playwright/tests/common/environment.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+          },
+        ],
+      });
+
+      await sliceMachinePage.gotoDefaultPage();
+
+      await sliceMachinePage.environmentSelector.selectEnvironment(
+        environments[0].name,
+      );
+      await expect(sliceMachinePage.topBorder).toHaveCSS(
```
xrutayisire: ⚠️ I feel like it's a bit of implementation details here. Can we not have a role as you did for the dot with an invisible text? I feel it would be the same here as the dot. WDYT?

xrutayisire (changes requested):
Super nice tests with some improvements for ours users in the codebase on top of the test! Great work for your first tests, I hope it was easy and will be more and more easy with time! 🤩

packages/slice-machine/src/components/PageLayout/PageLayout.tsx
bapmrl: ⚠️ UI components shouldn't contain domain/business/app concerns in my opinion. I think we should assign `data-cy`/`data-testid` through spread props in the `AppLayout` component. That way, we could very well move `PageLayout` to the Editor's design system in the future.

packages/slice-machine/src/components/SideNav/SideNavEnvironmentSelector.tsx
```diff
@@ -68,7 +70,11 @@ export const SideNavEnvironmentSelector: FC<SideNavEnvironmentSelectorProps> = (
         ) : undefined}
 
         {variant === "default" ? (
-          <Text component="span" className={styles.activeEnvironmentName}>
+          <Text
+            component="span"
+            className={styles.activeEnvironmentName}
+            data-cy="active-environment-name"
```
bapmrl, line 77: 💡 Same comment, but it's harder to fix there because `SideNav` already has domain/business/app concerns. Let's keep it this way for now. I think we will have to refactor this whole component at some point, because it appears Page Builder will have nearly the same menu in the future. It could be a good opportunity to simplify it as well (because it's far too complicated for what it does as a UI component).

packages/slice-machine/src/components/PageLayout/PageLayout.tsx
angeloashmore: > UI components shouldn't contain domain/business/app concerns in my opinion.

I agree as it makes components reusable.

> I think we should assign data-cy/data-testid through spread props in the AppLayout component. That way, we could very well move PageLayout to the Editor's design system in the future.

Is `data-cy`/`data-testid` something we need to make available to all components? If so, should we be adding the prop to all of our `Prop` types?

We could make a `TestingProps` type that we could intersect with our other types:

```ts
type ExampleProps = FC<PropsWithChildren & TestingProps>
```

In this specific case, we can't spread props to the top border component since it isn't the top-level element. We could implement something like a `borderTopProps` prop, but that feels inelegant. Perhaps we need to move the border out to its own component?

playwright/pages/SliceMachinePage.ts
angeloashmore: Works for me! Good catch.

playwright/pages/components/EnvironmentSelector.ts
angeloashmore: Great idea! I'll update the comments with tickets.

playwright/pages/components/EnvironmentSelector.ts
angeloashmore: Done 🙂

playwright/pages/components/EnvironmentSelector.ts
angeloashmore: Makes sense to me. 👍 I added a short section in the document.

playwright/pages/components/EnvironmentSelector.ts
angeloashmore: I originally tried to test more for behavior rather than the specific dot, but it didn't really work well. I changed the implementation to `checkHasSelectedEnvironmentDotKind()` with the following:

```ts
  async checkHasSelectedEnvironmentDotKind(kind: "prod" | "stage" | "dev") {
    const humanReadableKindMap = {
      prod: "Production",
      stage: "Staging",
      dev: "Development",
    } as const;

    await expect(this.page.getByTestId("active-environment-dot")).toBeVisible();
    await expect(this.page.getByTestId("active-environment-dot")).toHaveText(
      `${humanReadableKindMap[kind]} environment`,
    );
  }
```

playwright/pages/components/EnvironmentSelector.ts
angeloashmore: I replaced the status check with a more direct `data-cy` locator: https://github.com/prismicio/slice-machine/pull/1237#discussion_r1430798695

playwright/tests/common/environment.spec.ts
angeloashmore: The environment endpoint returns `prod`, `stage`, and `dev` environments, so the array should never be empty (it will always have at least the `prod` environment).

It is only treated as "empty" if the user is logged out since we are unable to fetch a list of the repository's environments.

playwright/tests/common/environment.spec.ts
angeloashmore: We could change the border and test to use `role="status"`, but then we wouldn't be testing the color. Instead, we would be testing the status.

I don't think we should mark the top border as a status item, since we already have another status indicator providing the same information (the dot). Thus, we would be adding text purely for testing purposes since it would not be announced.

In this case, would be make more sense to use a computed `data-cy` (e.g. `data-cy="top-border-prod"`) or `data-environment-kind="prod"`? We don't check the color of the dot either, so maybe we should just check for kind instead.

packages/slice-machine/src/components/SideNav/SideNavEnvironmentSelector.tsx
angeloashmore: I agree - we should refactor the whole menu in a future PR.

playwright/pages/components/EnvironmentSelector.ts
angeloashmore: (I'm resolving the conversation so we can track what was addressed, but feel free to re-open it if you have any other suggestions!)

packages/slice-machine/src/components/PageLayout/PageLayout.tsx
xrutayisire: I think anyway we need that management with TestingProps I needed it also for the Add IconButton within `WindowTabsList`. We can also find other ways but I guess we need a way to pass testing props to pure generic UI component. WDYT @bapmrl?

playwright/tests/common/environment.spec.ts
xrutayisire: Hum and since you don't mock the active env, what if the active env is "stage" for example?

playwright/tests/common/environment.spec.ts
xrutayisire: Honestly, I guess for this first test it's ok. Product is unsure of this part, so if one day we change the color or something like this we could make the improvement at the same time. WDYT?

packages/slice-machine/src/components/PageLayout/PageLayout.tsx
bapmrl: I'm fine with both `TestingProps` or passing them through spread `...otherProps` ^^

Regarding `borderTopProps`, I don't have any better idea, but I'm not sure it's worth creating a dedicated sub-component just for the sake of testing.

angeloashmore: @bapmrl With `...otherProps`, wouldn't we still need to type the props with `data-cy: string`? It would also need to be added up the tree to top-level consumers.

This feels dirty, but we could globally add `data-cy` as a prop to `JSX.ElementType`:

```ts
declare namespace JSX JSX {
  interface ElementClass {
    "data-cy"?: string;
  }
}
```

Then when we use `...othersProps`, it should pass down automatically.

xrutayisire: I did that @angeloashmore btw for the moment:

Screenshot 2023-12-19 at 20 27 40

WDYT about something like this, the name is not known then?

bapmrl: Hello!

@angeloashmore, I think `data-*` props don't need to be typed. TypeScript will let them through, but the underlying component needs to pass down the props in the correct places.

@xrutayisire, this is really unclear code. What are you trying to achieve?

xrutayisire: So here, I want to be able to target directly the icon button from within `WindowTabsList`. 

For this specific use case, the best will be to add a label to the button, as Angelo suggested in the past. My point is to show that if you want to target a specific nested component with a test id, you will need to find a way to pass the id without mentioning in props the "data-cy" for example.

angeloashmore: I solved this specific issue by changing the extra `<div>` used to draw the top border into an actual border on the layout's `<div>`. As a result, there is no need to locate a child element.

However, this is still a general issue that would need to be addressed eventually.

I prefer generalizing how we handle test props using something like `...otherProps` (so we don't hard-code what a test prop is) or `testProps?: Record<string, unknown>` (so that we can isolate test-specific props and potentially remove them in prod).

Using a convention like `testingProps` allows us to extend it to child components, like `iconButtonTestProps`.

```tsx
<WindowTabsList
  onAddNewTab={() => console.log("Added new tab")}
  testProps={{ "data-cy": "foo" }}
  iconButtonTestProps={{ "data-cy": "bar" }}
>
  {/* ... */}
</WindowTabsList>
```

playwright/tests/common/environment.spec.ts
angeloashmore: This case led me down a larger refactor: https://github.com/prismicio/slice-machine/pull/1253

In the case where an invalid environment is selected, Slice Machine will revert to the production environment. The above PR has a test for that scenario.

playwright/tests/common/environment.spec.ts
angeloashmore: Works for me 👍

playwright/README.md
```diff
@@ -179,14 +182,21 @@ this.customTypesLink = this.menu.getByText("Custom types", {
...
 
-3. Don't use `.locator()`
+### Prefer using `.toBeVisible()` when testing the presense of visble elements
+
+`.toBeVisible()` ensures the element is present and visible to users. Explicitly checking for visibility prevents a false-positive test where the element is in the DOM, but cannot be seen or interacted by the user.
+
+> [!NOTE]
+> This preference can be ignored if the targeted element is hidden.
+
+If the element is purposely hidden, use
```
xrutayisire: You miss something here, correct?

xrutayisire (approved):
LGTM, super work! Lot of improvements thanks to this PR! And a lot of discussions ahead. I think we should merge it ASAP so we can go on and discussions can continue to happen async for data test props 🙂

playwright/README.md
angeloashmore: Haha oops, thanks for catching that. I must have gotten distracted half way through the sentence. 😅

PR #1239, opened 2023-12-11
test(playwright): Convert Cypress tests to Playwright -1 

xrutayisire:
## Context

- Completes DT-1855

## The Solution

The goal is to be able to remove Cypress ASAP. For that, I will first focus on converting all Cypress tests that we want to keep. 
In the meantime and later on, we will continue to cover the rest of the app with new e2e tests, even if they didn't exist in Cypress.

## Impact / Dependencies

- None

- [x] The CI is successful.


playwright/tests/pageTypes/pageTypeBuilder.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+          "data.meta_description",
+        ),
+      ).toBeVisible();
+
+      await expect(
+        pageTypesBuilderPage.getStaticZoneListItemFieldName("Meta Image"),
+      ).toBeVisible();
+      await expect(
+        pageTypesBuilderPage.getStaticZoneListItemFieldId("data.meta_image"),
+      ).toBeVisible();
```
angeloashmore, line 26: This pattern of checking the field name and field ID occurs several times. Is it possible to make checking a field easier?

For example:
```ts
pageTypesBuilderPage.checkStaticZoneField({
  name: "Meta Image",
  id: "meta_image"
})
```

playwright/tests/pageTypes/pageTypesTable.spec.ts
```diff
@@ -0,0 +1,129 @@
+import { expect } from "@playwright/test";
+
+import { test } from "../../fixtures";
+import { generateRandomId } from "../../utils/generateRandomId";
+
+test.describe("Page types table", () => {
+  test.run()(
+    "I can create a reusable page type",
+    async ({ pageTypesTablePage, pageTypesBuilderPage }) => {
+      await pageTypesTablePage.goto();
```
angeloashmore, line 10: `.goto()` seems to be called at the beginning of every test. Should we use [`beforeEach()`](https://playwright.dev/docs/api/class-test#test-before-each-1) to reduce the boilerplate?

playwright/tests/slices/slicesList.spec.ts
```diff
@@ -64,19 +65,43 @@ test.describe("Slices", () => {
       await slicesListPage.clickCard(newSliceName);
 
       await expect(
-        sliceBuilderPage.breadcrumb.getByText(newSliceName),
+        sliceBuilderPage.getBreadcrumbLabel(newSliceName),
       ).toBeVisible();
     },
   );
 
+  // See: #791
```
angeloashmore: What does #791 refer to? Maybe we can add a URL here to make it extra clear.

```ts
// See: https://github.com/prismicio/slice-machine/issues/791
```

angeloashmore (approved):
LGTM! Just had a few small comments to consider.

playwright/tests/slices/slicesList.spec.ts
xrutayisire: Agree, since we have both Linear and GitHub, yes it refers to GitHub

playwright/tests/pageTypes/pageTypesTable.spec.ts
xrutayisire: I almost never use `beforeEach`, I know it's for that but also create a bit more complexity when you read tests. For me, a good pattern is always to be able to read and understand a test without having to remember some part that is done before for you somewhere else in the file. I feel the duplication is not a big problem in tests, and I tend to duplicate tests over adding more complexity. At the contrary of what I would do in a prod code where duplication of code is prone to bugs, of course.

My comment here can also refer to the comment above, as I tend to duplicate to simplify code most of the time.

If you can quickly read [this article](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing) you will see what kind of dev I'm for test haha. Mostly I'm talking about the [Nesting](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing#nesting) part. I really like Kent C. Dodds on testing subject and I mostly agree with his articles.

I think for the comment above, it makes more sense because we are actually describing enough what the check behind do with the function name, and it's our pattern with POM, so I like it and will do that 🙂

I'm more afraid of bringing a `beforeEach` here because I would like us to write tests in a `flat` manner, where everything can be understood from a single test. 

WDYT?

playwright/tests/pageTypes/pageTypeBuilder.spec.ts
xrutayisire: Good one, replied on the next comment 🙂

playwright/tests/pageTypes/pageTypesTable.spec.ts
angeloashmore: I totally agree about being able to understand the whole test within the test's function. Your reasoning to keep `beforeEach()` out of the tests makes sense to me. 👍

Same for the field name and ID checks. I prefer duplication as well over premature abstraction. It's a balance, and I think we're still on the positive side. 🙂

PR #1242, opened 2023-12-13
test: add E2E variation tests

bapmrl:
Completes DT-1822.

### DIFF
--- a/playwright/pages/SliceBuilderPage.ts
+++ b/playwright/pages/SliceBuilderPage.ts
@@ -1,11 +1,19 @@
 import { expect, Locator, Page } from "@playwright/test";
 
+import { AddVariationDialog } from "./components/AddVariationDialog";
+import { RenameVariationDialog } from "./components/RenameVariationDialog";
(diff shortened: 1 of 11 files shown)


playwright/pages/SliceBuilderPage.ts
```diff
@@ -1,11 +1,19 @@
 import { expect, Locator, Page } from "@playwright/test";
 
+import { AddNewVariationDialog } from "./components/AddNewVariationDialog";
```
xrutayisire: ❓Why not `AddVariationDialog` just a bit shorter since it's for sure `New`?
I don't really have a preference, more a question here 😉

xrutayisire (approved):
I forgot a super small question during our call.
Otherwise, LGTM super nice tests! 🤩 
It's cool to see we are starting to get back on our feet for e2e tests, and I really hope it will help us ship with more confidence!

playwright/pages/SliceBuilderPage.ts
bapmrl: I've removed the `New` part 🙂

PR #1244, opened 2023-12-14
fix(adapter-svelte): support previews in `@slicemachine/init` (DT-1867)

angeloashmore:
## Context

DT-1867

## The Solution

Generate the necessary files. The file contents are based off the [SvelteKit starter files](https://github.com/prismicio-community/sveltekit-starter-prismic-minimal).

## Impact / Dependencies

This PR only affects projects that:

1. Have an existing or vanilla SvelteKit project.
2. Run `@slicemachine/init` to bootstrap their Prismic project without a prismic starter.

- [ ] The CI is successful.

### DIFF
--- a/packages/adapter-sveltekit/src/hooks/documentation-read.ts
+++ b/packages/adapter-sveltekit/src/hooks/documentation-read.ts
@@ -32,8 +32,8 @@ export const documentationRead: DocumentationReadHook<PluginOptions> = async (
 			dataFileContent = source`
 				import { createClient } from "$lib/prismicio";
 
-				export async function load({ params, fetch }) {
-					const client = createClient({ fetch });
+				export async function load({ params, fetch, cookies }) {
+					const client = createClient({ fetch, cookies });
 
 					const page = await client.getByUID("${model.id}", params.uid);
 
@@ -56,8 +56,8 @@ export const documentationRead: DocumentationReadHook<PluginOptions> = async (
 			dataFileContent = source`
 				import { createClient } from "$lib/prismicio";
 
-				export async function load({ params, fetch }) {
-					const client = createClient({ fetch });
+				export async function load({ params, fetch, cookies }) {
+					const client = createClient({ fetch, cookies });
 
 					const page = await client.getSingle("${model.id}");
 
--- a/packages/adapter-sveltekit/src/hooks/project-init.ts
+++ b/packages/adapter-sveltekit/src/hooks/project-init.ts
@@ -55,7 +55,7 @@ const createPrismicIOFile = async ({
 	if (isTypeScriptProject) {
 		contents = source`
 			import * as prismic from "@prismicio/client";
-			import { CreateClientConfig, enableAutoPreviews } from '@prismicio/svelte/kit';
+			import { type CreateClientConfig, enableAutoPreviews } from '@prismicio/svelte/kit';
 			import config from "../../slicemachine.config.json";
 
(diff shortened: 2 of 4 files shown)


packages/adapter-sveltekit/src/hooks/project-init.ts
```diff
@@ -224,6 +224,42 @@ const createPreviewRouteMatcherFile = async ({
...
+		`+server.${extension}`,
+	);
+
+	if (await checkHasProjectFile({ filename, helpers })) {
+		return;
+	}
+
+	const contents = source`
+		import { redirectToPreviewURL } from '@prismicio/svelte/kit';
+		import { createClient } from '$lib/prismicio';
```
xrutayisire, line 246: Interesting about what $ will do here?

packages/adapter-sveltekit/src/hooks/documentation-read.ts
```diff
@@ -32,8 +32,8 @@ export const documentationRead: DocumentationReadHook<PluginOptions> = async (
 			dataFileContent = source`
 				import { createClient } from "$lib/prismicio";
 
-				export async function load({ params, fetch }) {
-					const client = createClient({ fetch });
+				export async function load({ params, fetch, cookies }) {
```
xrutayisire, line 35: We need to add the cookies manually now?

packages/adapter-sveltekit/src/hooks/project-init.ts
```diff
@@ -55,7 +55,7 @@ const createPrismicIOFile = async ({
 	if (isTypeScriptProject) {
 		contents = source`
 			import * as prismic from "@prismicio/client";
-			import { CreateClientConfig, enableAutoPreviews } from '@prismicio/svelte/kit';
+			import { type CreateClientConfig, enableAutoPreviews } from '@prismicio/svelte/kit';
```
xrutayisire, line 58: Is it required now, or it's just cleaner code?

xrutayisire (approved):
LGTM, 3 quick questions just for my knowledge more than a review 🙂

packages/adapter-sveltekit/src/hooks/project-init.ts
angeloashmore: Required now. SvelteKit 2 adds `verbatimModuleSyntax` to the generated `tsconfig.json` file, which requires declaring `type` when importing types.

angeloashmore: (Good question!)

packages/adapter-sveltekit/src/hooks/documentation-read.ts
angeloashmore: Yes, the client needs the request's cookies to check if a preview is active and to retrieve the preview ref. Content is still fetched on the server during a preview, which requires the preview ref.

Next.js projects don't need to explicitly pass cookies because Next.js exposes a request's cookies via the `cookies()` function, which reads data implicitly using `AsyncLocalStorage` (aka magic).

packages/adapter-sveltekit/src/hooks/project-init.ts
angeloashmore: SvelteKit comes with a built-in `$lib` path alias for `src/lib`. It is idiomatic to use the alias, and is helpful for us since we can generate files without computing relative paths.

PR #1245, opened 2023-12-15
Use profile shortId to match Dev env users

hypervillain:
Fix for the Env Switch, as the Core API returns users shortId rather than full userID


xrutayisire (approved):
LGTM

PR #1246, opened 2023-12-15
test(playwright): Convert Cypress tests to Playwright - 2

xrutayisire:
## Context

- Completes DT-1856

## The Solution

Sorry for the big PR 😬

This PR include:
- Allow parallel tests
- Remove any reference to login since we don't need it anymore (yet)
- Run tests on prod version not dev to speed up execution
- Improve mock procedures function to allow mocking multiple times the same procedure
- Comment Cypress CI since it's crashing without alpha now
- Remove converted Cypress tests
- Rename "Got It" to "Got it" where it's not done
- Improve README best practices
- Handle mock with functions to generate more than one
- Create tests for Review dialog
- Creates tests for more push use cases
- Creates tests for update available in side menu bar
- Create tests for tutorial video tooltip

## Impact / Dependencies

- None

- [x] The CI is successful.


playwright/pages/components/ReviewDialog.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Components
+     */
+    // Handle components here
+
+    /**
+     * Static locators
+     */
+    this.messageTextarea = this.dialog.getByPlaceholder("Tell us more...", {
```
angeloashmore, line 21: Do we have a label to use instead of placeholder? If we do, could we use `getByLabel()` instead of `getByPlaceholder()`?

If not, it's something we may want to raise as an accessibility issue.

playwright/mocks/generateLibraries.ts
```diff
@@ -0,0 +1,86 @@
+import { SharedSlice } from "@prismicio/types-internal/lib/customtypes";
+
+import { generateRandomId } from "../utils";
+
+type GenerateTypesArgs = { nbSlices: number };
```
angeloashmore: What does `nbSlices` mean? "nb" is unclear - is there a clearer name?

playwright/mocks/generateTypes.ts
```diff
@@ -0,0 +1,48 @@
+import { CustomType } from "@prismicio/types-internal/lib/customtypes";
+
+import { Library } from "./generateLibraries";
+import { generateRandomId } from "../utils";
+
+type GenerateTypesArgs = {
+  nbTypes: number;
```
angeloashmore: Same comment as above: is there a clearer name than "nb"?

playwright/tests/common/sideNav.spec.ts
```diff
@@ -0,0 +1,148 @@
+import { expect } from "@playwright/test";
+
+import { test } from "../../fixtures";
+import { generateLibraries, generateTypes } from "../../mocks";
+
+test.describe("Side nav", () => {
```
angeloashmore, line 6: This isn't a comment specific to this PR: Do you think the `describe` groups are necessary?

We seem to have one `describe` per `.spec.ts` file. The Playwright UI list could be shallower if we removed the `describe` groups.

angeloashmore (approved):
Looks good to me! I just had a few questions I left as comments throughout the code.

Could you explain how we are able to remove the `loggedIn` test config? Should we be mocking manager responses for anything that requires authentication?

playwright/mocks/generateLibraries.ts
xrutayisire: Yes sorry my bad here, `slicesCount` is better!

playwright/mocks/generateTypes.ts
xrutayisire: `typesCount`, thanks, you're right

playwright/tests/common/sideNav.spec.ts
xrutayisire: YESSSS! I had the same thought and I want to remove them ALL!

Because we prefer to have a small file, and so create another file instead of another describe, we can and should remove them. 
I will do it right after the two PRs, or just before merging the second one so I don't have conflict with myself 😅

xrutayisire:
@angeloashmore 

> Could you explain how we are able to remove the loggedIn test config? Should we be mocking manager responses for anything that requires authentication?

Yes, for standalone test at the moment. It finally set a common ground for all standalone tests that we are writing now.

We mock everything that query the outside world that we want to test for a specific test. So if you test something related to an authentication in wroom, yes mock. 

DX is much nicer like that since you don't care about anything external, email, password, etc... Let me know what you think and we could come back to something different.

playwright/pages/components/ReviewDialog.ts
xrutayisire: Sadly there is no label and that's not great. It's an old component and when we will redo it we need to change it.

PR #1247, opened 2023-12-15
fix: pin Slice Machine dependency versions

angeloashmore:
## Context

DT-1871

## The Solution

- Change all `workspace:^` versions to `workspace:*`.
- To fix a bug with r19 mismatching, pin r19 to the current latest version (0.1.8).

## Impact / Dependencies

The release process will change as `workspace:` will use exact versions rather than `^` versions.

This is part 1 of pinning all dependencies. See DT-1870 for details on part 2.

- [x] The CI is successful.


xrutayisire (approved):
LGTM, thanks 🙂

PR #1250, opened 2023-12-19
test(playwright): Convert Cypress tests to Playwright - 3

xrutayisire:
## Context

- Completes DT-1876

## The Solution

This PR include:
- Create tests for slice simulator tooltip
- Create first tests for changelog page
- Create tests for page type tabs
- Create tests for slice creation restriction

## Impact / Dependencies

- Need the PR `xru/convert-cypress-tests-2` to be merged first

- [x] The CI is successful.


playwright/tests/changelog/changelog.spec.ts
```diff
@@ -1,5 +1,64 @@
...
+      );
+
+      await changelogPage.goto();
+      await changelogPage.checkReleaseNotes(releaseNotes);
+      await expect(changelogPage.breakingChangesWarning).toBeVisible();
+    },
+  );
+
+  test.run()(
+    "I cannot see a warning if the selected release note don't have a breaking changes",
```
angeloashmore: Suggested edit:

> I cannot see a warning if the selected release does not contain a breaking change

playwright/tests/slices/slicesList.spec.ts
```diff
@@ -48,6 +48,23 @@ test.describe("Slices list", () => {
...
+    "I cannot create a slice with a name that already exists",
+    async ({ slicesListPage, slice }) => {
+      await slicesListPage.goto();
+      await slicesListPage.openCreateDialog();
+
+      const { nameInput, submitButton } = slicesListPage.createSliceDialog;
+
+      await nameInput.fill(slice.name);
+      await expect(submitButton).toBeDisabled();
+      await submitButton.click({ force: true });
```
angeloashmore, line 61: Rather than forcing a button click, should we test for form submission (e.g. pressing Enter)?

angeloashmore (commented):
Looks great! I just suggested a tiny wording change and had a question about forcing a form submission.

playwright/tests/changelog/changelog.spec.ts
xrutayisire: Thanks, it's better

playwright/tests/slices/slicesList.spec.ts
xrutayisire: That was a good idea, BUT sadly again, the way form management is done is that the error message is only displayed after a click outside of the input --"
This will need to be redone with new form management.

PR #1251, opened 2023-12-19
ci(deploy): Deploy alpha and beta in CI

xrutayisire:
## Context

- Completes DT-1771

## The Solution

This job is possible thanks to the great work done by Baptiste on the previous PR! 🙏

- Create a new workflow to easily handle condition with "on" keyword.
- Trigger on
  - push on main (beta)
  - opening, reopening, updating a pull request
  - never on a tag
  - not on a commit that start with "release: " to prevent double deployment of a beta with a latest version (manual now but automatic in the future)
- `fetch-depth: 0` so publish script have git history
- `ref: ${{ github.head_ref || github.ref }}` so publish script can access head on a detached event (pull reuqests events are detached)
- `git config` to handle author

Question: should I leave `--tolerate-republish`? It could help us re-trigger the CI if something went wrong, and since we are talking about a beta / alpha here it should be ok to do that. WDYT?

## Impact / Dependencies

- It takes less than 6m, in total, to have a beta / alpha

- [ ] The CI is successful.


bapmrl:
Continued in https://github.com/prismicio/slice-machine/pull/1235/commits/92a3ddfb444014cab81e5536595b9b8e230b0b1e.

PR #1252, opened 2023-12-20
Test sm update command

hypervillain:
Add Playwright tests for npm and yarn update sm to latest


xrutayisire (approved):
LGTM!

We could reuse part of the same string for the update command that are in common or even for the tests. But as I shared with you and Angelo already, duplication in tests can even be better to reduce overhead, and it's super easy to improve it if necessary in the future depending on our needs.

Super nice to see your first Playwright tests 🎉

PR #1254, opened 2023-12-22
test(playwright): enhance manager procedure mocking

angeloashmore:
## Context

This PR enhance's `@slicemachine/manager` mocking in Playwright tests.

This PR can be used as is or as a base for a refined implementation.

## The Solution

- Add a `procedures` fixture with a `procedures.mock()` method.
- Maintain a list of procedure handlers to allow precise mock orchestration.
- Allow tests to register procedure mocks one at a time.

## Impact / Dependencies

- Existing tests using procedure mocks needed to be updated.
- PRs #1256, #1250, #1253 will need to be updated. There will likely be conflicts.

- [ ] The CI is successful.

🐕


playwright/fixtures/index.ts
```diff
@@ -308,6 +314,13 @@ export const defaultTest = (
       // Propagate the modified page to the test
       await use(page);
     },
+
+    /**
+     * Mocks
+     */
+    procedures: async ({ page }, use) => {
+      await use(await MockManagerProcedures.init(page));
```
xrutayisire, line 322: Here this is broken in typescript with any :/

 I guess it's because of the confusion with class for typescript.
 
 Maybe we should create a classic TS function to handle this? It could remove the problem with the constructor. WDYT?

playwright/utils/MockManagerProcedures.ts
```diff
@@ -0,0 +1,113 @@
...
+  constructor(DO_NOT_USE_CONSTRUCTOR: typeof FORCE_INIT) {
+    if (DO_NOT_USE_CONSTRUCTOR !== FORCE_INIT) {
+      throw new Error(
+        "`new MockManagerProcedures()` is forbidden. Use `await MockManagerProcedures.init(page)` instead.",
+      );
+    }
+  }
+
+  mock(
+    path: string,
```
xrutayisire, line 47: I think I would keep the object, you may not need the handler for example and so positioning argument can be painful, WDYT?

xrutayisire (commented):
Thank you to have started this, so for me I tried this way and first the typescript problem is blocking, and I don't see a beautiful way to fix it :/
So maybe we should convert it to a function with two sub function inside?

I explain:

```ts
// Path: playwright/utils/mockManagerProcedures.ts

async function handleRoute() {}

function mockManagerProcedures() {
  const procedures: Map<string, MockManagerProcedure[]> = new Map();

  init: (page: Page) => Promise<void>;
  mock: ({ path, handler?, config? }) => void;
}

// Path: playwright/fixtures/index.ts

export type DefaultFixtures = {
  mockProcedure: ({ path, handler?, config? }) => void
};

export const defaultTest = () => {
  return baseTest.extend<DefaultFixtures>({
    mockProcedure: async ({ page }, use) => {
      const { init, mock } = mockManagerProcedures();
      await init(page);
      
      await use(mock);
    }
  });
```

WDYT?

playwright/fixtures/index.ts
angeloashmore: It might be because I renamed `mockManagerProcedures.ts` to `MockManagerProcedures.ts` and eslint/tsserver isn't picking it up. I have full typing on my end.

Git properly shows the rename that was done via `git mv`. If it's still giving you trouble, maybe we can just rename the file something else temporarily.

A classic function is possible, but seems a bit messy considering we already have a class that encapsulates the mocking logic. The alternative was:

```ts
const mockManagerProcedures = new MockManagerProcedures()
await mockManagerProcedures.setup(page)
```

The `setup`/`init` methods are necessary because constructors cannot return Promises, thus they cannot use `async`/`await`.

playwright/utils/MockManagerProcedures.ts
angeloashmore: Requiring a handler feels more explicit to me. By defining a `() => undefined` handler, the test tells readers that calling the manager procedure will return undefined. Without a handler, that functionality is implicit.

I typically prefer objects as well since they document argument names and allow for easier argument omission. In this case, I figured the `path` and `handler` arguments are always needed, allowing a more compact API.

The positional arguments also make reading the call more straightforward in my opinion.

```ts
procedures.mock("project.fetchActiveEnvironment", () => ({ environment }));

// Or a more compact version without a function when
// a function isn't needed (not implemented yet):
procedures.mock("project.fetchActiveEnvironment", { environment });
```

Compared to:

```ts
procedures.mock({
  path: "project.fetchActiveEnvironment",
  handler: () => ({ environment }),
});

// Or a more compact version without a function when
// a function isn't needed (not implemented yet):
procedures.mock({
  path: "project.fetchActiveEnvironment",
  // The `handler` name no longer makes sense since it isn't a function.
  handler: { environment },
});
```

"Handler" is more accurate than "data" since the function may throw an error.

I don't mind either option, but I prefer the non-object version. If we feel strongly about making the handler optional, then I agree, an object is better.

playwright/fixtures/index.ts
angeloashmore: Another thing to consider re: using a plain function:

We may want to check the list of mocked procedures. Returning the full `MockProceduresClass` allows us to read the instance's `procedures` property. Providing a single `mockProcedures` fixture wouldn't allow that, unless we returned an object.

playwright/utils/MockManagerProcedures.ts
xrutayisire: You convince me! 🙂 Let's go your way

xrutayisire (approved):
LGTM with the typescript problem 🚀
Super work! Thanks!

PR #1258, opened 2023-12-28
feat(auto-save): Auto-save on custom / page type & slice builder

xrutayisire:
## Context

- Completes DT-1872, DT-1832, DT-1831, DT-1833

## The Solution

Auto-save for page / custom type:
- Remove `selectedCustomTypeReducer` reducer
- Remove `watchSelectedCustomTypeSagas` saga
- Remove old redux sectors 
- Remove old redux actions
- Make the custom type builder working without redux with it's own data locally
- Support Enter key to submit a new field
- Ensure that when renaming a tab, the existing name is already filled
- Create a local context for custom type builder
- Ensure you cannot add a field if the form is already displayed
- Persist save to local custom type builder context
- Create a custom hook `useAutoSave` to ensure we have the proper way to handle multiple changes
- Create a component to display saving status
- Create domain functions to manage custom type model
- Add unit tests for custom type domain functions
- Add e2e tests for custom / page type builder 
- Add option to delay response for e2e mocking procedures management

## TO DO

- Discussions with Benjamin on the design of toaster on error
-  Discussions with Benjamin for the delay on error retry

## Impact / Dependencies

- The local custom type builder will not be updated automatically on window focus or anything else

- [x] The CI is successful.

## Preview

https://github.com/prismicio/slice-machine/assets/19946868/f772090b-424a-4544-8bd5-8655a04d5292


packages/slice-machine/src/features/autoSave/useAutoSave.tsx
```diff
@@ -0,0 +1,114 @@
...
+  retryDelay?: number;
+  retryMessage?: string;
+};
+
+type AutoSaveState = {
+  pendingSave: (() => Promise<unknown>) | null;
+  nextSave: (() => Promise<unknown>) | null;
+};
+
+export type AutoSaveStatusType = "saving" | "saved" | "error";
```
bapmrl: ⚠️ In my opinion, there's no need to append `Type` to a `type`.

💡 Maybe `failed` would be better than `error`? That way, you'd be using past tense like `saved`.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
bapmrl, line 1: 💡 I wonder if a new `core/autoSave` folder would be a better fit. Indeed, I feel like we're putting too many things into `features`. For me, `features` should ideally be "optional" parts/pages/modules and `core` should contain the shared domain bits that allow the `features` to work.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+type UseAutoSaveArgs = {
+  debounceDelay?: number;
+  errorMessage?: string;
+  retryDelay?: number;
+  retryMessage?: string;
+};
+
+type AutoSaveState = {
+  pendingSave: (() => Promise<unknown>) | null;
```
bapmrl: ⚠️ Let's favor `undefined` over `null` for consistency.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+export type AutoSaveStatusType = "saving" | "saved" | "error";
+
+export const useAutoSave = (args: UseAutoSaveArgs = {}) => {
+  const {
+    debounceDelay = 500,
+    errorMessage = "An error happened while saving",
+    retryDelay = 2000,
+    retryMessage = "Retry",
+  } = args;
```
bapmrl, line 42: 💡 A line break here would improve readability.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
```diff
@@ -0,0 +1,114 @@
...
+};
+
+type AutoSaveState = {
+  pendingSave: (() => Promise<unknown>) | null;
+  nextSave: (() => Promise<unknown>) | null;
+};
+
+export type AutoSaveStatusType = "saving" | "saved" | "error";
+
+export const useAutoSave = (args: UseAutoSaveArgs = {}) => {
```
bapmrl: ⚠️ I think it's important to always define the return type of top-level functions in order to help developers quickly understand what the function does.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
```diff
@@ -0,0 +1,114 @@
...
+    debounceDelay = 500,
+    errorMessage = "An error happened while saving",
+    retryDelay = 2000,
+    retryMessage = "Retry",
+  } = args;
+  const [autoSaveState, setAutoSaveState] = useState<AutoSaveState>({
+    pendingSave: null,
+    nextSave: null,
+  });
+  const [autoSaveStatusInternal, setAutoSaveStatusInternal] =
```
bapmrl: ⚠️ Naming is confusing as there are 2 adjacent "states" (`autoSaveState` on line 27 and `autoSaveStatusInternal` on line 31). Maybe `autoSaveStack` (instead of `autoSaveState`) and `autoSaveState` (instead of `autoSaveStatusInternal`) would work better?

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
```diff
@@ -0,0 +1,114 @@
...
+    useState<AutoSaveStatusType>("saved");
```
bapmrl: ⚠️ That `saved` initial state is confusing. `undefined` would probably better convey nothing happened yet and wouldn't make subsequent lines more complicated.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const [autoSaveStatus, setAutoSaveStatus] = useState<AutoSaveStatusType>(
+    autoSaveStatusInternal,
+  );
+
+  const setNextSave = useCallback((saveAction: () => Promise<unknown>) => {
```
bapmrl: 💡 Why not `nextSave` rather than `saveAction`? It's the first time you use the word "action" in that module.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
```diff
@@ -0,0 +1,114 @@
...
+          pendingSave: prevState.nextSave,
+          nextSave: null,
+        }));
+      } catch (error) {
+        setAutoSaveStatusInternal("error");
+        console.error(errorMessage, error);
+
+        const toastId = uniqueId();
+        toast.error(
+          <Box gap={8} alignItems="center">
```
bapmrl: 💡 To improve readability, this toast element could be defined in another function (but still in that module).

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
```diff
@@ -0,0 +1,114 @@
...
+export type AutoSaveStatusType = "saving" | "saved" | "error";
+
+export const useAutoSave = (args: UseAutoSaveArgs = {}) => {
+  const {
+    debounceDelay = 500,
+    errorMessage = "An error happened while saving",
+    retryDelay = 2000,
+    retryMessage = "Retry",
+  } = args;
+  const [autoSaveState, setAutoSaveState] = useState<AutoSaveState>({
```
bapmrl: 💡 I feel like this state could use `useRef` instead as it doesn't seem necessary to re-render the app when it changes.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
```diff
@@ -0,0 +1,114 @@
...
+        }));
+      } catch (error) {
+        setAutoSaveStatusInternal("error");
+        console.error(errorMessage, error);
+
+        const toastId = uniqueId();
+        toast.error(
+          <Box gap={8} alignItems="center">
+            <Text>{errorMessage}</Text>
+            <Button
```
bapmrl: ❓ Does this toast meet the Figma design? It seems very little styling was applied.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
```diff
@@ -0,0 +1,114 @@
...
+    retryDelay = 2000,
+    retryMessage = "Retry",
+  } = args;
+  const [autoSaveState, setAutoSaveState] = useState<AutoSaveState>({
+    pendingSave: null,
+    nextSave: null,
+  });
+  const [autoSaveStatusInternal, setAutoSaveStatusInternal] =
+    useState<AutoSaveStatusType>("saved");
+  const [autoSaveStatus, setAutoSaveStatus] = useState<AutoSaveStatusType>(
```
bapmrl: ⚠️ Naming is confusing. Maybe `debouncedAutoSaveState` and `setDebouncedAutoSaveState` would work better?

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
```diff
@@ -0,0 +1,114 @@
...
+        nextSave: null,
+      }));
+
+      executeSave().catch((error) =>
+        console.error("Error in save effect:", error),
+      );
+    }
+  }, [autoSaveStatusInternal, autoSaveState, executeSave]);
+
+  useEffect(() => {
```
bapmrl, line 97: ❓ If I understand well, this side effect is just a convenience for display purposes and doesn't change the behaviour of the overall `useAutoSave` hook?

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  const executeSave = useCallback(async () => {
+    if (autoSaveState.pendingSave) {
+      setAutoSaveStatusInternal("saving");
+
+      try {
+        await autoSaveState.pendingSave();
+
+        setAutoSaveStatusInternal("saved");
+        setAutoSaveState((prevState) => ({
```
bapmrl: ❓ Is this necessary as it's already done in the `useEffect` below?

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
```diff
@@ -0,0 +1,114 @@
...
+        toast.error(
+          <Box gap={8} alignItems="center">
+            <Text>{errorMessage}</Text>
+            <Button
+              onClick={() => {
+                toast.dismiss(toastId);
+                setAutoSaveStatusInternal("saving");
+                setTimeout(() => {
+                  void executeSave();
+                }, retryDelay);
```
bapmrl: ❓ Why introduce a delay on a user action? It could be confusing if nothing happens for 2 seconds.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
```diff
@@ -0,0 +1,114 @@
...
+        const toastId = uniqueId();
+        toast.error(
+          <Box gap={8} alignItems="center">
+            <Text>{errorMessage}</Text>
+            <Button
+              onClick={() => {
+                toast.dismiss(toastId);
+                setAutoSaveStatusInternal("saving");
+                setTimeout(() => {
+                  void executeSave();
```
bapmrl: ❓ I'm not sure I understand all the execution branches that could happen. For instance, what happens if:
1. There's an error.
2. The user doesn't click on the retry button, but performs another action in the UI.
3. The user clicks on the retry button.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
```diff
@@ -0,0 +1,114 @@
...
+      }
+    }
+  }, [autoSaveState, errorMessage, retryDelay, retryMessage]);
+
+  useEffect(() => {
+    if (
+      autoSaveStatusInternal !== "saving" &&
+      (autoSaveState.pendingSave || autoSaveState.nextSave)
+    ) {
+      setAutoSaveState((prevState) => ({
```
bapmrl: ❓ I'm not sure I understand well this side effect. It seems it will run if `autoSaveStatusInternal` is `error` and the `pendingSave` corresponding to that error would be replaced by `nextSave`. Is it intended? Should the toast be dismissed as `executeSave` will automatically be called?

bapmrl (changes requested):
Nice `useAutoSave` hook 😊 Just have some questions and comments as I don't understand all the intricacies of it.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: I did that because I have a visual component named `AutoSaveStatus` that also refer to the type. So I wanted to be clear about the difference. WDYT?

Good idea for `failed`, I will rename it 🙂

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Oh, I see. I really like it, we kind of discussed a bit that at the beginning I remember. But maybe it will also create a problem with too many first level folder in src with a dev never knowing where to put stuff?
For me, I "feel" it's ok, but I wanted to warn about that.

If we create a `core` folder, we could put:
- autoSave
- documentation
- environments
- inAppGuide
- legacyState
- slicesTemplates

And so in features we leave:
- changes
- cusotmTypes
- labs
- slices

WDYT?

Btw I guess since it's already a +3000/-3000 PR I can do that right after my PRs are merged and only do the autoSave now. So I ease the review for Hugo. Ok for you?

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Yep my bad, I prefer also to use undefined, was too fast here. Good catch, thanks 🙂
I will also add it to the best practices notion page.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: True! 👍

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Yep good catch, I will do it, and also it's already in the best practices 🎉

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Hum not sure to understand the proposed naming too. Mostly because there is a diff between "state" and "status" in the naming

It was a bit hard here for the naming.
So the goal is really differentiate the state that is purely internal with a debounce so autoSave hook to transmit that a save is in progress if the success is superfast.

It means we need 2 state for the status. One internal and one share externally. Maybe I could also try to rewrite it with only one state but 2 properties, WDYT?

Something like this (inlined here):

```ts
  const [autoSaveStack, setAutoSaveStack] = useState<AutoSaveStack>({
    pendingSave: null,
    nextSave: null,
  });
  const [autoSaveStatus, setAutoSaveStatus] = useState<{
    debounced: AutoSaveStatusType;
    immediate: AutoSaveStatusType;
  }>({
    debounced: "saved",
    immediate: "saved", // or "undebounced" or ??
  });
```

With that it's maybe clear what is the purpose of each useState and reduces the overall complexity.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Hum I'm not sure here, by default it's saved, when you open anything, you have the "saved" values displayed. Why not explicitly saying it's `saved` since it is?

Adding undefined will still add complexity for me. At least within the file and maybe even more because it would mean handling undefined everywhere outside of the hook. If as you said `undefined === saved` then saved is for sure simpler for me. WDYT? We can discuss that if necessary.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Agree, I can rename that 👍

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Good idea, I will create a function just below the hook 👍

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Hum, I don't think so.

So `autoSaveState` contain `nextSave`, I'm using `setAutoSaveState` to disatch the `setNextSave` and so I need rerender so the `useEffect` detect a new nextSave and triggers the save.

WDYT?

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: I have no design 😬 
Referring to my message from yesterday in Slack I need to sync with Benjamin to quickly apply the design he wants 🙂

I will add it as a requirement of the PR 👍

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Replied above with my proposition of one state for both.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Yes, exactly, it's just because we don't want to have a status that change too fast and flash in front of the user whatever you do with it.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Something will happen, it will spin. So the goal is to prevent a retry every one ms haha
And also to explain and tell, YES it's retrying. So you click on retry, its telling Saving... and then crash again if necessary. Without this, when there is a permanent fail, it's quite annoying to see that it's really retrying.

You can maybe test it with and without (I can do a small demo)? WDYT?

Btw it's something that Benjamin need to validate also.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Haha, that's something I totally tested (NO), so it's adding a double error if you really have a problem or it's saving the second thing you're doing BUT keep the retry visible.

I need to rework this! THANKS!

Screenshot 2024-01-16 at 12 00 46

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Ok, this one is the one that actually trigger every next save. 

Meaning, when we want to perform a next save we check, is there any pendingSave or nextSave WHILE the status is not "saving" and so doing something. If yes, let's run the nextSave or just remove the pending one.

About the error, yes I think I need to dismiss somewhere to prevent the error above

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Hum good thinking. I will try to see if it's possible and test it. It may simplify a bit the useEffect bellow even maybe.

But I'm not sure. I will try and come back to you 🙂

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
```diff
@@ -0,0 +1,93 @@
...
+  useAutoSave,
+} from "@src/features/autoSave/useAutoSave";
+import { getFormat } from "@src/domain/customType";
+import { updateCustomType } from "@src/apiClient";
+
+import { CUSTOM_TYPES_MESSAGES } from "../customTypesMessages";
+
+type CustomTypeContext = {
+  customType: CustomType;
+  autoSaveStatus: AutoSaveStatusType;
```
bapmrl: ❓ It means for now we have one autosave state for custom types. Is it going to be a different state for slices? How are we going to combine both autosave states to display the error toasts and status messages consistently/once?

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
```diff
@@ -0,0 +1,93 @@
...
+  setCustomType: Dispatch<SetStateAction<CustomType>>;
+};
+
+type CustomTypeProviderProps = {
+  children: ReactNode | ((value: CustomTypeContext) => ReactNode);
+  defaultCustomType: CustomType;
```
bapmrl: 💡 I think `initialCustomType` would be more in line with the parameter name of the `useState` hook.

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { CUSTOM_TYPES_MESSAGES } from "../customTypesMessages";
+
+type CustomTypeContext = {
+  customType: CustomType;
+  autoSaveStatus: AutoSaveStatusType;
+  setCustomType: Dispatch<SetStateAction<CustomType>>;
+};
+
+type CustomTypeProviderProps = {
+  children: ReactNode | ((value: CustomTypeContext) => ReactNode);
```
bapmrl, line 29: ❓ Are we really going to use both the function form and the `useCustomType` hook?

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
```diff
@@ -0,0 +1,93 @@
...
+  defaultCustomType: CustomType;
+};
+
+const CustomTypeContextValue = createContext<CustomTypeContext>({
+  autoSaveStatus: "saved",
+  customType: {} as CustomType,
```
bapmrl: ⚠️ I think it's a weird optimization. If the consumer of that context didn't use the `CustomTypeProvider`, he'll get a `customType` that doesn't conform to the `CustomType` interface. I think it's both safer and clearer to initialize the context with `undefined` and throw if the provider cannot be found.

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
```diff
@@ -0,0 +1,93 @@
...
+  setCustomType: () => void 0,
+});
+
+export function CustomTypeProvider(props: CustomTypeProviderProps) {
+  const { children, defaultCustomType } = props;
```
bapmrl: 💡 Adding a new line after the props spreading would make the code clearer in my opinion.

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const isMounted = useRef(false);
```
bapmrl: 💡 Maybe you could use the [useIsFirstRender](https://github.com/prismicio/editor/blob/e4c8075c20146e885be3a4c8df0e0fe6522d7e5f/packages/editor-support/src/React.ts#L66-L75) hook from `@prismicio/editor-support`?

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
```diff
@@ -0,0 +1,93 @@
...
+          }
+
+          // Update available custom types store with new custom type
+          saveCustomTypeSuccess(customType);
+        });
+      } else {
+        isMounted.current = true;
+      }
+    },
+    // Prevent saveCustomTypeSuccess to trigger infinite loop
```
bapmrl: 💡 Maybe "from triggering an infinite loop" would be better?

bapmrl (changes requested):
Again, excellent work Xavier 😊

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
bapmrl: Okay, what do you think about naming the component `AutoSaveStatusIndicator` then? 😊

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
bapmrl: Okay for me, but I'm not sure about the partitioning you propose. We'll discuss it again ^^

Also, what would you think of moving all the "old" code into a separate top-level `old` directory? 😜

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
xrutayisire: For me, I was planning to have 2 providers. As the provider, don't have the save logic since it's the hook and only contain things purely related to custom types or slices.

There are some duplications, but I would say healthy one that help us maintain the project without risking crashing everything.

You can see what I'm planning to do here: https://github.com/prismicio/slice-machine/blob/xru/auto-save-slice-builder/packages/slice-machine/src/features/slices/sliceBuilder/SliceBuilderProvider.tsx
And you can see that there are difference related to slices.

So for me 2 providers that reuse the hook yes.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
bapmrl: Nice idea! Maybe `actual` would be clearer than `immediate`?

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
xrutayisire: Yes good point! I think naming is my nemesis! 😈

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
bapmrl: You're right. Let's keep `saved` as the initial state.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
bapmrl: Okay, I see. Let's keep it that way then ^^

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
xrutayisire: Yes! Because as soon as we instance it we want to use the automatically updated custom type and autoSave status. And for me having to create a sub component is just more heavy.

See: https://github.com/prismicio/slice-machine/blob/xru/auto-save-type-builder/packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypesBuilderPage.tsx#L76

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
bapmrl: Okay, I understand. Thanks!

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
xrutayisire: Yeah, I was never a fan of how it's done in React. I would just love to have a provider where I initialize with undefined but every time I'm getting the context typescript knows that it's full. And React handle the error if I didn't setup the Prodiver.

xrutayisire: Will do it 👍

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
xrutayisire: I will add it, maybe it's something I will look for now. I tend to group it like that.

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
xrutayisire: 100%, I didn't even know about it. It's something we (expect you haha) are not doing. More looking for simple stuff in editor packages. I will use it!!

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
xrutayisire: Agree 👍

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: That's a good idea, yes!

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
bapmrl: Okay, I just wonder how the 2 `autoSaveStatus` will be combined to provide a clear visual indicator to the users.

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
bapmrl: Okay, great!

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Let's go with `actual` thanks

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: What old code you refer to? I'm not a fan of this, then you start to have again more confusion and folders. The old folders and old code. I feel It's just more pain to maintain.

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
xrutayisire: Hum, they will never be combined, since it's display from within the builder you never have both in the same view.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
```diff
@@ -0,0 +1,114 @@
...
+        pendingSave: prevState.nextSave,
+        nextSave: null,
+      }));
+
+      executeSave().catch((error) =>
+        console.error("Error in save effect:", error),
+      );
+    }
+  }, [autoSaveStatusInternal, autoSaveState, executeSave]);
```
hypervillain: I believe every property of `autoSaveState` should be listed in here

xrutayisire: Hey, it's not necessary here since it's immutable, we are talking about a state so each time the values changed the state itself 100% change. You never modify a state sub value directly.

The linter is pretty good on that and tell us here that it's good.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Done

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Following our discussion, will remove 2 folders after my PRs

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Done

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Done

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Done

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Done

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Done

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Designed updated with Benjamin

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Fixed, I don't trigger the next save if they is an error. The retry will do it only. But any changes done after a fail will still but stacked.

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
xrutayisire: Done

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
xrutayisire: Done

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
xrutayisire: Done

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
xrutayisire: Done

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: So I did it differently at the end. With the new requirement from Benjamin that want the opposite of a status that only show if it takes longer than x ms. The goal is to have a status that always update to display the loading state but will take 300ms longer to remove that it's saved or failed. So that we can always smoothly see that something happened.

With that said, at the end it was not better to mix both state as we really want to have two lifecycles. So I have one status delayed and one actual. I think it's clearer now.

I let you check 🙂

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Updated with my proposition above

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: Yes! 

So I fixed the bug of the retry, and I really took time to check that one.

When a save fail, we want to trigger a retry with the latest data we have (saw with Benjamin), meaning we don't want to just reply the pending. We may want to retry with the next, but if there is no next, meaning the user did not do any action in the meantime, we will retry with the pending.
Because of this it's important in case of a success to remove the pending, if not the useEffect will detect that there is a pending operation and will try to do it again, creating an endless loop.

I added comments for the tricky parts, btw.

packages/slice-machine/src/features/autoSave/useAutoSave.tsx
xrutayisire: useEffect updated with the fix on the error and it should be clearer now. With a comment explaining the code.

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
xrutayisire: Done

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
```diff
@@ -0,0 +1,96 @@
...
+};
+
+const CustomTypeContextValue = createContext<CustomTypeContext | undefined>(
+  undefined,
+);
+
+export function CustomTypeProvider(props: CustomTypeProviderProps) {
+  const { children, initialCustomType } = props;
+
+  const isFirstRender = useIsFirstRender();
```
hypervillain: 👍

packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
```diff
@@ -0,0 +1,96 @@
...
+          }
+
+          // Update available custom types store with new custom type
+          saveCustomTypeSuccess(customType);
+        });
+      }
+    },
+    // Prevent saveCustomTypeSuccess from triggering an infinite loop
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [customType, setNextSave],
```
hypervillain: Is it better to remove it from the deps or the wrap it with a useCallback?

xrutayisire: I could try to use `useStableCallback` from the editor support, good idea!

xrutayisire: It's working with `useStableCallback`:

Screenshot 2024-01-17 at 14 55 57

I did it in the second PR with both providers to freeze this one, since the review is done now.

PR #1261, opened 2024-01-10
chore: add new workflow to publish-stable

michelchau:
Completes DT-1773.

### DIFF
--- a/.github/workflows/publish-stable.yml
+++ b/.github/workflows/publish-stable.yml
@@ -1,0 +1,106 @@
+name: Publish stable
+
+on:
+    workflow_dispatch:
+        inputs:
+          slicemachine/adapter-next:
+            description: slicemachine/adapter-next
+            default: 'None'
+            type: choice
+            options:
+              - patch
+              - minor
+              - major
+              - "None"
+          slicemachine/adapter-nuxt:
+            description: slicemachine/adapter-nuxt
+            default: 'None'
+            type: choice
+            options:
+              - patch
+              - minor
+              - major
+              - "None"
+          slicemachine/adapter-nuxt2:
+            description: slicemachine/adapter-nuxt2
+            default: 'None'
+            type: choice
+            options:
+              - patch
+              - minor
+              - major
+              - "None"
+          slicemachine/init:
+            description: slicemachine/init
+            default: 'None'
+            type: choice
+            options:
+              - patch
+              - minor
+              - major
+              - "None"
+          slicemachine/manager:
+            description: slicemachine/manager
+            default: 'None'
+            type: choice
+            options:
+              - patch
+              - minor
+              - major
+              - "None"
+          slicemachine/plugin-kit:
+            description: slicemachine/plugin-kit
+            default: 'None'
+            type: choice
+            options:
+              - patch
+              - minor
+              - major
+              - "None"
+          slice-machine-ui:
+            description: slice-machine-ui
+            default: 'None'
+            type: choice
+            options:
+              - patch
+              - minor
+              - major
+              - "None"
+          start-slicemachine:
+            description: start-slicemachine
+            default: 'None'
+            type: choice
+            options:
+              - patch
+              - minor
+              - major
+              - "None"
+jobs:
+  publish-stable:
+    name: Publish stable
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          ref: ${{ github.head_ref || github.ref }}
+      - name: Use Node.js
+        uses: ./.github/actions/setup-node
+      - name: Install dependencies
+        run: yarn install
+      - name: Build essentials
+        uses: ./.github/actions/build-essential
+      - name: Run publish script
+        env:
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+          YARN_NPM_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: |
+          package=$(echo '${{ toJSON(github.event.inputs) }}' | jq -c)
+          keys=$(echo $package | jq -r 'keys | .[]')
+          package_output=""
+          for key in $keys; do
+            value=$(echo $package | jq -r --arg k "$key" '.[$k]')
+            package_output+="@$key:$value "
+          done
+          yarn publish stable --dry-run $package_output
--- a/.github/workflows/publish-stable.yml
+++ b/.github/workflows/publish-stable.yml
@@ -76,7 +76,7 @@ on:
           - decline
 jobs:
   publish-stable:
-    if: 'startsWith(github.ref, ''refs/heads/main'')'
+    if: ${{ startsWith(github.ref, 'refs/heads/main')}}
     name: Publish stable
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-stable.yml
```diff
@@ -0,0 +1,106 @@
+name: Publish stable
+
+on:
+    workflow_dispatch:
+        inputs:
+          slicemachine/adapter-next:
```
bapmrl: 💡 Would it be possible to have a direct match with the workspaces names? In this case, it would be `@slicemachine/adapter-next`. That way, the input names would use the `@` scopes consistently.

.github/workflows/publish-stable.yml
```diff
@@ -0,0 +1,106 @@
...
+            description: slicemachine/adapter-next
+            default: 'None'
```
bapmrl: 💡 I've thought about it and I think `decline` would be a better name than `None` as it's also used by [Yarn](https://yarnpkg.com/cli/version).

.github/workflows/publish-stable.yml
```diff
@@ -0,0 +1,106 @@
...
+            type: choice
+            options:
```
bapmrl: 💡 Would it be possible to have the options appear in the following order: `major`, `minor`, `patch` and finally `decline`? It would be more intuitive as SemVer follows the same `<major>.<minor>.<patch>` order.

.github/workflows/publish-stable.yml
```diff
@@ -0,0 +1,106 @@
...
+          slicemachine/adapter-nuxt2:
+            description: slicemachine/adapter-nuxt2
+            default: 'None'
+            type: choice
+            options:
+              - patch
+              - minor
+              - major
+              - "None"
+          slicemachine/init:
```
bapmrl: ⚠️ You forgot `@slicemachine/adapter-sveltekit` between `@slicemachine/adapter-nuxt2` and `@slicemachine/init`.

.github/workflows/publish-stable.yml
```diff
@@ -0,0 +1,106 @@
+name: Publish stable
+
+on:
+    workflow_dispatch:
+        inputs:
+          slicemachine/adapter-next:
+            description: slicemachine/adapter-next
```
bapmrl: ❓ Does the `description` correspond to the name of the input displayed in GitHub's popover?

.github/workflows/publish-stable.yml
```diff
@@ -0,0 +1,106 @@
...
+              - major
+              - "None"
+jobs:
+  publish-stable:
+    name: Publish stable
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
+        with:
```
bapmrl: 💡 I'm not sure we need the `fetch-depth` and `ref` options, at least probably not the `ref` option as this workflow will only run on the `main` branch.

.github/workflows/publish-stable.yml
bapmrl, line 1: 💡 Is there a GitHub option to only allow this workflow to run on the `main` branch?

⚠️ What about security? For now, we don't want contributors outside Prismic to be able to publish releases via this workflow.

⚠️ Could you please format the file with Prettier? If you don't have it configured in your editor, then the [online playground](https://prettier.io/playground) should give a similar result.

.github/workflows/publish-stable.yml
```diff
@@ -0,0 +1,106 @@
...
+      - name: Install dependencies
+        run: yarn install
+      - name: Build essentials
+        uses: ./.github/actions/build-essential
+      - name: Run publish script
+        env:
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+          YARN_NPM_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: |
+          package=$(echo '${{ toJSON(github.event.inputs) }}' | jq -c)
```
bapmrl: 💡 If I understand well those lines, you could use the following names instead to be consistent with the `publish` script:
- `package` → `inputs`.
- `keys` → `package_names`.
- `package_output` → `specifiers`.
- `value` → `package_release_type`.

.github/workflows/publish-stable.yml
```diff
@@ -0,0 +1,106 @@
...
+          keys=$(echo $package | jq -r 'keys | .[]')
+          package_output=""
+          for key in $keys; do
+            value=$(echo $package | jq -r --arg k "$key" '.[$k]')
+            package_output+="@$key:$value "
```
bapmrl: ⚠️ I'm not sure it's the case, but I think you are including the `decline`d values when you should not ^^ Indeed, the `script` expects only the packages to be published. The other packages should be filtered out.

bapmrl (changes requested):
Excellent!

I've added some comments though. Tell me if you prefer to address them asynchronously or if you'd like us to call 😊

.github/workflows/publish-stable.yml
michelchau: We can't use @ on the inputs variable.

.github/workflows/publish-stable.yml
michelchau: I added some condition for main branch only.

.github/workflows/publish-stable.yml
```diff
@@ -0,0 +1,108 @@
...
+        default: decline
+        type: choice
+        options:
+          - major
+          - minor
+          - patch
+          - decline
+jobs:
+  publish-stable:
+    if: 'startsWith(github.ref, ''refs/heads/main'')'
```
bapmrl: ❓ Why use `'` twice? Also, I've noticed in the `publish-unstable` workflow we used `${{ }}` around the whole expression (instead of a single `'`). Which option is better?

.github/workflows/publish-stable.yml
```diff
@@ -0,0 +1,108 @@
...
+        env:
+          SENTRY_AUTH_TOKEN: '${{ secrets.SENTRY_AUTH_TOKEN }}'
+          YARN_NPM_AUTH_TOKEN: '${{ secrets.NPM_TOKEN }}'
+        run: |
+          inputs=$(echo '${{ toJSON(github.event.inputs) }}' | jq -c)
+          package_names=$(echo $inputs | jq -r 'keys | .[]')
+          specifiers=""
+          for package in $package_names; do
+            package_release_type=$(echo $inputs | jq -r --arg k "$package" '.[$k]')
+            if
```
bapmrl: ❓Is it normal to have 2 `if`s in a row?

.github/workflows/publish-stable.yml
```diff
@@ -0,0 +1,108 @@
...
+          SENTRY_AUTH_TOKEN: '${{ secrets.SENTRY_AUTH_TOKEN }}'
+          YARN_NPM_AUTH_TOKEN: '${{ secrets.NPM_TOKEN }}'
+        run: |
+          inputs=$(echo '${{ toJSON(github.event.inputs) }}' | jq -c)
+          package_names=$(echo $inputs | jq -r 'keys | .[]')
+          specifiers=""
+          for package in $package_names; do
+            package_release_type=$(echo $inputs | jq -r --arg k "$package" '.[$k]')
+            if 
+            if ["$package_name == slicemachine/*" && "package_release_type" != "decline"]; then
```
bapmrl: ❓ I can't find a `package_name` variable. Should the `package` variable in the `for` loop be renamed to `package_name`?

.github/workflows/publish-stable.yml
```diff
@@ -0,0 +1,108 @@
...
+        uses: actions/checkout@v3
+      - name: Use Node.js
+        uses: ./.github/actions/setup-node
+      - name: Install dependencies
+        run: yarn install
+      - name: Build essentials
+        uses: ./.github/actions/build-essential
+      - name: Run publish script
+        env:
+          SENTRY_AUTH_TOKEN: '${{ secrets.SENTRY_AUTH_TOKEN }}'
```
bapmrl: ❓ Why add extra single quotes `'` around the `${{`? We didn't do that in the `publish-unstable` workflow.

bapmrl (commented):
Excellent job! I've just added 4 small questions and we'll be good.

PR #1264, opened 2024-01-11
chore: remove the dry-run flag from the publish-stable workflow

bapmrl:
The workflow passed on `main` with the `--dry-run` flag, so it should work fine next time we want to release.


xrutayisire (approved):
LGTM, super nice to have that now!

Thank you so much @michelchau and @bapmrl

PR #1265, opened 2024-01-17
chore: introduce CSS Modules

bapmrl:
Completes DT-1892.


package.json
```diff
@@ -65,6 +66,10 @@
     "**/packages/slice-machine/**/*.@(js|jsx|ts|tsx|)": [
       "prettier --write --ignore-unknown",
       "yarn workspace slice-machine-ui lint:precommit"
+    ],
+    "**/packages/slice-machine/**/*.module.css": [
+      "prettier --write --ignore-unknown",
+      "yarn workspace slice-machine-ui stylelint:precommit"
```
angeloashmore, line 72: I think there was a discussion a while ago about removing pre-commit hooks considering we do these checks in CI. Is that still planned? If yes, I wonder if we should not add new pre-commit hooks.

(I prefer no pre-commit hooks since it slows the commit process down, but I don't have a full understanding of the history.)

packages/slice-machine/package.json
```diff
@@ -12,18 +12,22 @@
     "node": ">=14"
   },
   "scripts": {
-    "build": "cross-env NODE_OPTIONS=--max_old_space_size=8192 next build && next export",
-    "dev": "next dev",
-    "dev-cypress": "concurrently \"next dev\" \"npm run dev --prefix ../../e2e-projects/cypress-next-app\"",
+    "build": "yarn typed-css-modules && cross-env NODE_OPTIONS=--max_old_space_size=8192 next build && next export",
+    "dev": "concurrently \"yarn typed-css-modules:watch\" \"next dev\"",
+    "dev-cypress": "concurrently \"yarn typed-css-modules:watch\" \"next dev\" \"npm run dev --prefix ../../e2e-projects/cypress-next-app\"",
```
angeloashmore: We can add names and colors to the `concurrently` output using `--names` and `--prefix-colors`:

```
    "dev": "concurrently \"yarn typed-css-modules:watch\" \"next dev\" --names \"typed-css-modules,next\" --prefix-colors auto",
    "dev-cypress": "concurrently \"yarn typed-css-modules:watch\" \"next dev\" \"npm run dev --prefix ../../e2e-projects/cypress-next-app\" --names \"typed-css-modules,next,cypress-next-app\" --prefix-colors auto",
```

The names and colors make it easier to understand where a particular line comes from:

image

packages/slice-machine/stylelint.config.js
```diff
@@ -0,0 +1,17 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { cssTheme } = require("@prismicio/editor-ui");
+
+module.exports = {
+  defaultSeverity: "warning",
+  reportNeedlessDisables: true,
+  reportInvalidScopeDisables: true,
+  reportDescriptionlessDisables: true,
+  rules: {
+    "declaration-property-value-allowed-list": cssTheme,
```
angeloashmore, line 10: Does this line ensure we follow the theme tokens? Like colors, spacing, sizing, etc.?

packages/slice-machine/package.json
```diff
@@ -123,8 +127,13 @@
     "remark-gfm": "3.0.1",
     "semver": "7.3.8",
     "storybook": "7.0.10",
+    "stylelint": "15.11.0",
+    "stylelint-config-css-modules": "4.3.0",
+    "stylelint-config-prettier": "9.0.5",
+    "stylelint-config-recommended": "13.0.0",
```
angeloashmore: Is it possible to use the latest versions of these packages?

packages/slice-machine/package.json
```diff
@@ -123,8 +127,13 @@
...
     "swr": "1.3.0",
     "theme-ui": "0.15.5",
+    "typed-css-modules": "0.8.0",
```
angeloashmore: Same as above: is it possible to use the latest version of this package?

angeloashmore (approved):
Mostly everything looks good to me! I just had a few questions and suggestions, but nothing blocking. I'm looking forward to the simpler styling system. 🙂

package.json
bapmrl: I think most team members, if not all, are okay to remove all our pre-commit hooks. However, I don't think it was listed somewhere so I've just created a Linear issue in our backlog: DT-1906. In the meantime, I think consistency has value and as TypeScript files are linted before committing, I think it's okay to do the same for CSS Modules.

packages/slice-machine/package.json
bapmrl: I've made the change. Thank you!

packages/slice-machine/stylelint.config.js
bapmrl: Yes, exactly 🙂 It works fine, but I miss the TypeScript auto completion.

packages/slice-machine/package.json
bapmrl: Yes, I've made the change and `stylelint-config-prettier` is no longer necessary as Stylelint has deprecated their formatting rules 🙂

packages/slice-machine/package.json
bapmrl: Yes, I've made the change.

PR #1268, opened 2024-01-18
feat(manager): add `GitManager`

angeloashmore:
## Context

DT-1891

## The Solution

Add `GitManager` (accessible via `manager.git`) with the following methods:

- `createGitHubAuthState()`
- `fetchOwners()`
- `fetchRepos()`
- `fetchLinkedRepos()`
- `linkRepo()`
- `unlinkRepo()`
- `checkHasWriteAPIToken()`
- `updateWriteAPIToken()`
- `deleteWriteAPIToken()`

The tests introduced in this PR cover ~80% of the code, which is sufficient for a first release.

## Impact / Dependencies

This PR allows us to implement the Git integration UI.

This PR also introduces a drastically simpler method for writing manager tests (inspired by @xrutayisire's Playwright setup). Only the new `GitManager` tests use the new method, but we can migrate existing tests to the new method over time.

- [ ] The CI is successful.


packages/manager/src/managers/git/GitManager.ts
```diff
@@ -0,0 +1,296 @@
...
+import fetch from "../../lib/fetch";
+import { decode } from "../../lib/decode";
+
+import { API_ENDPOINTS } from "../../constants/API_ENDPOINTS";
+
+import { UnauthorizedError, UnexpectedDataError } from "../../errors";
+
+import { BaseManager } from "../BaseManager";
+
+import { GitRepo, GitRepoSpcifier, Namespace } from "./types";
```
bapmrl: ⚠️ Typo: `GitRepoSpecifier`.

packages/manager/src/managers/git/GitManager.ts
```diff
@@ -0,0 +1,296 @@
...
+
+type CreateGitHubAuthStateReturnType = {
```
bapmrl: ❓ Why isn't this prefixed with `GitManager`?

packages/manager/src/managers/git/types.ts
```diff
@@ -0,0 +1,18 @@
+export type Namespace = {
+	id: string;
+	name: string;
+	type: "user" | "team" | null;
```
bapmrl, line 8: ❓ Should it be `undefined` instead of `null`? I personally avoid using `null` whenever I can to avoid mixed-uses.

packages/manager/src/managers/git/GitManager.ts
```diff
@@ -0,0 +1,296 @@
...
+				default:
+					throw new Error("Failed to ulink repos.");
+			}
+		}
+	}
+
+	async #fetch(
+		url: URL,
+		config?: {
+			method?: "POST" | "PUT" | "DELETE";
```
bapmrl, line 442: 💡 It seems we never use `POST` so maybe it could be omitted?

packages/manager/src/managers/git/GitManager.ts
```diff
@@ -0,0 +1,296 @@
...
+			body?: unknown;
+		},
+	) {
+		const authenticationToken = await this.user.getAuthenticationToken();
+
+		return await fetch(url, {
+			method: config?.method,
+			body: config?.body ? JSON.stringify(config?.body) : undefined,
```
bapmrl: 💡 No need for `?.` in `config?.body`.

packages/manager/src/managers/git/GitManager.ts
```diff
@@ -0,0 +1,296 @@
...
+			API_ENDPOINTS.SliceMachineV1,
+		);
+		const res = await this.#fetch(url);
+
+		if (!res.ok) {
+			switch (res.status) {
+				case 401:
+					throw new UnauthorizedError();
+				default:
+					throw new Error("Failed to create GutHub auth state.");
```
bapmrl: ⚠️ Typo: `GitHub`.

packages/manager/src/managers/git/GitManager.ts
```diff
@@ -0,0 +1,296 @@
...
+			}
+		}
+
+		const json = await res.json();
+		const { value, error } = decode(
+			t.type({
```
bapmrl, line 119: 💡 Would it be better to derive/infer the type from the codec, instead of duplicating it inline?

packages/manager/src/managers/git/GitManager.ts
```diff
@@ -0,0 +1,296 @@
...
+		return value;
+	}
+
+	async fetchOwners(): Promise<GitManagerFetchOwnersReturnType> {
+		const url = new URL("./git/owners", API_ENDPOINTS.SliceMachineV1);
+		const res = await this.#fetch(url);
+
+		if (!res.ok) {
+			switch (res.status) {
+				case 403:
```
bapmrl, line 144: ⚠️ It seems `sm-api` can return a `401` on that route. See: https://github.com/prismicio/sm-api/pull/6/files#diff-748068ec62c6d7a24e4ef4afd2b0c01154623b97cb43fa4e6b1bc60424ce5f20R47. Maybe it would be worth going through the other routes as well?

packages/manager/src/managers/git/GitManager.ts
```diff
@@ -0,0 +1,296 @@
...
+	async fetchRepos(
+		args: GitManagerFetchReposArgs,
+	): Promise<GitManagerFetchReposReturnType> {
+		const url = new URL("./git/repos", API_ENDPOINTS.SliceMachineV1);
+		url.searchParams.set("provider", args.provider);
+		url.searchParams.set("owner", args.owner);
+		if (args.query) {
+			url.searchParams.set("q", args.query);
+		}
+		if (args.page) {
```
bapmrl: ⚠️ I think this expression will be `false` when `page` equals `0`. We don't want that, do we? That's why I don't like expressions that don't explicitely have a `boolean` type.

packages/manager/src/managers/git/GitManager.ts
```diff
@@ -0,0 +1,296 @@
...
+				},
+			},
+		});
+
+		if (!res.ok) {
+			switch (res.status) {
+				case 403:
+					throw new UnauthorizedError();
+				default:
+					throw new Error("Failed to ulink repos.");
```
bapmrl: ⚠️ Typo: `unlink`.

packages/manager/test/__testutils__/createAPIFixture.ts
```diff
@@ -0,0 +1,101 @@
...
+export type SerializableValueArray =
+	| SerializableValue[]
+	| readonly SerializableValue[];
+export type SerializableValuePrimitive =
+	| string
+	| number
+	| boolean
+	| Date
+	| null;
+export type SerializableValue =
```
bapmrl, line 21: ⚠️ `SerializableValue` doesn't appear to be used anywhere else. I don't think it's a good idea to artificially increase the API surface area of a module.

packages/manager/test/__testutils__/createAPIFixture.ts
```diff
@@ -0,0 +1,101 @@
+import { SetupServer } from "msw/lib/node";
+import { rest } from "msw";
+
+import { SliceMachineManager } from "../../src";
+
+type MockOptions = {
+	method?: keyof typeof rest;
+	statusCode?: number;
+	checkAuthentication?: boolean;
+	searchParams?: Record<string, string>;
```
bapmrl, line 10: 💡 Would `Partial<Record<string, string>>` be better as not all keys have a corresponding value?

packages/manager/test/__testutils__/createAPIFixture.ts
```diff
@@ -0,0 +1,101 @@
...
+
+	for (const key in apiEndpoints) {
+		api[`mock${key}` as keyof typeof api] = (path, response, options) => {
+			const apiEndpoint = apiEndpoints[key as keyof typeof apiEndpoints];
+			const method = options?.method ?? "get";
+
+			const handler = rest[method](
+				new URL(path, apiEndpoint).toString(),
+				async (req, res, ctx) => {
+					// TODO: Enable by default after fixing "Error [ERR_STREAM_PREMATURE_CLOSE]: Premature close" error.
```
bapmrl: ❓ Should a matching Linear issue be created to not lose track of that TODO?

packages/manager/test/__setup__.ts
```diff
@@ -130,14 +137,61 @@ vi.mock("module", async () => {
 	} as typeof actual;
 });
 
+vi.mock("@segment/analytics-node", async () => {
+	const { Analytics }: typeof import("@segment/analytics-node") =
+		await vi.importActual("@segment/analytics-node");
+
+	Analytics.prototype.track = vi.fn(
```
bapmrl: ❓ I'm not sure, but isn't this kind of globally mutating the `prototype` of the actual implementation? Could it be an issue?

bapmrl (changes requested):
Excellent work! I really like the new `createAPIFixture` function.

packages/manager/src/managers/git/GitManager.ts
angeloashmore: It should be. 👍

packages/manager/src/managers/git/types.ts
angeloashmore: In this case, `null` is treated as "unknown," which can happen if a Git provider returns a user type that we don't support. It shouldn't happen in theory, but it is technically possible.

Would you prefer a different special value, like "_unknown"? Or a comment?

packages/manager/src/managers/git/GitManager.ts
angeloashmore: It was mistakenly left out of `createGitHubAuthState()`'s `fetch()` call. We have a method that needs it now.

packages/manager/src/managers/git/GitManager.ts
angeloashmore: If you mean using the codec to produce the return type (or vice versa), I prefer keeping them separate. It allows us to change the return type without affecting the codec; the codec is purely about decoding the network request's response, not necessarily the method's return type.

WDYT? Do you see a better way to structure/think about it?

packages/manager/src/managers/git/GitManager.ts
angeloashmore: Thanks, good catch. Each method handles `401` now by throwing an `UnauthenticatedError`.

packages/manager/src/managers/git/GitManager.ts
angeloashmore: I think we want that behavior since `page=0` is not valid. However, maybe we need to be stricter, such as only supporting positive numbers.

I changed the condition to:

```ts
if (args.page && args.page > 0) { ... }
```

If you want to enforce boolean checking, do you think we should add that to the ESLint config?

packages/manager/test/__testutils__/createAPIFixture.ts
angeloashmore: I prefer exporting any top-level types that a function uses since those types may want to be referenced by a consumer. For example, `SerializableValue` may be used to declare a constant to be passed to the function call:

```ts
const fixture = createAPIFixture({ manager, mswServer });

const response = { foo: "bar" } satisfies SerializableValue;
fixture.mockSliceMachineV1("./git/owners", response);
```

The above example is contrived, but this practice can be useful in more specific cases.

When a type is not exported, consumers are forced to extract a type from another type, sometimes via `typeof` or without the ability to use generics:

```ts
const fixture = createAPIFixture({ manager, mswServer });

const response = { foo: "bar" } satisfies Parameters<typeof fixture['mockSliceMachineV1']>[1];
fixture.mockSliceMachineV1("./git/owners", response);
```

That being said, there is no need to export all of `SerializableValue`'s child types, like `SerializableValuePrimitive`. I removed `export` from those types.

packages/manager/test/__testutils__/createAPIFixture.ts
angeloashmore: In this case, no, we want consumers to always define a value for a search param. `searchParams` verifies that a given set of search params are included in the request. There shouldn't be a reason to include a parameter without a value.

Note that `undefined` is an invalid `URLSearchParams` param value; `""` is used instead:

```ts
api.mockSliceMachineV1(
	"./git/github/create-auth-state",
	{ key, expiresAt: expiresAt.toISOString() },
	{
		searchParams: {
			// Checks for ?foo=bar
			foo: "bar",
			// Checks for ?baz
			baz: "",
		},
	},
);
```

packages/manager/test/__testutils__/createAPIFixture.ts
angeloashmore: Done! For reference, here is the issue: DT-1919

packages/manager/test/__setup__.ts
angeloashmore: Good point. Yes it probably mutates the actual implementation, which we don't want. I changed the mocks to use `vi.spyOn` + `mockImplementation`:

```ts
const actual: typeof import("@segment/analytics-node") =
	await vi.importActual("@segment/analytics-node");

vi.spyOn(actual.Analytics.prototype, "track").mockImplementation(
	(_params, callback) => {
		callback?.();
	},
);

// Repeat for other mocked methods.

return actual;
```

packages/manager/src/managers/git/types.ts
bapmrl: Oh, I see. I think `null` is fine then, but maybe a comment wouldn't hurt to explain why it can be missing?

packages/manager/src/managers/git/GitManager.ts
bapmrl: Hmmm, okay I see and understand the rationale. I think it's fine like that ^^

packages/manager/src/managers/git/GitManager.ts
bapmrl: > I changed the condition to:
> 
> ```ts
> if (args.page && args.page > 0) { ... }
> ```
Thank you! The condition is clearer now.
> If you want to enforce boolean checking, do you think we should add that to the ESLint config?

I've just tried, but ESLint returns the following error:
```
Error: Error while loading rule '@typescript-eslint/strict-boolean-expressions': You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.
```
It seems we need to enable [Linting with Type Information](https://typescript-eslint.io/linting/typed-linting) and I don't think it's the right time to do that ^^

packages/manager/test/__testutils__/createAPIFixture.ts
bapmrl: Okay, thank you! It's great to read your answers, I'm learning a lot.

packages/manager/test/__testutils__/createAPIFixture.ts
bapmrl: You're right! Thank you.

bapmrl (approved):
I'm approving 😊 There's just a small conversation remaining. You can add a comment for the `null` value if you also think it's useful.

packages/manager/src/managers/git/types.ts
angeloashmore: Added a comment. 👍

PR #1270, opened 2024-01-22
refactor(changes): Remove pushChanges saga and store

xrutayisire:
## Context

- Completes DT-1907

## The Solution

- Remove pushChanges saga
- Remove pushChanges store 
- Remove push changes loading from within redux
- Remove soft and hard limit drawers modal from within redux
- Create a pushChanges action replacing the saga
- Export Limit type from manager instead of recreating it in Slice Machine UI
- Remove unused ReferencesErrorDrawer
- Fix react warning about missing key in changes table

## Impact / Dependencies

- No ReferencesErrorDrawer anymore
- No sortDocumentLimits anymore

- [x] The CI is successful.


packages/slice-machine/components/DeleteDocumentsDrawer/index.tsx
```diff
@@ -1,3 +1,2 @@
 export { SoftDeleteDocumentsDrawer } from "./SoftDeleteDocumentsDrawer";
 export { HardDeleteDocumentsDrawer } from "./HardDeleteDocumentsDrawer";
-export { ReferencesErrorDrawer } from "./ReferencesErrorDrawer";
```
xrutayisire, line 3: This ReferencesErrorDrawer was not used at all anymore and since a long time (10 months at least), correct?

packages/slice-machine/src/modules/pushChangesSaga/index.ts
```diff
@@ -1,191 +0,0 @@
...
-  >
->;
-
-export const changesPushCreator = createAsyncAction(
-  "PUSH_CHANGES.REQUEST",
-  "PUSH_CHANGES.RESPONSE",
-  "PUSH_CHANGES.FAILURE",
-)<ChangesPushSagaPayload, undefined, Limit | InvalidCustomTypeResponse>();
-
-export const sortDocumentLimits = (limit: Readonly<Limit>) => ({
```
xrutayisire, line 53: I don't see the need for this function anymore. 

It was used on this line:
```ts
yield put(changesPushCreator.failure(sortDocumentLimits(response)));
```

But I don't see the usage of that, do you have an idea? I may have missed something important here

packages/slice-machine/src/modules/pushChangesSaga/index.ts
```diff
@@ -1,191 +0,0 @@
...
-
-    // Display success toaster
-    yield put(
-      openToasterCreator({
-        content: "All slices and types have been pushed",
-        type: ToasterType.SUCCESS,
-      }),
-    );
-  } catch (error) {
-    if (isUnauthenticatedError(error) || isUnauthorizedError(error)) {
```
xrutayisire, line 143: I removed this support in the new code. 

First, it was not working, and even if you were unauthorized it was displaying the classic error. And also because you cannot see the push button without being logged in anyway. WDYT?

packages/slice-machine/components/DeleteDocumentsDrawer/index.tsx
hypervillain: Correct! It seems you left some related code inside `cypress/pages/changes/drawers`

xrutayisire: Yes, I can remove it, but I don't touch Cypress at all anymore usually as everything is skipped, it will be removed when doing the global removal of Cypress. That I'm planning to do super soon btw. WDYT?

packages/manager/src/index.ts
```diff
@@ -13,7 +13,7 @@ export { createSliceMachineManager } from "./managers/createSliceMachineManager"
 export { createSliceMachineManagerMiddleware } from "./managers/createSliceMachineManagerMiddleware";
 export type { CreateSliceMachineManagerMiddlewareArgs } from "./managers/createSliceMachineManagerMiddleware";
 
-export type { Environment } from "./managers/prismicRepository/types";
+export type { Environment, Limit } from "./managers/prismicRepository/types";
```
angeloashmore: Could we use a more descriptive name than `Limit`, such as `PushChangesLimit`?

I know this type already existed before this PR, but now that it is part of the public API, its name should be self-describing.

packages/slice-machine/components/DeleteDocumentsDrawer/HardDeleteDocumentsDrawer.tsx
```diff
@@ -57,14 +45,14 @@ export const HardDeleteDocumentsDrawer: React.FunctionComponent<{
 
   return (
     <SliceMachineDrawerUI
-      isOpen={isDeleteDocumentsDrawerOpen}
+      isOpen={modalData.type === "HARD"}
```
angeloashmore, line 48: Should we be using an enum rather than the string value?

packages/manager/src/index.ts
xrutayisire: Thanks I agree!

xrutayisire: Done

packages/slice-machine/components/DeleteDocumentsDrawer/HardDeleteDocumentsDrawer.tsx
xrutayisire: Hum, at the end typescript check it for us. I tried to use the enum from manager, but it seems I cannot reference an enum like this directly. I'm not sure creating it again in SM is better than the type checking we have already. WDYT?

packages/slice-machine/components/DeleteDocumentsDrawer/index.tsx
hypervillain: That works for me

packages/slice-machine/components/DeleteDocumentsDrawer/HardDeleteDocumentsDrawer.tsx
hypervillain: If TS can infer it, a string value is perfectly fine IMO

angeloashmore: We lose the ability to rename things quickly via the language server by using a string literal. The difference between an enum and a string literal is DX and code-as-documentation. If we were to use the enum, we would make it a top-level export from the manager.

In practice, there's no difference and we don't need to make any changes here. Just figured I'd raise the question. 🙂

xrutayisire: I totally agree that if it's free, let's go for it. But when I exported it globally, I ended up with a failed build in Slice Machine UI. I didn't have time to check really why. Maybe you have an idea? I'm not sure we did that before.

PR #1273, opened 2024-01-24
feat(experiment): Setup Amplitude experiment & Add first dev collab experiment

xrutayisire:
## Context

- Completes DT-1916

## The Solution

- Setup Amplitude Experiment with `@amplitude/experiment-js-client`
- Add two new function to telemetry manager
  - `initExperiment`
  - `experimentVariant`
- Do not allow experiment if telemetry is deactivated
- Create a new hook to query any experiment in the future
- Create a new component for the dev collab experiment
- Display it only if experiment is on
-  Use data from the variant payload

## Impact / Dependencies

- None

## Preview

Screenshot 2024-01-24 at 12 58 12

Screenshot 2024-01-24 at 12 58 20


packages/manager/src/managers/telemetry/TelemetryManager.ts
```diff
@@ -268,4 +277,44 @@ export class TelemetryManager extends BaseManager {
...
+			});
+
+			await this._experiment
+				.start({
+					user_properties: {
+						Repository: repositoryName,
+					},
+				})
+				.catch((error) => {
+					console.error("Error starting experiment", error);
```
hypervillain: why do you log this error and not the one below in prod?

hypervillain (commented):
Is there any way for us to know that an Experiment is over and that we can safely remove DevCollaborationExperiment?

packages/manager/src/managers/telemetry/TelemetryManager.ts
xrutayisire: Good one! I think I just don't need a catch, better to just use the try catch for both, thanks

xrutayisire:
> Is there any way for us to know that an Experiment is over and that we can safely remove DevCollaborationExperiment?

@hypervillain  Yes, Mabel or Côme tell us it's over 😬
We can also check it ourselves in Amplitude Experiment, but it's always best to ask before the removal.

For me, it should also be, as soon as they archive an experiment, they tell us and create a ticket to remove the dead code.

hypervillain (approved):

PR #1276, opened 2024-01-26
feat(slice-machine): add OwnerSelect in FieldSet's stories

bapmrl:


packages/slice-machine/src/components/FieldSet/FieldSet.stories.tsx
```diff
@@ -206,3 +216,27 @@ export const WithBoxContent = {
   },
   parameters: { controls: { hideNoControlsWarning: true, include: [] } },
 };
+
+type OrganizationSelectProps = { disabled?: boolean };
+
+function OrganizationSelect(props: OrganizationSelectProps): ReactNode {
```
angeloashmore: "Organization" is a GitHub-specific term and is only one kind of owner type (the other is "User"). The API uses the term "Owner" to represent an organization or user. Thus, we could replace references to "Organization" with "Owner". What do you think?

packages/slice-machine/src/components/FieldSet/FieldSet.stories.tsx
```diff
@@ -206,3 +216,27 @@ export const WithBoxContent = {
...
+  return (
+    <Select
+      {...props}
+      color="grey"
+      constrainContentWidth
+      flexContent
+      placeholder="Organization"
+      renderStartIcon={() => <GitHubIcon color={tokens.color.greyLight11} />}
```
angeloashmore, line 230: The icon appears further away from the text than it should. I saw you opened a thread here: [Open Slack](https://prismic-team.slack.com/archives/C0441S8SKQR/p1706279773276939)

Will we eventually change `<Select>` to use a smaller gap?

packages/slice-machine/src/components/FieldSet/FieldSet.stories.tsx
```diff
@@ -206,3 +216,27 @@ export const WithBoxContent = {
...
+      size="large"
+      sx={{ width: "calc(50% - 8px)" }}
+    >
+      <SelectItem size="large" value="organization-1">
+        Organization
```
angeloashmore: We may want to add icons to the select items to represent the Git provider. In this case, we would probably want a GitHub logo to appear next to each item. In the future, we would show different provider icons, like the GitLab and Bitbucket logos.

angeloashmore (approved):
Looks great! I really like how composable the FIeldSet components are. 🙂

I left few comments/suggestions.

packages/slice-machine/src/components/FieldSet/FieldSet.stories.tsx
bapmrl: I agree! I'll make the change.

packages/slice-machine/src/components/FieldSet/FieldSet.stories.tsx
bapmrl: I hope soon, but the conversation involves several team members with different priorities 😄

packages/slice-machine/src/components/FieldSet/FieldSet.stories.tsx
bapmrl: Shouldn't we display a "user" or "organization" icon instead? ^^ For instance, if the user just selected GitHub in the previous step, I would expect the `Select` to only contain items that are either GitHub users or GitHub organizations, no?

bapmrl:
I'm merging @angeloashmore, but we can continue the conversation about the icons to display next to each `SelectItem` 🙂

packages/slice-machine/src/components/FieldSet/FieldSet.stories.tsx
angeloashmore: We should probably support multiple connected accounts at once, like the Vercel "New Project" screen works:

image

They force switching Git providers, but I think we can support multiple providers at once. Thus, we would need to a way to differentiate providers.

I don't think owner type is as important as provider type since a user may add two accounts with the same name from different providers. The different between a user and team owner has less of an impact, but we could still display a "team" label next to team accounts.

bapmrl: I've updated the code to display the provider's icon in each item ^^

PR #1279, opened 2024-01-30
fix: ESM error when using SST

angeloashmore:
## Context

This PR fixes an issue when `@slicemachine/manager` is used in an SST application.

## The Solution

Move `prettier` into a dynamic import. Prettier declares `__dirname` in its code, which conflicts with SST's `__dirname` definition.

## Impact / Dependencies

N/A

- [ ] The CI is successful.

### DIFF
--- a/packages/manager/src/lib/format.ts
+++ b/packages/manager/src/lib/format.ts
@@ -1,4 +1,4 @@
-import prettier from "prettier";
+import type prettier from "prettier";
 import { stripIndent } from "common-tags";
 
 type FormatOptions = {
@@ -16,6 +16,8 @@ export const format = async (
 	filePath: string,
 	options?: FormatOptions,
 ): Promise<string> => {
+	const prettier = await import("prettier");
+
 	let formatted = stripIndent(source);
 
 	const prettierOptions = await prettier.resolveConfig(filePath);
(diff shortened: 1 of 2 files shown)


packages/manager/src/lib/format.ts
```diff
@@ -16,6 +16,8 @@ export const format = async (
 	filePath: string,
 	options?: FormatOptions,
 ): Promise<string> => {
+	const prettier = await import("prettier");
```
bapmrl, line 19: ❓ Can dynamically loading the `prettier` module impact the performance of the `format` function? I'm thinking of what could happen with a large amount of consecutive calls.

bapmrl (approved):

packages/manager/src/lib/format.ts
angeloashmore: Dynamic imports shouldn't significantly impact performance unless the module contains a lot of code. Prettier is large, so it's possible it could cause slow-downs, but I don't think it would be significant enough to worry about it.

Typically only the first load takes longer. I think Node.js's import cache speeds up subsequent imports.

angeloashmore:
It turns out this change didn't work. Instead, we need to include `prettier` in `node_modules` in the `sm-api` SST app.

PR #1280, opened 2024-01-30
refactor(slices): Remove deleteSliceSaga

xrutayisire:
## Context

- Completes DT-1751

## The Solution

- Remove deleteSliceSaga
- Rework slices reducer
- Rework slices success action in slice machine actions
- Update DeleteSliceModal
- Remove unnecessary old unit tests

## Impact / Dependencies

- None


hypervillain (approved):
Very pleasing. Do you think it would be interesting to explain the "success" actions of `useSliceMachineActions` if someone else refactors redux-saga or it should be self explanatory?

xrutayisire:
Thank you! For me the message you added back then in SliceMachineActions is perfect to understand the concept. I feel reading this give you enough knowledge of the problem we solve. WDYT?

PR #1281, opened 2024-01-31
refactor(slices): Remove renameSliceSaga

xrutayisire:
## Context

- Completes DT-1750

## The Solution

- Remove renameSliceSaga
- Remove watchSliceSagas
- Add a new rename domain function
- Add a new feature action renameSlice

## Impact / Dependencies

- None


packages/slice-machine/components/Forms/RenameSliceModal/RenameSliceModal.tsx
```diff
@@ -1,50 +1,65 @@
...
 
-  const handleOnSubmit = (values: SliceModalValues) => {
-    renameSlice(libName, sliceId, values.sliceName);
+  const handleOnSubmit = async (values: SliceModalValues) => {
+    if (slice) {
+      await renameSlice({
+        slice,
+        newSliceName: values.sliceName,
+        onSuccess: (renamedSlice) => {
+          renameSliceSuccess(
```
hypervillain: Is there a reason for not passing `renamedSlice` directly? Eg. we don't want `renameSliceSuccess` to how these properties are stored?

packages/slice-machine/components/Forms/RenameSliceModal/RenameSliceModal.tsx
```diff
@@ -1,50 +1,65 @@
...
-      formId={`rename-slice-modal-${sliceId}`}
-      close={closeModals}
+      formId={`rename-slice-modal-${slice?.model.id ?? ""}`}
+      close={onClose}
       buttonLabel="Rename"
-      onSubmit={handleOnSubmit}
+      onSubmit={(values) => void handleOnSubmit(values)}
       initialValues={{
-        sliceName: sliceName,
+        sliceName: slice?.model.name ?? "",
```
hypervillain: hmm, is it possible to check for slice higher up in the component? This would make this line and `formId` easier to read

packages/slice-machine/components/Forms/RenameSliceModal/RenameSliceModal.tsx
xrutayisire: You're right! sliceId so `renamedSlice.model.id` is not even used as I using model.id in the reducer. So I can remove second paramater. But why I dont just pass renamedSlice is because we are working with SliceSM and not ComponentUI for the reducer and so I need the libName (first param)

packages/slice-machine/components/Forms/RenameSliceModal/RenameSliceModal.tsx
xrutayisire: Yes good idea, will do

hypervillain (approved):

PR #1282, opened 2024-02-01
Set dev env as active env in start-slicemachine

hypervillain:


packages/start-slicemachine/src/StartSliceMachineProcess.ts
```diff
@@ -136,6 +139,18 @@ export class StartSliceMachineProcess {
...
 	}
 
+	private async _setDevEnvAsActiveEnvironment(): Promise<void> {
+		const { environments } =
+			await this._sliceMachineManager.prismicRepository.fetchEnvironments();
+
+		const maybeDevEnvironment = environments?.find((env) => env.kind === "dev");
+		if (maybeDevEnvironment !== undefined) {
+			await this._sliceMachineManager.project.updateEnvironment({
+				environment: maybeDevEnvironment?.domain,
```
xrutayisire: Since there is a condition above, no need for `?`, right?

packages/start-slicemachine/src/StartSliceMachineProcess.ts
```diff
@@ -124,6 +124,9 @@ export class StartSliceMachineProcess {
 				userID: profile.shortId,
 				intercomHash: profile.intercomHash,
 			});
+			try {
+				await this._setDevEnvAsActiveEnvironment();
+			} catch (_) {}
```
xrutayisire: What do you think about displaying a console.error only on DEV mode for us, so we have a warning in DEV mode this fail one day. I guess it should not throw an error if you just don't have a dev env, it will throw if there is a real problem, no?

PR #1286, opened 2024-02-06
Fix telemetry check which prevents init from running

hypervillain:
We realised `checkIsTelemetryEnabled` sometimes broke, which triggered a catch that would try to track the error on Sentry, making the init crash because the check for telemetry happens before its instantiation. I could not reproduce so I safeguarded every line

### DIFF
--- a/packages/init/src/SliceMachineInitProcess.ts
+++ b/packages/init/src/SliceMachineInitProcess.ts
@@ -133,10 +133,7 @@ export class SliceMachineInitProcess {
 					)}\n`,
 				);
 			}
-			await this.manager.telemetry.initTelemetry({
-				appName: pkg.name,
-				appVersion: pkg.version,
-			});
+
 			try {
 				await setupSentry();
 			} catch (error) {
@@ -144,6 +141,12 @@ export class SliceMachineInitProcess {
 				// because of failed tracking set up. We probably couldn't determine the
 				// Sentry environment.
 			}
+
+			await this.manager.telemetry.initTelemetry({
+				appName: pkg.name,
+				appVersion: pkg.version,
+			});
+
 			await this.manager.telemetry.track({
 				event: "command:init:start",
 				repository: this.options.repository,
--- a/packages/init/src/SliceMachineInitProcess.ts
+++ b/packages/init/src/SliceMachineInitProcess.ts
@@ -112,6 +112,13 @@ export class SliceMachineInitProcess {
 		);
 
 		try {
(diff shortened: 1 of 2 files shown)


packages/init/src/SliceMachineInitProcess.ts
```diff
@@ -133,17 +133,20 @@ export class SliceMachineInitProcess {
 					)}\n`,
 				);
 			}
-			await this.manager.telemetry.initTelemetry({
-				appName: pkg.name,
-				appVersion: pkg.version,
-			});
+
 			try {
 				await setupSentry();
```
xrutayisire: Do you think we could put Sentry even higher up? 

I would go the highest place in this function, WDYT?

packages/manager/src/managers/telemetry/TelemetryManager.ts
```diff
@@ -90,7 +90,7 @@ export class TelemetryManager extends BaseManager {
 		};
 
 		if (isTelemetryEnabled) {
-			await this.initExperiment();
+			this.initExperiment();
```
xrutayisire, line 93: Thanks for that 😅

packages/manager/src/managers/telemetry/TelemetryManager.ts
```diff
@@ -272,15 +272,17 @@ export class TelemetryManager extends BaseManager {
...
 		try {
-			root = await this.project.getRoot();
+			root = await this.project
+				.getRoot()
+				.catch(() => this.project.suggestRoot());
+
+			return readPrismicrc(root).telemetry !== false;
 		} catch {
-			root = await this.project.suggestRoot();
+			return false;
```
xrutayisire: Dump question, maybe we could default to true? If we didn't find the `prismicrc` with telemetry `false`, then it's true by default. WDYT?

xrutayisire (changes requested):
Super cool to have this easy solution! 🙌

packages/manager/src/managers/telemetry/TelemetryManager.ts
hypervillain: No idea I'm not very well versed in GDPR stuff :) but why not!

xrutayisire: I would say it's already what we are doing if we don't see a prismicrc file anyway.

xrutayisire (approved):
LGTM!

PR #1287, opened 2024-02-07
feat(slice-machine-ui): enable multiple paragraphs in rich text fields by default

angeloashmore:
## Context

DT-1841

## The Solution

Change the default field to use the `multi` property in place of the `single` property.

## Impact / Dependencies

All rich text fields will have "Allow multiple paragraphs" checked by default.

- [x] The CI is successful.


xrutayisire (approved):
LGTM 🙂

PR #1291, opened 2024-02-08
chore: support GitHub staging app

angeloashmore:
## Context

Completes DT-1963

## The Solution

- We have access to a new app that works exclusively with the staging SM API

   **Prismic.io (Staging)**: https://github.com/apps/prismic-io-staging

- Add `GitManager.prototype.getProviderAppInstallURL()`. The method returns an app installation URL based on the provider, Slice Machine mode (prod/staging), and user.

   This encapsulation allows us to hide provider-specific details, like needing to generate a state value for GitHub auth, from Slice Machine UI.

## Impact / Dependencies

We can now develop Git integrations against the staging evnrionment.

- [x] The CI is successful.


packages/slice-machine/src/features/settings/git/GitProvider.ts
```diff
@@ -5,16 +5,12 @@ import { managerClient } from "@src/managerClient";
...
 export const gitProviderToConfig = {
   gitHub: {
     connect: async () => {
-      const state = await managerClient.git.createGitHubAuthState();
-      const url = new URL(
-        `https://github.com/apps/${GITHUB_APP_SLUG}/installations/new`,
-      );
-      url.searchParams.set("state", state.key);
+      const url = await managerClient.git.getProviderAppInstallURL({
+        provider: "gitHub",
```
bapmrl: 💡 Are the 3 `connect` functions going to be same, except for that `provider` argument? If so, we could maybe replace them with a single `installGitProviderApp` function:
```
export async function installGitProviderApp(provider: GitProvider): Promise<void> {
  const url = await managerClient.git.getProviderAppInstallURL({ provider });
  window.open(url, "git-provider-app-installation");
}
```

Note that I don't know what the `target` parameter (of the `window.open` function) is used for ^^

packages/manager/src/managers/git/constants.ts
bapmrl, line 1: ❓ Why declare these `const` here and not in `packages/manager/src/constants`?

packages/slice-machine/src/features/settings/git/GitProvider.ts
```diff
@@ -5,16 +5,12 @@ import { managerClient } from "@src/managerClient";
 
 export type GitProvider = keyof typeof gitProviderToConfig;
```
bapmrl: 💡 Maybe we could import the `GitProvider` type of `@slicemachine/manager` and augment it with `bitbucket | gitLab`?

packages/slice-machine/src/features/settings/git/GitProvider.ts
```diff
@@ -5,16 +5,12 @@ import { managerClient } from "@src/managerClient";
...
 
-const GITHUB_APP_SLUG = "prismic-io";
-
 export const gitProviderToConfig = {
```
bapmrl, line 8: 💡 Maybe the `supported` property could be turned into a `isGitProviderSupported` function that would simply have to search the `GitProvider` record of `@slicemachine/manager`?

packages/manager/src/managers/git/GitManager.ts
```diff
@@ -436,6 +442,36 @@ export class GitManager extends BaseManager {
...
 	}
 
+	async getProviderAppInstallURL(args: {
+		provider: GitProvider;
+	}): Promise<string> {
+		const mode = this.getMode();
+
+		switch (args.provider) {
+			case GitProvider.GitHub: {
+				const appSlug =
```
bapmrl: 💡 Adding a new mode would result in `appSlug` defaulting to staging. Would a record be easier to maintain?

packages/manager/src/managers/SliceMachineManager.ts
```diff
@@ -172,6 +175,20 @@ export class SliceMachineManager {
 		return API_ENDPOINTS;
 	}
 
+	getMode(): (typeof APPLICATION_MODE)[keyof typeof APPLICATION_MODE] {
+		const mode = process.env.SM_ENV || APPLICATION_MODE.Production;
```
bapmrl: ❓ The knowledge of parsing `process.env.SM_ENV` (and defaulting to `APPLICATION_MODE.Production`) is kind of duplicated in `packages/manager/src/constants/API_ENDPOINTS.ts` and `packages/manager/src/constants/API_TOKENS.ts`. Should these constants be turned into functions, that would take as a parameter a parsed application mode?

bapmrl (approved):
Nice change 🙂

packages/slice-machine/src/features/settings/git/GitProvider.ts
angeloashmore: Yes, they should be the same. How does this look to you?

https://github.com/prismicio/slice-machine/blob/4eba1b48beefce863269543783340d539d29d119/packages/slice-machine/src/features/settings/git/GitProvider.ts

The `target` parameter defines how a URL is opened: [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#target).

In our case, we set the target to a static string so the same window/tab is used each time a "Connect" button is clicked. That should be a rare occurrence, but it prevents having more than one active connect flow at a single time.

packages/manager/src/managers/git/constants.ts
angeloashmore: I figured they were constants specific to `GitManager`. We probably should be consistent by placing them in the top-level `constants` directory instead. I'll move them. 👍

packages/slice-machine/src/features/settings/git/GitProvider.ts
angeloashmore: Yes, good idea. I changed the import.

packages/slice-machine/src/features/settings/git/GitProvider.ts
angeloashmore: Something like this?

https://github.com/prismicio/slice-machine/blob/4eba1b48beefce863269543783340d539d29d119/packages/slice-machine/src/features/settings/git/GitProvider.ts

packages/manager/src/managers/git/GitManager.ts
angeloashmore: I changed it to work like `API_ENDPOINTS`, where we can set values based on `SM_ENV` and other environment variables as needed.

https://github.com/prismicio/slice-machine/blob/4eba1b48beefce863269543783340d539d29d119/packages/manager/src/constants/GIT_PROVIDER_APP_SLUGS.ts

Does that work for you?

packages/manager/src/managers/SliceMachineManager.ts
angeloashmore: I removed `getMode()` and replaced it with something that works more like the existing API_ENDPOINTS constant:

https://github.com/prismicio/slice-machine/blob/4eba1b48beefce863269543783340d539d29d119/packages/manager/src/constants/GIT_PROVIDER_APP_SLUGS.ts

I agree that `API_ENDPOINTS` and the new `GIT_PROVIDER_APP_SLUGS` could be a function rather than a constant. Let's save that for a different PR (unless you feel strongly about it).

packages/slice-machine/src/features/settings/git/GitProvider.ts
bapmrl: Excellent change and thanks for the explanation 🙂

packages/slice-machine/src/features/settings/git/GitProvider.ts
bapmrl: Excellent! Thanks 😊

packages/manager/src/managers/SliceMachineManager.ts
bapmrl: Nope, it can wait 🙂

bapmrl (approved):
You made really nice changes! Thank you 🙂

packages/manager/src/managers/git/GitManager.ts
bapmrl: Yes, it's simpler 👍

PR #1294, opened 2024-02-15
fix(slice-machine): reset ErrorBoundary after auth

bapmrl:
Completes DT-1952.


packages/slice-machine/src/ErrorBoundary.tsx
```diff
@@ -0,0 +1,43 @@
...
+          if (errorBoundary === null || errorRef.current === undefined) return;
+          errorRef.current = undefined;
+          errorBoundary.reset();
+        },
+        /* eslint-disable-next-line react-hooks/exhaustive-deps --
+         * Whenever `authStatus` changes, we want to `reset` the `errorBoundary`
+         * if an authentication or authorization error has been caught.
+         **/
+        [authStatus],
+      )}
```
angeloashmore: This code is a bit hard to understand. Is the trick that `useAuthStatus()` triggers a re-render when auth state changes, causing `ref` to be called again?

What do you think of this implementation, which I think is more idiomatic:

```tsx
// packages/slice-machine/src/ErrorBoundary.tsx

import { ErrorBoundary as EditorUiErrorBoundary } from "@prismicio/editor-ui";
import { type ComponentProps, type FC, useRef } from "react";

import { useAuthStatus } from "@src/hooks/useAuthStatus";

// TODO(DT-1979): Export the `ErrorBoundaryProps` type from `@prismicio/editor-ui`.
type ErrorBoundaryProps = ComponentProps<typeof EditorUiErrorBoundary>;

export const ErrorBoundary: FC<ErrorBoundaryProps> = (props) => {
  const errorBoundaryRef = useRef<EditorUiErrorBoundary>(null);

  useAuthStatus({
    onChange: () => errorBoundaryRef.current?.reset(),
  });

  return <EditorUiErrorBoundary {...props} ref={errorBoundaryRef} />;
};
```

```ts
// packages/slice-machine/src/hooks/useAuthStatus.ts

import { useEffect, useRef } from "react";
import { useSelector } from "react-redux";

import { getAuthStatus } from "@src/modules/environment";
import { AuthStatus } from "@src/modules/userContext/types";

type UseAuthStatusArgs = {
  onChange?: (status: AuthStatus) => void;
};

export function useAuthStatus({ onChange }: UseAuthStatusArgs = {}) {
  const { authStatus } = useSelector((store: SliceMachineStoreType) => ({
    authStatus: getAuthStatus(store),
  }));

  // A ref prevents calling `onChange` when its reference changes.
  const onChangeRef = useRef(onChange);
  onChangeRef.current = onChange;

  useEffect(() => {
    onChangeRef.current?.(authStatus);
  }, [authStatus]);

  return authStatus;
}
```

Note the addition of the `onChange` argument to `useAuthState()`. It uses a `useEffect()`, which I know we try to avoid, but I think this is a proper use case for it.

Disclaimer: I didn't test this code.

angeloashmore (commented):
I left a refactoring suggestion below, but everything else looks good! 👍

packages/slice-machine/src/ErrorBoundary.tsx
bapmrl: Hmmm, that's an interesting comment ^^

>This code is a bit hard to understand.

I understand. That's not something we're used to doing, but it's been in the [official documentation](https://legacy.reactjs.org/docs/refs-and-the-dom.html#callback-refs) for a long time.

>Is the trick that useAuthStatus() triggers a re-render when auth state changes, causing ref to be called again?

Yes, that's the trick. You can find an explanation [here](https://tkdodo.eu/blog/avoiding-use-effect-with-callback-refs#callback-refs). Basically, `ref` will be called on each render, but not if `authStatus` hasn't changed.

>What do you think of this implementation, which I think is more idiomatic:

It's interesting, but I don't think it's better than my initial proposal because:
1. The concern of resetting the `ErrorBoundary` is now spread across 2 different modules.
2. `useAuthStatus` now has the additional responsibility of alerting the outside world when it changes (with the `onChange` callback). Yet, Redux already does that by re-rendering the components that use it via `useSelector`.
3. `useAuthStatus` now needs to have an `useStableCallback` (what you did with `useRef(onChange)`) to avoid triggering `onChange` too often.
4. There's no guarantee that `errorBoundaryRef` will be set when `onChange` is called. It might, but that depends on `ErrorBoundary`'s implementation.

What do you think @angeloashmore? Are you convinced?

bapmrl:
I've just pushed [this change](https://github.com/prismicio/slice-machine/compare/0a5b0b4e48dd7c345a188120cbcc33bd662d8389..721837ddace22a1536ee201135c6d7d9cf6347bd) to simplify a bit the implementation 🙂

packages/slice-machine/src/ErrorBoundary.tsx
angeloashmore: Hm, it still feels like a clever use of `ref`, in the non-obvious way. It requires detailed knowledge of React.

We are trying to react upon a change in `authState`, not the re-render of ErrorBoundary, which is why I feel attaching the reset logic to `authState` makes more sense. The linked example where a DOM element's height is saved as a side-effect seems more fitting since it is reacting to the re-render of that `ref`ed element.

You are right that `onChange` + `useEffect` feels a bit inelegant. Event handlers aren't always a good option.

Here's a modified version that contains all logic within ErrorBoundary. The previous proposed changes to `useAuthStatus` can be ignored.

```tsx
import { ErrorBoundary as EditorUiErrorBoundary } from "@prismicio/editor-ui";
import { type ComponentProps, type FC, useRef } from "react";

import { useAuthStatus } from "@src/hooks/useAuthStatus";

// TODO(DT-1979): Export the `ErrorBoundaryProps` type from `@prismicio/editor-ui`.
type ErrorBoundaryProps = ComponentProps<typeof EditorUiErrorBoundary>;

export const ErrorBoundary: FC<ErrorBoundaryProps> = (props) => {
  const errorBoundary = useRef<EditorUiErrorBoundary>(null);
  const authStatus = useAuthStatus();

  const previousAuthStatus = useRef(authStatus);
  if (previousAuthStatus.current !== authStatus) {
    errorBoundary.current?.reset();
  }
  previousAuthStatus.current = authStatus;

  return <EditorUiErrorBoundary {...props} ref={errorBoundary} />;
};
```

This implementation ensures the Suspense cache is cleared on any auth status change, not just ones picked up by an ErrorBoundary, and more specifically, ones that correctly throw `UnauthorizedError` or `UnauthenticatedError`. It changes the original intent of this PR slightly, but it might be more correct. Please feel free to challenge that assumption if you think it's incorrect.

I'll leave it at that and let you decide. Both implementations should get the job done. 🙂

Should we stick with the current ErrorBoundary implementation, is it possible to retain the ability to use a custom `onError` and `ref`? It could be as simple as calling `props.onError` and `props.ref` within the already present callbacks.

angeloashmore:
@bapmrl Followed up here: https://github.com/prismicio/slice-machine/pull/1294#discussion_r1492890858

packages/slice-machine/src/ErrorBoundary.tsx
bapmrl: Hello @angeloashmore,

>Here's a modified version that contains all logic within ErrorBoundary. The previous proposed changes to useAuthStatus can be ignored.

I've already tried something like what you propose, but I was forced to add a `setTimeout` to avoid the following React error: ``Warning: Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state``. I think the `setTimeout` makes it even less satisfactory than the `useCallback` version I propose 😅

Here's the implementation I tried:
```typescript
export const ErrorBoundary: FC<ErrorBoundaryProps> = (props) => {
  const errorRef = useRef<unknown>();
  const errorBoundaryRef = useRef<EditorUiErrorBoundary>(null);

  const authStatus = useAuthStatus();
  const prevAuthStatus = useRef(authStatus);
  if (authStatus !== prevAuthStatus.current) {
    const error = errorRef.current;
    if (isUnauthenticatedError(error) || isUnauthorizedError(error)) {
      setTimeout(() => {
        errorRef.current = undefined;
        errorBoundaryRef.current?.reset();
      });
    }
  }
  prevAuthStatus.current = authStatus;

  return (
    <EditorUiErrorBoundary
      {...props}
      onError={(error) => {
        errorRef.current = error;
      }}
      ref={errorBoundaryRef}
    />
  );
};
```

Otherwise, the following implementation, based on `useEffect`, works but is really similar to the `useCallback` version:
```typescript
export const ErrorBoundary: FC<ErrorBoundaryProps> = (props) => {
  const errorRef = useRef<unknown>();
  const errorBoundaryRef = useRef<EditorUiErrorBoundary>(null);

  const authStatus = useAuthStatus();
  useEffect(() => {
    const error = errorRef.current;
    if (isUnauthenticatedError(error) || isUnauthorizedError(error)) {
      errorRef.current = undefined;
      errorBoundaryRef.current?.reset();
    }
  }, [authStatus]);

  return (
    <EditorUiErrorBoundary
      {...props}
      onError={(error) => {
        errorRef.current = error;
      }}
      ref={errorBoundaryRef}
    />
  );
};
```

>This implementation ensures the Suspense cache is cleared on any auth status change, not just ones picked up by an ErrorBoundary, and more specifically, ones that correctly throw UnauthorizedError or UnauthenticatedError. It changes the original intent of this PR slightly, but it might be more correct. Please feel free to challenge that assumption if you think it's incorrect.

I don't think the `ErrorBoundary` component should have the responsibility of handling successful `authStatus` state transitions. I think this pull request should be seen as a temporary workaround to address the specific case of automatically resetting failed error boundaries. Later, we should delete the current implementation of `authStatus` and imagine a more robust and simpler authentication system. It will then inform us on how to implement a better `ErrorBoundary` component.

>I'll leave it at that and let you decide. Both implementations should get the job done. 🙂

I'll merge this pull request and, as I've said in the previous paragraph, we can revisit it later when the authentication system will be a bit more sane.

>Should we stick with the current ErrorBoundary implementation, is it possible to retain the ability to use a custom onError and ref? It could be as simple as calling props.onError and props.ref within the already present callbacks.

I don't think we should add more code without a clear use case. It can be added later if needed.

PR #1296, opened 2024-02-16
refactor(changelog): Remove getChangelogSaga

xrutayisire:
## Context

- Completes DT-1744, DT-1794

## The Solution

- Remove `getChangelogSaga`
- Remove all Redux involvement in fetching all data relative to changelog page or new versions
- Remove old changelog loading key
- Remove unused `updatesViewed` object in store as today we always display the update box
- Export necessary types from manager to use in SM UI
- Decide to NOT touch the UI part (theme-ui, React components modeling, ...) to focus on the 2 tickets above
- Fetching of the RN is now done only when you click to see one
- Rework `useAdapterName` as we should throw if we cannot get it and so we should always have a value
- Remove unnecessary tests
- Remove unnecessary screenshot loading key
- Restore missing loading state for screenshots upload

## Impact / Dependencies

- Fix the Sentry problem we have with tons of errors because of the RNs fetching


hypervillain (approved):
Looks good!

PR #1298, opened 2024-02-19
feat(slice-machine): disable push changes with connected Git repo

bapmrl:
Completes DT-1901.

### DIFF
--- a/packages/slice-machine/src/features/changes/PushChangesButton.tsx
+++ b/packages/slice-machine/src/features/changes/PushChangesButton.tsx
@@ -1,0 +1,83 @@
+import { Button, Skeleton, Tooltip } from "@prismicio/editor-ui";
+import {
+  type ComponentPropsWithoutRef,
+  type ElementRef,
+  type FC,
+  type ReactNode,
+  Suspense,
+  forwardRef,
+} from "react";
+
+import { gitProviderToConfig } from "@src/features/settings/git/GitProvider";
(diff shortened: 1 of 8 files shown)


packages/slice-machine/src/features/changes/PushChangesButton.tsx
```diff
@@ -0,0 +1,79 @@
...
+  );
+};
+
+const PushChangesButtonUsingLinkedGitRepos: FC<PushChangesButtonProps> = (
+  props,
+) => {
+  const { linkedGitRepos } = useLinkedGitRepos();
+  return linkedGitRepos.length > 0 ? (
+    <Tooltip
+      content="This repository has been set up to use the Git workflow."
```
angeloashmore: Is there room to add an extra bit of information?

> This project uses the Git workflow. Update Prismic by pushing your changes to ${provider}.

`${provider}` would be "GitHub," "GitLab," etc.

angeloashmore (changes requested):
It works well! I added a wording suggestion, which I will let you and @benjammartin decide.

It looks like E2E tests are broken as of this change, however. Do you have an idea on how to fix it?

packages/slice-machine/src/features/changes/PushChangesButton.tsx
bapmrl: I've made the change @angeloashmore 😊

angeloashmore (approved):
Looks good!

bapmrl:
Thanks @angeloashmore!

PR #1302, opened 2024-02-21
Revert "Set dev env as active env in start-slicemachine (#1282)"

xrutayisire:
This reverts commit 3a9b25386a1b17c058f6874cad2b98beaba5f19c.

## Context

After some discussions with a user, we saw that this management brings more confusion than something else. We wanted to push the sandbox to be used more but it looks like a bug when you are using Slice Machine.


hypervillain:
That's exactly the feeling I had while testing something else. I got switched to dev env and thought the feature was broken

hypervillain (approved):

PR #1304, opened 2024-02-21
feat(tracking): Add tracking on slice machine start

xrutayisire:
## Context

- Completes DT-1983

## To Do

- Test for new manager function to detect git provider

## The Solution

- Added a new manager function to detect git provider 
- Added a new event type for slice machine start
- Added all the properties that can be useful for us in order to see how our users use our product
- Fixed a bug that was adding `_includeEnvironmentKind` to the event payload

## Impact / Dependencies

- None, the start process is not longueur as it's done after

### DIFF
--- a/packages/manager/src/managers/git/GitManager.ts
+++ b/packages/manager/src/managers/git/GitManager.ts
@@ -1,5 +1,6 @@
 import * as t from "io-ts";
 import * as tt from "io-ts-types";
+import { execaCommandSync } from "execa";
 
 import fetch from "../../lib/fetch";
 import { decode } from "../../lib/decode";
@@ -461,6 +462,23 @@ export class GitManager extends BaseManager {
 					`Git provider not supported: ${args.provider}.`,
 				);
 			}
+		}
+	}
+
+	async detectGitProvider(): Promise<string> {
+		try {
+			const remoteUrl = execaCommandSync("git remote get-url origin");
+			const domainRegex = /(?:https?:\/\/|git@)([^:/]+)[/:]/i;
+			const match = remoteUrl.stdout.match(domainRegex);
+			const domain = match && match.length > 0 ? match[1] : "";
+
+			return domain;
+		} catch (error) {
+			if (import.meta.env.DEV) {
+				console.error("Failed to detect Git provider:", error);
+			}
+
+			return "_unknown";
 		}
 	}
 
--- a/packages/manager/src/managers/git/GitManager.ts
+++ b/packages/manager/src/managers/git/GitManager.ts
@@ -470,7 +470,7 @@ export class GitManager extends BaseManager {
 			const remoteUrl = execaCommandSync("git remote get-url origin");
 			const domainRegex = /(?:https?:\/\/|git@)([^:/]+)[/:]/i;
 			const match = remoteUrl.stdout.match(domainRegex);
-			const domain = match && match.length > 0 ? match[1] : "";
+			const domain = match?.[1] ?? "";
 
 			return domain;
 		} catch (error) {
--- a/packages/start-slicemachine/src/StartSliceMachineProcess.ts
+++ b/packages/start-slicemachine/src/StartSliceMachineProcess.ts
@@ -90,6 +90,21 @@ export class StartSliceMachineProcess {
 				// noop - We don't want to stop the user from using Slice Machine
 				// because of failed tracking set up. We probably couldn't determine the
 				// Sentry environment.
+
+				if (import.meta.env.DEV) {
+					console.error("Error setting up Sentry:", error);
+				}
+			}
+
+			try {
+				this._trackStart();
+			} catch (error) {
+				// noop - We don't want to stop the user from using Slice Machine
+				// because of failed start event tracking.
+
+				if (import.meta.env.DEV) {
+					console.error("Error tracking start event:", error);
+				}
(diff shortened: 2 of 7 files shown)


packages/manager/src/managers/git/GitManager.ts
```diff
@@ -464,6 +465,23 @@ export class GitManager extends BaseManager {
...
 	}
 
+	async detectGitProvider(): Promise<string> {
+		try {
+			const remoteUrl = execaCommandSync("git remote get-url origin");
+			const domainRegex = /(?:https?:\/\/|git@)([^:/]+)[/:]/i;
+			const match = remoteUrl.stdout.match(domainRegex);
+			const domain = match && match.length > 0 ? match[1] : "";
+
+			return domain;
```
angeloashmore, line 475: We use `"gitHub" | "gitLab" | "bitbucket"` as provider identifiers, which are the camelcase versions of each provider's name.

https://github.com/prismicio/slice-machine/blob/9388a4b033a69b61abb4b57e6831abf722237519/packages/manager/src/constants/GIT_PROVIDER.ts#L5

We only support GitHub today, but we plan to support GitLab and Bitbucket in the future.

We'll probably want consistency in a general purpose method like this one. Is it possible for us to map the values for providers we know about?

packages/start-slicemachine/src/StartSliceMachineProcess.ts
```diff
@@ -212,4 +220,70 @@ export class StartSliceMachineProcess {
...
+			isAdapterUpdateAvailable,
+			isLoggedIn,
+			isSliceMachineUpdateAvailable,
+			isTypeScriptProject,
+			nodeVersion: process.versions.node,
+			numberOfCustomTypes: customTypes?.models.length,
+			numberOfSlices: slices?.models.length,
+			osPlatform: os.platform(),
+			// Ensure we escape the "@" character to prevent it from being interpreted as an email address.
+			packageManager: packageManager?.replace("@", "[at]"),
```
angeloashmore, line 304: Does Segment do something weird with `@`?

packages/start-slicemachine/src/lib/safelyExecute.ts
```diff
@@ -0,0 +1,9 @@
+export async function safelyExecute<T>(
```
angeloashmore, line 1: This is pedantic, but `safelyExecute` implies the function will execute something. Instead, this functions awaits something that was already executed in a Promise.

How about `safelyAwait` or `safelyTry`?

Or change `operation` to a function that executes the function when `safetyExecute` is called:

```ts
safelyExecute(() => this._sliceMachineManager.simulator.getLocalSliceSimulatorURL())
```

angeloashmore (approved):
Looks good! I have some small comments, none of which block the PR from being merged. I'll let you decide if the suggestions are necessary.

packages/start-slicemachine/src/StartSliceMachineProcess.ts
```diff
@@ -124,16 +126,22 @@ export class StartSliceMachineProcess {
...
 			await Promise.allSettled([
 				// noop - We'll try again when needed.
 				this._sliceMachineManager.user.refreshAuthenticationToken(),
 				// noop - We'll try again before uploading a screenshot.
 				this._sliceMachineManager.screenshots.initS3ACL(),
 			]);
 		}
+
+		try {
+			this._trackStart();
```
angeloashmore: Is there a reason we delay tracking to the end of the process? Since it happens in the background and shouldn't affect start up performance significantly, I wonder if we should move this immediately after initializing telemetry and plugins

packages/manager/src/managers/git/GitManager.ts
xrutayisire: I did that at the beginning, but it was less powerful than this. The goal here is really to just get the domain and let the domain tell us what they us. 

Thanks to this and not having a fixed list, it will help to get all possible providers / custom self hosted solution. I would differentiate the tracking compare to the git integration then. WDYT?

I will merge that but we can always change after without a problem!

packages/start-slicemachine/src/StartSliceMachineProcess.ts
xrutayisire: Yes, I will precise my comment. Segment treat email with * and does this "y***@b***" for example

packages/start-slicemachine/src/lib/safelyExecute.ts
xrutayisire: 100% right, I like your second proposal 🙂

packages/start-slicemachine/src/StartSliceMachineProcess.ts
xrutayisire: You're right it's even better now!

packages/manager/src/managers/git/GitManager.ts
angeloashmore: When we need to use `detectGitProvider` for other uses, we can revisit. This version works well for tracking, indeed. 🙂

packages/start-slicemachine/src/StartSliceMachineProcess.ts
angeloashmore: Ah, got it! To anonymize the data, I assume.

PR #1305, opened 2024-02-22
test: connect Git repository

bapmrl:
Add Playwright tests to cover the new `ConnectGitRepository` feature.

### DIFF
--- a/playwright/pages/SettingsPage.ts
+++ b/playwright/pages/SettingsPage.ts
@@ -1,0 +1,76 @@
+import { type Locator, type Page, expect } from "@playwright/test";
+
+import { GitRepositoryConnectDialog } from "./components/GitRepositoryConnectDialog";
+import { GitRepositoryDisconnectDialog } from "./components/GitRepositoryDisconnectDialog";
+import { SliceMachinePage } from "./SliceMachinePage";
+
+export class SettingsPage extends SliceMachinePage {
+  readonly breadcrumbLabel: Locator;
+  readonly installGitHubButton: Locator;
+  readonly gitRepositoryConnectDialog: GitRepositoryConnectDialog;
+  readonly gitRepositoryDisconnectDialog: GitRepositoryDisconnectDialog;
+  readonly unauthenticatedErrorTitle: Locator;
+  readonly unauthorizedErrorTitle: Locator;
+  readonly unknownErrorTitle: Locator;
+
+  constructor(page: Page) {
+    super(page);
+
+    /**
+     * Components
+     */
+
+    /**
+     * Static locators
+     */
+    this.breadcrumbLabel = this.breadcrumb.getByText("Settings", {
+      exact: true,
+    });
+    this.installGitHubButton = page.getByRole("button", {
+      name: "GitHub",
+      exact: true,
+    });
+    this.gitRepositoryConnectDialog = new GitRepositoryConnectDialog(page);
+    this.gitRepositoryDisconnectDialog = new GitRepositoryDisconnectDialog(
+      page,
+    );
+    this.unauthenticatedErrorTitle = page.getByText(
+      "It seems like you are logged out",
+      { exact: true },
+    );
+    this.unauthorizedErrorTitle = page.getByText(
+      "It seems like you do not have permission",
+      { exact: true },
+    );
+    this.unknownErrorTitle = page.getByText("Unable to fetch data", {
+      exact: true,
+    });
+  }
+
+  /**
+   * Dynamic locators
+   */
+  getConnectGitRepoButton(name: string): Locator {
+    return this.page
+      .locator("div", { hasText: name })
+      .getByRole("button", { name: "Connect", exact: true });
+  }
+
+  getDisconnectGitRepoButton(name: string): Locator {
+    return this.page
+      .locator("div", { hasText: name })
+      .getByRole("button", { name: "Disconnect", exact: true });
+  }
+
+  /**
+   * Actions
+   */
+  async goto() {
+    await this.page.goto("/settings");
+    await expect(this.breadcrumbLabel).toBeVisible();
+  }
+
+  /**
+   * Assertions
+   */
+}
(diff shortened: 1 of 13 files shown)


playwright/utils/errors.ts
```diff
@@ -0,0 +1,11 @@
+export class SliceMachineError extends Error {
```
xrutayisire, line 1: Super nice to export that kind of utils!

playwright/tests/settings/settings.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+    procedures.mock(
+      "git.getProviderAppInstallURL",
+      () => providerAppInstallURL,
+      { execute: false },
+    );
+
+    const gitHubPagePromise = settingsPage.page.context().waitForEvent("page");
+    await settingsPage.installGitHubButton.click();
+    const gitHubPage = await gitHubPagePromise;
+    await gitHubPage.waitForLoadState();
```
xrutayisire, line 37: I don't really understand the goal here. For me, we should try to test as a user can test. Can you not expect an element to be visible or not here? What the user should see after the click, that's waht we want to expect

playwright/pages/SettingsPage.ts
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * Actions
+   */
+  async goto() {
+    await this.page.goto("/settings");
+    await expect(this.breadcrumbLabel).toBeVisible();
+  }
+
+  /**
+   * Assertions
```
xrutayisire, line 80: If we want to strictly be consistent, I added a comment for every empty block ^^ Like "Handle xxxxxx here". Maybe we don't want it, but then we have some or without. Not a big deal at all.

playwright/pages/SettingsPage.ts
```diff
@@ -0,0 +1,76 @@
...
+  readonly gitRepositoryDisconnectDialog: GitRepositoryDisconnectDialog;
+  readonly unauthenticatedErrorTitle: Locator;
+  readonly unauthorizedErrorTitle: Locator;
+  readonly unknownErrorTitle: Locator;
+
+  constructor(page: Page) {
+    super(page);
+
+    /**
+     * Components
```
xrutayisire, line 20: Components here for the Dialogs. So we differentiate classic locators and class components

```ts
this.gitRepositoryConnectDialog = new GitRepositoryConnectDialog(page);
this.gitRepositoryDisconnectDialog = new GitRepositoryDisconnectDialog(
  page,
);
```

playwright/pages/SettingsPage.ts
```diff
@@ -0,0 +1,76 @@
...
+      .locator("div", { hasText: name })
+      .getByRole("button", { name: "Disconnect", exact: true });
+  }
+
+  /**
+   * Actions
+   */
+  async goto() {
+    await this.page.goto("/settings");
+    await expect(this.breadcrumbLabel).toBeVisible();
```
xrutayisire, line 76: After some check, it seems it's because of the Next Error Popup grrrr

Not 100% sure but you could verify with having the screenshot.

Screenshot 2024-02-23 at 00 43 52

xrutayisire: So breadcrumbLabel exist but hidden

xrutayisire: I would suggest if it's the modal that you just share from `SliceMachinePage` a function to close the modal manually... It's super hacky, but we should not have this modal for expected error. Then you just have to call the closeNextErrorMocal modal or something like that after loading the page.

Maybe you would also like to invest time in removing the modal but since it's Next and we don't want to keep it I would really try to keep it minimal here.

xrutayisire: Btw today we run Playwright with SM in dev mode so we have the latest version of the package. It could be different if we use the alpha version for example. But not done at all yet. More reason not to invest too much here I guess.

playwright/pages/SettingsPage.ts
```diff
@@ -0,0 +1,76 @@
...
+      exact: true,
+    });
+  }
+
+  /**
+   * Dynamic locators
+   */
+  getConnectGitRepoButton(name: string): Locator {
+    return this.page
+      .locator("div", { hasText: name })
```
xrutayisire: These two locators feels a bit off. Sorry I don't have time to look for the solution, but I think it should be closer to what we are doing in `TypesTablePage` for the `getRow` function.

No usage of implementation details should be done in tests. Here "div" for example. Normally we should almost never need `locator`. I added a best practice in the test README for that ^^

xrutayisire (changes requested):
It's a super good work! I love the tests, feel natural to read them, even with the mocking! 

Left some comments and tried to help you for the crash in CI 🙌

playwright/pages/SettingsPage.ts
angeloashmore: We could build packages rather than run the dev scripts. Would that solve this problem, and technically get us more accurate testing?

playwright/tests/settings/settings.spec.ts
bapmrl: I've just added a check on the GitHub page's title, but I cannot check the `Button`'s loading state as the "GitHub" label disappears. To circumvent this, I would have to add a `data-cy` attribute, but I find it contrived (if we want to write tests from the user's point of view).

playwright/pages/SettingsPage.ts
bapmrl: I don't think it's useful, on top of the block comment, but I've added them back and we could remove them all in another pull request.

playwright/pages/SettingsPage.ts
bapmrl: You're right. I've changed the implementation of `FieldSetList` to use an `ol` and of `FieldSetListItem` to use a `li`. Thus, I can use `this.page.getByRole("listitem")` instead of `this.page.locator("div")`.

playwright/pages/SettingsPage.ts
xrutayisire: Did you check @bapmrl if the cause of the bug is the Next error modal?

bapmrl: It was due to the `TelemetryManager` not returning an eligible `git-integration-experiment`. I've mocked the `TelemetryManager.getExperimentVariant` method and the tests now pass.

xrutayisire (approved):
LGTM 🙌

PR #1306, opened 2024-02-22
test: refactor test options

bapmrl:
Use Playwright's [fixture-options](https://playwright.dev/docs/test-fixtures#fixtures-options) instead of our own implementation.

### DIFF
--- a/playwright/README.md
+++ b/playwright/README.md
@@ -68,48 +68,49 @@ To open a downloaded CI test report from anywhere in your computer:
 
 ## Creating tests
 
-### `test.run()`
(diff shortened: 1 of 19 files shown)


xrutayisire (approved):
Of course this one is based on the other and will have to wait to check if nothing break in CI, but I LOVE IT!

Thank you so much for the investment now as the more we were doing the custom way the harder it would have been to change. 

Your way is the clean easy way! So nice 🚀

bapmrl:
Thank you @xrutayisire 😊

PR #1309, opened 2024-02-28
test(playwright): rename `data-cy` to `data-testid`

angeloashmore:
## Context

Completes DT-1839

## The Solution

- Rename `data-cy` to `data-testid` in all files.
- Remove the `testIdAttribute` Playwright setting that changed `data-testid` to `data-cy`; the default value is `data-testid`.
- Rename `dataCy` props to `testId`.

## Impact / Dependencies

N/A

- [ ] The CI is successful.

### DIFF
--- a/packages/slice-machine/components/ModalFormCard/index.tsx
+++ b/packages/slice-machine/components/ModalFormCard/index.tsx
@@ -104,7 +104,7 @@ function ModalCard<Values extends FormikValues>({
           return (
             <Form
               id={formId}
-              {...(dataCy != null ? { "data-cy": dataCy } : null)}
+              {...(dataCy != null ? { "data-testid": dataCy } : null)}
             >
               <Card
                 borderFooter
--- a/packages/slice-machine/components/ModalFormCard/index.tsx
+++ b/packages/slice-machine/components/ModalFormCard/index.tsx
@@ -48,7 +48,7 @@ type ModalCardProps<T extends FormikValues> = {
   cardProps?: ComponentPropsWithoutRef<typeof Card>;
   omitFooter?: boolean;
   isLoading?: boolean;
-  dataCy?: string;
+  testId?: string;
   actionMessage?: ((props: FormikProps<T>) => ReactNode) | ReactNode;
 };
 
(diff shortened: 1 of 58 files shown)


packages/slice-machine/test/pages/slices.test.tsx
```diff
@@ -337,13 +337,17 @@ describe("slices", () => {
       },
     });
 
-    const createOneButton = document.querySelector('[data-cy="create-slice"]');
+    const createOneButton = document.querySelector(
```
xrutayisire: I guess we can just getByTestId, right?

packages/slice-machine/test/pages/slices.test.tsx
```diff
@@ -337,13 +337,17 @@ describe("slices", () => {
...
+      '[data-testid="create-slice"]',
+    );
     await act(async () => {
       // @ts-expect-error TS(2345) FIXME: Argument of type 'Element | null' is not assignabl... Remove this comment to see the full error message
       fireEvent.click(createOneButton);
     });
 
-    const nameInput = document.querySelector('[data-cy="slice-name-input"]');
+    const nameInput = document.querySelector(
```
xrutayisire: I guess we can just getByTestId, right?

packages/slice-machine/components/ModalFormCard/index.tsx
```diff
@@ -104,7 +104,7 @@ function ModalCard<Values extends FormikValues>({
           return (
             <Form
               id={formId}
-              {...(dataCy != null ? { "data-cy": dataCy } : null)}
+              {...(dataCy != null ? { "data-testid": dataCy } : null)}
```
xrutayisire: I guess we could rename all dataCy to clean old code also, WDYT?

xrutayisire (approved):
Super nice, small comments for improvements. I guess if you decide not to do it, you will have a good reason so either way good to merge after.

packages/slice-machine/test/pages/slices.test.tsx
angeloashmore: Good catch! We already use Testing Library in this test.

packages/slice-machine/components/ModalFormCard/index.tsx
angeloashmore: Yes, good point. I didn't know about these `dataCy` props. I renamed the props to `testId`, which I think is more declarative:

```tsx
<Foo testId="foo-thing" />
```

PR #1310, opened 2024-02-28
test(playwright): complete page type tests

angeloashmore:
## Context

Completes DT-1835

## The Solution

- Test deleting a page type.
- Test for the blank slate message when the project has no page types.

## Impact / Dependencies

N/A

- [ ] The CI is successful.

### DIFF
--- a/playwright/tests/pageTypes/pageTypesTable.spec.ts
+++ b/playwright/tests/pageTypes/pageTypesTable.spec.ts
@@ -154,8 +154,25 @@ test.run()(
     await expect(pageTypesTablePage.blankSlateCreateAction).toBeVisible();
   },
 );
-    await pageTypesTablePage.goto();
 
-    await expect(pageTypesTablePage.blankSlate).toBeVisible();
+test.run()(
+  "I can create a new page type from the blank slate",
+  async ({ pageTypesTablePage, pageTypesBuilderPage, procedures }) => {
+    procedures.mock("getState", ({ data }) => ({
+      ...(data as Record<string, unknown>),
+      customTypes: [],
+    }));
+    await pageTypesTablePage.goto();
+    await pageTypesTablePage.blankSlateCreateAction.click();
+
+    const name = "Page Type " + generateRandomId();
+    await pageTypesTablePage.createTypeDialog.createType(name, "reusable");
+
+    // TODO(DT-1801): Production BUG - When creating a page type, don't redirect
+    // to the builder page until the page type is created
+    procedures.unmock("getState");
+    await pageTypesBuilderPage.goto(name);
+
+    await expect(pageTypesBuilderPage.getBreadcrumbLabel(name)).toBeVisible();
   },
 );
--- a/playwright/tests/pageTypes/pageTypesTable.spec.ts
+++ b/playwright/tests/pageTypes/pageTypesTable.spec.ts
@@ -13,10 +13,6 @@ test("I can create a reusable page type", async ({
   const name = "Page Type " + generateRandomId();
   await pageTypesTablePage.createTypeDialog.createType(name, "reusable");
 
-  // TODO(DT-1801): Production BUG - When creating a page type, don't redirect
-  // to the builder page until the page type is created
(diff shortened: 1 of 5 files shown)


playwright/tests/pageTypes/pageTypesTable.spec.ts
```diff
@@ -120,3 +120,35 @@ test.run()(
...
+  "I can see the blank slate message when I don't have any custom types",
+  async ({ pageTypesTablePage, procedures }) => {
+    procedures.mock(
+      "customTypes.readAllCustomTypes",
+      () => ({ models: [], errors: [] }),
+      { execute: false },
+    );
+    await pageTypesTablePage.goto();
+
+    await expect(pageTypesTablePage.blankSlate).toBeVisible();
```
xrutayisire: Test don't pass 👀.

But I was thinking to test closer to the user without testId, maybe we should more verify the blank slate description and button? With that no need to a testId for the blankSlate, WDYT?

We could also check that clicking on the button open the create modal so we ensure user is not blocked on the blank slate forever. WDYT?

xrutayisire (changes requested):
Super cool to continue tech debt roadmap and coverage here!
I added two propositions of changes, let me know.

Note: Baptiste merged a big PR that is super cool removing the need of test.run, so you have conflicts and need to update your new tests. It's really better and exactly how we are supposed to do in Playwright. Baptiste updated the README so you should quickly get it. If not, don't hesitate to update the README also 🙂

bapmrl:
> Note: Baptiste merged a big PR that is super cool removing the need of test.run, so you have conflicts and need to update your new tests.

You just have to replace `test.run()` with `test` @angeloashmore 🙂

playwright/tests/pageTypes/pageTypesTable.spec.ts
angeloashmore: Oops, it passed locally because I manually deleted the existing custom type before I started. 👀

Testing the message and button is a good idea. I think we still need a test ID to locate the blank slate message and button, however, unless you have a different idea of how to target them.

Here is the updated test: https://github.com/prismicio/slice-machine/blob/cdc4166cb32958340a8a7afedb391741cc1cb6f6/playwright/tests/pageTypes/pageTypesTable.spec.ts#L142-L156

And the new test verifying the blank slate's action button: https://github.com/prismicio/slice-machine/blob/cdc4166cb32958340a8a7afedb391741cc1cb6f6/playwright/tests/pageTypes/pageTypesTable.spec.ts#L158-L178

xrutayisire: I'm never sure if we really need to locate the item first. It's more if the text can be at two different places why not. But here for example it seems the description and create button can only be there within the blank slate. It's even more powerful with the id and I don't have a strong opinion so LGTM 🙂

xrutayisire: Btw relative to the todo you wrote. I think I may have fixed the problem so you may not need that overhead and the redirection should be perfect! Thanks to the removal of the saga, I forgot about these todo! 👀 Could you check if there are required and remove them if not?

angeloashmore: The description is probably fine to check if it is present anywhere on the page, but the "Create" button could be confused with the non-blank-slate "Create" button. We don't currently show two Create buttons, but I'm thinking we should be explicit about which button we are checking for in case we have a regression somewhere.

Regarding the TODO: You are totally right! I tested it without the extra navigation before, and it failed. But now it works without it. There must have been something else wrong with the test back then.

I'll remove the other uses of `goto()`.

xrutayisire: In this case I guess it's nice to be specific yes as we know we have another button with the same label, so all good with the testid 👍

Super cool! Something less to take care haha

xrutayisire (approved):
LGTM after the changes we discussed about 🙂

PR #1311, opened 2024-02-28
fix(sideNav): Rename side nav Tutorial / Academy link to Learn Prismic

xrutayisire:
## Context

- DT-1987

## The Solution

- Rename Tutorial (Nuxt, SvelteKit) and Academy (Next) to "Learn Prismic".
- Added an e2e tests to check the link redirection
- Added titles for our Describe

## Impact / Dependencies

- None

### DIFF
--- a/playwright/tests/common/sideNav.spec.ts
+++ b/playwright/tests/common/sideNav.spec.ts
@@ -108,7 +108,7 @@ test("I cannot see the updates available warning", async ({
   await expect(pageTypesTablePage.menu.updatesAvailableTitle).not.toBeVisible();
 });
 
-test.describe(() => {
+test.describe("Tutorial tooltip", () => {
   test.use({
     onboarded: false,
     reduxStorage: {
@@ -152,4 +152,18 @@ test.describe(() => {
       sliceMachinePage.menu.tutorialVideoTooltipTitle,
     ).not.toBeVisible();
   });
+});
+
+test("I can access the Academy from the 'Learn Prismic' link", async ({
+  pageTypesTablePage,
+}) => {
+  await pageTypesTablePage.goto();
+  await expect(pageTypesTablePage.menu.learnPrismicLink).toBeVisible();
+
+  await pageTypesTablePage.menu.learnPrismicLink.click();
+
+  const newTab = await pageTypesTablePage.page.waitForEvent("popup");
+  await newTab.waitForLoadState();
+
+  await expect(newTab).toHaveURL(/https:\/\/prismic.io\/academy*/);
 });
--- a/playwright/tests/common/sideNav.spec.ts
+++ b/playwright/tests/common/sideNav.spec.ts
@@ -154,16 +154,16 @@ test.describe("Tutorial tooltip", () => {
   });
 });
 
-test("I can access the Academy from the 'Learn Prismic' link", async ({
-  pageTypesTablePage,
+test('I can access the Academy from the "Learn Prismic" link', async ({
+  sliceMachinePage,
 }) => {
-  await pageTypesTablePage.goto();
-  await expect(pageTypesTablePage.menu.learnPrismicLink).toBeVisible();
+  await sliceMachinePage.gotoDefaultPage();
+  await expect(sliceMachinePage.menu.learnPrismicLink).toBeVisible();
 
-  await pageTypesTablePage.menu.learnPrismicLink.click();
+  await sliceMachinePage.menu.learnPrismicLink.click();
 
-  const newTab = await pageTypesTablePage.page.waitForEvent("popup");
+  const newTab = await sliceMachinePage.page.waitForEvent("popup");
   await newTab.waitForLoadState();
 
-  await expect(newTab).toHaveURL(/https:\/\/prismic.io\/academy*/);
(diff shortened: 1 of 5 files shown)


playwright/tests/common/sideNav.spec.ts
```diff
@@ -153,3 +153,17 @@ test.describe(() => {
     ).not.toBeVisible();
   });
 });
+
+test("I can access the Academy from the 'Learn Prismic' link", async ({
```
bapmrl: 💡 I would tend to favor double quotes in output strings (that are visible in some user interface).

playwright/tests/common/sideNav.spec.ts
```diff
@@ -153,3 +153,17 @@ test.describe(() => {
...
+  pageTypesTablePage,
+}) => {
+  await pageTypesTablePage.goto();
```
bapmrl: ❓ Why not use the `sliceMachinePage` instead? We don't need to access a specific page for this test.

playwright/tests/common/sideNav.spec.ts
```diff
@@ -153,3 +153,17 @@ test.describe(() => {
...
+  await expect(pageTypesTablePage.menu.learnPrismicLink).toBeVisible();
+
+  await pageTypesTablePage.menu.learnPrismicLink.click();
+
+  const newTab = await pageTypesTablePage.page.waitForEvent("popup");
```
bapmrl: 💡 For consistency, you could remove the `context()` call from the `settingsPage.page.context().waitForEvent("page")` expression in the `playwright/tests/settings/settings.spec.ts` module.

playwright/tests/common/sideNav.spec.ts
```diff
@@ -153,3 +153,17 @@ test.describe(() => {
...
   });
 });
+
+test("I can access the Academy from the 'Learn Prismic' link", async ({
+  pageTypesTablePage,
+}) => {
+  await pageTypesTablePage.goto();
+  await expect(pageTypesTablePage.menu.learnPrismicLink).toBeVisible();
+
+  await pageTypesTablePage.menu.learnPrismicLink.click();
```
bapmrl: 💡 I think it's safer to `click()` after `waitForEvent("popup")` has been called.

playwright/tests/common/sideNav.spec.ts
```diff
@@ -153,3 +153,17 @@ test.describe(() => {
...
+
+  const newTab = await pageTypesTablePage.page.waitForEvent("popup");
+  await newTab.waitForLoadState();
+
+  await expect(newTab).toHaveURL(/https:\/\/prismic.io\/academy*/);
```
bapmrl: 💡 I think it would be simpler to just check if the `title()` contains `Prismic Academy`. Real users wouldn't know the specific URL.

bapmrl (approved):
Nice work 🙂

playwright/tests/common/sideNav.spec.ts
xrutayisire: 100% right, my bad!

playwright/tests/common/sideNav.spec.ts
xrutayisire: True!

playwright/tests/common/sideNav.spec.ts
xrutayisire: Hum, I don't understand, I directly await for the new tab event to be done, the line will never resolve before the click.

playwright/tests/common/sideNav.spec.ts
xrutayisire: We cannot do that, it's two different way to handle in Playwright side.

See: https://playwright.dev/docs/pages#handling-new-pages
Screenshot 2024-03-05 at 10 59 56

playwright/tests/common/sideNav.spec.ts
xrutayisire: Good idea, I was afraid of having crash just because they changed something within the page, but the title seems as much stable as the URL.

xrutayisire: But just so you know for me URLs are part of user's world, title is better, but as long as the URL can be visible in browser we can consider that's testing it is not implementation details.

xrutayisire: Done 🙂

playwright/tests/common/sideNav.spec.ts
xrutayisire: Done 🙂

playwright/tests/common/sideNav.spec.ts
xrutayisire: Done 🙂

playwright/tests/common/sideNav.spec.ts
bapmrl: Okay, no problem ^^

playwright/tests/common/sideNav.spec.ts
bapmrl: My bad, in my test suite covering the Settings page I'm not `await`ing the `waitForEvent` call. Your use case is different :)

bapmrl:
Looks good to merge 👍

PR #1314, opened 2024-03-05
refactor(redux-saga): Remove toasters sagas & saveMock saga

xrutayisire:
## Context

- Completes DT-1967, DT-1994, DT-1838

## The Solution

- Replace all usage of openToaster with a direct toast.<type>
- Remove watchToasterSagas
- Remove saveSliceMockSaga saga

## Impact / Dependencies

- None


packages/slice-machine/components/Simulator/index.tsx
```diff
@@ -9,14 +9,15 @@ import {
...
+import { toast } from "react-toastify";
 
 import { defaultSharedSliceContent } from "@src/utils/editor";
 
 import { BaseStyles, Box, Flex, Spinner } from "theme-ui";
 
 import Header from "./components/Header";
 
-import { telemetry } from "@src/apiClient";
+import { saveSliceMock, telemetry } from "@src/apiClient";
```
angeloashmore, line 20: A note for future tech debt work we can plan later: `@src/apiClient` should be replaced with direct manager usage. In this case, `saveSliceMock` is effectively an alias for `managerClient.slices.updateSliceMocks`.

https://github.com/prismicio/slice-machine/blob/973eb9558e666f97897d050c524289cd68c2a36b/packages/slice-machine/src/apiClient.ts#L201-L209

For context, `@src/apiClient` looked very different before we had the manager. As a first step, we converted that file to use the manager. Now, we seem better prepared to remove that file altogether.

angeloashmore (approved):
LGTM! Simplifying `openToaster` to `toast.xxx` is a nice improvement.

packages/slice-machine/components/Simulator/index.tsx
xrutayisire: I like the idea!

PR #1315, opened 2024-03-05
feat(adapter-next): support Next.js dynamic functions in the slice simulator

angeloashmore:
## Context

DT-1995

## The Solution

When using the App Router with a Server Component, render the simulator's `<SliceZone>` component on the server. Rendering on the server allows slices written as Server Components to access dynamic functions, like `headers()` and `cookies()`.

Rendering `<SliceZone>` on the server requires removing the render prop from the classic boilerplate. To get around the render prop, state is stored in the URL and requires a `getSlices()` helper in the page's component.

The following flow is used:

```mermaid
flowchart TD
    A[Use the `state` URL parameter to render the SliceZone] -.->|Simulated content is updated| B
    B[Receive slice data from the parent] --> C
    C[Store the slices in the URL as a URL parameter,\ncompressed with lz-string] --> D
    D[Revalidate the path's data] -->|Triggers a re-render| A
```

The new App Router-specific boilerplate looks like this:

```tsx
// src/app/slice-simulator/page.tsx

import {
  SliceSimulator,
  SliceSimulatorParams,
  getSlices,
} from "@slicemachine/adapter-next/simulator";
import { SliceZone } from "@prismicio/react";

import { components } from "../../slices";

export default function SliceSimulatorPage({
  searchParams,
}: SliceSimulatorParams) {
  const slices = getSlices(searchParams.state);

  return (
    <SliceSimulator>
      <SliceZone slices={slices} components={components} />
    </SliceSimulator>
  );
}
```

All projects using the App Router _should_ use the new style that supports Next.js' dynamic functions. Thus, generated code from `@slicemachine/adapter-next` have been updated with the new boilerplate. Prismic's Next.js starters will be updated once this PR is released.

### Differences between the Pages Router and the App Router

When `SliceSimulator` is imported in the Pages Router or a Client Component (one that uses the `"use client"` directive), a client-side-only version is provided. **This version of `SliceSimulator` is the same as before this PR**.

When `SliceSimulator` is imported in the App Router with a Server Component (one that is not marked with the `"use client"` directive), `SliceSimulator` is a special version that stores state in the URL as a search parameter and requires use of the `getSlices()` helper.

## Impact / Dependencies

Projects that need to support Next.js' dynamic functions will need to change the code for their `/slice-simulator` page. The code is mostly generic and can be copy/pasted into most projects.

There are no breaking changes in this PR. Projects that do not modify their `/slice-simulator` will continue to work as it does today. That means they will not support Next.js' dynamic functions.

- [x] The CI is successful.


hypervillain:
Thanks a lot for the PR description, it helped a lot. Can you explain me why we need to provide a `getSlices` helper that throws inside the client version of the simulator? Is it in case someone copied the wrong template (eg. switched from app router to "classic" router)?

angeloashmore:
> Can you explain me why we need to provide a `getSlices` helper that throws inside the client version of the simulator? Is it in case someone copied the wrong template (eg. switched from app router to "classic" router)?

Yes, that is exactly the reason. If someone copy-and-pastes code meant for a Server Component into a Client Component, we can tell them early that the code is incorrect.

We also need to ensure we have the same export names in both the client module (`/simulator`) and the server module (`/simulator/react-server`). If we only provided `getSlices()` from the server module, TypeScript wouldn't pick it up and would probably throw on build.

Note that the switch between the client and server modules happens automatically as part of Next.js' build pipeline due to how the `package.json` `exports` property is configured:

https://github.com/prismicio/slice-machine/blob/6387341a80105ccdfd5b64b82334fbd457bb8c5c/packages/adapter-next/package.json#L24-L28

[This RFC explains the feature](https://github.com/reactjs/rfcs/blob/c15bc9df5afa8fd1dca6e5fd1c2ed073f7a9bd79/text/0227-server-module-conventions.md#react-server-conditional-exports).

PR #1316, opened 2024-03-08
test(playwright): test the sidebar's repository and "Invite team" links

angeloashmore:
## Context

DT-1824

## The Solution

- Add a test for the repository link
- Add a test for the "Invite team" link

The "Invite team" test requires checking the opened window's URL rather than its title like the other tests. The title on the `/settings/users` page does not contain anything specific to the repository's user settings.

## Impact / Dependencies

N/A

- [ ] The CI is successful.

### DIFF
--- a/playwright/tests/common/sideNav.spec.ts
+++ b/playwright/tests/common/sideNav.spec.ts
@@ -60,6 +60,28 @@ test("I can navigate through all menu entries", async ({
   );
 });
 
+test("I can access the repository using the open icon", async ({
+  sliceMachinePage,
+  procedures,
+}) => {
+  procedures.mock("getState", ({ data }) => {
+    const result = data as { env: { manifest: { apiEndpoint: string } } };
+    result.env.manifest.apiEndpoint =
+      "https://example-prismic-repo.cdn.prismic.io/api/v2";
+    return result;
+  });
+
+  await sliceMachinePage.gotoDefaultPage();
+  await expect(sliceMachinePage.menu.repositoryLink).toBeVisible();
+
+  await sliceMachinePage.menu.repositoryLink.click();
+
+  const newTab = await sliceMachinePage.page.waitForEvent("popup");
+  await newTab.waitForLoadState();
+
+  await expect(newTab).toHaveTitle("prismic.io - Example Prismic Repo");
+});
+
 test("I access the changelog from Slice Machine version", async ({
   pageTypesTablePage,
   changelogPage,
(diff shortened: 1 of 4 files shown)


playwright/tests/common/sideNav.spec.ts
```diff
@@ -60,6 +60,28 @@ test("I can navigate through all menu entries", async ({
...
+      "https://example-prismic-repo.cdn.prismic.io/api/v2";
+    return result;
+  });
+
+  await sliceMachinePage.gotoDefaultPage();
+  await expect(sliceMachinePage.menu.repositoryLink).toBeVisible();
+
+  await sliceMachinePage.menu.repositoryLink.click();
+
+  const newTab = await sliceMachinePage.page.waitForEvent("popup");
```
xrutayisire: I made an error on my code like that that Baptiste tried to show me, but that I didn't see.

This is working, but not totally like the doc is doing: https://playwright.dev/docs/pages#handling-popups

As you can see they tell you to create the Promise without await before the click that trigger the popup event. I think it's better if we do like the doc wants us to do.

I fixed it already in my opened PR for the Academy test.

Screenshot 2024-03-08 at 10 17 58

playwright/tests/common/sideNav.spec.ts
```diff
@@ -167,3 +189,28 @@ test('I can access the Academy from the "Learn Prismic" link', async ({
...
+      "https://example-prismic-repo.cdn.prismic.io/api/v2";
+    return result;
+  });
+
+  await sliceMachinePage.gotoDefaultPage();
+  await expect(sliceMachinePage.menu.inviteTeamLink).toBeVisible();
+
+  await sliceMachinePage.menu.inviteTeamLink.click();
+
+  const newTab = await sliceMachinePage.page.waitForEvent("popup");
```
xrutayisire: Same comment as above

xrutayisire (approved):
2 comments that are the same that I think we should handle.

After that, LGTM, nice to have more coverage! You can merge after.

xrutayisire:
And btw small unit test to update before merging since it's crashing

playwright/tests/common/sideNav.spec.ts
angeloashmore: Ah, interesting! Fixed here: https://github.com/prismicio/slice-machine/pull/1316/commits/ae0a66b222fe809ab2cfaaac89715841064b1bec

PR #1317, opened 2024-03-13
feat(slice-machine): make the menu responsive

bapmrl:
Completes DT-1800.


packages/slice-machine/components/Navigation/ChangesItem.tsx
```diff
@@ -22,25 +31,50 @@ export const ChangesItem: FC = () => {
...
+      align="start"
       open={open}
       openDelay={3000}
       onClose={onClose}
       side="right"
       sideOffset={24}
-      collisionPadding={280}
       trigger={
-        <Box padding={{ bottom: 24 }}>
+        <Box alignItems={isSmall ? "center" : undefined} flexDirection="column">
```
angeloashmore, line 49: In the future, do you think we will have a built-in system for managing responsive styles? For example:

```ts
alignItems={{ small: "center" }}
```

packages/slice-machine/src/components/SideNav/SideNav.module.css
```diff
@@ -0,0 +1,318 @@
+.block {
+  all: unset;
+  box-sizing: border-box;
+  font-family: var(--font-family-body);
```
angeloashmore, line 4: Should `font-family` be in`.root` instead? `.block` would imply a class primarily adding styles for `display: block`.

packages/slice-machine/src/components/SideNav/SideNav.module.css
```diff
@@ -0,0 +1,318 @@
...
+  margin-top: 8px;
+  &:first-child {
+    margin-top: 0px;
+  }
+}
+
+.listTitle {
+  composes: blockWithDisplayRevert;
+  color: var(--grey11);
+  font-size: calc(12rem / 16);
```
angeloashmore, line 122: Interesting technique to specify `rem` using `px` numbers. Is that what this is doing?

packages/slice-machine/src/components/SideNav/SideNav.tsx
```diff
@@ -118,18 +118,24 @@ export const SideNavLink: FC<SideNavLinkProps> = ({
...
-      <div className={styles.linkContent}>
-        <span className={styles.linkText}>{title}</span>
-        {RightElement}
-      </div>
-    </>,
+}) => {
+  const isSmall = useMediaQuery(breakpoints.small);
+  return (
+    <Tooltip content={title} side="right" visible={isSmall}>
+      {createElement(
```
angeloashmore: I don't see `createElement` being used directly often. I see this pattern instead:

```tsx
const MyComponent = ({ component: Comp }) => {
  return <Comp>{/* ... */}</Comp>;
}
```

Do you see an advantage to `createElement` over JSX?

angeloashmore (approved):
LGTM!

As a general question, do we have a standard practice for naming CSS classes? Maybe a list of good and bad examples would help me and others know how we should be structuring our CSS files.

packages/slice-machine/components/Navigation/ChangesItem.tsx
bapmrl: I don't know. It's something that was supported by `sprinkles`, but we would have to implement our own version of it. What do you think @AlexGalays?

packages/slice-machine/src/components/SideNav/SideNav.module.css
bapmrl: I've tried to minimize the amount of changes I made while converting the file to CSS Modules. I'm not found of how it was implemented, but I think improving that component is a topic for another time ^^

packages/slice-machine/src/components/SideNav/SideNav.module.css
bapmrl: Yes, I think so ^^ It's what the Page Builder team is doing.

packages/slice-machine/src/components/SideNav/SideNav.tsx
bapmrl: I didn't want to change the existing implementation while converting the component, but you're right, it's better 🙂 I've removed both uses of `createElement`.

bapmrl:
> LGTM!
> 
> As a general question, do we have a standard practice for naming CSS classes? Maybe a list of good and bad examples would help me and others know how we should be structuring our CSS files.

From what I've seen, the Page Builder team uses `camelCasing` and they append a dash for variants. For example, the `red` variant of a button would be `buttonColor-red`.

PR #1318, opened 2024-03-13
refactor(redux-saga): Remove watchCheckSetup sagas

xrutayisire:
## Context

- Completes DT-1966

## The Solution

- Remove ALL code related to Simulator setup
- From the Modal displayed to the adapters with the markdown
- Enforce the fact the simulator is mandatory for all adapters as it's now a required big part of Prismic with the new PB

## To-do

- Documentation needs to be updated
- Add documentation when an error happen

## Impact / Dependencies

- No verification of adapters supporting the simulator


angeloashmore (approved):
It all looks good to me! It's great that we can remove all of this code.

When the simulator fails to load, we display a "troubleshooting instructions" link that points to <https://prismic.io/docs/slice-machine>. Perhaps we should change it to <https://prismic.io/docs/slice-machine#simulate-slices>.

PR #1319, opened 2024-03-13
test(playwright): update slice screenshot from the Changes page

angeloashmore:
## Context

DT-1826

## The Solution

- Add a test that changes a slice's screenshot from the Changes page

We need to use a real slice (via the `slice` fixture) because the adapter processes the request. Doing it this way is preferred over mocking and validating every procedure that this action touches because we can verify it actually works as a user would use it.

## Impact / Dependencies

N/A

- [x] The CI is successful.

### DIFF
--- a/playwright/tests/changes/changes.spec.ts
+++ b/playwright/tests/changes/changes.spec.ts
@@ -1,4 +1,7 @@
-import { CustomType } from "@prismicio/types-internal/lib/customtypes";
+import {
+  CustomType,
+  SharedSlice,
+} from "@prismicio/types-internal/lib/customtypes";
 import { expect } from "@playwright/test";
 
 import { test } from "../../fixtures";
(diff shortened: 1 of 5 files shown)


playwright/tests/changes/changes.spec.ts
```diff
@@ -96,6 +96,28 @@ test("I can see the changes I have to push", async ({
   await changesPage.checkCustomTypeStatus(customType.id, "New");
 });
 
+test("I can update screenshots", async ({ changesPage, procedures, slice }) => {
+  procedures.mock("getState", ({ data }) => ({
+    ...(data as Record<string, unknown>),
+    clientError: undefined,
+  }));
+  // Necessary to ensure the page is not logged out.
+  procedures.mock("git.fetchLinkedRepos", () => []);
```
xrutayisire: Btw I think it's related to the project your testing on having or not the amplitude experiment activated. If. not activated, it will not even call the manager procedure.

Are the other tests crashing locally for you with the repository you're testing?
Another way could be to mock the procedure for the amplitude experiment before all tests to ensure it's always working.

xrutayisire (approved):
I left one comment on a way we could improve the problem you had. I let you judge if it's better 🙂

LGTM otherwise

playwright/tests/changes/changes.spec.ts
angeloashmore: I decided to disable all experiments by default, since that seems to be the most stable way to run tests: https://github.com/prismicio/slice-machine/pull/1319/commits/2040fdc291d22cf1303951995676817e8beb608b

What do you think? Should we disable them in all tests? If so, is that the correct place? I worry disabling the experiment there is not obvious for others.

xrutayisire: Super good proposal! I like it! I would do it in the fixture, WDYT? So it's for all but we should be able to override it when we really want to test the experiment right?

angeloashmore: Okay, works for me!

Experiments are disabled in tests here: https://github.com/prismicio/slice-machine/blob/2040fdc291d22cf1303951995676817e8beb608b/playwright/fixtures/index.ts#L251-L252

Documentation about enabling experiments was added here: https://github.com/prismicio/slice-machine/blob/512da7167297f7e43477e651905f57db477b1ac5/playwright/README.md#enable-amplitude-experiments-as-needed

PR #1320, opened 2024-03-13
test(playwright): custom types slice zone

angeloashmore:
## Context

DT-1830

## The Solution

Copy and convert the slice zone tests written for page types into ones for custom types.

## Impact / Dependencies

N/A

- [x] The CI is successful.


xrutayisire (approved):
LGTM!

It's something I'm not yet 100% about, the duplication. For tests, I think it's fine as you did, and we could improve it in the future. But it less us keep clear readability, so I like what you did.

PR #1321, opened 2024-03-13
test(playwright): complete slice list tests

angeloashmore:
## Context

DT-1817

- [x] The CI is successful.

### DIFF
--- a/packages/slice-machine/pages/slices.tsx
+++ b/packages/slice-machine/pages/slices.tsx
@@ -137,7 +137,7 @@ const SlicesIndex: React.FunctionComponent = () => {
                           .
                         </>
                       }
-                      data-testid="blank-slate"
+                      data-testid="slices-table-blank-slate"
                     />
                   </Flex>
                 ) : (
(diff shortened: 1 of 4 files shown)


packages/slice-machine/pages/slices.tsx
```diff
@@ -137,6 +137,7 @@ const SlicesIndex: React.FunctionComponent = () => {
                           .
                         </>
                       }
+                      data-testid="blank-slate"
```
xrutayisire:
```suggestion
                      data-testid="slices-table-blank-slate"
```

I would suggest something specific to slices here, it's a specific blank test, and we should try to keep test id unique across the app on things that are unique.

xrutayisire (approved):
I left you one small comment, let you find the best name 🙂

After that LGTM to merge!

packages/slice-machine/pages/slices.tsx
angeloashmore: Perfect, good idea.

PR #1322, opened 2024-03-14
test: remove Cypress

angeloashmore:
## Context

DT-1875

## The Solution

Remove Cypress!

A few Cypress-related entries were removed from the root `.gitignore`. Although some local projects may contain the previously ignored files, we can easily check that they are not merged into `main` during reviews.

## Impact / Dependencies

All Cypress tests have been replaced by Playwright tests.

In the future, all Vitest tests that use React may be replaced by Playwright tests. In many cases, those tests already exist in our Playwright suite.

- [ ] The CI is successful.


.github/workflows/ci.yml
```diff
@@ -228,38 +228,3 @@ jobs:
           name: html-report--attempt-${{ github.run_attempt }}
           path: playwright/playwright-report
           retention-days: 14
-
-  e2e-cypress:
```
xrutayisire, line 232: I can see one reference to Cypress that should be remove in `.github/actions/restore-cache/action.yml` for ` ~/.cache/Cypress`.

package.json
```diff
@@ -48,10 +42,6 @@
     "@types/semver": "7.3.13",
     "chalk": "4.1.2",
     "concurrently": "7.6.0",
-    "cypress": "12.8.0",
```
xrutayisire, line 51: Put the comment here to have the nice comment PR review management 😁

I can see one `data-cy` left, I think it was because of bad merge timing management. ` data-cy="push-changes"` in `packages/slice-machine/src/features/changes/PushChangesButton.tsx`. I don't even think it's used so you could remove it.

xrutayisire (approved):
Added 2 comments that we should handle before merging.

Otherwise, it's sooo coool! Finally, no Cypress anymore and a much better management with Playwright! 

So happy to see this PR happening ✨🥳

.github/workflows/ci.yml
angeloashmore: Ooh, good catch!

package.json
angeloashmore: Removed!

PR #1324, opened 2024-03-18
chore: track adding and editing fields

angeloashmore:
## Context

Completes DT-2013

## The Solution

- Add new tracking event types: `field:added`, `field:settings-opened`.
- Send `field:added` when a new field is added.
- Send `field:settings-opened` when a field's settings are opened.

## Impact / Dependencies

These events will help us understand which fields are added and how often the defaults settings are changed.

- [x] The CI is successful.

### DIFF
--- a/packages/manager/src/managers/telemetry/types.ts
+++ b/packages/manager/src/managers/telemetry/types.ts
@@ -13,8 +13,9 @@ export const SegmentEventType = {
 	pageView: "page-view",
 	users_invite_button_clicked: "users-invite-button-clicked",
 	openVideoTutorials: "open-video-tutorials",
+	field_added: "field:added",
+	field_settingsOpened: "field:settings-opened",
 	customType_created: "custom-type:created",
-	customType_fieldAdded: "custom-type:field-added",
 	customType_sliceZoneUpdated: "custom-type:slice-zone-updated",
 	customType_openAddFromTemplates: "custom-type:open-add-from-templates",
 	customType_saved: "custom-type:saved",
@@ -47,9 +48,9 @@ export const HumanSegmentEventType = {
 	[SegmentEventType.users_invite_button_clicked]:
 		"Slice Machine Users Invite Button Clicked",
 	[SegmentEventType.openVideoTutorials]: "SliceMachine Open Video Tutorials",
+	[SegmentEventType.field_added]: "SliceMachine Field Added",
+	[SegmentEventType.field_settingsOpened]: "SliceMachine Field Settings Opened",
 	[SegmentEventType.customType_created]: "SliceMachine Custom Type Created",
-	[SegmentEventType.customType_fieldAdded]:
-		"SliceMachine Custom Type Field Added",
 	[SegmentEventType.customType_sliceZoneUpdated]:
 		"SliceMachine Slicezone Updated",
 	[SegmentEventType.customType_openAddFromTemplates]:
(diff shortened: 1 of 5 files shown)


packages/manager/src/managers/telemetry/types.ts
```diff
@@ -170,24 +171,36 @@ type OpenVideoTutorialsSegmentEvent = SegmentEvent<
...
-	typeof SegmentEventType.customType_created,
+type FieldAddedSegmentEvent = SegmentEvent<
+	typeof SegmentEventType.field_added,
 	{
 		id: string;
 		name: string;
-		format: CustomTypeFormat;
-		type: "repeatable" | "single";
-		origin: "onboarding" | "table";
+		type: string;
```
bapmrl: 💡 Would it be better to use a narrower string literal type, if it exists?

packages/manager/src/managers/telemetry/types.ts
```diff
@@ -170,24 +171,36 @@ type OpenVideoTutorialsSegmentEvent = SegmentEvent<
...
+		isInAGroup: boolean;
+		contentType: "page type" | "custom type" | "slice";
```
bapmrl, line 183: ❓ Akin to `CustomTypeFormat`, would a dedicated `ContentType` would be useful somewhere else? If not, I think it's fine to keep this locally defined string literal type.

packages/slice-machine/lib/models/common/widgets/Group/ListItem/index.jsx
```diff
@@ -141,9 +151,20 @@ const CustomListItem = ({
     setCustomType(newCustomType);
   };
 
+  /** @param {[string, import("@prismicio/types-internal/lib/customtypes").NestableWidget]} field */
```
bapmrl, line 154: ❓ What's this comment used for?

packages/slice-machine/lib/builders/common/Zone/index.jsx
```diff
@@ -48,9 +50,26 @@ const Zone = ({
   const [selectModalData, setSelectModalData] = useState({ isOpen: false });
   const [newFieldData, setNewFieldData] = useState(null);
 
+  /** @param {[string, import("@prismicio/types-internal/lib/customtypes").NestableWidget]} field */
```
bapmrl, line 53: ❓ What's this comment used for?

packages/slice-machine/lib/builders/common/Zone/index.jsx
```diff
@@ -48,9 +50,26 @@ const Zone = ({
...
+      name: model.config.label,
+      type: model.type,
+      isInAGroup: false,
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+      contentType:
+        zoneType === "customType"
+          ? zoneTypeFormat === "page"
+            ? "page type"
+            : "custom type"
+          : zoneType,
```
bapmrl: 💡 As this is JavaScript code, I think putting `slice` here would be clearer, no?

bapmrl (approved):
LGTM!

packages/manager/src/managers/telemetry/types.ts
angeloashmore: Good idea! I found a `FieldType` export from `@prismicio/types-internal` that fits here. It has a few extra values (e.g. `"Slice"`, `"Choice"`), but that won't affect us here.

packages/manager/src/managers/telemetry/types.ts
angeloashmore: I don't think we would use a dedicated `ContentType` anywhere else, and an enum feels like over-engineering. Let's keep it as a string literal union, and we can always introduce that type later if we find another need for it.

packages/slice-machine/lib/models/common/widgets/Group/ListItem/index.jsx
angeloashmore: ESLint complains about using `any` values. I had to type the function parameter using JSDoc in order to use the `field` argument.

See the following screenshot without the JSDoc annotation:

image

I figured the type annotation was better than multiple `eslint-disable-next-line` comments. WDYT?

packages/slice-machine/lib/builders/common/Zone/index.jsx
angeloashmore: See https://github.com/prismicio/slice-machine/pull/1324#discussion_r1531211305.

packages/slice-machine/lib/builders/common/Zone/index.jsx
angeloashmore: Sure, let's do it. It's safer. 🙂

packages/slice-machine/lib/models/common/widgets/Group/ListItem/index.jsx
bapmrl: Ok, nice ^^

PR #1326, opened 2024-03-28
fix(adapter-next): prevent `revalidatePath` error in the slice simulator

angeloashmore:
## Context

The Next.js slice simulator can throw an error when used in a React Server Component:

```
Uncaught (in promise): Error: Invariant: static generation store missing in revalidateTag _N_T_/slice-simulator
```

This PR prevents that error.

## The Solution

Refactor `<SliceSimulator>` to properly pass the `revalidatePath` function as a prop. Passing as a prop seems to prevent the error.

The refactor allowed for removing duplicate code.

## Impact / Dependencies

Existing projects with the Server Components simulator boilerplate do not need to change their code. The Next.js simulator's API did not change as of this PR.

```tsx
// src/app/slice-simulator/page.tsx

import {
  SliceSimulator,
  SliceSimulatorParams,
  getSlices,
} from "@slicemachine/adapter-next/simulator";
import { SliceZone } from "@prismicio/react";

import { components } from "@/slices";

export default function SliceSimulatorPage({
  searchParams,
}: SliceSimulatorParams) {
  const slices = getSlices(searchParams.state);

  return (
    <SliceSimulator>
      <SliceZone slices={slices} components={components} />
    </SliceSimulator>
  );
}
```

Likewise, simulators in the Pages Router are unaffected.

- [ ] The CI is successful.


angeloashmore:
The correct fix is to update Next.js to the latest version. Next.js `v14.1.4` is confirmed to work.

PR #1327, opened 2024-03-29
chore: add `play` script

angeloashmore:
## Context

We often need a running Slice Machine instance complete with a Prismic repository. We have projects in the `e2e-projects` directory to serve that purpose, but they have the following limitations:

- Tied to a specific Prismic repository.
- Require access to the Prismic repository.
- Are version controlled.
- Are stateful since they are paired to a real Prismic repository.

## The Solution

We can use a script to quickly generate new projects with their own Prismic repositories. Projects are generated using Prismic's minimal starters, just like customers would use.

These projects are called "playgrounds" since they allow us to play with Slice Machine while developing.

This PR introduces a new `yarn play` command. The CLI looks like this:

```sh
# Start the most recent playground or create one if none exist.
yarn play

# Start an existing playground or create one if it doesn't exist.
yarn play my-playground-name

# Force create a new playground
yarn play --new
yarn play --new my-playground-name

# Specify a framework (default: next)
yarn play -f sveltekit
yarn play --framework sveltekit

# If `-f` is passed with a playground name and the playground
# already exists, halt. If the playground does not exist, create
# a new one with that framework.
yarn play -f sveltekit my-playground-name

# Don't start the project
yarn play --no-start

# Dry run
yarn play -n
yarn play --dry-run
```

Both Slice Machine and the playground project are started automatically (unless disabled via `--no-start`).

## Impact / Dependencies

This should make it easier to develop features. It could also streamline how we run end-to-end tests with Playwright.

- [x] The CI is successful.


scripts/utils/commandUtils.ts
```diff
@@ -30,8 +30,13 @@ export function exec(
 export async function exec(
   file: string,
   args?: string[],
-  options?: ExecaOptions & { dryRun: boolean },
+  options?: ExecaOptions & { dryRun?: true },
```
bapmrl: ⚠️ This overload can be simplified. Indeed, when `dryRun` is `true`, the function will return `undefined`. It means we could write instead:

```typescript
export function exec(
  file: string,
  args?: string[],
  options?: ExecaOptions & { dryRun: true },
): void;
```

I've checked all cases with the following assignments:
```typescript
const a = exec("file", [], {}); // ExecaChildProcess
const b = exec("file", [], { dryRun: false }); // ExecaChildProcess
const c = exec("file", [], { dryRun: true }); // void
const d = exec("file", [], { dryRun: false as boolean }); // Promise<void | ExecaReturnValue>
const e = exec("file", [], { dryRun: true as boolean }); // Promise<void | ExecaReturnValue>
```

scripts/play.ts
```diff
@@ -0,0 +1,380 @@
+import mri from "mri";
+import chalk from "chalk";
+import * as fs from "node:fs/promises";
+import * as fsSync from "node:fs";
+import * as path from "node:path";
+import * as crypto from "node:crypto";
+
+import { exec } from "./utils/commandUtils";
+
+const PLAYGROUNDS_ROOT = path.join(process.cwd(), "playgrounds");
```
bapmrl: ❓ Would [this](https://stackoverflow.com/a/50052194) more modern approach work?

💡 It seems "root" refers to both the `playgrounds` parent directory and the directory of a bootstrapped project. Would renaming `PLAYGROUNDS_ROOT` to `PLAYGROUNDS_DIR` help better differentiate the 2 concepts?

scripts/play.ts
```diff
@@ -0,0 +1,380 @@
...
+
+type SliceMachineConfig = {
+  localSliceSimulatorURL?: string;
+  apiEndpoint?: string;
+};
+
+/**
+ * The root command.
+ */
+async function run(): Promise<void> {
```
bapmrl: 💡 `run` is a pretty generic name for a function. Would `play` be better?

scripts/play.ts
```diff
@@ -0,0 +1,380 @@
...
+const PLAYGROUNDS_ROOT = path.join(process.cwd(), "playgrounds");
+const DEFAULT_FRAMEWORK = "next" satisfies Args["framework"];
+
+run();
+
+type Args = {
+  "dry-run": boolean;
+  start: boolean;
+  new: boolean;
+  framework?: "next" | "nuxt" | "sveltekit" | (string & {});
```
bapmrl: ❓ What does this `string & {}` type mean?

scripts/play.ts
```diff
@@ -0,0 +1,380 @@
...
+
+  const playgroundRoot = path.join(PLAYGROUNDS_ROOT, playgroundName);
+  const playgroundExists = await pathExists(playgroundRoot);
+
+  if (playgroundExists) {
+    if (args.new) {
+      console.error(
+        chalk.red`A playground named ${playgroundName} already exists. Exiting.`,
+      );
+      process.exit(1);
```
bapmrl: ⚠️ I've read in several places (like [this one](https://kostasbariotis.com/why-you-should-not-use-process-exit)) that this is not a good Node.js practice. That's why in the `publish` script I'm throwing `CommandError`s and wrapping the function body between `process.on("uncaughtException", handleUncaughtException)` and `process.off("uncaughtException", handleUncaughtException)`.

scripts/play.ts
```diff
@@ -0,0 +1,380 @@
...
+    }
+
+    if (didProvidePlaygroundName) {
+      console.info(`Using playground: ${chalk.blue(playgroundName)}`);
```
bapmrl: 💡 `blue` is already being used by the dry run mode. Maybe it would be better to choose another color?

scripts/play.ts
```diff
@@ -0,0 +1,380 @@
...
+  },
+) {
+  if (options?.dryRun) {
+    console.info(`Update ${filePath} with ${JSON.stringify(contents)}`);
+
+    return;
+  }
+
+  const existingContents = JSON.parse(await fs.readFile(filePath, "utf8"));
+  const newContents = deepMerge(existingContents, contents);
```
bapmrl, line 536: ❓ Can't we just use `structuredClone`?

scripts/play.ts
```diff
@@ -0,0 +1,380 @@
...
+type Args = {
+  "dry-run": boolean;
+  start: boolean;
+  new: boolean;
+  framework?: "next" | "nuxt" | "sveltekit" | (string & {});
+};
+
+type PackageJSON = {
+  name: string;
+  scripts: Record<string, string>;
```
bapmrl: 💡 I think `Partial<Record<string, string>>` would be more correct as not all keys will have a value.

scripts/play.ts
```diff
@@ -0,0 +1,380 @@
...
+    "semla",
+    "strudel",
+    "taiyaki",
+    "toast",
+  ];
+
+  const adjective = ADJECTIVES[Math.floor(Math.random() * ADJECTIVES.length)];
+  const pastry = PASTRIES[Math.floor(Math.random() * PASTRIES.length)];
+
+  return appendRandomHash(`${adjective}-${pastry}`);
```
bapmrl: 💡 What would you think of ``return `${adjective}-${pastry}-${generateRandomHash()}`;``? That way, the dashes (/format) would only be specified in this `createRandomName` function.

scripts/play.ts
```diff
@@ -0,0 +1,380 @@
...
+    dryRun: options.dryRun,
+  });
+
+  await updateJSONFile<PackageJSON>(
+    path.join(root, "package.json"),
+    {
+      name,
+      scripts: {
+        dev: 'concurrently --prefix-colors auto "yarn:dev:website" "yarn:dev:slicemachine"',
+        "dev:slicemachine": `SM_ENV=staging ../../packages/start-slicemachine/bin/start-slicemachine.js`,
```
lihbr: Can we use [`cross-env`](https://www.npmjs.com/package/cross-env) here? Could be added alongside `concurrently` above?

scripts/play.ts
```diff
@@ -0,0 +1,380 @@
...
+    path.join(root, "slicemachine.config.json"),
+    {
+      localSliceSimulatorURL: "http://localhost:8001/slice-simulator",
+      apiEndpoint: `https://${name}.cdn.wroom.io/api/v2`,
+    },
+    { dryRun: options.dryRun },
+  );
+
+  await exec(
+    "npx",
```
bapmrl: ⚠️ We're using Yarn in this monorepo ^^ I don't think we should mix package managers. Would `yarn dlx` work in your case?

playgrounds/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "packageManager": "yarn@3.6.0",
+  "private": true,
+  "workspaces": [
+    "*"
+  ],
+  "resolutions": {
```
bapmrl, line 7: ❓ Why do we need all these `resolutions`?

playgrounds/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "packageManager": "yarn@3.6.0",
+  "private": true,
+  "workspaces": [
+    "*"
```
bapmrl, line 5: ❓ Couldn't we just add `playgrounds/*` to the root `package.json` file? I'm sure you've found issues I'm not aware of ^^ Maybe this is to isolate changes we don't want to push to Git?

bapmrl (commented):
Quick incomplete review as this is still a draft ^^ Thanks for starting this highly useful work!

scripts/play.ts
angeloashmore: Yep, good idea!

scripts/play.ts
angeloashmore: > ❓ Would [this](https://stackoverflow.com/a/50052194) more modern approach work?

It's slightly different in that `process.cwd()` is the working directory, while `import.meta.url` is the file's URL. The former should decouple the script's location from its internal logic, while `import.meta.url` forces the script to know its location.

In this case, `import.meta.url` will work since we know the script will always live in `scripts`. It will let us replace `path.join` with something more platform-agnostic.

Edit: Unless you meant to just replace `path.join` with `new URL()`, in which case, yes! In either case, `import.meta.url` might make the script more predictable, so I'll change to that.

> 💡 It seems "root" refers to both the `playgrounds` parent directory and the directory of a bootstrapped project. Would renaming `PLAYGROUNDS_ROOT` to `PLAYGROUNDS_DIR` help better differentiate the 2 concepts?

Yes, good call. I was getting confused myself. `root` refers to the playground's directory, and it's already scoped under function names like `createPlayground`, so I think the name `dir` is clear enough. What do you think?

scripts/play.ts
angeloashmore: I figured `run` make sense here since this is a script. `main` could be another option.

Names like `run` and `main` differentiate itself from the script's domain logic, like `createPlayground`. `play` could be confused with something underlying logic.

To me, `run` (or `main`) a clear way to declare "this is the top-level function that runs the script."

I don't feel too strongly about it, so if you think `play` is still a better option, I will change it. 🙂

scripts/play.ts
angeloashmore: It means "any string." The `{}` is kind of [a TypeScript hack](https://github.com/Microsoft/TypeScript/issues/29729#issuecomment-567871939) to ensure `string` doesn't clobber the other string literal union type.

```ts
type WithoutHack = "next" | "nuxt" | "sveltekit" | string
// => string

type WithHack = "next" | "nuxt" | "sveltekit" | (string & {})
// => "next" or "nuxt" or "sveltekit" or any string
```

We need the "or any string" part of this type since users can pass in any string. Literal types are still useful for autocomplete, however.

scripts/play.ts
angeloashmore: Good idea! I hadn't thought of that issue, but it makes sense. I like the solution you came up with.

scripts/play.ts
angeloashmore: Changed to green.

scripts/play.ts
angeloashmore: I opted for a deep merge function since I would like to provide partial values to `updateJSONFile`. A deep merge is necessary if the partial value replaces an object, like `scripts`.

`structuredClone`, from my understanding, would simply copy a given value without merging capabilities. Is that correct?

We could probably simplify the `updatePackageJSON` and `updateSliceMachineConfig` functions (wrappers around `updateJSONFile` introduced after your review) since we know exactly which properties we want to edit. For example, we could manually merge `scripts`. What do you think?

scripts/play.ts
angeloashmore: Good point. We don't access the scripts, but your version is technically more correct.

scripts/play.ts
angeloashmore: Good idea! 🙂

scripts/play.ts
angeloashmore: It does work, but it takes noticeably longer since `yarn dlx` doesn't cache its installations like `npx` does. It's not too bad though. Let's go with `yarn dlx` and if we get annoyed with the creation time, we have ways to optimize it.

playgrounds/package.json
angeloashmore: We need to link to the local development Slice Machine packages. Without the resolutions, the published versions are used.

Originally, this script linked the packages in each playground's `package.json`, but this version lets us declare it once and have it automatically apply to each playground.

playgrounds/package.json
angeloashmore: That's what I tried first, and it works well, except that installing playground dependencies affects the root-level `yarn.lock`. Playgrounds are temporary and user-specific, thus they shouldn't be committed.

Similarly, I tried adding `workspaces: ["../packages"]` to the `playgrounds/package.json`, but there were several version conflicts. I wasn't sure how to work around that and figured it's simpler to use `resolutions` + `link:`.

Unfortunately, Yarn does not link `bin` entries when using `link:`. That's why we need to override the `slicemachine` playground script:

```sh
cross-env SM_ENV=staging ../../packages/start-slicemachine/bin/start-slicemachine.js
```

scripts/play.ts
```diff
@@ -0,0 +1,402 @@
...
+      "--quiet",
+      "@slicemachine/init@latest",
+      `--repository="${name}"`,
+      "--no-start-slicemachine",
+    ],
+    {
+      cwd: dir,
+      env: {
+        // Use wroom.io
+        SM_ENV: "staging",
```
xrutayisire: Should we be able to run the play script with a prod repo? I feel it can still be useful to test, WDYT?

scripts/play.ts
```diff
@@ -0,0 +1,402 @@
...
+        "https://github.com/prismicio-community/sveltekit-starter-prismic-minimal.git",
+        dir,
+        { dryRun: options.dryRun },
+      );
+
+      break;
+    }
+
+    default: {
+      console.error(chalk.red`Framework not supported: ${options.framework}`);
```
xrutayisire: We don't support Nuxt today with this script? What do we miss?

scripts/play.ts
angeloashmore: Yep, you can now. I figured we could add that functionality in later, but it wasn't hard to add now. 🙂

```sh
yarn play --new --stage=prod
```

Playgrounds are paired to their stage since the repository only lives on that stage. Something like this is invalid:

```sh
yarn play --stage=prod my-existing-playground
```

scripts/play.ts
angeloashmore: We do now!

```sh
yarn play --new -f nuxt
```

scripts/play.ts
bapmrl: >In either case, import.meta.url might make the script more predictable, so I'll change to that.

I think it will make it more predictable too.

>I think the name dir is clear enough. What do you think?

I agree 🙂

scripts/play.ts
bapmrl: Neither do I feel strongly about it ^^ But I prefer `main` other `run`. It feels more usual if you want to convey the idea it's the main entry point.

scripts/play.ts
bapmrl: >structuredClone, from my understanding, would simply copy a given value without merging capabilities. Is that correct?

You're right, sorry ^^ I think it's great as you implemented it.

playgrounds/package.json
bapmrl: Okay, thanks for the explanation ^^ It makes sense!

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+
+import {
+  CommandError,
+  exec,
+  handleUncaughtException,
+} from "./utils/commandUtils";
+
+const PLAYGROUNDS_ROOT = new URL("../playgrounds/", import.meta.url);
+const DEFAULT_FRAMEWORK = "next" satisfies Args["framework"];
+const DEFAULT_STAGE = "staging" satisfies Args["stage"];
```
bapmrl: 💡 Would `environment` be more usual than `stage`? It would also recall the `SM_ENV` environment variable.

playgrounds/package.json
```diff
@@ -0,0 +1,18 @@
...
+  ],
+  "resolutions": {
+    "@slicemachine/adapter-next": "link:../packages/adapter-next",
```
bapmrl, line 8: 💡 Would the [portal protocol](https://yarnpkg.com/protocol/portal) be a better fit?

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+    if (args.framework) {
+      console.warn(
+        chalk.yellow`The --framework/-f argument is ignored when a playground already exists.`,
+      );
+    }
+  } else {
+    console.log(`Creating a new playground: ${chalk.green(playgroundName)}`);
+
+    await createPlayground(playgroundName, playgroundDir, {
+      framework: args.framework ?? DEFAULT_FRAMEWORK,
```
bapmrl: 💡 `DEFAULT_FRAMEWORK` could be used the same way as `DEFAULT_STAGE`. It means `framework` could be required in the `Args` type.

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+  // Ensure dependencies are isolated from the monorepo.
+  if (!(await pathExists(new URL("./yarn.lock", PLAYGROUNDS_ROOT)))) {
+    await fs.writeFile(new URL("./yarn.lock", PLAYGROUNDS_ROOT), "");
+  }
+
+  const args = mri<Args>(process.argv.slice(2), {
+    boolean: ["help", "dry-run", "start", "new"],
+    string: ["framework", "stage"],
+    alias: { h: "help", n: "dry-run", f: "framework" },
+    default: {
```
bapmrl, line 72: ⚠️ `help` should have a `false` default value to satisfy the `Args` type.

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+  "../../packages/start-slicemachine/bin/start-slicemachine.js";
+
+run();
+
+type Args = {
+  "dry-run": boolean;
+  start: boolean;
+  new: boolean;
+  framework?: "next" | "nuxt" | "sveltekit";
+  stage: "staging" | "prod" | "development";
```
bapmrl: 💡 Should all stages use their full name for the sake of consistency? It means `prod` would become `production`.

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+
+    return;
+  }
+
+  if (!["prod", "staging", "development"].includes(args.stage)) {
+    throw new CommandError(`Unsupported stage: ${args.stage}`);
+  }
+
+  if (
+    args.framework &&
```
bapmrl: 💡 This `args.framework &&` expression wouldn't be necessary if `framework` was specified as a required argument in the `mri` call.

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+
+  let [playgroundName] = args._;
+  const didProvidePlaygroundName = Boolean(playgroundName);
+
+  if (!playgroundName) {
+    if (args.new) {
+      playgroundName = createRandomName();
+    } else {
+      playgroundName =
+        (await getLastModifiedEntry(PLAYGROUNDS_ROOT, {
```
bapmrl: 💡 I think an intermediary `const lastModifiedEntryName` would help readability as this is a quite complex expression.

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+    playgroundName = `play-${playgroundName}`;
+  }
+
+  const playgroundDir = new URL(`./${playgroundName}/`, PLAYGROUNDS_ROOT);
+  const playgroundExists = await pathExists(playgroundDir);
+
+  if (playgroundExists) {
+    if (args.new) {
+      throw new CommandError(
+        `A playground named ${playgroundName} already exists. Exiting.`,
```
bapmrl: 💡 I think it's hard to read the word `Exiting` just after `exists`. I would remove it as the error will already be quite visible (in red).

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+        `Reusing most recent playground: ${chalk.green(playgroundName)}`,
+      );
+    }
+
+    console.info('  (Use "--new" to create a new playground)');
+    console.info();
+
+    if (args.framework) {
+      console.warn(
+        chalk.yellow`The --framework/-f argument is ignored when a playground already exists.`,
```
bapmrl, line 154: 💡 I guess this is also applicable to the `stage` argument?

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+
+    if (didProvidePlaygroundName) {
+      console.info(`Using playground: ${chalk.green(playgroundName)}`);
+    } else {
+      console.info(
+        `Reusing most recent playground: ${chalk.green(playgroundName)}`,
+      );
+    }
+
+    console.info('  (Use "--new" to create a new playground)');
```
bapmrl, line 149: 💡 Should this message only be displayed if the `new` argument is `false`?

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+
+    /**
+     * The framework used to bootstrap the playground.
+     *
+     * @defaultValue `"next"`
+     */
+    framework?: Args["framework"];
+  } = {},
+) {
+  switch (options.framework ?? "next") {
```
bapmrl: 💡 I think all fallbacks should use the constants you defined (like `DEFAULT_FRAMEWORK`).

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+      break;
+    }
+
+    default: {
+      throw new CommandError(`Unsupported framework: ${options.framework}`);
+    }
+  }
+
+  if (!options.dryRun) {
+    await fs.rm(new URL("./.git", dir), { recursive: true });
```
bapmrl: ❓ Why do we need to do this?

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+      },
+      { dryRun: options.dryRun },
+    );
+  }
+
+  // Update Slice Machine configuration with the correct API endpoint (if needed).
+  if (options.stage === "staging") {
+    await updateSliceMachineConfig(
+      dir,
+      { apiEndpoint: `https://${name}.cdn.wroom.io/api/v2` },
```
bapmrl, line 304: ❓ Shouldn't we specify the `repositoryName` parameter instead and let the Slice Machine manager adapt its `API_ENDPOINTS` (i.e., as specified in `packages/manager/src/constants/API_ENDPOINTS.ts`) depending on the `SM_ENV` environment variable?

scripts/play.ts
bapmrl: ❓ Are we still using `npx`? ^^

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+      dir,
+      { apiEndpoint: apiEndpoint.toString() },
+      { dryRun: options.dryRun },
+    );
+  }
+
+  await exec(
+    "npx",
+    [
+      "@slicemachine/init@latest",
```
bapmrl: ❓ Should we use `@slicemachine/init`'s source code as defined in the monorepo instead of the latest version published on npm?

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+      },
+      stdio: "inherit",
+      dryRun: options.dryRun,
+    },
+  );
+
+  // The Nuxt adapter edits the `endpoint` option as part of `@slicemachine/init`.
+  // We need to change it to the API endpoint to support different stages.
+  if (options.framework === "nuxt") {
+    const sliceMachineConfig: SliceMachineConfig = JSON.parse(
```
bapmrl, line 340: 💡 Instead of reading the `slicemachine.config.json` file, can't you reuse the `apiEndpoint` variable defined earlier?

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+): Promise<void> {
+  await exec("git", ["clone", "--depth=1", url, fileURLToPath(dir)], {
+    stdio: "inherit",
+    dryRun: options.dryRun,
+  });
+}
+
+/**
+ * Returns a random project name. The name is not guaranteed to be unique. Use
+ * `appendRandomHash` to virtually guarantee uniqueness.
```
bapmrl: ⚠️ The `appendRandomHash` function no longer exists.

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+  return `${adjective}-${pastry}-${hash}`;
+}
+
+/**
+ * Returns the name of the last modified file or directory in a given directory.
+ *
+ * @param dir - The directory to search.
+ * @param options - Options that determine the function's behavior.
+ */
+async function getLastModifiedEntry(
```
bapmrl: 💡 Would `getLastModifiedEntryName` be a better name for this function?

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+    return;
+  }
+
+  const existingContents = JSON.parse(await fs.readFile(filePath, "utf8"));
+  const newContents = deepMerge(existingContents, contents);
+
+  await fs.writeFile(filePath, JSON.stringify(newContents, null, 2));
+}
+
+function deepMerge(a: any, b: any): Record<PropertyKey, unknown> {
```
bapmrl: 💡 Would this compile with `unknown` instead of `any`? Just to avoid using `any` ^^

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+ * @param contents - Contents to add to the JSON file.
+ * @param options - Options that determine the function's behavior.
+ */
+async function updateJSONFile<TSchema extends Record<string, unknown>>(
+  filePath: URL,
+  contents: Partial<TSchema>,
+  options: DryRunOption = {},
+) {
+  if (options.dryRun) {
+    console.info(`Update ${filePath} with ${JSON.stringify(contents)}`);
```
bapmrl, line 530: 💡 I think it would be nicer to use `chalk.blue` for consistency with `exec` ^^

scripts/play.ts
```diff
@@ -0,0 +1,550 @@
...
+      stage: args.stage,
+      dryRun: args["dry-run"],
+    });
+  }
+
+  if (args.start) {
+    await exec("yarn", ["dev"], {
+      cwd: playgroundDir,
+      env: {
+        CONCURRENTLY_PREFIX: `${chalk.white.dim(playgroundName)} [{name}]`,
```
bapmrl: ⚠️ This is not easily readable when using macOS in light mode.

bapmrl (changes requested):
This is a really nice advancement to easily work on Slice Machine! Thank you Angelo 😊

⚠️ I just had the following error when trying to access `localhost:9999` in Chrome:
```
play-tall-semla-02707b3 [slicemachine] 
play-tall-semla-02707b3 [slicemachine] > play-tall-semla-02707b3@0.1.0 slicemachine
play-tall-semla-02707b3 [slicemachine] > ../../packages/start-slicemachine/bin/start-slicemachine.js
play-tall-semla-02707b3 [slicemachine] 
play-tall-semla-02707b3 [next] 
play-tall-semla-02707b3 [next] > play-tall-semla-02707b3@0.1.0 next:dev
play-tall-semla-02707b3 [next] > next dev --port=8001
play-tall-semla-02707b3 [next] 
play-tall-semla-02707b3 [next]    ▲ Next.js 14.1.4
play-tall-semla-02707b3 [next]    - Local:        http://localhost:8001
play-tall-semla-02707b3 [next] 
play-tall-semla-02707b3 [slicemachine] [HPM] Proxy created: /  -> http://localhost:3000
play-tall-semla-02707b3 [slicemachine] 
play-tall-semla-02707b3 [slicemachine]  Slice Machine v1.26.0  → Running at http://localhost:9999
play-tall-semla-02707b3 [slicemachine] 
play-tall-semla-02707b3 [next]  ✓ Ready in 1431ms
play-tall-semla-02707b3 [slicemachine] [HPM] Error occurred while proxying request localhost:9999/ to http://localhost:3000/ [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)
play-tall-semla-02707b3 [slicemachine] [HPM] Error occurred while proxying request localhost:9999/favicon.ico to http://localhost:3000/ [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)
play-tall-semla-02707b3 [next]  ○ Compiling / ...
play-tall-semla-02707b3 [next]  ✓ Compiled / in 1237ms (854 modules)
play-tall-semla-02707b3 [next]  ○ Compiling /icon.png ...
play-tall-semla-02707b3 [next]  ✓ Compiled /icon.png in 509ms (541 modules)
play-tall-semla-02707b3 [slicemachine] [HPM] Error occurred while proxying request localhost:9999/ to http://localhost:3000/ [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)
play-tall-semla-02707b3 [slicemachine] [HPM] Error occurred while proxying request localhost:9999/favicon.ico to http://localhost:3000/ [ECONNREFUSED] (https://nodejs.org/api/errors.html#errors_common_system_errors)
```

package.json
bapmrl, line 1: ⚠️ Is there a way to add the generated playgrounds to the `clean` script?

scripts/play.ts
angeloashmore: Let's go with `main` then. 🙂

scripts/play.ts
angeloashmore: I had to change it back because `yarn dlx` was running into a dependency issue with `readable-web-to-node-stream`, a dependency of `file-type`, which we use in `@slicemachine/manager`. It's weird though: I didn't see it when I first tried it.

I reverted back to `npx` because it works reliably, but we will need to find a proper fix.

scripts/play.ts
angeloashmore: Yes, great callout. It's renamed to `--environment`/`-e`.

playgrounds/package.json
angeloashmore: It doesn't seem like we can, at least not without other changes. Using `portal:` causes version mismatch issues:

```
➤ YN0000: ┌ Link step
➤ YN0071: │ Cannot link @slicemachine/manager into start-slicemachine@portal:../packages/start-slicemachine::locator=root-workspace-0b6124%40workspace%3A. dependency @prismicio/mocks@npm:2.0.0 conflicts with parent dependency @prismicio/mocks@npm:2.1.1
➤ YN0071: │ Cannot link @slicemachine/manager into slice-machine-ui@portal:../packages/slice-machine::locator=root-workspace-0b6124%40workspace%3A. dependency @prismicio/mocks@npm:2.0.0 conflicts with dependency @prismicio/mocks@npm:2.1.1 from sibling portal start-slicemachine
```

I'm not 100% sure what the cause is. I don't think keeping all of our dependencies on the same version is the right solution since playgrounds should be able to use any version.

scripts/play.ts
angeloashmore: Done! It required changing some logic since the script prints a message if a `-f` flag is provided and not used. Before providing a default value to `mri`, we could check if `args.framework` contained a value. Now we have a separate helper to check if a flag was provided:

```ts
/**
 * Determines if a CLI argument was provided.
 *
 * @param key - The key or an array of keys to search for.
 * @param args - An array of arguments to search.
 */
function didProvideArgument(key: string | string[], args: string[]): boolean {
  return args.some((arg) =>
    [key].flat().some((key) => new RegExp(`^${key}=?`).test(arg)),
  );
}
```

scripts/play.ts
angeloashmore: Thanks, good catch. I wish `mri` had better type support.

scripts/play.ts
angeloashmore: Yes, we can make that change. I figured `prod` was quicker to type, but let's go for consistency here instead. Providing an environment will be rare, I think.

scripts/play.ts
angeloashmore: Removed now that `framework` has a default value.

scripts/play.ts
angeloashmore: Done! 👍 I named it `mostRecentPlaygroundName` to better represent what the value is.

scripts/play.ts
angeloashmore: Good call. 👍

scripts/play.ts
angeloashmore: Yes, there is now an error for `--environment`/`-e`.

scripts/play.ts
angeloashmore: This line should only be reached if `--new` is false since we throw above if `--new` is true: https://github.com/prismicio/slice-machine/blob/bfb737b97ecaa25ad0f3797b5416a2c39f9036f4/scripts/play.ts#L117-L121

scripts/play.ts
angeloashmore: Thanks, everything uses the constants now. I also added a `DEFAULT_WROOM_URL` constant as a fallback for the `development` environment's API endpoint.

scripts/play.ts
angeloashmore: 😮 It looks like we don't need to! I thought a Git repo within another would cause issues if it wasn't marked as a submodule, but it seems to work okay. This is better since it lets us track changes and quickly revert a playground.

We still need to remove the `origin` remote since we don't want the starters to be changed.

Do you think we should remove all Git history and initialize a new repo?

scripts/play.ts
angeloashmore: We need to add an `apiEndpoint` property for the website to call the correct domain. Slice Machine should already handle it automatically, but starters assume `prismic.io` is used unless a full endpoint is provided.

scripts/play.ts
angeloashmore: I felt it was safer to use the published known-working version rather than the development build. It's possible we could break `@slicemachine/init` locally and not be able to start a new playground.

If we want to test the local version of `@slicemachine/init`, we can call it directly somewhere with more control rather than through `yarn play`.

What do you think?

scripts/play.ts
angeloashmore: I felt this was more contained. I had what you suggested before, and it required a `let apiEndpoint` above, exposing some logic contained in an `if` statement to the outside context. It felt messy to me.

I think we could also change the Nuxt adapter to not overwrite the `endpoint` property as well, allowing us to remove this chunk of code. @lihbr, I'll probably need your help with this. 🙂

scripts/play.ts
angeloashmore: Thanks, changed to this description:

```
/**
 * Returns a random project name. A random 4-character hash is appended to the
 * name to prevent name collisions.
 */
```

scripts/play.ts
angeloashmore: Yes! It's been changed.

scripts/play.ts
angeloashmore: No, it does not compile without other changes, unfortunately. We would need to do some runtime checking to ensure it is an object. Although that is technically more correct, I didn't think it was necessary for an internal script.

However, we can still change it to something slightly more specific than `any`, like `object` or `{}`. Both are not good practices, but they might be better than `any`. What do you think?

scripts/play.ts
angeloashmore: Done!

scripts/play.ts
angeloashmore: I changed it to `chalk.grey`. Could you test it again and let me know if this is better?

I would like the prefix to be a different color and of lower visual priority than the script names (`[{name}]`).

package.json
angeloashmore: We could, but I wonder if we should ask for confirmation before deleting all playgrounds. It's possible a playground could contain a lot of work or a specific set up, and I wouldn't want someone to accidentally lose it all.

angeloashmore: Maybe we could create a simple custom `scripts/clean.ts` with a prompt?

scripts/play.ts
xrutayisire: Maybe it can be an option?
I think being able to test with the source code will allow us to use the play script in the CI and so test that the init works and that all e2e tests work after an init from the source code, like users will do, WDYT?

angeloashmore:
> @bapmrl: ⚠️ I just had the following error when trying to access `localhost:9999` in Chrome…

You may not have had the Slice Machine development server running. You need to run `yarn dev` in a separate terminal in order to run the playground.

I considered running `yarn dev` into the `yarn play` script, but I wasn't sure if that added complexity was worth it. What do you think?

scripts/play.ts
bapmrl: Okay 🙂

playgrounds/package.json
bapmrl: Okay, it seems `link:` works anyway ^^ Thanks!

scripts/play.ts
bapmrl: I like what you did!

My understanding is that:
1. Each playground has a separate Git history that starts from the last commit made on the starter.
2. You then commit the configuration changes as `chore: init playground`.
3. You remove the remote to ensure the starter won't get accidentally changed.

It seems great 🙂

scripts/play.ts
bapmrl: Okay, I see. Thanks for the explanation!

scripts/play.ts
bapmrl: > I felt it was safer to use the published known-working version rather than the development build. It's possible we could break `@slicemachine/init` locally and not be able to start a new playground.
> 
> If we want to test the local version of `@slicemachine/init`, we can call it directly somewhere with more control rather than through `yarn play`.
> 
> What do you think?

I feel like the `play` script should help us work daily on Slice Machine. Hence, as a developer, I would expect it to fail if I broke the `@slicemachine/init` package. The CI and E2E tests will help us ensure it doesn't happen and if it does, we can still use the Git history to pinpoint the issue. Also, `@slicemachine/init` won't be called when playgrounds are reused so it's unlikely we would be blocked. What do you think? ^^

bapmrl: > Maybe it can be an option?

What would be the use case for the option? Do we need to be able to use the published `@slicemachine/init` package? Maybe I'm missing something regarding the E2E tests?

scripts/play.ts
bapmrl: Okay, I understand ^^ Thanks!

scripts/play.ts
bapmrl: Okay, let's keep `any` then ^^

scripts/play.ts
bapmrl: It's better now, thank you!

Screenshot 2024-04-03 at 14 58 30

package.json
bapmrl: Oh, you're right. We'll probably have changes we don't want to lose.

Consequently, I think we can either:
1. Keep it simple and not change anything.
2. Update the `clean` script so that it only deletes well known build/cache directories in the playgrounds. For example, it could delete the nested `.next` directories. Of course, it shouldn't accidentally delete content in the `playgrounds/node_modules` directory or other configuration files.

bapmrl:
> > @bapmrl: ⚠️ I just had the following error when trying to access `localhost:9999` in Chrome…
> 
> You may not have had the Slice Machine development server running. You need to run `yarn dev` in a separate terminal in order to run the playground.
> 
> I considered running `yarn dev` into the `yarn play` script, but I wasn't sure if that added complexity was worth it. What do you think?

You're right. I forgot to run `yarn dev` and doing it automatically in the `play` script would make it too complex. Let's keep it this way. It's perfect!

scripts/play.ts
```diff
@@ -0,0 +1,569 @@
...
+  const args = mri<Args>(process.argv.slice(2), {
+    boolean: ["help", "dry-run", "start", "new"],
+    string: ["framework", "environment"],
+    alias: { h: "help", n: "dry-run", f: "framework", e: "environment" },
+    default: {
+      "dry-run": false,
+      help: false,
+      start: true,
+      new: false,
+      framework: "next",
```
bapmrl: ⚠️ This could be `DEFAULT_FRAMEWORK`.

scripts/play.ts
xrutayisire: With what you said, maybe we can only use a version relative to the source code and so no need for an option.

We just need to be sure the way we are doing the init with a local version is the same as a normal init so we have a real reproduction of a prod use case.

bapmrl (approved):
❤️ This is truly excellent and will greatly help us!

I'm approving as there are just 4 comments remaining. Don't forget the little suggestion for the `exec` function ^^

scripts/utils/commandUtils.ts
angeloashmore: Thanks, you are right. 🙂

scripts/play.ts
angeloashmore: Okay, this all works for me. 🙂 We are also using the local `start-slicemachine` script, so my argument doesn't work since we could also break that script during dev.

You can see the change here: https://github.com/prismicio/slice-machine/pull/1327/commits/ce650e660cc8e8ca5de83bf3ecb0797fa5886796

package.json
angeloashmore: To expedite merging this PR, let's not change the `clean` script for now. We can revisit it in a future PR. I like the second idea, perhaps in addition to the ability to delete playgrounds.

scripts/play.ts
angeloashmore: Oops! Fixed.

PR #1329, opened 2024-04-03
refactor: Breadcrumb component

angeloashmore:
## Context

DT-2027

## The Solution

- Change `Breadcrumb`'s API to a pair of components: `Breadcrumb` and `BreadcrumbItem`.
- Convert the component to CSS Modules.

Because we don't have a robust CSS reset and our `Text` component is quite strict, styling links is difficult. This refactor *does not* take into account links, just as the previous implementation did not.

### Example

```tsx
<Breadcrumb>
  <BreadcrumbItem>Slices</BreadcrumbItem>
  <BreadcrumbItem>CallToAction</BreadcrumbItem>
  <BreadcrumbItem active>WithButton</BreadcrumbItem>
</Breadcrumb>
```

## Impact / Dependencies

Slice Machine components use `AppLayoutBreadcrumb`, a small wrapper around `Breadcrumb`. `AppLayoutBreadcrumb` was edited to expose the same API as before while using the new `Breadcrumb` API internally.

- [ ] The CI is successful.


packages/slice-machine/test/__setup__.ts
```diff
@@ -20,7 +21,11 @@ declare module "vitest" {
   }
 }
 
-const mswServer = setupServer();
+const mswServer = setupServer(
```
xrutayisire: _❓ question_

What is this change for?

packages/slice-machine/components/AppLayout/index.tsx
```diff
@@ -89,11 +89,23 @@ export const AppLayoutBackButton: FC<AppLayoutBackButtonProps> = ({
...
-  <Box flexGrow={1}>
-    <Breadcrumb {...props} />
-  </Box>
-);
+type AppLayoutBreadcrumbProps = {
+  folder: string;
+  page?: string;
+};
+
+export const AppLayoutBreadcrumb: FC<AppLayoutBreadcrumbProps> = (props) => {
```
xrutayisire: _✍️ asking for a potential change_

How I see this little wrapper today is more about the control of the style relative to the AppLayout.

I suggest we could still keep the idea to have:

```
<Box flexGrow={1}>
    <Breadcrumb {...props} />
  </Box>
```

The goal would be to only have children has props from Breadcrumb I guess, but could take other props that are extended from `Breadcrumb` component.

That way when we want to use our Breadcrumb within the app we remove totally the usage of folder / page that is anyway a bit strange. 

We would use it like:

```
<AppLayoutBreadcrumb>
  <BreadcrumbItem>Slices</BreadcrumbItem>
  <BreadcrumbItem>{slice.model.name}</BreadcrumbItem>
</AppLayoutBreadcrumb>
```

WDYT?

packages/slice-machine/src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -1,35 +1,62 @@
...
+type BreadcrumbContext = {
+  separator?: ReactNode;
 };
 
-export const Breadcrumb: FC<BreadcrumbProps> = ({
-  folder,
-  page,
-  separator = "/",
-}) => {
+const BreadcrumbContext = createContext<BreadcrumbContext>({});
```
xrutayisire: _❓ question_

I love the idea to have the separator like this so you can easily change it, but I'm also wondering if it's not overkill, as we don't know yet if one day we will want to change it.

Before it was just a props not used, but now it's a context. Now that you have it, it's tempting to keep it anyway, but it also creates a bit more code never really used.

WDYT about that, I don't have a strong opinion on this?

packages/slice-machine/src/components/Breadcrumb/Breadcrumb.module.css
```diff
@@ -0,0 +1,23 @@
...
+  margin: 0px;
+  padding: 0px;
+}
+
+.item {
+  display: flex;
+  gap: 4px;
+  align-items: center;
+
+  &:last-child {
```
xrutayisire: _❓question_

I was wondering if we would have like to have the last separator not even in the DOM or just not displayed is enough. I fell I prefer not having it in the DOM of course, but I don't see a real reason was this simplicity should not be used. It's a lot more simple to just do this, I guess. WDYT?

packages/slice-machine/src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -1,35 +1,62 @@
...
+
+type BreadcrumbProps = PropsWithChildren<{
+  separator?: ReactNode;
+}>;
+
+export function Breadcrumb(props: BreadcrumbProps) {
```
xrutayisire: _✍️ asking for a potential change_

We started to use the same pattern in SM with FC ([see here](https://www.notion.so/prismic/Slice-Machine-UI-s-Coding-Guidelines-d377c3e33faa44b3a5e0ad84a942616d?pvs=4#899b8887ae4e42e9b4576a0f96539534)). I guess it's just better to keep consistency, even if FC it's not a strong need for us.

packages/slice-machine/src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -1,35 +1,62 @@
...
+      </ol>
+    </nav>
+  );
+}
+
+type BreadcrumbItemProps = PropsWithChildren<{
+  active?: boolean;
+}>;
+
+export function BreadcrumbItem(props: BreadcrumbItemProps) {
```
xrutayisire: _✍️ asking for a potential change_

We started to use the same pattern in SM with FC ([see here](https://www.notion.so/prismic/Slice-Machine-UI-s-Coding-Guidelines-d377c3e33faa44b3a5e0ad84a942616d?pvs=4#899b8887ae4e42e9b4576a0f96539534)). I guess it's just better to keep consistency, even if FC it's not a strong need for us.

packages/slice-machine/src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -1,35 +1,62 @@
...
-  separator = "/",
-}) => {
+const BreadcrumbContext = createContext<BreadcrumbContext>({});
+
+type BreadcrumbProps = PropsWithChildren<{
+  separator?: ReactNode;
+}>;
+
+export function Breadcrumb(props: BreadcrumbProps) {
+  const { separator = "/", children, ...restProps } = props;
```
xrutayisire: _✍️ asking for a potential change_

I can see you used `...otherProps` bellow but `...restProps` here. I think it could be better to keep the same kind of naming for that. Editor and SM started to use `otherProps`, so I think we should keep it like this ([see here](https://www.notion.so/prismic/Slice-Machine-UI-s-Coding-Guidelines-d377c3e33faa44b3a5e0ad84a942616d?pvs=4#847fba9f52cf435ea3664fe600f72a4d))

packages/slice-machine/src/components/Breadcrumb/Breadcrumb.module.css
```diff
@@ -0,0 +1,23 @@
+.items {
```
xrutayisire, line 1: _✍️ asking for a potential change_

Shouldn't we use reset in this file instead of trying to reset manually styles?
Maybe it's something we would like to document in the code guidelines 🧐

playwright/pages/SliceBuilderPage.ts
```diff
@@ -73,7 +73,7 @@ export class SliceBuilderPage extends BuilderPage {
    * Dynamic locators
    */
   override getBreadcrumbLabel(sliceName: string) {
-    return this.breadcrumb.getByText(`Slices / ${sliceName}`, { exact: true });
+    return this.breadcrumb.getByText(sliceName, { exact: true });
```
xrutayisire: _✍️ asking for a potential change_

It's a bit less specific here, right? If a slice name is "Slices", then it would always resolve, no? I guess it could be safer to keep the level of check as before. WDYT?

packages/slice-machine/src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -1,35 +1,62 @@
...
+      <Text
+        component="span"
+        color={active ? "grey12" : "grey11"}
+        className={clsx(active && styles.active)}
+      >
+        {children}
       </Text>
-    </div>
+      <div role="presentation" className={styles.separator}>
+        <Text component="span" color="grey11">
```
xrutayisire: _✍️ asking for a potential change_

I see that we had a style issue problem before, what I see in the Figma design is that the separator is in `grey12`. Maybe @benjammartin you could confirm what I'm saying? ([see here, Breadcrumb component](url))

packages/slice-machine/src/components/Breadcrumb/Breadcrumb.module.css
```diff
@@ -0,0 +1,23 @@
...
+  display: flex;
+  gap: 4px;
```
xrutayisire, line 3: _✍️ asking for a potential change_

For both `gap: 4px`, I think we had a style issue before also. It seems in the design the gap is 4px BUT the total space is 6px. Maybe @benjammartin you could confirm what I'm saying ^^? ([see here, Breadcrumb component](https://www.figma.com/file/wb7ujCJ4j7EiP7aDDGRYXB/SliceMachine%E2%84%A2-UI?type=design&node-id=2110-267713&mode=dev))

Screenshot 2024-04-04 at 11 09 20

xrutayisire (changes requested):
Super work, it's a good improvement, in the right direction. We will be able to handle Links and things like this in the future more easily thanks to this PR. 🚀

Left you some comments to see what you think.

packages/slice-machine/src/components/Breadcrumb/Breadcrumb.module.css
benjammartin: Well spotted. It's a mistake on my part in the design. The gap is indeed 4px.

xrutayisire: Thank you! So no need to change anything on the code 🙂

packages/slice-machine/src/components/Breadcrumb/Breadcrumb.tsx
benjammartin: Second mistake. It's grey11.

xrutayisire: Thank you again! So no need to change here also 🙂

packages/slice-machine/components/AppLayout/index.tsx
angeloashmore: Yep, that sounds good to me. I've replaced all uses of `folder` + `page` with `<BreadcrumbItem>`.

packages/slice-machine/src/components/Breadcrumb/Breadcrumb.tsx
angeloashmore: Due to underlying weirdness in how the separate was rendered (the last one was hidden with CSS), testing was overly difficult using the context method. I decided to replace it with `Children.map` and conditionally render the separator.

We don't have a need to customize the separator, so let's leave it out. It is trivial to re-introduce a custom separator with the latest refactor should we need it in the future.

packages/slice-machine/src/components/Breadcrumb/Breadcrumb.module.css
angeloashmore: Right, in trying to avoid using React.Children, I couldn't find another solution. CSS was the most straightforward way.

But I removed that and used React.Children instead. In this instance, I think it is a good use of it and shouldn't cause issues in the future. Now we don't render the separator in the DOM at all.

packages/slice-machine/src/components/Breadcrumb/Breadcrumb.tsx
angeloashmore: Oh cool! I didn't realize we had a style document like that. I checked both the editor and SM code base and there isn't consistency, even within each project.

I'll refactor this to the FC style and use it going forward. 👍

packages/slice-machine/src/components/Breadcrumb/Breadcrumb.tsx
angeloashmore: Oops! I'm accustomed to using `restProps`, but I saw we are using `otherProps` elsewhere. It's fixed!

packages/slice-machine/src/components/Breadcrumb/Breadcrumb.module.css
angeloashmore: From what I understand, we don't have a standard reset strategy. `all: unset` is not a reset strategy, [as discussed here](https://prismic-team.slack.com/archives/C0441S8SKQR/p1706789739634429).

I feel this method is straightforward, but let me know if you have a different idea. Ideally, we would have a global reset somewhere that removes default browser styling (i.e. not what a normalizer does).

xrutayisire: I even forgot this discussion, maybe we will need to continue this discussion to write info the best practices. LGTM

packages/slice-machine/test/__setup__.ts
angeloashmore: MSW prints the following error when running some tests (e.g. `CustomTypeBuilderPage.test.tsx`):

```
[MSW] Error: captured a request without a matching request handler:

  • GET https://api.lab.amplitude.com/sdk/v2/vardata

If you still wish to intercept this unhandled request, please create a request handler for it.
Read more: https://mswjs.io/docs/getting-started/mocks
[Experiment] Retry falied:  Error: [MSW] Cannot bypass a request when using the "error" strategy for the "onUnhandledRequest" option.
    at applyStrategy ([redacted]/slice-machine/node_modules/msw/src/utils/request/onUnhandledRequest.ts:234:1)
    at onUnhandledRequest ([redacted]/slice-machine/node_modules/msw/src/utils/handleRequest.ts:73:3)
    at handleRequest ([redacted]/slice-machine/node_modules/msw/src/utils/handleRequest.ts:127:3)
    at [redacted]/slice-machine/node_modules/msw/src/node/SetupServerApi.ts:107:9
```

This seems to happen because `@amplitude/experiment-node-server` is not mocked in `slice-machine-ui` tests. My Vitest breadcrumb tests were failing while I was writing this PR, and mocking the endpoint seemed to fix it.

However, it's working now without the extra network mock, so I guess we should just leave it as it was. My code may have been in a weird state at the time.

angeloashmore: Besides, we are planning to remove most of the Vitest tests anyway, so I don't think we need to worry too much about these errors/warnings.

playwright/pages/SliceBuilderPage.ts
angeloashmore: I refactored the tests to this:

```ts
async checkBreadcrumbItems(items: string[]) {
  await expect(this.breadcrumb).toBeVisible();
  await expect(this.breadcrumb).toHaveText(items.join("/"));
}
```

The `Breadcrumb` component is now a set of elements, where before it was a single string, so we can't check perform a simple "getByText" anymore.

This version lets us check every breadcrumb item and its order, just as before. 👍

xrutayisire (approved):
LGTM! Super nice API now 🥳

PR #1330, opened 2024-04-03
feat: warn and inform users on development environment usage (DT-2092)

lihbr:
## Context

Running Slice Machine with `SM_ENV=development` allows any endpoints to be set through environment variables. This is useful for Prismic developers to develop locally and potentially dangerous for users ([more details on Linear](https://linear.app/prismic/issue/DT-2092/aadev-and-aauser-im-clearly-aware-im-using-the-development-environment)).

## The Solution

This PR adds a clear warning that Slice Machine has been started in development mode, prompts the configured endpoints (useful for Prismic developers), and incentivize other users to exit immediately.

## Impact / Dependencies

I didn't use chalk or any formatting as we want to keep the manager free of such dependencies (the manager is not an application). I think the warning is big and unusual enough to catch attention as-is?

- [x] The CI is successful.

## [OPT] Preview

image

Get a similar result by running the following inside an e2e project:
```bash
npx cross-env \
NODE_ENV=development \
SM_ENV=development \
wroom_endpoint=https://prismic.io/ \
authentication_server_endpoint=https://auth.prismic.io/ \
customtypesapi_endpoint=https://customtypes.prismic.io/ \
user_service_endpoint=https://user-service.prismic.io/ \
acl_provider_endpoint=https://0yyeb2g040.execute-api.us-east-1.amazonaws.com/prod/ \
npx start-slicemachine
```

### DIFF
--- a/packages/manager/src/constants/API_ENDPOINTS.ts
+++ b/packages/manager/src/constants/API_ENDPOINTS.ts
@@ -41,11 +41,11 @@ export const API_ENDPOINTS: APIEndpoints = (() => {
 
 			if (missingAPIEndpoints.length > 0) {
 				console.error(
-					`You are running Slice Machine in development mode (SM_ENV=${APPLICATION_MODE.Development}) where API endpoints are configured via environment variables.
+					`You are running Slice Machine in development mode (SM_ENV=${
+						APPLICATION_MODE.Development
+					}) where API endpoints are configured via environment variables.
 
-The following endpoints were not configured: ${missingAPIEndpoints.join(
-						", ",
-					)}.
+The following endpoints were not configured: ${missingAPIEndpoints.join(", ")}.
 
 Configure them before continuing.`,
 				);
@@ -53,10 +53,14 @@ Configure them before continuing.`,
 				process.exit(1);
 			}
 
-			console.warn(`You are running Slice Machine in development mode (SM_ENV=${APPLICATION_MODE.Development}).
+			console.warn(`You are running Slice Machine in development mode (SM_ENV=${
+				APPLICATION_MODE.Development
+			}).
 
 The following API endpoints were configured via environment variables:
-${Object.entries(apiEndpoints).map(([name, endpoint]) => `  - ${name}: ${endpoint}`).join("\n")}
+${Object.entries(apiEndpoints)
+	.map(([name, endpoint]) => `  - ${name}: ${endpoint}`)
+	.join("\n")}
 
 If you didn't intend to run Slice Machine this way stop it immediately and unset the SM_ENV environment variable.`);
 
--- a/packages/manager/src/constants/API_ENDPOINTS.ts
+++ b/packages/manager/src/constants/API_ENDPOINTS.ts
@@ -41,13 +41,24 @@ export const API_ENDPOINTS: APIEndpoints = (() => {
 
 			if (missingAPIEndpoints.length > 0) {
 				console.error(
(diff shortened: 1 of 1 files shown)


packages/manager/src/constants/API_ENDPOINTS.ts
```diff
@@ -41,14 +41,29 @@ export const API_ENDPOINTS: APIEndpoints = (() => {
...
 
+			console.warn(`You are running Slice Machine in development mode (SM_ENV=${
+				APPLICATION_MODE.Development
+			}).
+
+The following API endpoints were configured via environment variables:
+${Object.entries(apiEndpoints)
+	.map(([name, endpoint]) => `  - ${name}: ${endpoint}`)
+	.join("\n")}
+
```
angeloashmore, line 64: To be extra clear to users, maybe we could add a line like this if we detect the endpoints are different than prod or staging:

> These endpoints are different than Slice Machine's normal endpoints and are not trusted.

packages/manager/src/constants/API_ENDPOINTS.ts
```diff
@@ -41,14 +41,29 @@ export const API_ENDPOINTS: APIEndpoints = (() => {
 
 			if (missingAPIEndpoints.length > 0) {
 				console.error(
-					`You are running Slice Machine in development mode where API endpoints are configured via environment variables. The following endpoints were not configured: ${missingAPIEndpoints.join(
-						", ",
-					)}. Configure them before continuing.`,
+					`You are running Slice Machine in development mode (SM_ENV=${
```
angeloashmore, line 44: I have a suggestion outside the scope of this PR: `@slicemachine/manager` is a Node.js library, not a CLI. Rather than calling `console.error`, maybe we should be throwing errors instead. Then, in a CLI like `@slicemachine/init`, we could catch these errors and log them to the console.

An exception to this might be the `console.warn` below since we don't want to accidentally not warn users about potentially malicious endpoints.

packages/manager/src/constants/API_ENDPOINTS.ts
```diff
@@ -41,14 +41,29 @@ export const API_ENDPOINTS: APIEndpoints = (() => {
...
+			console.warn(`You are running Slice Machine in development mode (SM_ENV=${
+				APPLICATION_MODE.Development
+			}).
+
+The following API endpoints were configured via environment variables:
+${Object.entries(apiEndpoints)
+	.map(([name, endpoint]) => `  - ${name}: ${endpoint}`)
+	.join("\n")}
+
+If you didn't intend to run Slice Machine this way stop it immediately and unset the SM_ENV environment variable.`);
```
angeloashmore: > If you didn't intend to run Slice Machine this way stop it…

Missing a "," after "this way":

> If you didn't intend to run Slice Machine this way, stop it…

angeloashmore (approved):

packages/manager/src/constants/API_ENDPOINTS.ts
lihbr: Thinking about it twice, I'm not sure how another package could catch such error since the error would be thrown upon import of the manager because `API_ENDPOINTS.ts` is not a function we run, it's just executed upon load.

I think we'll leave that out for now as I don't see a simple path towards something like that now 🤔

packages/manager/src/constants/API_ENDPOINTS.ts
lihbr: I added it after the list, thanks!

I don't think we want to run some logic to differentiate weird endpoints from others (feeling like it would be a bit overkill) as users should not be using this mode anyway.

PR #1331, opened 2024-04-05
fix: don't hardcode repository name with Nuxt

lihbr:
## Context

With Nuxt 2/3 we were hardcoding the repository name inside Nuxt config instead of sourcing it from Slice Machine configuration. This PR fixes it.

## The Solution

While [magicast](https://github.com/unjs/magicast) can not handle adding import statements, it still has some limitations regarding logical expressions, so I had to use the old (and not so clean 🙈) search & replace trick.

- [x] The CI is successful.

### DIFF
--- a/packages/adapter-nuxt/test/plugin-project-init.test.ts
+++ b/packages/adapter-nuxt/test/plugin-project-init.test.ts
@@ -72,6 +72,43 @@ describe("Prismic module", () => {
 			"
 		`);
 	});
+
+	test("overwrites existing endpoint", async (ctx) => {
+		const log = vi.fn();
+		const installDependencies = vi.fn();
+
+		const nuxtConfigPath = path.join(ctx.project.root, "nuxt.config.js");
+		await fs.writeFile(
+			nuxtConfigPath,
+			`
+export default defineNuxtConfig({
+  modules: ["@nuxtjs/prismic"],
+
+  prismic: {
+    endpoint: "example-prismic-repo",
+  },
+})
+`.trim(),
+		);
+
+		await ctx.pluginRunner.callHook("project:init", {
+			log,
+			installDependencies,
+		});
+
+		await expect(fs.readFile(nuxtConfigPath, "utf-8")).resolves
+			.toMatchInlineSnapshot(`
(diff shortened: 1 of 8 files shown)


packages/adapter-nuxt/test/plugin-project-init.test.ts
angeloashmore, line 1: 💡 Perhaps we should test that the code overwrites an existing config. The code already handles it correctly, but it's always good to confirm.

<details><summary>Here's a version I wrote while I was reviewing the PR. You can use it as is or write your own.</summary>
<p>

```ts
	test("overwrites existing endpoint", async (ctx) => {
		const log = vi.fn();
		const installDependencies = vi.fn();

		const nuxtConfigPath = path.join(ctx.project.root, "nuxt.config.js");
		await fs.writeFile(
			nuxtConfigPath,
			`
export default defineNuxtConfig({
  modules:["@nuxtjs/prismic"],

  prismic: {
    endpoint: "example-prismic-repo"
  }
})
`.trim(),
		);

		await ctx.pluginRunner.callHook("project:init", {
			log,
			installDependencies,
		});

		await expect(fs.readFile(nuxtConfigPath, "utf-8")).resolves
			.toMatchInlineSnapshot(`
			"import { apiEndpoint, repositoryName } from \\"./slicemachine.config.json\\";
			export default defineNuxtConfig({
			  modules: [\\"@nuxtjs/prismic\\"],

			  prismic: {
			    endpoint: apiEndpoint || repositoryName,
			  },
			});
			"
		`);
	});
```

</p>
</details>

angeloashmore (approved):
Looks good! I took a brief look into how we could do this without the `.replace()`, and it was more code/effort than it was worth.

Just one small suggestion below.

---

Side note: Adding support for LogicalExpression to Magicast probably wouldn't be too much work. It's not our priority right now, but it could be a fun side quest. 🙂

lihbr:
Thanks for the review! Updated tests indeed, that'll be safer~

> Side note: Adding support for LogicalExpression to Magicast probably wouldn't be too much work. It's not our priority right now, but it could be a fun side quest.

That'd be fun yes, I'll think about it ☺️

PR #1332, opened 2024-04-06
chore(slice-machine-ui): automatically sort imports using `eslint-plugin-simple-import-sort`

angeloashmore:
## Context

We currently manually sort imports. Not all files are sorted. Imports are not consistently sorted between files.

Manually sorting imports has the following downsides:
- Moving imports manually adds unnecessary lines to a diff.
- It requires brain power to find the correct location for an import.
- An order that makes sense for one person may not make sense for another.

## The Solution

Use a tool to automatically sort all imports.

This PR adds `eslint-plugin-simple-import-sort`, the same plugin used by the Page Builder. It lets us sort imports using `eslint --fix`

We can start using this solution in the `slice-machine-ui` package. If we find the solution works well, we can port it to the other packages in the monorepo.

## Impact / Dependencies

N/A

- [x] The CI is successful.


xrutayisire:
Slack thread for reminder: https://prismic-team.slack.com/archives/C02L3FN3AJK/p1712611867217019

What's the status of this @angeloashmore? What would you like to do with the following recommendations we gave?

xrutayisire (approved):
LGTM! Super nice to have that and stop managing manually our imports 🥳

PR #1333, opened 2024-04-09
feat: introduce "Repeatable Group" field

angeloashmore:
> [!IMPORTANT]
> We are waiting for final copy ([related thread](https://prismic-team.slack.com/archives/C01QH6YR6AX/p1712722630093849)) and design ([related thread](https://prismic-team.slack.com/archives/C02G5CFFLR2/p1712722201562769)) before merging this PR.

## Context

DT-2056

## The Solution

- Add a new "Repeatable Group" field based on the existing group widget. The field supports `repeat: true | undefined`.
- Change the existing group widget to only support `repeat: false`.
- Remove the ability to toggle the `repeat` config value through the field editor.

## Impact / Dependencies

All existing group fields are supported. Existing group fields will be locked into their current configuration (i.e. repeatable or single). The majority of groups will remain a `repeatable` group.

- [ ] The CI is successful.

## [OPT] Preview

**Adding a field to a custom type:**
image

**Editing a group field:**
image


xrutayisire:
Since it's a Draft I'm waiting for the go to review it 🙌

packages/slice-machine/lib/models/common/widgets/ctBuilderArray.ts
```diff
@@ -3,6 +3,7 @@ import * as Widgets from "./withGroup";
 export default [
   Widgets.UID,
   Widgets.Group,
+  Widgets.RepeatableGroup,
```
xrutayisire, line 5: ✍️ potential change

I think I would put it before UID, or at least ensure it's on the same line? WDYT, maybe it's something we should ask design. I feel hard to see that there are two groups now.

Screenshot 2024-04-11 at 18 11 05

packages/slice-machine/lib/models/common/widgets/RepeatableGroup/ListItem/index.jsx
```diff
@@ -0,0 +1,285 @@
+import { findWidgetByConfigOrType } from "@builders/utils";
```
xrutayisire: ✍️ potential change

Can we not reuse ListItem component from the non-repeatable Group?

packages/slice-machine/lib/models/common/widgets/RepeatableGroup/Form.jsx
```diff
@@ -0,0 +1,29 @@
+import { Col, Flex as FlexGrid } from "@components/Flex";
```
xrutayisire: ✍️ potential change

Can we not reuse Form component from the non-repeatable Group?

xrutayisire (changes requested):
Nice work!

I was wondering if we cannot just prevent having a duplication of the files. Added 2 comments about that. If it's really the same component, we could say that the repeatable group reuse the one from group.

packages/slice-machine/lib/models/common/widgets/RepeatableGroup/ListItem/index.jsx
angeloashmore: Ah, I totally forgot to do that! Thank you for mentioning it.

packages/slice-machine/lib/models/common/widgets/ctBuilderArray.ts
angeloashmore: Good catch, it should be before UID so the group fields are on the same line.

From a more general view, should we even have UID as an option? We already add a UID field to repeatable custom/page types. If someone adds a single type, they shouldn't be using a UID since the type itself is the identifier.

angeloashmore: If you think that's worth discussing, should we bring it up to Côme?

angeloashmore: (I moved the UID field after the group fields in the meantime.)

packages/slice-machine/lib/models/common/widgets/RepeatableGroup/Form.jsx
angeloashmore: We can, I just forgot! It's done now.

packages/slice-machine/lib/models/common/widgets/ctBuilderArray.ts
xrutayisire: Côme was not here today, I think it should be good that Benjamin and Côme knows yes. 🙂

xrutayisire (approved):
LGTM, love it!

I'm just wondering if we can merge or not before having for example the icon, because this one is not protected from the amplitude experiment. WDYT?

angeloashmore:
I'd like to merge it now, but how about we wait until we hear back from Sam on the naming and descriptions. Waiting will avoid user confusion if we happen to change the name.

But I agree, we probably don't need to wait for the icon before merging. We can update that in the future.

xrutayisire:
I agree, the discussion between Côme and Sam need to finish before merging 👍

angeloashmore:
We decided **not** to pursue this change. Instead, we will keep the group field as it exists today.

We will likely change the group field's description to highlight that it can be repeated.

In the future, we may explore an object-like field, similar to a non-repeatable group field without an array.

PR #1334, opened 2024-04-11
feat: hide the Repeatable Zone in slices that do not have `items` fields

angeloashmore:
## Context

DT-2062

## The Solution

- Do not show the Repeatable Zone if a slice variation does not have an `items` property, or the property is empty (`{}`).
- Rename the Static Zone to "Fields".
- Update the blank slate message to not use "Static" and "Repeatable".
- Add a confirmation dialog when the last Repeatable Zone field is deleted. Deleting it will remove the Repeatable Zone permanently.

Several changes to the Playwright setup were required to test the new behavior:

- Add a new `repeatableZoneSlice` fixture that includes one field in the Repeatable Zone. This allows the zone to appear.
- Use `@slicemachine/manager/client` to create `repeatableZoneSlice`. Since the UI is incapable of creating a slice with a Repeatable Zone, it needed to be created outside the UI.
- Use `@prismicio/mock` to generate the mock model. This is not required, but it reduces the amount of code needed in the fixtures.
- Convert all Repeatable Zone-related tests to use `repeatableZoneSlice`.

We can choose to remove `@prismicio/mock` if we think the overhead is not necessary. Instead, we can inline a model as an object literal.

## Impact / Dependencies

New slices will no longer show the Repeatable Zone.

Existing slices that have fields in the Repeatable Zone will continue to show the Repeatable Zone. The zone will be removed if its fields are removed.

Existing slices that do not have fields in the Repeatable Zone will no longer show the Repeatable Zone.

- [ ] The CI is successful.

## [OPT] Preview

**With no fields**
image

**With fields in `primary` and no fields in `items`**
image

**With fields in `items`**
image

**Dialog shown when deleting the last `items` field**
image

### DIFF
--- a/packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
+++ b/packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
@@ -31,6 +31,7 @@ import Zone from "@lib/builders/common/Zone";
 import EditModal from "@lib/builders/common/EditModal";
 import { AnyWidget } from "@lib/models/common/widgets/Widget";
 import { telemetry } from "@src/apiClient";
+import { useGroupsInSlicesExperiment } from "@src/features/slices/sliceBuilder/useGroupsInSlicesExperiment";
 
 const dataTipText = ` The non-repeatable zone
   is for fields<br/> that should appear once, like a<br/>
@@ -46,6 +47,7 @@ const FieldZones: FC = () => {
     isDeleteRepeatableZoneDialogOpen,
     setIsDeleteRepeatableZoneDialogOpen,
   ] = useState(false);
+  const groupsInSlicesExperiment = useGroupsInSlicesExperiment();
 
   // We won't show the Repeatable Zone if no items are configured.
   const hasItems = Boolean(
@@ -54,6 +56,7 @@ const FieldZones: FC = () => {
 
   const _onDeleteItem = (widgetArea: WidgetsArea) => (key: string) => {
     if (
+      groupsInSlicesExperiment.eligible &&
       widgetArea === WidgetsArea.Items &&
       variation.items &&
       Object.keys(variation.items).length <= 1
@@ -216,7 +219,7 @@ const FieldZones: FC = () => {
         testId="static-zone-content"
         isRepeatableCustomType={undefined}
       />
-      {hasItems ? (
+      {!groupsInSlicesExperiment.eligible || hasItems ? (
         <Zone
           zoneType="slice"
           zoneTypeFormat={undefined}
@@ -257,8 +260,8 @@ const FieldZones: FC = () => {
               {slice.model.name} slice.
             </strong>
             <div>
-              If you need repeatable fields again, use a group field in place of
-              the repeatable zone.
+              If you need repeatable fields again, use a repeatable group field
+              in place of the repeatable zone.
             </div>
           </Box>
           <DialogActions
--- a/packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
+++ b/packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
@@ -1,10 +1,17 @@
-import { FC } from "react";
+import { FC, useState } from "react";
 import { flushSync } from "react-dom";
 import { DropResult } from "react-beautiful-dnd";
 import {
   FieldType,
   NestableWidget,
 } from "@prismicio/types-internal/lib/customtypes";
+import {
+  Box,
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogHeader,
+} from "@prismicio/editor-ui";
 
 import { ensureDnDDestination } from "@lib/utils";
 import { transformKeyAccessor } from "@utils/str";
@@ -16,6 +23,7 @@ import { useSliceState } from "@src/features/slices/sliceBuilder/SliceBuilderPro
 import {
   addField,
   deleteField,
+  deleteRepeatableZone,
   reorderField,
   updateField,
 } from "@src/domain/slice";
@@ -34,8 +42,26 @@ const dataTipText2 = `The repeatable zone is for a group<br/>
 
 const FieldZones: FC = () => {
   const { slice, setSlice, variation } = useSliceState();
+  const [
+    isDeleteRepeatableZoneDialogOpen,
+    setIsDeleteRepeatableZoneDialogOpen,
+  ] = useState(false);
+
+  // We won't show the Repeatable Zone if no items are configured.
+  const hasItems = Boolean(
+    variation.items && Object.keys(variation.items).length > 0,
+  );
 
   const _onDeleteItem = (widgetArea: WidgetsArea) => (key: string) => {
+    if (
+      widgetArea === WidgetsArea.Items &&
+      variation.items &&
+      Object.keys(variation.items).length <= 1
+    ) {
+      setIsDeleteRepeatableZoneDialogOpen(true);
+      return;
+    }
+
     const newSlice = deleteField({
       slice,
       variationId: variation.id,
@@ -154,13 +180,20 @@ const FieldZones: FC = () => {
     flushSync(() => setSlice(newSlice));
   };
 
+  const onDeleteRepeatableZone = () => {
+    const newSlice = deleteRepeatableZone({ slice, variationId: variation.id });
+
+    setSlice(newSlice);
+    setIsDeleteRepeatableZoneDialogOpen(false);
+  };
+
   return (
     <List>
       <Zone
         zoneType="slice"
         zoneTypeFormat={undefined}
         tabId={undefined}
-        title="Non-Repeatable Zone"
+        title="Fields"
         dataTip={dataTipText}
         // eslint-disable-next-line @typescript-eslint/ban-ts-comment
         fields={variation.primary}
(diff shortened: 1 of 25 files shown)


packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
```diff
@@ -154,13 +180,20 @@ const FieldZones: FC = () => {
...
+  };
+
   return (
     <List>
       <Zone
         zoneType="slice"
         zoneTypeFormat={undefined}
         tabId={undefined}
-        title="Non-Repeatable Zone"
+        title="Fields"
```
xrutayisire, line 199: With Lucie proposal change with the usage of `primary` I'm wondering if here it could be good to indicate that it's corresponding to primary. Could help new users to understand. WDYT? 

Like `Fields (primary)` 

But maybe it's not interesting, if you like it, we could ask Sam. If you don't, let's not, I'm not super confident on this, mostly wanted to share.

packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
```diff
@@ -34,8 +42,26 @@ const dataTipText2 = `The repeatable zone is for a group<br/>
...
+    variation.items && Object.keys(variation.items).length > 0,
+  );
 
   const _onDeleteItem = (widgetArea: WidgetsArea) => (key: string) => {
+    if (
+      widgetArea === WidgetsArea.Items &&
+      variation.items &&
+      Object.keys(variation.items).length <= 1
+    ) {
+      setIsDeleteRepeatableZoneDialogOpen(true);
```
xrutayisire, line 64: ✨ _love it_

That's super smart to open the modal and explain what's happening!
I guess it's something you can share with Côme and Benjamin, It's a very good idea.

packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
```diff
@@ -183,32 +216,61 @@ const FieldZones: FC = () => {
...
+      >
+        <DialogHeader icon="delete" title="Delete field" />
+        <DialogContent>
+          <Box padding={24} gap={12} flexDirection="column">
+            <strong>
+              This action will permanently remove the Repeatable Zone from the{" "}
+              {slice.model.name} slice.
+            </strong>
+            <div>
+              If you need repeatable fields again, use a group field in place of
```
xrutayisire: _✍️ potential change_

`If you need repeatable fields again, use a repeatable group field in place of the repeatable zone.`

packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
```diff
@@ -183,32 +216,61 @@ const FieldZones: FC = () => {
...
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment, @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-member-access
-        renderHintBase={({ item }) => `item${transformKeyAccessor(item.key)}`}
-        renderFieldAccessor={(key) =>
-          // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
-          `slice.items[i]${transformKeyAccessor(key)}`
-        }
-        testId="slice-repeatable-zone"
-        isRepeatableCustomType={undefined}
-      />
+      {hasItems ? (
```
xrutayisire: _✍️ potential change_

Here I think we should add a condition with the Amplitude experiment value. It would let us merge branches without having to wait everything. WDYT?

packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
```diff
@@ -34,8 +42,26 @@ const dataTipText2 = `The repeatable zone is for a group<br/>
...
+    setIsDeleteRepeatableZoneDialogOpen,
+  ] = useState(false);
+
+  // We won't show the Repeatable Zone if no items are configured.
+  const hasItems = Boolean(
+    variation.items && Object.keys(variation.items).length > 0,
+  );
 
   const _onDeleteItem = (widgetArea: WidgetsArea) => (key: string) => {
+    if (
```
xrutayisire, line 58: _✍️ potential change_

Here I think we should add a condition with the Amplitude experiment value. It would let us merge branches without having to wait everything. WDYT?

packages/slice-machine/lib/builders/common/Zone/components/EmptyState/index.tsx
```diff
@@ -13,7 +13,7 @@ const ZoneEmptyState = ({
     <Heading as="h5">Add a new field here</Heading>
     <Box sx={{ my: 2 }}>
       <Text sx={{ color: "textClear" }}>
-        Add a field to your {zoneName} Zone
+        Fields hold content, like text and images.
```
xrutayisire: _✍️ potential change_

After testing your work, I feel like we could be a bit more explicit about the fact there is a field type for repeatable content. 
Like `Fields hold content, like text, images, repeatable group of fields.`
WDYT?

packages/slice-machine/src/domain/slice.ts
```diff
@@ -281,3 +286,23 @@ export function rename(slice: ComponentUI, newSliceName: string): ComponentUI {
     },
   };
 }
+
+export function deleteRepeatableZone(args: DeleteRepeatableZoneArgs) {
```
xrutayisire, line 290: _✍️ potential change_

I know that `domain/slice.ts` isn't unit tested at all, not like `domain/customType.ts` that is fully unit tested. But what do you feel about adding a test for `deleteRepeatableZone` function? For me `domain` folder is one that is perfect for unit tests. WDYT?

playwright/fixtures/index.ts
```diff
@@ -48,11 +53,19 @@ type Fixtures = {
   reusableCustomType: { name: string };
   singleCustomType: { name: string };
   slice: { name: string };
+  repeatableZoneSlice: { name: string };
```
xrutayisire, line 56: _✨ love it_

Love the name, we should start to use that name to talk about a slice that use the items property!

playwright/fixtures/index.ts
```diff
@@ -158,6 +171,27 @@ export const test = baseTest.extend<Options & Fixtures>({
 
     await use({ name: sliceName });
   },
+  repeatableZoneSlice: async ({ sliceLibrary, manager, createMock }, use) => {
+    const variation = createMock.model.sharedSliceVariation({
```
xrutayisire: _✨ love it_

Super happy to bring @prismic/mock to slice-machine! I will count on you to tell us when we should have used it for future PRs 😛

playwright/fixtures/index.ts
```diff
@@ -158,6 +171,27 @@ export const test = baseTest.extend<Options & Fixtures>({
...
+      itemsFields: {
+        existing_field: createMock.model.richText(),
+      },
+    });
+    const model = createMock.model.sharedSlice({ variations: [variation] });
+
+    await manager.slices.createSlice({ libraryID: sliceLibrary.id, model });
```
xrutayisire: _✨ love it_

THANK YOU! That's so cool that I hate myself to not have done that before. We should NEVER use the UI to create our default state before a test. Thanks to that, it's really better AND faster to run tests.

We will need to copy your way for the others fixtures we have 🚀

playwright/fixtures/index.ts
```diff
@@ -158,6 +171,27 @@ export const test = baseTest.extend<Options & Fixtures>({
...
+
+    await use({ name: model.name });
+  },
+  sliceLibrary: async ({ manager }, use) => {
+    const config = await manager.project.getSliceMachineConfig();
+    const libraryID = config.libraries?.[0];
+    if (!libraryID) {
+      throw new Error("At least one library is required.");
```
xrutayisire: _✍️ potential change_

Maybe it could be good to explain where it should be defined like: `At least one library is required in your "slicemachine.config.json"`. WDYT?

playwright/fixtures/index.ts
```diff
@@ -158,6 +171,27 @@ export const test = baseTest.extend<Options & Fixtures>({
...
+        existing_field: createMock.model.richText(),
+      },
+    });
+    const model = createMock.model.sharedSlice({ variations: [variation] });
+
+    await manager.slices.createSlice({ libraryID: sliceLibrary.id, model });
+
+    await use({ name: model.name });
+  },
+  sliceLibrary: async ({ manager }, use) => {
```
xrutayisire: _✍️ potential change_

Maybe we should name it `firstSliceLibrary`? Or we should give a function like `getSliceLibrary` with the index of the slice library you want. It's really something not useful today but since we are hard-coding 0 I wanted to propose.

playwright/fixtures/index.ts
```diff
@@ -247,6 +281,19 @@ export const test = baseTest.extend<Options & Fixtures>({
     await newContext.close();
   },
 
+  /**
+   * Manager
+   */
+  // eslint-disable-next-line no-empty-pattern
+  manager: async ({}, use) => {
+    const client = createSliceMachineManagerClient({
+      // TODO: Experiment with config.webServer.url (config is the third parameter of a fixture function)
```
xrutayisire: I agree that it should work by taking the url from the config 🧐

xrutayisire (changes requested):
Awesome! Really love your work!

Some comments to answer but nothing big on my side 🙌

packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
angeloashmore: I think "Fields" is a better option to keep it consistent with slices that don't have the Repeatable Zone. I think there will be less confusion around why the configuration is different between slices that have and don't have a Repeatable Zone.

Alternatively, we could continue to use "Static Zone" and "Repeatable Zone" only when a Repeatable Zone is active. We would still revert to "Fields" when it is the only zone.

I asked Sam here: https://prismic-team.slack.com/archives/C02L3FN3AJK/p1712869154256539

packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
angeloashmore: Good idea! I asked them here: https://prismic-team.slack.com/archives/C02G5CFFLR2/p1712869312778609

packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
angeloashmore: Good catch!

packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
angeloashmore: That's right - I forgot about the experiment part. I added `useGroupsInSlicesExperiment()` and the correct checks.

packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
angeloashmore: Yep, totally forgot about the experiment! I added `useGroupsInSlicesExperiment()` and the correct checks.

packages/slice-machine/lib/builders/common/Zone/components/EmptyState/index.tsx
angeloashmore: I don't think we need to advertise the types of fields too much. "Group of fields" isn't really content; it's more of a content structure, so it sounds misplaced in this sentence.

@samlfair recommended this text:
> Fields are inputs for content (e.g. text, images, links)

I think that works well. Developers will quickly see that they can add groups once they click "Add a field".

What do you think? Should we open this question to the rest of the team?

xrutayisire: Let's do that, no need to ask 🙂

packages/slice-machine/src/domain/slice.ts
angeloashmore: Yep, good idea! I added `@prismicio/mock` like the other PR to focus the tests on what we want to test. Let me know if you are good with that. 🙂

https://github.com/prismicio/slice-machine/pull/1334/commits/ffd17386fa91073e832582c3883c6def4168fce8

playwright/fixtures/index.ts
angeloashmore: Haha sounds good! I'll keep an eye out for good uses of it.

playwright/fixtures/index.ts
angeloashmore: I was surprised how easy it was to implement! It really is faster than going through the UI.

I'll keep this PR focused on just the new feature, but we can convert the remaining tests as a tech debt item.

playwright/fixtures/index.ts
angeloashmore: Yep, that works for me. 👍

playwright/fixtures/index.ts
angeloashmore: Done!

```ts
{
  // ...
  manager: async ({}, use, config) => {
    const client = createSliceMachineManagerClient({
      serverURL: new URL("./_manager", config.project.use.baseURL).toString(),
    });

    await use(client);
  },
  // ...
}
```

playwright/fixtures/index.ts
angeloashmore: I think `firstSliceLibrary` (or `defaultSliceLibrary`?) works better than a `getSliceLibrary` function. We will most likely only use this value in fixtures to generate data, in which case we will only be using the first slice library.

If we are manipulating the UI, we should be using the actual UI instead to read values (I think, you tell me if that's incorrect).

I'll rename it to `firstSliceLibrary`, but let me know if you think otherwise.

packages/slice-machine/src/domain/slice.ts
xrutayisire: Super nice!

playwright/fixtures/index.ts
xrutayisire: I like `firstSliceLibrary` easy and devs will clearly understand what they are using 🙂

xrutayisire (approved):
LGTM! Super nice work 🙂

angeloashmore:
Installing `@prismicio/mock` and `@prismicio/client` into the `playwright` workspace caused issues, as shown in the above commits. Ultimately, I removed the dependencies and wrote the only mock model needed by hand.

We can restore `@prismicio/mock` in a different PR. I finally figured out what changes are necessary, but they require changes to tests outside the scope of this PR.

PR #1335, opened 2024-04-16
feat: support group fields in slices

angeloashmore:
## Context

DT-2059

## The Solution

Add the group field to the allowed list of slice fields. It is only included in the Fields zone, not the Repeatable Zone. Remember that all new slices will not have a Repeatable Zone soon.

This feature required changing how the group field works. Before this PR, adding and editing a group field directly edited the field's custom type. Because groups can now exist in slices, there is no custom type to update.

The group field's custom logic now uses a `saveItem` prop provided by the root model's list of fields. This allows the group field to save the field using the root model's logic.

Adding a group to a slice is hidden behind a feature flag as of this PR. The restriction will be removed in a future PR.

### Tracking updates

The group field sends a tracking event when one of its fields is added or edited. The tracking event sends the type of the root model (e.g. `"custom type"` or `"slice"`). This information was previously hard-coded to `"custom type"` since groups could only be added to custom/page types.

As of this PR, the tracking event can now correctly send `"slice" | "custom type" | "page type"`. This requires access to the model regardless of where the group field is used.

To access the model, a new `ModelContext` is introduced. The context holds the type of model (`type`) and the model itself (`model`). The context is populated above the field zones with the appropriate model.

### Data path and snippets updates

The data path and snippets (which use the data path) were updated to more closely reflect how the data is used.

**Group fields:**
- Before: `slice.primary.my_group.[...]` or `data.my_group.[...]`
- After: `slice.primary.my_group` or `data.my_group`

**Fields within groups:**
- Before: `slice.primary.my_group.my_field` or `data.my_group.my_field`
- After: `item.my_field`

Note that fields within groups always use `item`, which matches how they are used in `my_group.map(...)`:

```ts
slice.data.my_group.map(item => {
  // Do something with item.my_field
})
```

## Impact / Dependencies

Related: https://github.com/prismicio/slice-machine/pull/1334

- [ ] The CI is successful.

## [OPT] Preview

**A group field in a slice:**
image

**The group field is an available field option:**
image

**The group field is not available in the repeatable zone:**
Screenshot 2024-04-17 at 4 09 34 PM


lihbr (commented):
I wanted to go through it to familiarize myself with that part of the codebase a bit and get used to reading some more React/Playwright.

It's interesting how it now works with the `ModelProvider`! I like the pattern.

Anyway, it looks good to me but I'm sure the others might have more relevant things to say~

packages/slice-machine/lib/models/common/Slice.ts
```diff
@@ -55,7 +65,10 @@ export const Variations = {
       Variation.decode({
         ...variation,
         primary: variation.primary?.reduce(
-          (acc, { key, value }) => ({ ...acc, [key]: value }),
+          (acc, { key, value }) => ({
+            ...acc,
+            [key]: value.type === "Group" ? Groups.fromSM(value) : value,
```
xrutayisire: 💡#suggestion

Should we use `GroupFieldType` from `@prismicio/types-internal`:

```suggestion
            [key]: value.type === GroupFieldType ? Groups.fromSM(value) : value,
```

packages/slice-machine/lib/models/common/Slice.ts
```diff
@@ -80,7 +93,7 @@ export const Variations = {
         primary: Object.entries(variation.primary || {}).map(
           ([key, value]) => ({
             key,
-            value,
+            value: value.type === "Group" ? Groups.toSM(value) : value,
```
xrutayisire: 💡#suggestion

Should we use `GroupFieldType` from `@prismicio/types-internal`:

```suggestion
            value: value.type === GroupFieldType ? Groups.toSM(value) : value,
```

packages/slice-machine/lib/models/common/widgets/Group/ListItem/index.jsx
```diff
@@ -66,54 +64,52 @@ const CustomListItem = ({
...
+    });
 
     void telemetry.track({
       event: "field:added",
       id,
       name: label,
       type: widgetTypeName,
       isInAGroup: true,
-      contentType: customType.format === "page" ? "page type" : "custom type",
+      contentType:
```
xrutayisire: 💡 #suggestion

I find the code hard to read when it's a double ternary. Could we do something without? Maybe a function `getFieldContentType`? Something else?

Same for the other time in this file*

packages/slice-machine/lib/models/common/widgets/sliceBuilderArray.ts
```diff
@@ -1,31 +1,19 @@
...
-import { LinkWidget } from "./Link";
-import { LinkToMediaWidget } from "./LinkToMedia";
-import { NumberWidget } from "./Number";
-import { SelectWidget } from "./Select";
-import { StructuredTextWidget } from "./StructuredText";
-import { TextWidget } from "./Text";
-import { TimestampWidget } from "./Timestamp";
+import * as Widgets from "./withGroup";
 
 export default [
```
xrutayisire: 👍 #praise

Just simpler like that, like it!

packages/slice-machine/package.json
```diff
@@ -41,14 +41,14 @@
   "devDependencies": {
     "@emotion/react": "11.11.1",
     "@extractus/oembed-extractor": "3.1.8",
-    "@prismicio/client": "7.4.1",
+    "@prismicio/client": "^7.5.0-alpha.0",
```
xrutayisire: ⚠️ #issue

We should continue with standard practice without `^`, right?

```suggestion
    "@prismicio/client": "7.5.0-alpha.0",
```

Maybe you planned to do it when it's not an alpha.

packages/slice-machine/src/domain/slice.ts
```diff
@@ -207,18 +211,30 @@ export function addField(args: AddFieldArgs): ComponentUI {
     model: {
       ...slice.model,
       variations: slice.model.variations.map((v) => {
-        if (v.id === variationId) {
+        if (v.id !== variationId) {
+          return v;
+        }
+
+        if (widgetArea === WidgetsArea.Items) {
+          if (newField.type === "Group") {
```
xrutayisire: 💡#suggestion

Should we use `GroupFieldType` from `@prismicio/types-internal`:

```suggestion
          if (newField.type === GroupFieldType) {
```

packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
```diff
@@ -111,32 +107,19 @@ const FieldZones: FC = () => {
...
       widgetTypeName: FieldType;
     }) => {
-      // @ts-expect-error TS(7053) FIXME: Element implicitly has an 'any' type because expre... Remove this comment to see the full error message
-      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-      const widget = Widgets[widgetTypeName];
-      // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+      const widget = sliceBuilderWidgetsArray.find(
+        (sliceBuilderWidget) =>
+          sliceBuilderWidget.CUSTOM_NAME === widgetTypeName ||
```
xrutayisire: ❓ #question

I don't understand this check, why you need to check `CUSTOM_NAME`? By checking the usage it feels it's never used to verify the support of a field or not.

packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
```diff
@@ -111,32 +107,19 @@ const FieldZones: FC = () => {
...
+          sliceBuilderWidget.TYPE_NAME === widgetTypeName,
+      );
       if (!widget) {
-        console.log(
```
xrutayisire: 👍 #praise

Good one, this was duplicated with the next condition, correct? Good catch!

packages/slice-machine/src/domain/group.ts
```diff
@@ -0,0 +1,89 @@
+import { NestableWidget } from "@prismicio/types-internal/lib/customtypes";
+
+import { GroupSM } from "@lib/models/common/Group";
```
xrutayisire: ⚠️ #issue

I like the idea to create a new `group.ts` file to make it sharable.
But, I think we need to continue in the goal of trying to use only `prismic-types-internal` types. It's not the case for `domain/slice.ts` file, but it is for `domain/customTypes.ts`. 

What do you think of just ensuring here it's using `Group` prismic-types-internal` and when using the function you just convert `fromSM` to ensure it's working?

packages/slice-machine/src/domain/group.ts
xrutayisire, line 1: ⚠️ #issue

Can we add unit tests for this new file?

playwright/tests/slices/sliceBuilderFields.spec.ts
```diff
@@ -51,6 +51,159 @@ test("I can edit a rich text field in the static zone", async ({
   ).toBeVisible();
 });
 
+test("I can add a group field in the static zone", async ({
+  sliceBuilderPage,
+  slice,
+  procedures,
+}) => {
+  procedures.mock(
```
xrutayisire, line 59: 💡 #suggestion

Should we add a note in the ticket that will remove the experiment to ensure we don't forget in the tests?

https://linear.app/prismic/issue/DT-2105/aasmuser-i-dont-need-to-be-part-of-the-amplitude-experiment-to-create

playwright/tests/slices/sliceBuilderFields.spec.ts
```diff
@@ -51,6 +51,159 @@ test("I can edit a rich text field in the static zone", async ({
   ).toBeVisible();
 });
 
+test("I can add a group field in the static zone", async ({
```
xrutayisire, line 54: 👍 #praise

Super nice tests! Feel super nice to be covered like that!

packages/slice-machine/lib/models/common/widgets/Group/ListItem/index.jsx
```diff
@@ -226,8 +233,7 @@ const CustomListItem = ({
                           enterEditMode,
                           deleteItem: onDeleteItem,
                           renderFieldAccessor: (key) =>
-                            // eslint-disable-next-line @typescript-eslint/restrict-template-expressions, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/restrict-template-expressions
-                            `data.${groupItem.key}.${key}`,
+                            `item${transformKeyAccessor(item.key)}`,
```
xrutayisire: ❓#question

Why do we need this changes from `data.` to `item.`, what is the impact?

playwright/pages/SliceBuilderPage.ts
```diff
@@ -94,32 +100,59 @@ export class SliceBuilderPage extends BuilderPage {
...
+    return this.getListItem(fieldId, zoneType, groupFieldId).getByTestId(
+      "field-menu-button",
+    );
   }
 
-  getListItemFieldId(fieldId: string, zoneType: ZoneType) {
+  getListItemFieldId(
+    fieldId: string,
+    zoneType: ZoneType,
+    groupFieldId?: string,
```
xrutayisire: 💡 #suggestion

Should we do the same as below? When you pass a zoneType `repetable`, you cannot pass `groupFieldId`?

Same for all functions like this in this file* Maybe we can share something between all?

packages/slice-machine/src/domain/group.ts
xrutayisire, line 1: ⚠️ issue

Maybe I don't understand, but now we have a shared `group.ts` file that handle logic related to group field. We need to remove all the code that handled this in `domain/customTypes.ts` right?

packages/slice-machine/src/features/models/ModelProvider.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const ModelContext = createContext<ModelContextValue>({});
+
+type ModelProviderProps = PropsWithChildren<
+  | { type?: never; model?: never }
+  | { type: "customType"; model: CustomType }
+  | { type: "slice"; model: ComponentUI }
+>;
+
+export function ModelProvider(props: ModelProviderProps) {
```
xrutayisire: 💡#suggestion

I like the idea of a Model Provider to know what model we are editing, but should we just use something route based? And so not having the need to create a new context just for that?
At the end, we just need to know that for tracking 🧐 We could even embed that in the track management specific for group. 

I agree that this feel cleaner in a sense but also add new stuff. Route base could just be a simple custom hook `useCurrentModel` that we consume to know where we are `slices` or `custom-types`. WDYT?

Maybe @bapmrl you have a preference also for this kind of thing?

xrutayisire (changes requested):
Super work!

I added some comments that we can discuss 🙂

packages/slice-machine/lib/models/common/Slice.ts
angeloashmore: Good idea!

packages/slice-machine/lib/models/common/widgets/Group/ListItem/index.jsx
angeloashmore: I like the idea of a URL-based helper, as you suggested in https://github.com/prismicio/slice-machine/pull/1335#discussion_r1572109420. It's a much simpler solution than the `ModelContext` I created.

I replaced the ternary in this commit: https://github.com/prismicio/slice-machine/pull/1335/commits/6dffec7e368abeb7c1f9efe1a1cca9fcba3d0e8e

I decided to pass `window.location.pathname` to make it clear that the content type is determined using the URL's pathname.

packages/slice-machine/package.json
angeloashmore: Oops, that wasn't intentional. Thanks for catching it.

`@slicemachine/manager` and `@slicemachine/plugin-kit` use a mix of `^` and non-`^`. From what I remember, we decided packages will use `^` as it better matches what other packages use and won't confuse/frustrate users, while apps would use non-`^`.

That means `manager` and `plugin-kit` would use `^` and `slice-machine-ui` would not. What do you think? I'll update the other `package.json`s accordingly.

packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
angeloashmore: This came from supporting "Repeatable Group" and "Group." Since they are both group fields, they require a custom identifier. `CUSTOM_NAME` is already used for "Link to media," so it followed the convention.

This version is more accurate than one without the `CUSTOM_NAME` check since it's possible a "synthetic" field type (e.g. "Link to media") requires a specific `widget`. Let me know if I misunderstood the purpose of `widget` though!

packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
angeloashmore: Yep, that's right. There was some duplicate logic I was able to clean up. 🙂

packages/slice-machine/src/domain/group.ts
angeloashmore: **Edit**: Nevermind, ignore what I wrote below. I think I understand what you are suggesting better now. I will try something and report back. 🙂

This is tricky to do now because `domain/slice.ts` always works with `ComponentUI` (which essentially includes something like `SliceSM`). We would need to do `Groups.toSM` and `Groups.fromSM` in multiple areas.

I think we don't see type errors now because the slice-related code uses `any`. I suspect we have some incorrect code though.

I agree we should remove all `*SM` types as it's a big source of indirection. Would you be okay with me refactoring `domain/slice.ts`, and the related `domain/group.ts`, in a separate PR?

xrutayisire: Super good point! I forgot that `ComponentUI` make everything trickier :/ 
You're totally right, it's in a separate PR but even more this will be tackle when we will handle this specific tech debt topic. 👍

angeloashmore: I edited `domain/group.ts` to use `Group` instead of `GroupSM`. You can see the commit here: https://github.com/prismicio/slice-machine/pull/1335/commits/d377fe21578eb3fc0a25c60f3beadac36d4f8f4a

I needed to add `Groups.toSM` and `Groups.fromSM` when data goes in and out of the helpers. Although it seems a bit messy now, it should removing `*SM` types easier in the future.

packages/slice-machine/src/domain/group.ts
angeloashmore: Done! https://github.com/prismicio/slice-machine/pull/1335/commits/e96f967697c5136cf2f9d4891350bf79185e729a

playwright/tests/slices/sliceBuilderFields.spec.ts
angeloashmore: Good call. I added two checklist items to the ticket.

playwright/tests/slices/sliceBuilderFields.spec.ts
angeloashmore: I never would have though Playwright tests could be so simple and easy to write. It's a sign of a great setup. 😄

packages/slice-machine/lib/models/common/widgets/Group/ListItem/index.jsx
angeloashmore: Basically, the previous field accessor paths did not actually work if you copy/pasted them into a project. `item.` is what projects actually use (the name `item` is used by convention).

I tried to describe this in the PR description, which I'll copy below:

> ### Data path and snippets updates
> 
> The data path and snippets (which use the data path) were updated to more closely reflect how the data is used.
> 
> **Group fields:**
> - Before: `slice.primary.my_group.[...]` or `data.my_group.[...]`
> - After: `slice.primary.my_group` or `data.my_group`
> 
> **Fields within groups:**
> - Before: `slice.primary.my_group.my_field` or `data.my_group.my_field`
> - After: `item.my_field`
> 
> Note that fields within groups always use `item`, which matches how they are used in `my_group.map(...)`:
> 
> ```ts
> slice.data.my_group.map(item => {
>   // Do something with item.my_field
> })
> ```

playwright/pages/SliceBuilderPage.ts
angeloashmore: I went with this approach: https://github.com/prismicio/slice-machine/pull/1335/commits/f36caa935b7c978ab8c280d74f5eee810b604e54

```ts
getListItem<TZoneType extends ZoneType>(
  fieldId: string,
  zoneType: TZoneType,
  groupFieldId?: TZoneType extends "static" ? string : never,
): Locator
```

It lets us use the same practice between positional parameters and object parameters. It means less code sharing between functions, but I think it's extra clear this way. What do you think?

packages/slice-machine/src/domain/group.ts
angeloashmore: You're right, we can remove them. They are gone now!

packages/slice-machine/src/features/models/ModelProvider.tsx
angeloashmore: Since the data is strictly used for tracking, let's keep it simple and use the route to determine the custom type. The provider is too heavy-handed for this purpose and adds complexity.

angeloashmore: Commit: https://github.com/prismicio/slice-machine/pull/1335/commits/0b01b237e4f3935716315634376fd46ead83c85a

packages/slice-machine/lib/builders/SliceBuilder/FieldZones/index.tsx
angeloashmore: The `FieldType` parameter type is incorrect. `widgetTypeName` can also be Slice Machine-specific types like `"ContentRelationship"` and `"LinkToMedia"`.

I updated the type here: https://github.com/prismicio/slice-machine/pull/1335/commits/6f4c723c8ef4916a69b65bbe5cda862f05239727

packages/slice-machine/package.json
xrutayisire: Yes, for me what we have today where slice-machine is using pinned version 👍

playwright/pages/SliceBuilderPage.ts
xrutayisire: It's perfect!

xrutayisire (approved):
LGTM! Super work! 🚀🚀🚀🚀

PR #1338, opened 2024-04-18
refactor(slice-machine-ui): convert `ContentTabs` to CSS Modules

angeloashmore:
## Context

DT-2029

## The Solution

- Convert `ContentTabs` mostly 1:1 to CSS Modules.
- The trigger's `box-shadow` was replaced with an `::after` pseudo-element because `@prismicio/editor-ui`'s Stylelint config does not allow custom `box-shadow` values.

## Impact / Dependencies

N/A

- [x] The CI is successful.


packages/slice-machine/src/components/ContentTabs/ContentTabs.module.css
```diff
@@ -0,0 +1,66 @@
...
+  position: relative;
+
+  &:not(:last-child) {
+    margin-right: 16px;
+  }
+
+  &[data-state="active"] {
+    color: var(--purple9);
+
+    &::after {
```
xrutayisire: 💡 #suggestion

I checked the `editor-ui` and for very specific box-shadow they just are ignoring manually the rule. This ensures consistency by default, and you have to explicitly ignore to override. I think it's what we want here, WDYT?

Example: https://github.com/prismicio/editor/blob/main/packages/editor-ui/src/components/ColorInput/ColorSlider.module.css

xrutayisire: I guess no need of the position: relative if we change that, right?

xrutayisire (approved):
I let you handle my only comment with what you think 🙂 I think it's better to be aligned with editor practices.

LGTM for the rest! 👍

packages/slice-machine/src/components/ContentTabs/ContentTabs.module.css
angeloashmore: Oh cool, I missed that in the `editor-ui` repo. I had to use a `stylelint-disable` rather than `stylelint-disable-next-line` because Prettier formats the property into two lines. Should be good now!

xrutayisire:
LGTM! 👍

PR #1344, opened 2024-04-19
refactor(vanilla extract): Convert Vanilla Extract to CSS Modules for CodeBlock

xrutayisire:
## Context

- Completes DT-2025

## The Solution

- Simple 1-1 conversion to CSS modules


angeloashmore (approved):
👍 Looks good!

PR #1345, opened 2024-04-19
refactor(slice-machine-ui): convert `List` to CSS Modules

bapmrl:
Completes DT-2018.

**⚠️ TODO:**
- [x] Find a way to replace the selectors that are using `windowStyles`.


angeloashmore:
I completed this PR by adding an `inset` prop. It lets the `List` consumer decide whether the component is rendered with a border and radius rather than it being implicit depending on the DOM order.

```diff
  <Window>
    <WindowTabs>
      <WindowTabsContent>
-       <List />
+       <List inset />
      </WindowTabsContent>
    </WindowTabs>
  </Window>
```

xrutayisire:
> I completed this PR by adding an `inset` prop. It lets the `List` consumer decide whether the component is rendered with a border and radius rather than it being implicit depending on the DOM order.
> 
> ```diff
>   <Window>
>     <WindowTabs>
>       <WindowTabsContent>
> -       <List />
> +       <List inset />
>       </WindowTabsContent>
>     </WindowTabs>
>   </Window>
> ```

Sorry to be a pain in advanced haha
Could you ask PB if they think this is the best way? I personally like it, but I wonder what they did on their side since they converted a lot of Vanilla Extract components.

angeloashmore:
@xrutayisire Haha, not a pain at all!

[I asked in #frontend-engineering](https://prismic-team.slack.com/archives/C06N6107ZG8/p1716411987064509) and @dani-mp suggested we replace `inset={true}` with `border={false}`. The `border` prop is more easily understood than `inset` when you don't know the component's specifics.

xrutayisire (approved):
LGTM 🙂

angeloashmore:
Maybe we should wait a bit to give others a chance to reply in the thread and, if no changes are needed, merge tomorrow.

Edit: Or we can merge now if it's good with you!

xrutayisire:
As you want for the merge, not urgent true so anyway are ok on my side 👍

angeloashmore:
Okay, I'll go ahead and merge. I can open a new PR another time if we need to make changes.

PR #1346, opened 2024-04-22
improvement(sideNav): increase css mobile breakpoint

levimykel:
## Context
Completes [DT-2112](https://linear.app/prismic/issue/DT-2112)

Increases the mobile side navigation breakpoint from `small` to `medium`.

## The Solution
- Increase the breakpoint in CSS files from `767px` to `1011px`
- Update side nav components to include the medium breakpoint when using `useMediaQuery`

## Impact / Dependencies
- None


packages/slice-machine/components/Navigation/ChangesItem.tsx
```diff
@@ -32,6 +32,8 @@ export const ChangesItem: FC = () => {
   const router = useRouter();
   const { autoSyncStatus } = useAutoSync();
   const isSmall = useMediaQuery(breakpoints.small);
+  let isMedium = useMediaQuery(breakpoints.medium);
+  isMedium = isMedium || isSmall;
```
bapmrl: 💡 Sorry, I've thought about it, and I think your `const isBreakpoint = isSmall || isMedium` was clearer 🙂 It uses a `const` instead of a `let` and it doesn't mutate an already assigned variable.

bapmrl (changes requested):
Nice first PR on your first day in the team 👍

I just have 2 questions (❓) and 1 suggestion (💡).

1. ❓ Did you check it works in both the app and Storybook?
2. ❓ Have you shared a proposal on the `#page-builder-dev` channel to improve the API for using and composing breakpoints?

packages/slice-machine/src/components/SideNav/SideNav.tsx
```diff
@@ -119,9 +118,9 @@ export const SideNavLink: FC<SideNavLinkProps> = ({
   component: Comp = "a",
   ...otherProps
 }) => {
-  const isSmall = useMediaQuery(breakpoints.small);
+  const isMedium = useMediaQuery({ max: "medium" });
```
bapmrl: 💡 I think just `const visible` would make more sense, because `isMedium` doesn't really reflect what `{ max: "medium" }` means. Indeed, `{ max: "medium" }` matches both the `small` and `medium` breakpoints.

packages/slice-machine/src/components/SideNav/SideNavEnvironmentSelector.tsx
```diff
@@ -47,14 +46,14 @@ export const SideNavEnvironmentSelector: FC<SideNavEnvironmentSelectorProps> = (
     onSelect,
   } = props;
 
-  const isSmall = useMediaQuery(breakpoints.small);
+  const isMedium = useMediaQuery({ max: "medium" });
```
bapmrl: 💡 Same comment. Maybe `const collapsed` would make more sense?

packages/slice-machine/src/legacy/components/Navigation/ChangesItem.tsx
```diff
@@ -31,7 +30,7 @@ export const ChangesItem: FC = () => {
   const open = useOpenChangesHoverCard();
   const router = useRouter();
   const { autoSyncStatus } = useAutoSync();
-  const isSmall = useMediaQuery(breakpoints.small);
+  const isMedium = useMediaQuery({ max: "medium" });
```
bapmrl: 💡 Same comment. Maybe `const collapsed` would make more sense?

bapmrl (changes requested):
Nearly there 🙂

packages/slice-machine/src/components/SideNav/SideNav.tsx
levimykel: Yeah, I agree that this is better.

bapmrl (approved):
Excellent!

PR #1347, opened 2024-04-23
chore(slice-machine-ui): replace all import aliases with `@/*`

angeloashmore:
## Context

https://prismic-team.slack.com/archives/C02L3FN3AJK/p1712611867217019

## The Solution

This PR replaces all import aliases with `@/*`:

| Before        | After                   |
| ------------- | ----------------------- |
| `@src`        | `@`                     |
| `@components` | `@/legacy/components`   |
| `@lib`        | `@/legacy/lib`          |
| `@builders`   | `@/legacy/lib/builders` |
| `@models`     | `@/legacy/lib/models`   |
| `@utils`      | `@/legacy/lib/utils`    |
| `test` (no `@`)     | `@/../test`    |

The `@/../test` import is ugly and unintuitive, but should be temporary. The alias is only used in tests where we use a custom set of [Testing Library](https://testing-library.com/) functions. We plan to remove integration tests since they have been replaced by Playwright.

In addition to replacing import aliases, this PR moves the top-level `components` and `lib` directories into a new `src/legacy` directory. The `legacy` name serves as a sign to developers that the code is planned to be removed or modernized.

Lastly, this PR moves the top-level `pages` directory to `src/pages`.

All product code is now in `src`, accessible using the `@/*` alias.

## Impact / Dependencies

All new code should use the `@/*` import alias when importing code from the `src` directory. `tsserver`-powered autocomplete should use the alias automatically.

- [x] The CI is successful.


xrutayisire (approved):
LGTM! ❤️

I think it's will be super clear what should be well maintain or not. Even in src we will start maintaining it in a better way, no need for top folders like before

PR #1351, opened 2024-04-24
fix: restyle the field zone empty state

angeloashmore:
## Context

DT-2116

## The Solution

- Use a darker background color in the empty state.
- Update the empty state's text.
- Replace `theme-ui` with `@prismicio/editor-ui` and custom CSS.

## Impact / Dependencies

N/A

- [x] The CI is successful.

## Preview

This PR affects all builder screens: custom types, page types, and slices.

### Slices

**Without fields**:
localhost_9999_slices_ --src--slices_RichText_default

**With the repeatable zone**
Note: This state will not be possible in the near future once the Repeatable Zone is no longer available. Emptying the zone will remove it.
localhost_9999_slices_ --src--slices_RichText_default (3)

**With fields**:
localhost_9999_slices_ --src--slices_RichText_default (1)

**With fields and the repeatable zone**:
localhost_9999_slices_ --src--slices_RichText_default (2)

### Page/Custom Types

**Without fields**:
localhost_9999_custom-types

**With fields**:
localhost_9999_custom-types (1)

### DIFF
--- a/packages/slice-machine/src/components/List/List.css.ts
+++ b/packages/slice-machine/src/components/List/List.css.ts
@@ -50,6 +50,11 @@ export const header = style([
   }),
   {
     selectors: {
+      ["&:not(:first-child)"]: {
+        borderTopStyle: "solid",
+        borderTopColor: vars.color.greyLight6,
+        borderTopWidth: 1,
+      },
       [`${windowStyles.tabsContent} > ${root} > &:not(:first-child)`]: {
         marginTop: vars.space[16],
       },
--- a/packages/slice-machine/src/components/List/List.css.ts
+++ b/packages/slice-machine/src/components/List/List.css.ts
@@ -55,9 +55,6 @@ export const header = style([
         borderTopColor: vars.color.greyLight6,
         borderTopWidth: 1,
       },
-      [`${windowStyles.tabsContent} > ${root} > &:not(:first-child)`]: {
-        marginTop: vars.space[16],
-      },
     },
   },
 ]);
--- a/packages/slice-machine/src/legacy/lib/builders/common/Zone/components/ZoneEmptyState/ZoneEmptyState.module.css
+++ b/packages/slice-machine/src/legacy/lib/builders/common/Zone/components/ZoneEmptyState/ZoneEmptyState.module.css
@@ -1,4 +1,8 @@
 .root {
+  display: flex;
+  align-items: center;
+  flex-direction: column;
+  gap: 16px;
   background-color: var(--grey3);
   text-align: center;
   padding-top: 32px;
--- a/packages/slice-machine/src/legacy/lib/builders/common/Zone/components/ZoneEmptyState/ZoneEmptyState.module.css
+++ b/packages/slice-machine/src/legacy/lib/builders/common/Zone/components/ZoneEmptyState/ZoneEmptyState.module.css
@@ -1,0 +1,6 @@
+.root {
+  background-color: var(--grey3);
+  text-align: center;
+  padding-top: 32px;
+  padding-bottom: 32px;
+}
--- a/packages/slice-machine/src/legacy/lib/builders/common/Zone/components/ZoneEmptyState/ZoneEmptyState.module.css
+++ b/packages/slice-machine/src/legacy/lib/builders/common/Zone/components/ZoneEmptyState/ZoneEmptyState.module.css
@@ -1,10 +1,12 @@
 .root {
   display: flex;
-  align-items: center;
   flex-direction: column;
+  align-items: center;
+  justify-content: center;
   gap: 16px;
   background-color: var(--grey3);
   text-align: center;
   padding-top: 32px;
   padding-bottom: 32px;
+  min-height: 288px;
 }
--- a/packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/TabZone/index.tsx
+++ b/packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/TabZone/index.tsx
@@ -242,6 +242,7 @@ const TabZone: FC<TabZoneProps> = ({ tabId }) => {
               }
               testId="static-zone-content"
               isRepeatableCustomType={customType.repeatable}
+              emptyStateActionTestId="add-Static-field"
             />
           ) : undefined}
 
--- a/packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/TabZone/index.tsx
+++ b/packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/TabZone/index.tsx
@@ -215,6 +215,7 @@ const TabZone: FC<TabZoneProps> = ({ tabId }) => {
           {query.newPageType === undefined ? (
             <Zone
               zoneType="customType"
+              zoneTypeFormat={customType.format ?? "custom"}
               // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
               tabId={tabId}
               title="Static Zone"
--- a/packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
+++ b/packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
@@ -212,7 +212,9 @@ const FieldZones: FC = () => {
         }
         testId="static-zone-content"
         isRepeatableCustomType={undefined}
-        emptyStateHeading="No fields"
+        emptyStateHeading={
+          groupsInSlicesExperiment.eligible ? undefined : "No fields"
+        }
         emptyStateActionTestId="add-Static-field"
       />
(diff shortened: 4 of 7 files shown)


packages/slice-machine/src/components/List/List.css.ts
```diff
@@ -50,8 +50,10 @@ export const header = style([
   }),
   {
     selectors: {
-      [`${windowStyles.tabsContent} > ${root} > &:not(:first-child)`]: {
-        marginTop: vars.space[16],
+      ["&:not(:first-child)"]: {
```
bapmrl, line 53: :warning: I'm not sure about this change @angeloashmore.

Indeed, in Figma, I don't see anything indicating we no longer want 16 px of space when the `ListHeader` is used inside a `Window` ([here](https://www.figma.com/file/wb7ujCJ4j7EiP7aDDGRYXB/SliceMachine%E2%84%A2-UI?type=design&node-id=6938-202002&mode=design&t=Yc7ROraoOlrJEFOy-4), you can still see the vertical space). Also, I don't see anything indicating we want a top border.

I know @benjammartin wants to change the background color of the empty state, as shown [here](https://www.figma.com/file/wb7ujCJ4j7EiP7aDDGRYXB/SliceMachine%E2%84%A2-UI?type=design&node-id=11703-467431&mode=design&t=Yc7ROraoOlrJEFOy-4), but we need to know how it impacts the `List` as a whole and in all its use cases.

packages/slice-machine/src/components/List/List.css.ts
```diff
@@ -50,8 +50,10 @@ export const header = style([
...
+        borderTopStyle: "solid",
```
bapmrl, line 54: :warning: I think this change will display a solid border just after a dashed border if the consumer renders the following:
```
<List>
  {/* ... */}
  <ListItem />
  <ListHeader />
</List>
```

packages/slice-machine/src/legacy/lib/builders/common/Zone/components/EmptyState/index.tsx
```diff
@@ -1,33 +0,0 @@
-import { Button } from "@prismicio/editor-ui";
```
bapmrl, line 1: :bulb: Where's the closest `BaseStyles` ancestor? Maybe it can be moved to wrap only the components that are still using `theme-ui` in the builder? Maybe it can be removed? Maybe not ^^

packages/slice-machine/src/legacy/lib/builders/common/Zone/components/ZoneEmptyState/ZoneEmptyState.module.css
```diff
@@ -0,0 +1,12 @@
+.root {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  gap: 16px;
+  background-color: var(--grey3);
+  text-align: center;
+  padding-top: 32px;
```
bapmrl, line 9: :question: What should be the horizontal padding? Reducing the viewport size, the text touches the borders of the `List`.

packages/slice-machine/src/legacy/lib/builders/common/Zone/components/ZoneEmptyState/ZoneEmptyState.module.css
```diff
@@ -0,0 +1,12 @@
...
+  padding-bottom: 32px;
```
bapmrl, line 10: :question: Why 32 px of vertical padding? Where did you find this value? We need to ask @benjammartin.

bapmrl (changes requested):

packages/slice-machine/src/legacy/lib/builders/common/Zone/components/ZoneEmptyState/ZoneEmptyState.module.css
```diff
@@ -0,0 +1,12 @@
+.root {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  gap: 16px;
+  background-color: var(--grey3);
```
bapmrl, line 7: ❓ Do we want to change the background color of the empty state when it's used inside a `Window`?

packages/slice-machine/src/components/List/List.css.ts
angeloashmore: Slice Machine UI doesn't use `ListItem` currently. Should we remove the component, which would simplify `List`'s implementation?

For reference, the solid border was to prevent this UI where the new empty state is shown above a `ListHeader`:

image

Note the dark gray to light gray transition, which looks odd in my opinion.

packages/slice-machine/src/components/List/List.css.ts
angeloashmore: I suppose we need to determine how the empty state affects the styling of its surrounding components.

Here is an example without the border and with the margin where the empty state's background creates odd spacing:

image

This is what the component looks like with the border and without the margin:

localhost_9999_custom-types

angeloashmore: @benjammartin Could we get your design guidance?

packages/slice-machine/src/legacy/lib/builders/common/Zone/components/EmptyState/index.tsx
angeloashmore: There is a `BaseStyles` immediately around `ZoneEmptyState`. Since we aren't using `theme-ui` anymore, we can remove it, as you suggested. 🙂 

I don't think I should go trough every `BaseStyle` and refactor its position, but we can do it here for sure.

packages/slice-machine/src/legacy/lib/builders/common/Zone/components/ZoneEmptyState/ZoneEmptyState.module.css
angeloashmore: 32px comes from the original empty zone, before this redesign:

```tsx
<Box sx={{ textAlign: "center", my: 4 }}>
```

It can be adjusted per @benjammartin's recommendation, of course.

packages/slice-machine/src/legacy/lib/builders/common/Zone/components/ZoneEmptyState/ZoneEmptyState.module.css
angeloashmore: Good question. The original empty state didn't have horizontal padding either: 
image

16px seem to work well, but I'll let @benjammartin recommend something:
image

packages/slice-machine/src/legacy/lib/builders/common/Zone/components/ZoneEmptyState/ZoneEmptyState.module.css
angeloashmore: I'm not sure. Is there a reason we would want to do it in one place but not the other? cc @benjammartin

benjammartin: Yes, it's a temporary solution while waiting for the redesign of this screen. The idea is to highlight the empty area to encourage the user to add fields. There is a redesign of this section coming up.

packages/slice-machine/src/components/List/List.css.ts
bapmrl: > Slice Machine UI doesn't use `ListItem` currently. Should we remove the component, which would simplify `List`'s implementation?

Indeed, but at the time I think we were considering reworking the fields. That's why, in Storybook, I added `ListItem`. It helped me structure the `List`'s API and layout. We could remove it, but I don't think this is the real issue. We should first know what we want to do exactly in the UI ^^

packages/slice-machine/src/legacy/lib/builders/common/Zone/components/EmptyState/index.tsx
bapmrl: It's perfect ^^ You don't have to go through every `BaseStyles`.

samlfair:
@angeloashmore The text looks good; however, I think it could be improved. What do you think of this?

```md
**Your (slice|page type|custom type) has no fields yet**

A field is an input for content (e.g. text, images, links)

[Add a field]
```

This doesn't work for slices with the repeatable zone, so we could leave those ones as just `No fields`.

Also I would prefer the singular `A field is an input` over the plural `Fields are inputs`.

angeloashmore:
@samlfair I updated the messages:

**Custom type**
image

**Page type**
image

**Slice without repeatable zone**
image

**Slice with repeatable zone**
image

samlfair:
Looks great!

xrutayisire:
If I understood, we can merge @angeloashmore, I let you be sure and handle it 🙂

angeloashmore:
Great! We were still waiting on guidance from @benjammartin, but we can revisit it later.

PR #1356, opened 2024-04-26
test(playwright): speed up tests

angeloashmore:
## Context

Playwright tests are slower than they could be because we use the UI to generate fixtures. Playwright tests are currently the slowest part of the CI.

## The Solution

Use `@slicemachine/manager` to generate fixtures instead. Using the manager is nearly instant since we are not waiting on page loads or user interactions to complete.

This PR also introduces `@prismicio/mock` to generate mock models.

## Impact / Dependencies

Playwright tests will run more quickly, resulting in a faster feedback loop.

- [ ] The CI is successful.


angeloashmore:
I decided against these changes. You can find the reasonings here: [Notion](https://www.notion.so/prismic/dd598e35e1a0427d9e7ef10223b646aa)

PR #1362, opened 2024-05-01
refactor(slice-machine-ui): use `Table` from `@prismicio/editor-ui`

angeloashmore:
## Context

DT-2019

## The Solution

- Replace use of `Table` with `@prismicio/editor-ui`'s `Table`.
- Remove the empty space around the "Repeatable" and "Single" document icons.
- Remove the `Table` component set.
- Remove the `CustomTypesTablePage` Vitest tests. The tests broke as a result of this PR, and we already have Playwright tests to cover the page. It isn't worth our time to fix the Vitest version.

## Impact / Dependencies

N/A

- [ ] The CI is successful.

## Preview

image


packages/slice-machine/src/features/customTypes/customTypesTable/CustomTypesTable.module.css
```diff
@@ -0,0 +1,4 @@
+.ellipsis {
```
bapmrl: :warning: You can replace `ellipsis` with the `Text` component's `noWrap` prop.

packages/slice-machine/src/icons/ReusableIcon.tsx
```diff
@@ -1,23 +1,17 @@
...
-    <path d="M23 9H26L21 4V7C21 8.10457 21.8954 9 23 9Z" fill="#9E8CFC" />
-    <path
-      fillRule="evenodd"
-      clipRule="evenodd"
-      d="M16.0912 12C13.2794 12 11 14.2794 11 17.0912C11 19.903 13.2794 22.1824 16.0912 22.1824C17.9763 22.1824 19.6215 21.1577 20.501 19.6372C20.6563 19.3688 20.7878 19.0849 20.8926 18.7883C20.9846 18.5279 20.8482 18.2422 20.5878 18.1502C20.3275 18.0582 20.0418 18.1947 19.9498 18.455C19.8657 18.693 19.7601 18.9209 19.6354 19.1365C18.9273 20.3606 17.6049 21.1824 16.0912 21.1824C13.8317 21.1824 12 19.3507 12 17.0912C12 14.8317 13.8317 13 16.0912 13C17.4478 13 18.6507 13.66 19.3955 14.6782H17.6216C17.3455 14.6782 17.1216 14.9021 17.1216 15.1782C17.1216 15.4544 17.3455 15.6782 17.6216 15.6782H20.2998C20.576 15.6782 20.7998 15.4544 20.7998 15.1782V12.5C20.7998 12.2239 20.576 12 20.2998 12C20.0237 12 19.7998 12.2239 19.7998 12.5V13.6032C18.8717 12.6167 17.5536 12 16.0912 12Z"
-      fill="white"
-    />
+export const ReusableIcon: FC<SVGProps<SVGSVGElement>> = (props) => (
+  <svg viewBox="0 0 20 24" xmlns="http://www.w3.org/2000/svg" {...props}>
+    <g fill="none">
```
bapmrl: :question: Have you done something different to export the SVG? I'm asking because the elements have changed a bit. For example, the `g` element wasn't used before.

packages/slice-machine/src/features/customTypes/customTypesTable/CustomTypesTable.tsx
```diff
@@ -82,49 +87,80 @@ export const CustomTypesTable: FC<CustomTypesTableProps> = ({
...
-          <TableCell>Label</TableCell>
-          <TableCell>API ID</TableCell>
-          <TableCell>Limit</TableCell>
-          <TableCell />
-        </TableRow>
-      </TableHead>
-      <TableBody>
-        {sortedCustomTypes.map((customType: CustomType) => {
-          const { repeatable, label, id } = customType;
+    <div>
```
bapmrl, line 92: :bulb: I'm not asking you to update the `@prismicio/editor-ui` package, but it's unfortunate the `Table` component has `flex-grow: 1` on its `.root` class. It makes `Table` difficult to use in contexts where it shouldn't take the full height of the container. CC @AlexGalays.

packages/slice-machine/src/features/customTypes/customTypesTable/CustomTypesTable.tsx
```diff
@@ -82,49 +87,80 @@ export const CustomTypesTable: FC<CustomTypesTableProps> = ({
...
+                key={id}
+                onClick={() => {
+                  void router.push(
+                    CUSTOM_TYPES_CONFIG[format].getBuilderPagePathname(id),
+                  );
+                }}
+              >
+                <TableCell>
+                  {repeatable ? (
+                    <ReusableIcon width={20} />
```
bapmrl: :bulb: You could use the matching theme token here.

bapmrl (changes requested):
Excellent!

🕵️‍♂️ I've only noticed that the `Button`'s shadow is clipped, but I don't understand why. I don't think this is a blocking issue though.

Screenshot 2024-05-06 at 11 02 10

packages/slice-machine/src/features/customTypes/customTypesTable/CustomTypesTable.module.css
angeloashmore: Oh perfect! I didn't realize we had that built in.

packages/slice-machine/src/icons/ReusableIcon.tsx
angeloashmore: Good catch. I re-exported the SVGs without the extra embedded whitespace. The whitespace was causing sizing issues with `Table` since the component requires a set column template.

I ran the re-exported SVG through SVGO, and I guess it figured the `<g>` was necessary. I removed them manually and the icons seem to work the same as with them.

packages/slice-machine/src/features/customTypes/customTypesTable/CustomTypesTable.tsx
angeloashmore: I was surprised by that as well. There must be a reason why it has that by default, but it would be great to remove it. For now, the extra `<div>` prevents `Table` from growing.

packages/slice-machine/src/features/customTypes/customTypesTable/CustomTypesTable.tsx
angeloashmore: Done!

filip-prismic:
> Excellent!
> 
> 🕵️‍♂️ I've only noticed that the `Button`'s shadow is clipped, but I don't understand why. I don't think this is a blocking issue though.
> 
> Screenshot 2024-05-06 at 11 02 10

It is because TableCell has overflow class assigned by default and the outline spans beyond element bounds. Maybe we could provide prop to remove it, bcs in the cell like this you don't need it.
```css
.overflow {
  text-overflow: ellipsis;
  overflow: hidden;
  white-space: nowrap;
}
```

angeloashmore:
> It is because TableCell has overflow class assigned by default and the outline spans beyond element bounds. Maybe we could provide prop to remove it, bcs in the cell like this you don't need it.
> 
> ```css
> .overflow {
>   text-overflow: ellipsis;
>   overflow: hidden;
>   white-space: nowrap;
> }
> ```

Exposing a prop that controls the overflow feels like a leaky abstraction. Throwing a different idea out: what if we removed the 16px inline padding and moved it to the first and last cells in a row? That would prevent the overflow clipping.

Unfortunately, it means the first and last column sizes need to increase by 16px, which may have negative consequences:

```diff
- <Table columnLayout="28px 1fr 1fr 1fr 40px">
+ <Table columnLayout="44px 1fr 1fr 1fr 56px">
```

angeloashmore:
@bapmrl I fixed the issue by increasing the table column by 2px: https://github.com/prismicio/slice-machine/pull/1362/commits/2dc6d5a48fb2c3f8d62697e16d637d6ebaf82f8c

I think it's an okay solution for now. The change is barely visible:

image

packages/slice-machine/src/features/customTypes/customTypesTable/CustomTypesTable.tsx
xrutayisire: Arf, not super great true, did you asked why to editor @angeloashmore?

Anyway, should not block a merge we can quickly remove it the day they remove it also

xrutayisire (approved):
LGTM! Let's merge and keep open the two small things that are not sure good because. of current editor Table implementation.

filip-prismic:
> what if we removed the 16px inline padding and moved it to the first and last cells in a row? That would prevent the overflow clipping.

@angeloashmore  seems like patch for a problem we should not be dealing with. I would rather ask why the overflow is there for all the columns  by default in the first place. Seems weird to me :D 

But agree this one shouldn't be a blocker for you and 2px size change is a good enough right now.

angeloashmore:
@filip-prismic I quickly looked at it closer and I agree; it doesn't seem like the overflow styles are needed. It could be something handled by the cell's children instead.

But let's move forward with the 2px fix for now.

PR #1364, opened 2024-05-02
fix(slice-machine-ui): Sidebar's Gradient

bapmrl:
Completes DT-2121.

The `Box`'s implementation changed and we can no longer use the `style` prop to pass `marginBottom`. Hence, I'm now using an additional `div`. A better solution is specified in [PBD-1080](https://linear.app/prismic/issue/PBD-1080/merge-gradient-into-more-dx-friendly-scrollarea), but we don't need to implement it right now.

### DIFF
--- a/packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
+++ b/packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
@@ -1,4 +1,4 @@
-import { Box, Button, Gradient } from "@prismicio/editor-ui";
+import { Box, Button, Gradient, theme } from "@prismicio/editor-ui";
 import { useRouter } from "next/router";
 import { type FC, useState } from "react";
 import { toast } from "react-toastify";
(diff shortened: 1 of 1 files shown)


packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
```diff
@@ -73,32 +73,40 @@ export const Sidebar: FC = () => {
...
+         * the bottom of the page.
+         * TODO(PBD-1080): write the new `ScrollAreaEndGradient` component
+         * instead of using a `div`.
+         */}
+        <div
+          style={{
+            bottom: theme.space[0],
+            marginBottom: `-${theme.space[16]}`,
+            position: "sticky",
+          }}
```
angeloashmore, line 88: ❓ How do we decide whether to use a CSS class or a `style` prop?

❓ How do we decide when to use `theme` (e.g. `theme.space[16]`) or hard-coded values (e.g. 16px)?

angeloashmore (approved):
LGTM! I have two questions for my own knowledge, but the PR itself is good. 🙂

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
bapmrl: > ❓ How do we decide whether to use a CSS class or a `style` prop?

I'm personally not for writing `.module.css` files outside of `src/components`. I think most UI elements can be implemented as well designed React components. When styles are only contextual and cannot be re-used elsewhere, it's okay to use `Box`. When `Box` isn't suitable, it's okay to use a `div` (because we don't want to clobber `Box`'s API with concerns that could be better addressed by a dedicated component, in this case, `ScrollAreaEndGradient`).

> ❓ How do we decide when to use `theme` (e.g. `theme.space[16]`) or hard-coded values (e.g. 16px)?

I think it's better to use the `theme` whenever possible. It should be quite rare to use "non-standard" values.

PR #1365, opened 2024-05-07
feat(slice-machine-ui): add horizontal scroll for slice variations

levimykel:
## Context

[DT-1799](https://linear.app/prismic/issue/DT-1799/aauser-i-expect-variations-to-scroll-horizontally-on-medium-widths)

## The Solution

Add the `useMediaQuery` hook to switch from a vertical list to a horizontal scroll at the `medium` breakpoint. 

- [ ] The CI is successful.

## Preview

https://github.com/prismicio/slice-machine/assets/9385378/934bd618-0d33-4049-92e4-86b1d64d50a9


packages/slice-machine/src/icons/SliceVariationsIcon.tsx
```diff
@@ -0,0 +1,24 @@
+import { FC } from "react";
+
+// This is a placeholder until I can get the actual SVG code from Benjamin
+export const SliceVariationsIcon: FC = () => (
+  <svg
+    width="24"
+    height="24"
+    viewBox="0 0 32 32"
```
bapmrl: :bulb: Nice, but when you'll replace this placeholder with the real icon, I think it should have:
1. A `viewBox` width that matches the `svg`'s `width` prop.
1. A `viewBox` height that matches the `svg`'s `height` prop.

packages/slice-machine/src/icons/SliceVariationsIcon.tsx
```diff
@@ -0,0 +1,24 @@
+import { FC } from "react";
```
bapmrl: :bulb: Not everyone does it in the team, but if you want you can make your imports a bit more specific by adding the `type` keyword. That way, when you read your imports, you can quickly get a grasp of what's a value and what's a type. I think there's a TypeScript configuration to enforce that, but we haven't enabled (and discussed) it.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/index.tsx
```diff
@@ -20,6 +20,16 @@ import { Sidebar } from "./Sidebar";
...
   const { slice, actionQueueStatus } = useSliceState();
+  const horizontalScroll = useMediaQuery({ max: "medium" });
+
+  const contentDisplayProps = horizontalScroll
+    ? {
+        gridTemplateRows: "320px 1fr",
+      }
+    : {
+        gridTemplateColumns: "320px 1fr",
+        gap: 16 as const,
```
bapmrl: :warning: In Figma, there's a `16px` `gap` in both cases so why specify it only when `horizontalScroll` is `false`?

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/index.tsx
```diff
@@ -20,6 +20,16 @@ import { Sidebar } from "./Sidebar";
 
 export const SliceBuilder: FC = () => {
   const { slice, actionQueueStatus } = useSliceState();
+  const horizontalScroll = useMediaQuery({ max: "medium" });
+
+  const contentDisplayProps = horizontalScroll
+    ? {
+        gridTemplateRows: "320px 1fr",
```
bapmrl: :question: How did you find the `320px` value? There's `64px` between the `Divider` and the top of the `SliceCard`. The card is `240px` tall, so shouldn't it be `304px`?

Maybe you specified `320px` because you're excluding the margin between the bottom of the card and the top of `FieldZones`? If that's the case, using `304px` will allow you to always specify the same `16px` `gap` in the `Box` where you spread the `contentDisplayProps` object.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/index.tsx
```diff
@@ -20,6 +20,16 @@ import { Sidebar } from "./Sidebar";
 
 export const SliceBuilder: FC = () => {
   const { slice, actionQueueStatus } = useSliceState();
+  const horizontalScroll = useMediaQuery({ max: "medium" });
+
+  const contentDisplayProps = horizontalScroll
+    ? {
```
bapmrl: :bulb: If you want, you can manually remove the comma so that the objects take less vertical space.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/index.tsx
```diff
@@ -35,13 +45,8 @@ export const SliceBuilder: FC = () => {
...
       </AppLayoutHeader>
       <AppLayoutContent>
-        <Box
-          alignItems="flex-start"
-          display="grid"
-          gap={16}
-          gridTemplateColumns="320px 1fr"
-        >
-          <Sidebar />
+        <Box display="grid" alignItems="flex-start" {...contentDisplayProps}>
```
bapmrl: :question: Should it also use `alignItems="flex-start"` when `horizontalScroll` is `true`?

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
```diff
@@ -22,7 +24,12 @@ type DialogState =
   | { type: "DELETE_VARIATION"; variation: VariationSM; loading?: boolean }
   | undefined;
 
-export const Sidebar: FC = () => {
+type SidebarProps = {
+  horizontalScroll?: boolean;
```
bapmrl: :warning: I think a required prop would make more sense as there's not really a "default state" for the layout. Both `horizontal` and `vertical` have equal importance, no?

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
bapmrl, line 1: :bulb: Maybe we could find a better name than `Sidebar` for this component, as it's no longer always displayed on the side. Maybe `SliceBuilderNavigationMenu`? I'm sure there's a better (/simpler) name.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
```diff
@@ -31,82 +38,87 @@ export const Sidebar: FC = () => {
...
-              color="grey"
-              onClick={() => {
-                setDialog({ type: "ADD_VARIATION" });
+        {horizontalScroll ? (
+          <>
+            <Divider variant="edgeFaded" color="grey6" />
+            <Box alignItems="center" gap={8}>
+              <SliceVariationsIcon />
+              <Box flexGrow={1}>
+                {variationCount} variation{variationCount !== 1 && "s"}
```
bapmrl, line 55: :warning: You should use the `Text` component to apply the proper font family, font size, etcetera.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
```diff
@@ -31,82 +38,87 @@ export const Sidebar: FC = () => {
...
-            <Button
-              color="grey"
-              onClick={() => {
-                setDialog({ type: "ADD_VARIATION" });
+        {horizontalScroll ? (
+          <>
+            <Divider variant="edgeFaded" color="grey6" />
+            <Box alignItems="center" gap={8}>
+              <SliceVariationsIcon />
+              <Box flexGrow={1}>
```
bapmrl: :warning: I think the DOM structure isn't clear enough. Something like this would removing the need to make the `Text` super large (with `flexGrow={1}`):
```
<Box justifyContent="space-between">
  <Box>
    <SliceVariationsIcon />
    <Text>3 variations</Text>
  </Box>
  <Button startIcon="add" />
</Box>
```

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
```diff
@@ -31,82 +38,87 @@ export const Sidebar: FC = () => {
...
+              style={{
+                width: "calc(100% + 32px)",
+                height: "calc(240px + 16px)",
+                position: "relative",
+                overflowX: "scroll",
+              }}
+            >
+              <div
+                style={{
+                  position: "absolute",
```
bapmrl, line 77: :warning: This seems a bit complicated. Would a negative `margin-right` work and help you cancel the `PageLayoutContent`'s `padding-right`?

In an ideal world, we could add a `padded` prop (or something better named) to specify whether the `PageLayout` components should include the default padding or not. But I don't think you should change that in this pull request, as it would force you to update other components that aren't related to your work here.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
```diff
@@ -31,82 +38,87 @@ export const Sidebar: FC = () => {
...
+                  paddingRight: "32px",
+                }}
+              >
+                <Box flexDirection="row" gap={16}>
+                  <SharedSliceCards width="320px" />
+                </Box>
+              </div>
+              <Gradient sx={{ position: "fixed", right: 0, height: "100%" }} />
```
bapmrl: :warning: Unless I'm mistaken, I don't see a `Gradient` when `horizontalScroll` is `true`.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
```diff
@@ -31,82 +38,87 @@ export const Sidebar: FC = () => {
...
+                  setDialog({ type: "ADD_VARIATION" });
+                }}
+                startIcon="add"
+                color="grey"
+              />
+            </Box>
+            <div
+              style={{
+                width: "calc(100% + 32px)",
+                height: "calc(240px + 16px)",
```
bapmrl: :warning: This seems a bit complicated. Can't you use a vertical margin instead of a `calc`?

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
```diff
@@ -163,3 +175,56 @@ export const Sidebar: FC = () => {
     </>
   );
 };
+
+type SharedSliceCardsProps = {
+  width?: `${string}px`;
```
bapmrl: :warning: I think `width: 320 | undefined` would better convey the intent here.

However, the real fix would probably be to use Radix UI's `AspectRatio` component. Maybe this could be done in a follow up issue? I think it should be added (back) to the `@prismicio/editor-ui` package (and discussed quickly with the Editor) team.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
```diff
@@ -163,3 +175,56 @@ export const Sidebar: FC = () => {
...
+};
+
+export const SharedSliceCards = (props: SharedSliceCardsProps) => {
+  const { width } = props;
+  const [_dialog, setDialog] = useState<DialogState>();
```
bapmrl: :warning: It's more usual to write `const [, setDialog]` I think.

bapmrl (commented):
Nice work Levi!

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/index.tsx
bapmrl: I think my calculations are wrong here, but you'll get the idea ^^

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/index.tsx
levimykel: I originally did this to add space for the scrollbar under the variations, but I found a different way to do it so that gap is kept for both cases.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/index.tsx
levimykel: Yes, I understand. It had to do with the scrollbar, but I've adjust everything to include space for the scrollbar another way.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/index.tsx
levimykel: Yes, both cases need it otherwise it extends the fields box all the way to the bottom of the page.

Screenshot 2024-05-07 at 16 47 33

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
levimykel: Yeah, good point.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
levimykel: I agree with this. What do you think about `VariationsList`?

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
levimykel: I took this from another example somewhere, but your solution is better.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
levimykel: A negative `margin-right` works.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
levimykel: It's there. It's a vertical gradient that I added to give a subtle clue that the variations are scrollable. It's easier to see when you scroll:

https://github.com/prismicio/slice-machine/assets/9385378/fc360914-6741-4d15-ab43-dd5181b7597d

Should I remove it?

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
levimykel: Do you mean just a normal height like this:

```
style={{
  height: "252px",
  ...
}}
```

Or something else?

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
levimykel: After discussing with @bapmrl, I'm removing it.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
levimykel: I discussed this with @bapmrl. He suggested using a height & padding or margin, but it needs to be combined into height to keep the scrollbar down below the items. I can set the height to 240px, though, if we don't care about that.

https://github.com/prismicio/slice-machine/assets/9385378/fed4cd3e-a937-413e-8fe2-4f72fe9161ae

packages/slice-machine/src/icons/SliceVariationsIcon.tsx
levimykel: After discussing with @benjammartin, we decided to remove the icon.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/Sidebar/index.tsx
levimykel: After discussing with @benjammartin, we decided to remove the scrollbar and add scroll snapping.

xrutayisire (approved):
LGTM, super work!! 🚀

It may be another topic and much another PR if needed, but I would like to see if we could not handle a different breakpoint specially for this change.

Like medium breakpoint for the side nav bar, but for this change, a large breakpoint. WDYT?

Also @benjammartin 

Because here it's sad that it's not trigger sooner, since the work is super nice.

Screenshot 2024-05-14 at 13 31 51

benjammartin:
I think we should try 1280px. I tested it on the browser, and it looks okay.

PR #1366, opened 2024-05-07
chore(packages): Update prismic-types-internal to 2.4.0

xrutayisire:
- Update @prismicio/types-internal to 2.4.0


xrutayisire:
Waiting release of editor-fields before setting it ready for review

PR #1367, opened 2024-05-07
refactor(slice-machine-ui): convert `Card` to CSS modules

levimykel:
## Context

[DT-2017](https://linear.app/prismic/issue/DT-2017/convert-card-to-css-modules)

## The Solution
- Perform a 1:1 conversion from Vanilla Extract to CSS Modules.

## Impact / Dependencies
N/A

- [x] The CI is successful.
- [ ] If there could be backward compatibility issues, it has been discussed and planned.


packages/slice-machine/src/components/Card/Card.module.css
```diff
@@ -0,0 +1,185 @@
...
+}
+
+.size-medium {
+  height: 320px;
+}
+
+.solid {
+  background-color: var(--grey3);
+}
+.solid[data-state="checked"] {
```
angeloashmore: 💡 You could use the `&` selector to nest rules. Using `&` is usually more maintainable than not since the parent selector only needs to be declared once:

```css
.solid {
  background-color: var(--grey3);

  &[data-state="checked"] {
    border-color: var(--purple10);
  }
}
```

`&` can be used in the other nested selectors as well, like `outlined` and `interactive`.

packages/slice-machine/src/components/Card/Card.module.css
```diff
@@ -0,0 +1,185 @@
...
+  box-shadow: var(--box-shadow-purple-focus);
+}
+.interactive[data-disabled] {
+  opacity: 0.5;
+}
+.interactive:not([data-disabled]) {
+  cursor: pointer;
+}
+
+.interactive.solid:focus-visible {
```
angeloashmore: 💡 General selectors are generally preferred over specific ones since they avoid specificity issues. In this case, `.interactive.solid` is more specific than something like `.interactive-solid`.

Before:
```css
.interactive.solid {
  /* Your interactive-solid variant styles... */
}
```

After:
```css
.interactiveBase {
  /* Your interactive styles... */
}

.interactive-solid {
  composes: interactiveBase;
  /* Your interactive-solid variant styles... */
}
```

This change would require adding/restoring an entry to `clsx`.

I'll let you decide which one you prefer since I can see this component is fairly complex. Chaining the class names like you have now might be simpler than what I am suggesting.

packages/slice-machine/src/components/Card/Card.module.css
```diff
@@ -0,0 +1,185 @@
...
+.outlined > .footer {
+  padding-inline: 16px;
+}
+
+.footerTexts {
+  composes: column;
+  flex-basis: 0;
+  flex-grow: 1;
+  min-width: 0px;
+  margin-bottom: 4px;
```
angeloashmore, line 190: 👍 Nice simplification! I find this version more readable. The comment from before might be beneficial to keep since it give some context around this magic value.

angeloashmore (changes requested):
It's looking good! This is a pretty complicated component to convert. I left a couple of suggestions below. 🙂

packages/slice-machine/src/components/Card/Card.module.css
levimykel: Ah, cool. I didn't realize you could do that with CSS modules. I'll make that change.

packages/slice-machine/src/components/Card/Card.module.css
levimykel: I'll leave this with the chaining as I did it. I see and understand your point and will do it that way on other componentns. For this one, though, since it's complex, I think the chaining makes the styling simpler.

levimykel:
@angeloashmore I responded to your comments and updated the PR based on your feedback. Let me know if you have any other feedback. If it looks good to you, then I'll merge.

angeloashmore (approved):
Looks good! Nice changes. 🙂

PR #1368, opened 2024-05-12
feat(slice templates): Update slice templates with repeatable content to use a gorup instead of items

xrutayisire:
## Context

- Completes DT-2086

## The Solution

- Updated slice template AlternateGrid to use a group in primary instead of items
  - group id: "items"
- Updated slice template CustomerLogos to use a group in primary instead of items
  - group id: "logos"
- Updated slice template CallToAction to remove unnecessary group item in mocks value

## Impact / Dependencies

- Should be merged ONLY when we are sure next release will be for group in slice!

- [x] The CI is successful.

## Preview

Screenshot 2024-05-12 at 23 58 21
Screenshot 2024-05-12 at 23 58 31

### DIFF
--- a/packages/adapter-next/public/AlternateGrid/javascript.jsx
+++ b/packages/adapter-next/public/AlternateGrid/javascript.jsx
@@ -58,9 +58,9 @@ const PascalNameToReplace = ({ slice }) => {
 							</div>
 						)}
 					</div>
-					{slice.items.length > 0 && (
+					{slice.primary.items.length > 0 && (
 						<div className="es-alternate-grid__primary-content__items">
-							{slice.items.map((item, i) => (
+							{slice.primary.items.map((item, i) => (
 								<div key={`item-${i + 1}`} className="es-alternate-grid__item">
 									{isFilled.richText(item.title) && (
 										<div className="es-alternate-grid__item__heading">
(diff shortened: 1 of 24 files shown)


lihbr (approved):
LGTM, thanks for taking care of that!

packages/adapter-next/public/AlternateGrid/javascript.jsx
```diff
@@ -58,9 +58,9 @@ const PascalNameToReplace = ({ slice }) => {
 							</div>
 						)}
 					</div>
-					{slice.items.length > 0 && (
+					{slice.primary.items.length > 0 && (
```
levimykel, line 61: Is there a plan to eventually move away from `primary` and/or change it to something like `data` or `fields`?

levimykel (approved):
Everything looks good to me. I added a question for my own curiosity.

packages/adapter-next/public/AlternateGrid/javascript.jsx
xrutayisire: That's a good question, we discussed it a while back because we wanted to do it at some point.

Here is the conclusion: https://prismic-team.slack.com/archives/C02L3FN3AJK/p1712825788916059?thread_ts=1712751482.676019&cid=C02L3FN3AJK

levimykel: Thanks for the context.

PR #1369, opened 2024-05-12
feat(group in slices): Remove amplitude exp feature flag check

xrutayisire:
## Context

- Completes DT-2105

## The Solution

- Remove group in slice feature flag
- Remove mock related to the feature flag in e2e tests

## Impact / Dependencies

- Don't merge if you don't want the next release to be group in slice release

- [ ] The CI is successful.

### DIFF
--- a/packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
+++ b/packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
@@ -20,7 +20,6 @@ import {
   updateField,
 } from "@/domain/slice";
 import { useSliceState } from "@/features/slices/sliceBuilder/SliceBuilderProvider";
-import { useGroupsInSlicesExperiment } from "@/features/slices/sliceBuilder/useGroupsInSlicesExperiment";
 import EditModal from "@/legacy/lib/builders/common/EditModal";
 import Zone from "@/legacy/lib/builders/common/Zone";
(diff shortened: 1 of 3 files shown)


packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
```diff
@@ -212,12 +205,10 @@ const FieldZones: FC = () => {
         }
         testId="static-zone-content"
         isRepeatableCustomType={undefined}
-        emptyStateHeading={
-          groupsInSlicesExperiment.eligible ? undefined : "No fields"
-        }
+        emptyStateHeading={undefined}
```
levimykel, line 208: Is there a particular reason that we specify `undefined` for all the unused props here?

levimykel (approved):
It's a straightforward change, and I approve the PR. I did, though, leave a question for my own curiosity.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
xrutayisire: That's a very good question. I have no idea. These are ollllddd components written in JSX not TSX. 

Maybe it's done like this because since it's not typed, you don't know the props at all. Because of this way, you know exactly what you are passing or not. 
I preferred keeping that way since I like consistency, but we could change all that of course.

PR #1371, opened 2024-05-20
Add Master Slice Library [BETA] modal

jaen:
## Context

There's a new experiment called Master Slice Library and it needs a call to action in the Slice Machine. See this ticket: https://linear.app/prismic/issue/AGE-274/experiment-for-slice-library-in-slicemachine

## The Solution

Replaces a sidebar navigation link we don't want anymore with one that opens a call to action modal for Master Slice Library. The modal shows a video preview and links to an example slice library and code necessary to try it. Modal and code button have added tracking.

## Impact / Dependencies

This doesn't rely on anything else and is a purely visual change, so impact should be minimal.


packages/slice-machine/src/features/masterSliceLibrary/SliceLibraryPreviewModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+import { telemetry } from "@/apiClient";
+import { useMarketingContent } from "@/hooks/useMarketingContent";
+
+type MasterSliceLibraryPreviewModalProps = {
+  isOpen: boolean;
+  onClose: () => void;
+};
+
+export const MasterSliceLibraryPreviewModal: React.FunctionComponent<
```
xrutayisire:
```suggestion
export const MasterSliceLibraryPreviewModal: FC<
```

packages/slice-machine/src/hooks/useMarketingContent.ts
```diff
@@ -17,6 +17,14 @@ type MarketingContent = {
       };
     };
   };
+  masterSliceLibrary: MasterSliceLibraryMarketingContent;
```
xrutayisire: I love the idea! Super nice, and aligned with our objectives. No direct content written in the component 🙌

I just think we could keep the same structure with:

```suggestion
  masterSliceLibrary?: {
  exampleLinkUrl: string;
  codeLinkUrl: string;
  previewVideoUrl: string;
};
```

Like others, no need of showInNavigation boolean, you just check that the object is defined.

packages/slice-machine/src/hooks/useMarketingContent.ts
```diff
@@ -53,6 +69,12 @@ const DEFAULT_CONTENT: MarketingContent = {
         "Follow our Quick Start guide to learn the basics of Slice Machine",
     },
   },
+  masterSliceLibrary: {
```
xrutayisire: And so nothing needed here, related to my other comment above

packages/slice-machine/src/legacy/components/Navigation/index.tsx
```diff
@@ -44,6 +46,9 @@ const Navigation: FC = () => {
   const { repositoryName, repositoryDomain, repositoryUrl } =
     useRepositoryInformation();
   const gitIntegrationExperiment = useGitIntegrationExperiment();
+  const [isSliceLibraryDialogOpen, setIsSliceLibraryDialogOpen] =
+    useState<boolean>(false);
```
xrutayisire: I think you don't need that 👍

```suggestion
    useState(false);
```

packages/slice-machine/src/hooks/useMarketingContent.ts
jaen: Okay, then in the component I have this:
```typescript
const { masterSliceLibrary } = useMarketingContent();
const { exampleLinkUrl, codeLinkUrl, previewVideoUrl } = masterSliceLibrary;
```
which will not compile with that change.

How would you prefer me to address that? I could use `masterSliceLibrary!` because if the modal is displayed we can be sure those values are set, or I could do this:
```typescript
const { masterSliceLibrary } = useMarketingContent();

if (!masterSliceLibrary) return null;

const { exampleLinkUrl, codeLinkUrl, previewVideoUrl } = masterSliceLibrary;
```
Do you have any preference?

packages/slice-machine/src/legacy/components/Navigation/index.tsx
jaen: This was what I tried at first, but I could swear it gave me a type error — I'm not getting one now, so I'm not sure what that was ~~" It certainly makes more sense for it to be inferred, if it can be.

packages/slice-machine/src/hooks/useMarketingContent.ts
xrutayisire: Ah I see! 
Hum, for me a mix of both I would say then. Good catch!

So for me you could continue to define a type and then pass it as a props:

```ts
type MasterSliceLibraryPreviewModalProps = {
  isOpen: boolean;
  masterSliceLibrary: MasterSliceLibraryMarketingContent;
  onClose: () => void;
};
```
Then before:

```ts
<SideNavList position="bottom">
        {masterSliceLibrary !== undefined && (
          <SideNavListItem>
            <MasterSliceLibraryPreviewModal
              isOpen={isSliceLibraryDialogOpen}
              masterSliceLibrary={masterSliceLibrary}
              onClose={() => {
                setIsSliceLibraryDialogOpen(false);
              }}
            />
            <SideNavLink
              title="Master Slice Library"
              href={"/"}
              Icon={MasterSliceLibraryIcon}
              onClick={(e) => {
                void telemetry.track({
                  event: "slice-library:beta:modal-opened",
                });

                setIsSliceLibraryDialogOpen(true);

                // We don't want it to actually navigate anywhere, but `href` is required
                e.preventDefault();
              }}
            />
          </SideNavListItem>
        )}
```

xrutayisire: WDYT?

playwright/tests/common/sideNav.spec.ts
```diff
@@ -190,7 +190,7 @@ test('I can access the Academy from the "Learn Prismic" link', async ({
   await expect(newTab).toHaveTitle(/Prismic Academy/);
 });
 
-test('I can access the repository\'s users page from the "Invite team" link', async ({
+test('I can open a modal describing Master Slice Libraries by clicking the "Master Slice Library" button', async ({
```
xrutayisire, line 196: For the test, I let you read the README of our e2e tests: https://github.com/prismicio/slice-machine/blob/main/playwright/README.md

You will see that globally, the thing you miss the most is not targeting element directly but using our POM structure.

xrutayisire (changes requested):
Some small changes and it should be good 🚀
 
Doc about our coding guidelines for the future: https://www.notion.so/prismic/Slice-Machine-UI-s-Coding-Guidelines-d377c3e33faa44b3a5e0ad84a942616d

packages/slice-machine/src/hooks/useMarketingContent.ts
jaen: Personally, I'd prefer to not pass this information as props from the outside and keep getting the marketing content inside the component, but if that's what you prefer I can change the code like this.

xrutayisire: Then I prefer, like this 🙂

```
const { masterSliceLibrary } = useMarketingContent();

if (!masterSliceLibrary) return undefined; // we prefer undefined, but sometimes React don't want 

const { exampleLinkUrl, codeLinkUrl, previewVideoUrl } = masterSliceLibrary;
```

xrutayisire: It's just a double check that masterSliceLibrary is here, with the parent call but it's good for me 🙌

packages/slice-machine/src/features/masterSliceLibrary/SliceLibraryPreviewModal.tsx
jaen: Changed

packages/slice-machine/src/hooks/useMarketingContent.ts
jaen: Changed

packages/slice-machine/src/hooks/useMarketingContent.ts
jaen: As discussed in DMs, went with early return option.

packages/slice-machine/src/legacy/components/Navigation/index.tsx
jaen: Changed

playwright/tests/common/sideNav.spec.ts
jaen: As discussed in DMs we agreed that writing a page object for a one-off modal that's supposed to be there only for a few weeks does not make sense, so we can leave it as–is.

xrutayisire (approved):
LGTM! Nice work 🙂

PR #1375, opened 2024-05-28
refactor(slice-machine-ui): use `Window` from `@prismicio/editor-ui`

bapmrl:
Completes DT-2021.


levimykel (approved):
Nice work. Everything looks good to me. I can see that there are still some references to vanilla-extract in the `yarn.lock` file. I assume that's not a big deal, so this is good to go for me.

bapmrl:
> Nice work. Everything looks good to me. I can see that there are still some references to vanilla-extract in the `yarn.lock` file. I assume that's not a big deal, so this is good to go for me.

Thank you @levimykel!

Yes, it's normal. It's because the `@prismicio/editor-ui` package still has some references to `vanilla-extract`.

PR #1379, opened 2024-06-03
chore: update api endpoints

xrutayisire:
### Description

- Use new /push endpoint from sm-api to bulk update changes
- Rename the stage URL for sm-api
- Rename the stage URL for AwsAclProvider
- Rename the prod URL for AwsAclProvider

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/manager/src/managers/prismicRepository/types.ts
+++ b/packages/manager/src/managers/prismicRepository/types.ts
@@ -81,7 +81,7 @@ export type AllChangeTypes =
 	| CustomTypeInsertChange
 	| CustomTypeUpdateChange
 	| CustomTypeDeleteChange;
-export interface BulkBody extends Record<string, unknown> {
+export interface PushBody extends Record<string, unknown> {
 	confirmDeleteDocuments: boolean;
 	changes: AllChangeTypes[];
 }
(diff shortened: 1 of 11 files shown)


packages/manager/src/managers/prismicRepository/types.ts
```diff
@@ -80,7 +80,7 @@ export type AllChangeTypes =
 	| CustomTypeInsertChange
 	| CustomTypeUpdateChange
 	| CustomTypeDeleteChange;
-export interface BulkBody extends Record<string, unknown> {
+export interface PushBody extends Record<string, unknown> {
```
angeloashmore: ❓ #ask: Do we need the `extends Record<string, unknown>`?

angeloashmore (approved):
LGTM! Just a quick question about a type that may be able to simplify.

packages/manager/src/managers/prismicRepository/types.ts
xrutayisire: Haha clearly not necessary, good catch!

xrutayisire: Removed

dani-mp (approved):

PR #1383, opened 2024-06-06
feat(slice-machine-ui): add support for advanced groups

levimykel:
**Resolves**: [DT-2138](https://linear.app/prismic/issue/DT-2138/aadev-slice-machine-support-advanced-groups-with-current-ui)

### Description

This PR adds support for advanced groups. There are a few considerations:

- I needed to add a new NestedGroup field widget to avoid circular dependencies with the Group widget
- This PR is not yet complete. A preliminary review can be done to ensure it's heading in the right direction. It will need to be updated with the following:
  - Style changes based on @benjammartin's feedback.
  - Update e2e tests to include nested groups   

### Checklist

### Preview

The nested group in a custom type:

https://github.com/prismicio/slice-machine/assets/9385378/0d286b82-2b0a-415b-9993-567c80e2c239

The nested group in a slice:

https://github.com/prismicio/slice-machine/assets/9385378/815f4a3a-8bb2-4bed-9068-f44b8316eab2

### How to QA [^1]

[^1]:


package.json
```diff
@@ -68,7 +68,7 @@
   "resolutions": {
     "connected-next-router/react-redux": "8.0.7",
     "react-beautiful-dnd/react-redux": "8.0.7",
-    "@prismicio/types-internal": "2.4.1"
+    "@prismicio/types-internal": "2.5.0-alpha.4"
```
angeloashmore: 💡 #idea: It's possible we don't need this root-level resolution. Could we try the PR without it?

If you run into version incompatibilities with `@prismicio/types-internal`, you may need to update some packages to their alpha versions, like `@prismicio/mocks`.

packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/components/NewField.tsx
```diff
@@ -15,8 +15,8 @@ import {
 } from "@/legacy/lib/forms";
 import { validateId } from "@/legacy/lib/forms/defaults";
 import { InputType } from "@/legacy/lib/forms/fields";
+import * as Widgets from "@/legacy/lib/models/common/widgets/allWidgets";
```
angeloashmore: 💡 #idea: I wonder if it would be better to remove these widget arrays, like `groupBuilderArray.ts` and `sliceBuilderArray.ts`. Most of them are used in one place, which means we could inline the value directly where it is used.

The `allWidgets` array could be moved to `@/legacy/lib/models/common/widgets/index.ts` as the directory's root-level export.

If it works, it could be a cleaner approach.

packages/slice-machine/src/legacy/lib/models/common/NestedGroup.ts
```diff
@@ -0,0 +1,82 @@
+// import { NestableWidget } from "@prismicio/types-internal/lib/customtypes";
```
angeloashmore: ⚠️ #issue: This commented-out line can be removed.

packages/slice-machine/src/legacy/lib/models/common/NestedGroup.ts
angeloashmore, line 1: ⚠️ #issue: This file is almost the same as `Group.ts`, effectively only differing in the type of fields that are accepted. Is there a way we could share code?

Here is a good example of creating two types with parameterized values: https://github.com/prismicio/prismic-types-internal/blob/67dc5be82ef8aa0bab177567d1831e5a82fe1110/src/customtypes/widgets/Group.ts.

Note how `Group` and `NestedGroup` are created from `createGroup()`.

packages/slice-machine/src/legacy/lib/models/common/widgets/NestedGroup/Form.jsx
angeloashmore, line 1: ⚠️ #issue: This file is mostly the same as `Group/Form.jsx`. Is there a way we could share the code?

packages/slice-machine/src/legacy/lib/models/common/widgets/NestedGroup/ListItem/index.jsx
angeloashmore, line 1: ⚠️ #issue: This file is mostly the same as `Group/ListItem/index.jsx`. Is there a way we could share the code?

packages/slice-machine/src/legacy/lib/models/common/widgets/NestedGroup/ListItem/index.jsx
```diff
@@ -0,0 +1,288 @@
...
+                    }
+                    {provided.placeholder}
+
+                    {newFieldData && (
+                      <NewField
+                        {...newFieldData}
+                        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
+                        fields={groupItem.value.config.fields || []}
+                        onSave={(...args) => {
+                          console.log({ log: "onSave", ...args });
```
angeloashmore: ⚠️ #issue: We can remove this `console.log()`.

packages/slice-machine/src/legacy/lib/models/common/widgets/NestedGroup/index.ts
angeloashmore, line 1: ⚠️ #issue: This file is mostly the same as `Group/index.ts`. Is there a way we could share the code?

angeloashmore (changes requested):
Awesome work! I know it wasn't easy. 😄

The PR is heading in the right direction, and we can see that it works well. Next, we can streamline the code by reducing duplication, especially since `Group` and `NestedGroup` are so similar.

I noted some improvements below. Let me know if any of it is unclear or if you'd like to brainstorm solutions together.

Before refactoring the internal logic, it might be best to start with adding E2E tests. Tests will let us refactor with confidence. The expected design changes shouldn't drastically impact how the E2E tests work, unless the flow completely changes (unlikely).

lihbr:
Hey @levimykel, I know you're not completely done with your PR, but just in case it's not on your radar yet, I noticed the following issue while doing QA of other services on staging using your PR to run the Slice Machine part.

Basically, you cannot edit config of fields inside of groups (editing config inside of nested groups works fine).

<details open>
<summary>Reproduction</summary>

1. Create a custom type
2. Add a group field inside it
3. Add any field (i.e. text field) inside the group
4. Attempt to edit and save the field config
5. Currently yields a frontend error: `Error while parsing an SM group.`

</details>

<details>
<summary>Video clip if it helps</summary>

https://github.com/prismicio/slice-machine/assets/25330882/f0ddb127-5b20-4135-9e11-4b91592ad5ce

</details>

levimykel:
@lihbr Thanks for pointing this out 😄
I discovered this today as well and just pushed a fix for it.

packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/components/NewField.tsx
levimykel: I reworked this as you suggested and put the widget arrays inline. I agree that it's cleaner and safer this way.

package.json
levimykel: @angeloashmore When I tried removing this, things seemed to work, but all the CI tests failed.

levimykel: @xrutayisire helped me track this down today. The reason the CI tests were failing is because `@prismicio/editor-fields` has a dependency on `@prismicio/types-internal`.

packages/slice-machine/src/features/settings/git/useNestedGroupExperiment.ts
```diff
@@ -0,0 +1,8 @@
+import { useExperimentVariant } from "@/hooks/useExperimentVariant";
+
+type UseNestedGroupExperimentReturnType = { eligible: boolean };
+
+export function useNestedGroupExperiment(): UseNestedGroupExperimentReturnType {
```
xrutayisire: ⚠️ #issue
 
You put it under git folder that is specific to all git features. If on your side it's not related to a specific feature that I think it's not, you can put the hook directly in hooks folder.

levimykel: Good catch, I moved it.

package.json
angeloashmore: Ah, okay. In that case, we should update `@prismicio/editor-fields` to use the latest `@prismicio/types-internal`. We should avoid `resolutions` as much as possible since it could break packages.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
```diff
@@ -42,6 +40,25 @@ const dataTipText2 = `The repeatable zone is for a group<br/>
...
+  Widgets.Date,
+  Widgets.Timestamp,
+  Widgets.Embed,
+  Widgets.Number,
+  Widgets.GeoPoint,
+  Widgets.Color,
+  Widgets.Text,
+];
+
+const primaryWidgetsArray = [Widgets.Group, ...itemsWidgetsArray];
```
angeloashmore, line 60: 🎉 #nice: `itemsWidgetsArray` and `primaryWidgetsArray` is soooo much cleaner now.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/VariationsList/index.tsx
```diff
@@ -60,7 +60,9 @@ export const VariationsList: FC<VariationsListProps> = (props) => {
                 }}
                 startIcon="add"
                 color="grey"
-              />
+              >
+                Add a variation
```
angeloashmore, line 64: ⚠️ #issue: On the desktop version, the button reads "Add a new variation." The "Add a variation" version is aligned with the "Add a field" button, so could we update the desktop button to read "Add a variation"?

packages/slice-machine/src/legacy/lib/models/common/Group.ts
```diff
@@ -1,43 +1,66 @@
...
+export type NestedGroupSM = t.TypeOf<typeof NestedGroupSM>;
+
+const GroupFieldsSM = t.array(
+  t.type({
+    key: t.string,
+    value: t.union([NestableWidget, NestedGroupSM]),
   }),
 );
+
+export const GroupConfig = createGroupConfig(GroupFieldsSM);
```
angeloashmore: ⚠️ #issue: `GroupConfig` doesn't seem to be used anywhere. Can it be removed?

packages/slice-machine/src/legacy/lib/models/common/Group.ts
```diff
@@ -1,43 +1,66 @@
...
+      t.partial({
+        fieldset: StringOrNull,
+        icon: t.string,
+        description: t.string,
+        config: createGroupConfig(fields),
+      }),
+    ]),
+  );
+
+export const NestedGroupSM = createGroupSM(FieldsSM);
```
angeloashmore, line 36: 🎉 #nice: Great use of type builders!

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
```diff
@@ -86,7 +86,7 @@ const CustomListItem = ({
       group: Groups.fromSM(groupItem.value),
       previousFieldId: previousKey,
       newFieldId: newKey,
-      field: value,
+      field: value.config?.fields ? Groups.fromSM(value) : value,
```
angeloashmore: 💡 #idea: Checking for `value.type === GroupFieldType` is more straightforward. Checking `value.config.fields` is more like duck typing.

```ts
field: value.type === GroupFieldType Groups.fromSM(value) : value,
```

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/createWidget.ts
angeloashmore, line 1: 💡 #idea: `createGroupWidget.ts` is a less ambiguous name.

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/createWidget.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export function createWidget(
+  props: createGroupWidgetProps,
+): Widget<GroupSM, SchemaType>;
+
+export function createWidget(
+  props: createNestedGroupWidgetProps,
+): Widget<NestedGroupSM, SchemaType>;
+
+export function createWidget({
```
angeloashmore: 💡 #idea: `createGroupWidget` is a less ambiguous name.

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/createWidget.ts
```diff
@@ -0,0 +1,78 @@
...
+  customName: "NestedGroup";
+};
+type createWidgetProps = createGroupWidgetProps | createNestedGroupWidgetProps;
+
+export function createWidget(
+  props: createGroupWidgetProps,
+): Widget<GroupSM, SchemaType>;
+
+export function createWidget(
+  props: createNestedGroupWidgetProps,
```
angeloashmore: 💡 #idea: `args` rather than `props` is typically used for function parameter names (except when used in React components):

```ts
export function createWidget(
  args: CreateGroupWidgetArgs,
): Widget<GroupSM, SchemaType>;
```

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/createWidget.ts
```diff
@@ -0,0 +1,78 @@
...
+    config: yup.object().shape({
+      fields: yup.array(),
+      label: yup.string(),
+      placeholder: yup.string(),
+      repeat: yup.boolean().optional(),
+    }),
+  });
+export type SchemaType = ReturnType<typeof createSchema>;
+
+type createWidgetPropsBase<T extends TabField> = {
```
angeloashmore: ⚠️ #issue: Argument object types typically use PascalCase: `CreateWidgetPropsBase`.

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/createWidget.ts
```diff
@@ -0,0 +1,78 @@
...
+import { type GroupListItemProps } from ".";
+import Form, { FormFields } from "./Form";
+
+const Meta = {
+  icon: MdPlaylistAdd,
+  title: "Group",
+  description: "A repeatable set of fields",
+};
+
+export const createSchema = (typeRegex: RegExp) =>
```
angeloashmore: ⚠️ #issue: We can remove `export` since this function is not used elsewhere.

playwright/pages/shared/TypeBuilderPage.ts
```diff
@@ -228,10 +228,19 @@ export class TypeBuilderPage extends BuilderPage {
...
+      args;
 
-    if (groupFieldId) {
+    if (groupFieldId && grandparentGroupFieldId) {
+      await this.getListItem(groupFieldId, grandparentGroupFieldId)
+        .getByRole("button", {
+          name: "Add Field",
+          exact: true,
+        })
+        .click();
```
angeloashmore: 💡 #idea: I think we can simplify this `if...else` block to:

```ts
if (groupFieldId) {
  await this.getListItem(groupFieldId, grandparentGroupFieldId)
    .getByRole("button", {
      name: "Add Field",
      exact: true,
    })
    .click();
} else {
  await this.staticZoneAddFieldButton.click();
}
```

angeloashmore (changes requested):
Looking good! I left a bunch of small comments that I think you'll be able to handle quickly. After those changes, I think we are good to go!

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
levimykel: Agreed 😄

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/VariationsList/index.tsx
levimykel: I agree with this, and @benjammartin is good with it, so I made the change.

packages/slice-machine/src/legacy/lib/models/common/Group.ts
levimykel: Good catch, it can be removed.

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
levimykel: Yeah, that's better for sure.

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/createWidget.ts
levimykel: Thanks, good to know.

playwright/pages/shared/TypeBuilderPage.ts
levimykel: Good call.

packages/adapter-next/package.json
```diff
@@ -67,7 +67,7 @@
 	},
 	"dependencies": {
 		"@prismicio/simulator": "^0.1.4",
-		"@prismicio/types-internal": "^2.4.1",
+		"@prismicio/types-internal": "2.5.0",
```
xrutayisire: ⚠️ #issue

For adapters and plugin-kit, devx prefer that we keep "^" in front to allow automatic update in case of an update. We are only doing that for adapters and plugin-kit, as we consider that more as packages compare to SM that could be considered an app.

packages/plugin-kit/package.json
```diff
@@ -62,7 +62,7 @@
 		"audit": "yarn npm audit --environment production --severity high"
 	},
 	"dependencies": {
-		"@prismicio/client": "^7.5.0",
+		"@prismicio/client": "v7.6.0",
```
xrutayisire: ⚠️ #issue

`v` should be removed right? And we should keep "^" usage in adapters / plugin-kit.

xrutayisire (changes requested):
Small changes related to packages and let's go!

xrutayisire (approved):
LGTM!

PR #1388, opened 2024-06-14
feat: support nuxt 4

lihbr:
**Resolves**: /

### Description

This PR adds support for the upcoming Nuxt 4 version, it:
- Updates the Nuxt adapter to account for Nuxt 4 changes
- Updates `@slicemachine/init` to detect Nuxt 4 correctly

The main change of Nuxt 4 impacting us is that frontend-related directories will now be nested under an `app` directory (`./pages` is now `./app/pages`), so we have to account for that.

A detailed list of breaking changes and migration strategy is available here: https://nuxt.com/docs/getting-started/upgrade#migrating-to-nuxt-4

### Checklist

### Preview

/

### How to QA [^1]

Nuxt 4 is not released yet, for now you can perform the following:
1. Create a Nuxt 3 app with v4 compat
```
npx nuxi@latest init -t v4-compat
```
2. Run the init command
```
npx @slicemachine/init@alpha.lh-support-nuxt-4
```
3. Check that:
   - `slice-simulator.vue` is created under `app/pages`
   - `app/app.vue` is moved to `app/pages/app.vue`
   - Slice library is pointing to `./app/slices`   

[^1]:


packages/adapter-nuxt/src/hooks/project-init.ts
```diff
@@ -125,13 +115,37 @@ const createSliceSimulatorPage = async ({
...
+		helpers,
+	});
+
+	const hasAppDirectory = await checkHasAppDirectory({ helpers });
+	const hasSrcDirectory = await checkHasSrcDirectory({ helpers });
+
 	const filename = path.join(
-		srcPagesDirectoryExists ? "src/pages" : "pages",
+		// We first give priority to existing `pages` directory, then to `srcDir`.
+		appPagesDirectoryExists
```
xrutayisire: 💡 #idea

I always have a hard time to read nested ternary like this, WDYT about using a function? Or something that prevent nested ternary? 

```ts
function getPagesDirectory({ appPagesDirectoryExists, srcPagesDirectoryExists, pagesDirectoryExists, hasAppDirectory, hasSrcDirectory }) {
    if (appPagesDirectoryExists) {
        return "app/pages";
    } else if (srcPagesDirectoryExists) {
        return "src/pages";
    } else if (pagesDirectoryExists) {
        return "pages";
    } else if (hasAppDirectory) {
        return "app/pages";
    } else if (hasSrcDirectory) {
        return "src/pages";
    } else {
        return "pages";
    }
}

const filename = path.join(
    getPagesDirectory(appPagesDirectoryExists, srcPagesDirectoryExists, pagesDirectoryExists, hasAppDirectory, hasSrcDirectory),
    "slice-simulator.vue"
);
```

packages/adapter-nuxt/src/hooks/project-init.ts
```diff
@@ -169,14 +183,15 @@ const moveOrDeleteAppVue = async ({
...
-	const srcDirectoryExists = await checkHasProjectFile({
-		filename: "src",
-		helpers,
-	});
+	const hasAppDirectory = await checkHasAppDirectory({ helpers });
+	const hasSrcDirectory = await checkHasSrcDirectory({ helpers });
 
-	const filenameAppVue = path.join(srcDirectoryExists ? "src" : "", "app.vue");
+	const filenameAppVue = path.join(
+		hasAppDirectory ? "app" : hasSrcDirectory ? "src" : "",
```
xrutayisire: 💡 #idea

Same comment about double ternary here. Maybe we could create a utils function that could be use everywhere. Something that return `"app"`, `"src"` or `""`.

Then it would prevent to do the same check.

packages/adapter-nuxt/src/lib/checkHasAppDirectory.ts
```diff
@@ -0,0 +1,18 @@
...
+
+type CheckHasAppDirectoryArgs = Pick<
+	SliceMachineContext<PluginOptions>,
+	"helpers"
+>;
+
+export async function checkHasAppDirectory(
+	args: CheckHasAppDirectoryArgs,
+): Promise<boolean> {
+	return await checkHasProjectFile({
```
xrutayisire: ❓ #question

I'm wondering what is the advantage to use a specific function over directly using `checkHasProjectFile` with the filename "app"? Did this create more files for another direct function call?

packages/init/src/lib/framework.ts
```diff
@@ -94,14 +94,14 @@ export const FRAMEWORKS: Record<string, Framework> = {
 			"@slicemachine/adapter-nuxt2": npmDistributionTag,
 		},
 	},
-	"nuxt-3": {
-		name: "Nuxt 3",
-		sliceMachineTelemetryID: "nuxt-3",
+	nuxt: {
+		name: "Nuxt",
+		sliceMachineTelemetryID: "nuxt",
```
xrutayisire, line 99: 💡 #idea

I guess we would need to communicate to Côme about the change to ensure we have all dashboard updated 🧐

xrutayisire (approved):
Let's go Nuxt 4 🥳

I left some comments, mostly about the use of nested ternary, that I don't like for maintenance and readability.

packages/adapter-nuxt/src/hooks/project-init.ts
```diff
@@ -28,23 +29,12 @@ type InstallDependenciesArgs = {
...
-		await installDependencies({
-			dependencies: {
-				[NUXT_PRISMIC]: "rc",
-			},
-			dev: true,
-		});
-	}
+	await installDependencies({
+		dependencies: {
+			[NUXT_PRISMIC]: "^3.3.0",
```
angeloashmore, line 33: ⚠️ #issue: Should this be updated to `^4` before merging?

packages/adapter-nuxt/src/hooks/project-init.ts
angeloashmore: I like the idea of having a dedicated path builder. Something like:

```ts
export function joinAppPath(
  segments: string[],
  context: { helpers: SliceMachineHelpers },
): string {
  // Build a path conditionally with `app` and/or `src`.
}
```

We could use it in the Next.js adapter as well.

packages/adapter-nuxt/src/hooks/project-init.ts
angeloashmore: See: https://github.com/prismicio/slice-machine/pull/1388/files#r1642358764

packages/adapter-nuxt/src/lib/checkHasAppDirectory.ts
angeloashmore: This function was much longer before we had `@slicemachine/plugin-kit/fs`.

Today, it doesn't serve much purpose besides being more declarative.

@lihbr You can decide which is the better approach and we can adopt the same practice in the Next.js adapter.

packages/adapter-nuxt/test/plugin-project-init.test.ts
```diff
@@ -108,181 +108,181 @@ export default defineNuxtConfig({
...
-	test("creates a TypeScript Slice Simulator page file when TypeScript is enabled", async (ctx) => {
-		ctx.project.config.adapter.options.typescript = true;
-		const pluginRunner = createSliceMachinePluginRunner({
-			project: ctx.project,
-			nativePlugins: {
-				[ctx.project.config.adapter.resolve]: adapter,
-			},
-		});
-		await pluginRunner.init();
+			expect(contents).toBe(`<template>
```
angeloashmore, line 132: ❓ #question: Is there a reason you prefer `toBe` over `toMatchInlineSnapshot`?

packages/init/src/lib/framework.ts
```diff
@@ -94,14 +94,14 @@ export const FRAMEWORKS: Record<string, Framework> = {
...
 		wroomTelemetryID: "nuxt",
-		prismicDocumentation: "https://prismic.dev/init/nuxt-3",
+		prismicDocumentation: "https://prismic.dev/init/nuxt",
 		adapterName: "@slicemachine/adapter-nuxt",
 		compatibility: {
-			nuxt: "^3.0.0",
+			nuxt: "^3.0.0 || ^4.0.0 || ^4.0.0-rc.0",
```
angeloashmore: 💡 #idea: Is `^4.0.0` necessary if you are including the RC?

packages/adapter-nuxt/src/hooks/project-init.ts
angeloashmore, line 1: ❓ #question: Why do we need to move `app/app.vue` to `app/pages/index.vue`? Is it a better location than the default?

angeloashmore (changes requested):
Looks good! Everything functions as expected. I left some implementation notes and added to @xrutayisire's comments.

I'm going to mark the review as "Request changes" since I think we need to edit the dependency version when Nuxt 4 is out. Some of the refactoring suggestions, like `joinAppPath`, could also significantly clean up the code (something we'll need to do in the Next.js adapter as well).

(GitHub displays replies I made to @xrutayisire's comments below. It's weird. Sorry.)

packages/adapter-nuxt/src/hooks/project-init.ts
lihbr: No, the kit major version remains 4, there's no breaking changes on that end and it will support both Nuxt 3 and 4 for the time being ☺️

packages/adapter-nuxt/test/plugin-project-init.test.ts
lihbr: I don't prefer it, it's just that inline snapshot don't work in a `describe.each` context (as each snapshot could be different, which is not our case here)

packages/init/src/lib/framework.ts
lihbr: I don't know, I'm not sure there will be RCs (there will probably), so I added the latter in case. Cleaned up to just `^3.0.0 || ^4.0.0-rc.0` as `^4.0.0-rc.0` will also satisfies stable versions (like what you did for Next 15)

packages/adapter-nuxt/src/hooks/project-init.ts
lihbr: Nice question!

Nuxt can be a "one-page" framework, not as-in SPA, but really as-in "one page", allowing for minimal setup. When it does it completely disables the router and removes it from the bundle. Nuxt is in this mode when you only have an `app/app.vue` file.

It's only when you create a `pages` directory and remove the `app/app.vue` that Nuxt turns into a regular "multi-page" framework with a router and file system based routing.

On the Nuxt default starter there's no `pages` directory, just an `app/app.vue` file. When initializing the adapter we need to create a page for the Simulator, this forces us to have at least two pages (`/` and `/slice-simulator`) and therefore to create the `pages` directory. We move the starter `app/app.vue` file to `app/pages/index.vue` to turn Nuxt into the regular "multi-page" framework, otherwise it'll remain into its "one-page" mode.

packages/init/src/lib/framework.ts
lihbr: Definitely! I just thought it'd make sense to update it because moving forward `nuxt` will mean Nuxt 3 and above, like `next` for Next.js ☺️

packages/adapter-nuxt/src/hooks/project-init.ts
lihbr: Refactored, see below comment ☺️ Thanks for challenging it!

packages/adapter-nuxt/src/lib/checkHasAppDirectory.ts
lihbr: Removed the two from all adapters.

packages/adapter-nuxt/src/hooks/project-init.ts
lihbr: Added and tested such helper. Refactored Next and Nuxt kits to use it ☺️

packages/adapter-nuxt/src/hooks/project-init.ts
angeloashmore: Interesting! Makes sense. Thanks for the explanation!

packages/adapter-nuxt/test/plugin-project-init.test.ts
angeloashmore: Ah, okay. Good to know!

packages/adapter-nuxt/src/hooks/project-init.ts
angeloashmore: Oops, I mixed up `@nuxtjs/prismic` and `nuxt` itself. Sounds good!

angeloashmore:
@lihbr Now having seen the new `joinAppPath` in action, I find it a bit confusing. The logic seems tailored for Nuxt's `srcDir` behavior. It also mixes the "app" and "src" names in its arguments.

I know I suggested the helper, but I didn't have a clear idea how or if it would work.

### Next.js adapter

The App Router and Pages Router have different directory structures, and I don't think it makes sense to abstract path building into a single helper. `joinAppPath` didn't do that, but the function's name implies it.

Instead, we could have an adapter-specific helper to auto-prepend `src`. The `src` directory is optional in Next.js, and if it exists, it all source files go in it, including `app`.

I think we could have an adapter-specific helper in `lib`:

```ts
// packages/adapter-next/src/lib/buildSrcPath.ts

import * as path from "node:path";

import type { SliceMachineHelpers } from "@slicemachine/plugin-kit";
import { checkHasProjectFile } from "@slicemachine/plugin-kit/fs";

export async function buildSrcPath(args: {
	filename: string;
	helpers: SliceMachineHelpers;
}): Promise<string> {
	const hasSrcDirectory = await checkHasProjectFile({
		filename: "src",
		helpers: args.helpers,
	});

	return hasSrcDirectory ? path.join("src", args.filename) : args.filename;
}
```

```ts
// elsewhere, like `project-init.ts`

const filename = await buildSrcPath({
	filename: "prismicio.ts",
	helpers,
});
```

### Nuxt adapter

A similar strategy can be used with Nuxt-specific paths:

```ts
// TODO: If Nuxt provides a way to read `nuxt.config.js`'s `srcDir` value, we
// can use the exact value given. The current implementation does not support
// custom values different than "app" or "src".

import * as path from "node:path";

import { SliceMachineHelpers } from "@slicemachine/plugin-kit";
import { checkHasProjectFile } from "@slicemachine/plugin-kit/fs";

export async function buildSrcPath(args: {
	filename: string;
	helpers: SliceMachineHelpers;
}): Promise<string> {
	const hasAppDirectory = await checkHasProjectFile({
		filename: "app",
		helpers: args.helpers,
	});
	if (hasAppDirectory) {
		return path.join("app", args.filename);
	}

	const hasSrcDirectory = await checkHasProjectFile({
		filename: "src",
		helpers: args.helpers,
	});
	if (hasSrcDirectory) {
		return path.join("src", args.filename);
	}

	return args.filename;
}
```

The current versions of these helpers look similar, but their intent is slightly different. Nuxt, for example, should technically read `nuxt.config.js` to determine the configured `srcDir` and act accordingly. Next.js will always only check `src`. Thus, I think these helpers can be adapter-specific and not in `@slicemachine/plugin-kit/fs`.

I opened a PR with the changes here for you to review: #1392

I'm happy to be corrected if I misunderstood the changes! 🙂

lihbr:
Thank you so much for taking the time to refactor that code! Indeed I was sensing the new util was feeling a bit odd as you'd have to conceive "App" as the directory where you app code lives, not an `./app` directory.

I like how you took it to each project with more specific helpers 💪

xrutayisire (approved):
LGTM! Go Nuxt 4!

PR #1390, opened 2024-06-17
feat: improve code snippet hints for groups & nested groups

levimykel:
**Resolves**: [DT-2191](https://linear.app/prismic/issue/DT-2191/aadev-i-can-see-a-difference-between-group-and-nested-group-code)

This PR depends on PR #1383.

### Description

This PR improves the way code snippet hints are constructed for groups and nested groups. It adds context to `SnippetReadHookData` to provide more info to the plugin kits when rendering snippet hints. In this case, it adds a `subItemHintBase` property to the context to distinguish sub-items of groups & nested groups. This allows slice machine to fully control the hint base. 

### Checklist

### Preview

Here, you can see that the group uses the sub-item hint base "item" and the nested group uses "subItem":

Screenshot 2024-06-17 at 14 54 31

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/plugin-kit/src/hooks/snippet-read.ts
+++ b/packages/plugin-kit/src/hooks/snippet-read.ts
@@ -17,6 +17,11 @@ export type Snippet = {
 
 /**
  * Data provided to `snippet:read` hook handlers.
+ *
+ * The `itemName` string keeps slice machine as the single source of truth for
+ * the code snippet base names. Angelo suggested an improvement to `fieldPath`
+ * (linked below) that could be done in the future.
+ * https://github.com/prismicio/slice-machine/pull/1390#pullrequestreview-2123845118
  */
 export type SnippetReadHookData = {
 	fieldPath: string[];
--- a/packages/plugin-kit/src/hooks/snippet-read.ts
+++ b/packages/plugin-kit/src/hooks/snippet-read.ts
@@ -21,9 +21,7 @@ export type Snippet = {
 export type SnippetReadHookData = {
 	fieldPath: string[];
 	model: DynamicWidget;
-	context?: {
-		subItemHintBase?: string;
-	};
+	itemName?: string;
 };
 
 /**
--- a/packages/plugin-kit/src/hooks/snippet-read.ts
+++ b/packages/plugin-kit/src/hooks/snippet-read.ts
@@ -21,6 +21,9 @@ export type Snippet = {
 export type SnippetReadHookData = {
 	fieldPath: string[];
 	model: DynamicWidget;
+	context?: {
+		subItemHintBase?: string;
+	};
 };
 
 /**
--- a/packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/components/Hints/index.tsx
+++ b/packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/components/Hints/index.tsx
@@ -9,9 +9,15 @@ interface HintProps {
   show: boolean;
   item: Item;
   renderHintBase: RenderHintBaseFN;
+  subItemHintBase?: string;
 }
 
-const Hint: React.FC<HintProps> = ({ show, renderHintBase, item }) => {
+const Hint: React.FC<HintProps> = ({
+  show,
+  renderHintBase,
+  item,
+  subItemHintBase,
+}) => {
   const fieldPathString = renderHintBase({ item });
 
   // TODO: Call `swr`'s global `mutate` function when something changes to clear the cache.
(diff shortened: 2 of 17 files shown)


packages/plugin-kit/src/hooks/snippet-read.ts
```diff
@@ -21,6 +21,9 @@ export type Snippet = {
 export type SnippetReadHookData = {
 	fieldPath: string[];
 	model: DynamicWidget;
+	context?: {
+		subItemHintBase?: string;
+	};
```
angeloashmore: 💡 #idea: Two things about `subItemHintBase`:

1. I think we can promote `subItemHintBase` to a top-level argument, alongside `fieldPath` and `model`.
2. I think `itemName` is a clearer name, referring specifically to the `.map` function's parameter name.

packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/components/Hints/index.tsx
```diff
@@ -9,9 +9,15 @@ interface HintProps {
   show: boolean;
   item: Item;
   renderHintBase: RenderHintBaseFN;
+  subItemHintBase?: string;
```
angeloashmore: 💡 #idea: If we move forward with renaming the `subItemHintBase` hook argument to `itemName`, this prop could be renamed to `hintItemName`.

(Side thought: "hint" is not an obvious name for this feature. In the future, we should rename it to "code snippet.")

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/createWidget.ts
```diff
@@ -74,5 +76,6 @@ export function createWidget({
     CustomListItem: customListItem,
     TYPE_NAME: "Group",
     CUSTOM_NAME: customName,
+    SUB_ITEM_HINT_BASE: subItemHintBase,
```
angeloashmore: 💡 #idea: CONSTANT_CASE is usually reserved for constants. I'm not sure why it was used for `TYPE_NAME` and `CUSTOM_NAME`. `subItemHintBase` (or `hintItemName` per my other comment) could be used instead.

angeloashmore (approved):
It looks good to me. 🙂 I left a few naming suggestions below.

To offer an alternative solution, we could pass more in-depth information about the field to `snippet:read`:

```ts
/**
 * Data provided to `snippet:read` hook handlers.
 */
export type SnippetReadHookData = {
  fieldPath: (
    | string
    | {
        name: string;
        model:
          | CustomType
          | SharedSlice
          | StaticWidget
          | StaticSlice;
      }
  )[];
  model: StaticWidget;
};
```

Providing the model as part of the path data gives adapters the ability to change their snippets based on a field's position. A group field, for example, could automatically detect if it is nested and change from `item` to `subItem`—no user-provided name needed. The adapter would be less reliant on Slice Machine's logic, opening the door for us to display the same information in e.g. a CLI.

Implementing this idea in Slice Machine's current architecture would be difficult. But maybe it's something we can explore now or in the future.

levimykel:
> To offer an alternative solution, we could pass more in-depth information about the field to `snippet:read`

I discussed this with @xrutayisire, and we agree that your proposal is probably better. However, we decided to keep this as a potential future improvement.

packages/plugin-kit/src/hooks/snippet-read.ts
```diff
@@ -17,10 +17,16 @@ export type Snippet = {
 
 /**
  * Data provided to `snippet:read` hook handlers.
+ *
+ * The `itemName` string keeps slice machine as the single source of truth for
```
xrutayisire, line 21: 🎉 #nice

Super nice comment! Very clear!

packages/slice-machine/src/legacy/lib/models/common/widgets/Widget.ts
```diff
@@ -23,5 +24,13 @@ export interface Widget<F extends TabField, S extends AnyObjectSchema> {
   prepareInitialValues?: (props: F["config"]) => F["config"];
 }
 
+export type Widget<F extends TabField, S extends AnyObjectSchema> = F extends
```
xrutayisire, line 27: 🎉 #nice

I like, it forces slice machine to not forget about the `hintItemName`  for group.
In the future it will be even better because the source of truth will be adapters, but for today I think it's perfect!

xrutayisire (approved):
LGTM, super nice work! 🚀

Issue #1387, opened 2024-06-14
Types generation error using groups and content relatioship fields


amendezm:
### Versions

- slice-machine-ui: v2.2.0 
- node: v20.3.0 

### Reproduction

<details open>
<summary>Additional Details</summary>
<br>

![image](https://github.com/prismicio/slice-machine/assets/26444448/0c39b6fe-ff0a-4069-915b-2defa9abab39)

![image](https://github.com/prismicio/slice-machine/assets/26444448/c8072172-6823-46a1-8715-544b46c6c3cb)

![image](https://github.com/prismicio/slice-machine/assets/26444448/6f8a66a6-5b47-48fd-b06b-74e018c0ce01)

</details>

### Steps to reproduce

Just create a group field in a slice and then add a nested content relatioshinp field

### What is expected?

Types should be generated correctly

### What is actually happening?

Type errors

xrutayisire:
Hi,

Good catch, thank you for the issue. To unlock you, can you pin the dependency of `prismic-ts-codegen` and re try to add any field to your group. It should fix the problem.

Here is an example with override:

```
"overrides": {
    "prismic-ts-codegen": "0.1.20"
  }
```

We will release a new Slice Machine and adapter version with the fix beginning of next week 🙌
Don't forget to remove the overrides after the release, I will post here about it.

Thank you,
Xavier

amendezm:
Hi Xavier,

Thank you for the quick response and the solution! I've pinned the dependency of prismic-ts-codegen to version 0.1.20 as suggested, and it resolved the issue. I was able to add fields to my group successfully.

I appreciate the heads-up about the upcoming release. I'll keep an eye out for your update and will remove the overrides once the new versions are available.

Thanks again for your help!

Best regards,

Alexis Mendez

amendezm:
![image](https://github.com/prismicio/slice-machine/assets/26444448/dfe31c36-2d44-41f2-8806-bd2952d90704)

angeloashmore:
Hi @amendezm, thanks for letting us know it worked. 🙂

I looked into the issue and it most likely happens because your Slice Machine adapter needs an update. Could you try the following when you have some time?

1. Remove the `"overrides"` property in `package.json` suggested by @xrutayisire above.
2. Update your project's Slice Machine adapter to the latest version. The package will look like `@slicemachine/adapter-*` (e.g. `@slicemachine/adapter-next`).
[...]

PR #1391, opened 2024-06-17
fix: correctly generate types for groups in slices

angeloashmore:
**Resolves**: #1387

### Description

This PR fixes a bug where TypeScript types for groups in slices were not generated correctly. A groups interface would not be generated when it was added as a slice's primary field.

The bug happens because `prismic-ts-codegen`, the underlying library used to generate the types, was not updated to a version that supports groups in slices.

`prismic-ts-codegen` was updated in all adapters and `@slicemachine/plugin-kit`. `prismic-ts-codegen` was also moved to a dev dependency in all adapters because it is only used in tests.

### Checklist

### Preview

N/A

### How to QA [^1]

TODO

[^1]:

### DIFF
--- a/packages/adapter-next/package.json
+++ b/packages/adapter-next/package.json
@@ -76,8 +76,7 @@
 		"lz-string": "1.5.0",
 		"monocle-ts": "^2.3.13",
 		"newtype-ts": "^0.3.5",
-		"pascal-case": "^3.1.2",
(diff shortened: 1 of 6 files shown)


xrutayisire (approved):
LGTM! Thanks Angelo, didn't take the time today to handle that! 🙏

PR #1392, opened 2024-06-19
refactor: create adapter-specific `buildSrcDir` helpers

angeloashmore:
**Resolves**: N/A

### Description

This PR removes `joinAppPath` from `@slicemachine/plugin-kit/fs` and replaces it with adapter-specific `buildSrcPath` helpers.

- Next.js: prepends `src` to a path if the `src` directory exists.
- Nuxt 3/4: prepends `app` or `src` to a path if the `app` or `src` directory exists.
- Nuxt 2: prepends `src` to a path if the `src` directory exists.

These helpers are framework-specific. Although there is some overlap in behavior, they different enough that it doesn't make sense for them to live in `@slicemachine/plugin-kit/fs`.

**Note**: The Nuxt builders may fail if a project has a custom `srcDir` configured. The current helpers assume either `"app"`, `"src"`, or no `srcDir` is configured. If there is a way to read the Nuxt config to detect the actual `srcDir` value, we can more accurately prepend the path.

### Checklist

### Preview

N/A

### How to QA [^1]

N/A

[^1]:


lihbr (approved):
Thanks! Just fixed something on Nuxt, will add tests on the other PR 👌

Unfortunately, also it's not really possible to read Nuxt `srcDir` value as it can be set from many different ways (`nuxt.config.ts`, `.nuxtrc`, `~/.nuxtrc` (that one be odd), [Nuxt layers](https://nuxt.com/docs/getting-started/layers), [Nuxt modules](https://nuxt.com/docs/guide/concepts/modules))

If it becomes an issue I think we'll try to resolve it from Nuxt config at least, for now I think we're good like this ☺️

PR #1395, opened 2024-07-03
fix: Yarn PnP support

lihbr:
**Resolves**: https://community.prismic.io/t/unable-to-setup-prismic-with-yarn-pnp-mode/16218

### Description

Slice Machine doesn't work with Yarn PnP. This PR leverages Yarn PnP API when available to properly resolve module handled by Yarn PnP (well... I really don't enjoy the non-standard way PnP is doing things...)

The PR also explicitly declares `readable-web-to-node-stream` which is directly used by our bundled code and updates `@segment/analytics-node` to accommodate Yarn ghost dependencies requirement.

To get a hang of the changes, I've created this "minimal reproduction/fix" of sort highlighting the issue. Refer to the README for how to operate it.

[pnp-poc.zip](https://github.com/user-attachments/files/16086339/pnp-poc.zip)

### Checklist

   I don't think there's a good way to test that, I'll QA manually with the Alpha.

### How to QA [^1]

1. Create a Next.js website, kill the process when it gets to install dependencies
   ```bash
   npx create-next-app
   ```
2. Get in the new app, enable Yarn Berry, and install dependencies with Yarn
   ```bash
   cd my-app && yarn set version berry && yarn
   ```
3. Init Slice Machine with latest, it should fail
   ```
   yarn dlx @slicemachine/init@latest
   ```
4. Init Slice Machine with alpha, it should work
   ```
   yarn dlx @slicemachine/init@alpha.lh-fix-yarn-pnp-support
   ```
5. Start Slice Machine, it should work
   ```
   yarn slicemachine
   ```

[^1]:

### DIFF
--- a/packages/manager/src/lib/requireResolve.ts
+++ b/packages/manager/src/lib/requireResolve.ts
@@ -9,22 +9,22 @@ import _module, { createRequire } from "node:module";
  *
  * @returns - Resolved module path.
  */
-export default (id: string, from: string): string => {
-	let _id = id;
+export const requireResolve = (id: string, from: string): string => {
+	let resolvedID = id;
 
 	// Support Yarn PnP
 	if (
 		process.versions.pnp &&
 		"findPnpApi" in _module &&
-		_module.findPnpApi instanceof Function
+		typeof _module.findPnpApi === "function"
 	) {
 		const pnpApi = _module.findPnpApi(from);
 		if (pnpApi) {
-			_id = pnpApi.resolveRequest(id, from);
+			resolvedID = pnpApi.resolveRequest(id, from);
 		}
 	}
 
 	const require = createRequire(from);
 
-	return require.resolve(_id);
+	return require.resolve(resolvedID);
 };
--- a/packages/manager/src/lib/requireResolve.ts
+++ b/packages/manager/src/lib/requireResolve.ts
@@ -1,4 +1,4 @@
-import _module from "node:module";
+import _module, { createRequire } from "node:module";
 
 /**
  * Resolves a module path with `createRequire().resolve()` with Yarn PnP
@@ -24,7 +24,7 @@ export default (id: string, from: string): string => {
 		}
 	}
 
-	const require = _module.createRequire(from);
+	const require = createRequire(from);
 
 	return require.resolve(_id);
 };
--- a/packages/manager/src/lib/requireResolve.ts
+++ b/packages/manager/src/lib/requireResolve.ts
@@ -1,0 +1,30 @@
+import _module from "node:module";
+
+/**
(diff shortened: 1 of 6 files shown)


lihbr:
WIP: https://prismic-team.slack.com/archives/C02L3FN3AJK/p1720022475787989

packages/manager/src/lib/requireResolve.ts
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Resolves a module path with `createRequire().resolve()` with Yarn PnP
+ * support.
+ *
+ * @param id - Module to resolve.
+ * @param from - Location to resolve the module from.
+ *
+ * @returns - Resolved module path.
+ */
+export default (id: string, from: string): string => {
```
angeloashmore: ⚠️ #issue: Could we use a named export to be consistent with other files?

packages/manager/src/lib/requireResolve.ts
```diff
@@ -0,0 +1,30 @@
...
+	let _id = id;
+
+	// Support Yarn PnP
+	if (
+		process.versions.pnp &&
+		"findPnpApi" in _module &&
+		_module.findPnpApi instanceof Function
```
angeloashmore: 💡 #idea: `typeof _module.findPnpApi === "function"` seems more idiomatic to me. What do you think? Is there a benefit to using `instanceof`?

packages/manager/src/lib/requireResolve.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Resolves a module path with `createRequire().resolve()` with Yarn PnP
+ * support.
+ *
+ * @param id - Module to resolve.
+ * @param from - Location to resolve the module from.
+ *
+ * @returns - Resolved module path.
+ */
+export default (id: string, from: string): string => {
+	let _id = id;
```
angeloashmore: 💡 #idea: The `_` prefix usually denotes private data. In this case, `_id` is more like a temporary mutable value. How about naming it `resolvedID`?

xrutayisire (approved):
Super nice! Tested with yarn, yarn pnp, npm, let's go! 🚀

We are planning to release SM next week middle, end of the week. If you want to release before on your own to fix this bug, feel free to do it 🙂

packages/manager/src/lib/requireResolve.ts
lihbr: Copilot held my hand too far 🫠

packages/manager/src/lib/requireResolve.ts
lihbr: That also works, not sure why I went `instanceof` here

PR #1396, opened 2024-07-08
feat(init): display more prominent login cta message

jomifepe:
**Resolves**: DT-1926

### Description

> In the init, we have received several feedbacks showing that the login CTA is not clear. Users don't really read their terminal and don't understand that they have to click and leave the terminal to login. Some just freeze and just think it's loading forever.

### Checklist

### Preview
Screenshot 2024-07-09 at 16 40 30

Screenshot 2024-07-09 at 16 40 50

### How to QA [^1]

```bash
# build the script on packages/init
npm run dev

# if you already have a prismic session, logout by removing or backing up your config file
rm -rf ~/.prismic

# spin up a new playground to run the updated init script
yarn play --new
```

[^1]:

### DIFF
--- a/packages/init/src/SliceMachineInitProcess.ts
+++ b/packages/init/src/SliceMachineInitProcess.ts
@@ -575,12 +575,12 @@ ${
 
 					if (!isLoggedIn) {
 						parentTask.title = this.getLoggingInTitle(
-							`${chalk.cyan("Press any key to open the browser to login...")}`,
+							chalk.cyan("Press any key to open the browser to login..."),
 						);
 						await this.pressKeyToLogin();
 						await this.waitingForLogin(({ url }) => {
 							parentTask.title = this.getLoggingInTitle(
-								`${chalk.cyan("Opening browser, waiting for you to login...")}`,
+								chalk.cyan("Opening browser, waiting for you to login..."),
 								chalk.yellow(
 									"If your browser did not open automatically, please use the url below:",
 								),
@@ -639,11 +639,10 @@ ${
 		) => void,
 	): Promise<void> {
 		const sessionInfo = await this.manager.user.getLoginSessionInfo();
-		const { port, url } = sessionInfo;
 		await this.manager.user.nodeLoginSession({
-			port,
+			port: sessionInfo.port,
 			onListenCallback() {
-				open(url);
+				open(sessionInfo.url);
 				onListenCallback?.(sessionInfo);
 			},
 		});
--- a/packages/init/src/SliceMachineInitProcess.ts
+++ b/packages/init/src/SliceMachineInitProcess.ts
@@ -548,13 +548,20 @@ Continue with next steps in Slice Machine.
 		}
 	}
 
-	protected getLoggingInTitle(description?: string): string {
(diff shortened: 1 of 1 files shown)


packages/init/src/SliceMachineInitProcess.ts
```diff
@@ -606,12 +631,20 @@ Continue with next steps in Slice Machine.
...
+			>,
+		) => void,
+	): Promise<void> {
+		const sessionInfo = await this.manager.user.getLoginSessionInfo();
+		const { port, url } = sessionInfo;
 		await this.manager.user.nodeLoginSession({
 			port,
 			onListenCallback() {
 				open(url);
+				onListenCallback?.(sessionInfo);
```
dani-mp, line 640: Probably a question for a more long-term team member, but i wonder why we use a callback instead of the promise we already have here.

jomifepe: I suspect that, for this particular case, is because we want to open the browser once the we start listening for auth events, but we only resolve the promise once the authentication succeeds.

levimykel:
For some reason, the link is difficult for me to click in VS Code's terminal. It wasn't an issue in Warp. It is also adding lines sometimes when I resize the application.

https://github.com/prismicio/slice-machine/assets/9385378/b77c7ca2-32b1-4a43-9114-2facdb22557e

jomifepe:
> For some reason, the link is difficult for me to click in VS Code's terminal. It wasn't an issue in Warp. It is also adding lines sometimes when I resize the application.
> 
>  Screen.Recording.2024-07-08.at.18.25.06.mov

Unfortunately, that seems related to the spinning animation (done with [listr](https://github.com/SamVerschueren/listr)), it also happens for other pending actions and in other terminal emulators 😕

https://github.com/prismicio/slice-machine/assets/7235666/3707038a-5ab0-43f1-a4d3-6acb3c358586

Will keep taking a look to see if there's any way to prevent this.

Regarding the URL issue, I'm able to click it using warp, does it happen all the time or just when you resize and the lines get duplicated?

jomifepe:
> For some reason, the link is difficult for me to click in VS Code's terminal. It wasn't an issue in Warp. It is also adding lines sometimes when I resize the application.
> 
>  Screen.Recording.2024-07-08.at.18.25.06.mov

Move the first lines out of the task title, so that the animation is lower and the line separator is not the one that gets duplicated on resize, should look a bit better 😄

levimykel:
> Regarding the URL issue, I'm able to click it using warp, does it happen all the time or just when you resize and the lines get duplicated?

For me, the link issue is just in the VS Code terminal and it's happening all the time.

jomifepe:
> > Regarding the URL issue, I'm able to click it using warp, does it happen all the time or just when you resize and the lines get duplicated?
> 
> For me, the link issue is just in the VS Code terminal and it's happening all the time.

Yeah it seems to be flashing on the VS Code terminal and makes it hard to click, probably related to the loading animation, but I'm able to click it with some effort 😅 Copying also works, probably better than clicking.

levimykel (approved):
Looks great, nice work!

PR #1399, opened 2024-07-15
feat(onboarding): revamped onboarding flow mvp

jomifepe:
**Resolves**: DT-2218

### Description

→  Proposal: https://www.notion.so/prismic/Onboarding-Guide-Revamp-6120dcbc7f2342b58c8cae03df7efa28
→  Design: https://www.figma.com/design/IzGRpY6Gp5wDr7XCTwGl6G/Onboarding

This PR implements the core/base functionality of the revamped onboarding flow. As it's more focused on the core functionality and implementing the full flow, it's only using existing components. In the next iterations, more tailored components will be implemented to meet the design specification.

Backwards compatibility: 
- If the user has previously dismissed the current guide, this new one won't appear.

### Checklist

### Preview

https://github.com/user-attachments/assets/5c23a3cc-ac1c-485f-ada4-307a04b182db

### How to QA [^1]

```bash
# after checking out to this branch
yarn dev
yarn play
```

[^1]:


packages/slice-machine/src/features/onboarding/OnboardingGuide.module.css
```diff
@@ -0,0 +1,14 @@
+.onboardingCard {
```
BohdanOne: 💡 #idea: I think you could use Card component from the `editor-ui`

jomifepe: Yeah that sounds good, applied 👍

packages/slice-machine/src/hooks/useLocalStorageItem.ts
```diff
@@ -0,0 +1,62 @@
...
+export type UseLocalStorageReturnType<T> = [
+  T,
+  Dispatch<SetStateAction<T>>,
+  UseLocalStorageItemInfo,
+];
+
+const SLICE_MACHINE_STORAGE_PREFIX = "slice-machine";
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const isFunction = (value: any): value is (...args: any[]) => any => {
```
BohdanOne: ❓ #ask: wouldn't type `unknown` be suitable here instead of any?

packages/slice-machine/src/features/onboarding/helpers.tsx
```diff
@@ -0,0 +1,165 @@
...
+};
+
+export type OnboardingStepContent = Record<
+  OnboardingStepType,
+  OnboardingStepContentDefinition
+>;
+
+export type OnboardingStepStatuses = Record<OnboardingStepType, boolean>;
+
+const steps: OnboardingStep[] = [
```
BohdanOne: 💡 #idea: there is quite a lot of things going on in this file, maybe this piece of data could be moved to other file e.g. `OnboardingGuide.data.ts` or `OnboardingGuide.const.ts` or just 'constants.ts` ?

packages/slice-machine/src/features/onboarding/helpers.tsx
```diff
@@ -0,0 +1,165 @@
...
+    title: "Write Content",
+    description: "Add engaging content.",
+  },
+];
+
+export const useOnboardingStepsContent = (): OnboardingStepContent => {
+  const { repositoryUrl } = useRepositoryInformation();
+
+  return {
+    createPage: {
```
BohdanOne: 💡 #idea: likewise here, maybe content could be stored aside of the logic

packages/slice-machine/src/hooks/useLocalStorageItem.ts
jomifepe: Using `unknown` doesn't allow the type of `value` to be inferred correctly 😕

With `unknown`
Screenshot 2024-07-16 at 16 42 35

With `any`
Screenshot 2024-07-16 at 16 40 24

BohdanOne: thanks for explanation!

packages/slice-machine/src/features/onboarding/helpers.tsx
jomifepe: There isn't a clear pattern for organizing these, and since I didn't see any file with that naming pattern, I left it as is to get some feedback in the review.

BohdanOne: sure, the naming is actually least important to me, the point was to just move content & config out of the code to make the file less heavy
I'm also curious what are the patterns for that here :)

jomifepe: But I also like to see things separate, so I just committed a refactor 😄

packages/slice-machine/src/hooks/useLocalStorageItem.ts
jomifepe: More context: Doing an inline `typeof value === 'function'` should suffice here, but it doesn't work, and I think it's due to a bug in this typescript version, hence the function with a type guard. I will try to look into that soon and propose to update it, as we're a bit behind on 4.9.5

BohdanOne (commented):
🎉 #nice: I find much better to read after last commits that organised the code into separate files!

packages/slice-machine/src/features/onboarding/OnboardingGuide.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const OnboardingGuide = () => {
+  const { steps, completedStepCount, isComplete } = useOnboardingProgress();
+
+  if (isComplete) return null;
+
+  return (
+    <Card color="grey2" variant="outlined" paddingBlock={16}>
+      <CardContent>
+        <div>
```
dani-mp, line 18: Not a biggie but maybe you can remove this one by nesting the texts.

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+type DialogState = {
+  isOpen: boolean;
+  step: OnboardingStep;
+};
+
+export const OnboardingProgressStepper = () => {
+  const { completedStepCount, steps, isStepComplete } = useOnboardingProgress();
+
+  const [dialog, setDialog] = useReducer(
```
dani-mp: Not important, but I'd use setState here, given that we're not using actions/events in the reducer and just using it as a setter. I know that this allows you to partially update the state but with setState is fairly easy to do so (especially for such a small state) and it's more explicit.

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
```diff
@@ -0,0 +1,78 @@
...
+    (prev: DialogState, next: Partial<DialogState>) => ({ ...prev, ...next }),
+    { isOpen: false, step: steps[0] },
+  );
+
+  const openDialog = (step: OnboardingStep) => {
```
dani-mp: Maybe we can call this `showStep`?

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.module.css
```diff
@@ -0,0 +1,12 @@
...
+  display: flex;
+  flex-direction: column;
+  gap: 16px;
+}
+
+.contentHeader {
+  > *:first-child {
+    margin-bottom: 4px;
+  }
+}
```
dani-mp: I would do this with markdown and gap instead of with a selector on the first child. I find it confusing to wrap the content with an element named content header, for instance.

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
```diff
@@ -0,0 +1,71 @@
...
+            <div className={styles.contentHeader}>
+              <Text color="purple9" variant="bold">
+                Step {getStepIndex(step) + 1}
+              </Text>
+              <Text variant="h3">{stepContent.title ?? step.title}</Text>
+              {typeof stepContent.content === "string" ? (
+                <Text>{stepContent.content}</Text>
+              ) : (
+                <stepContent.content />
+              )}
```
dani-mp: Is there a way to avoid this? For instance, wrapping the string content already in <Text>

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
```diff
@@ -0,0 +1,71 @@
...
+            </div>
+            <Video src={stepContent.videoUrl} sizing="contain" autoPlay loop />
+          </div>
+        </ScrollArea>
+        <DialogActions
+          ok={{
+            text: isStepComplete(step) ? "Undo step" : "Mark as done",
```
dani-mp: I saw that when marking the step as done, you can see the button you just clicked changing to "Undo step" while the dialog is closing. Not sure there's an easy way to solve this, though, so maybe it's ok.

packages/slice-machine/src/hooks/useLocalStorageItem.ts
```diff
@@ -0,0 +1,62 @@
...
+): UseLocalStorageReturnType<T | undefined> {
+  const storageKey = `${SLICE_MACHINE_STORAGE_PREFIX}_${key}`;
+  const getSnapshot = () => localStorage.getItem(storageKey);
+
+  const subscribe = (listener: () => void) => {
+    window.addEventListener("storage", listener);
+    return () => void window.removeEventListener("storage", listener);
+  };
+
+  const serializedItem = useSyncExternalStore(subscribe, getSnapshot);
```
dani-mp: `subscribe` here changes in every render, which makes `useSyncExternalStore` to subscribe/unsubscribe for every render. You can solve it by just moving the `subscribe` declaration out of the body of the component.

packages/slice-machine/src/hooks/useLocalStorageItem.ts
```diff
@@ -0,0 +1,62 @@
...
+    try {
+      const newValue = JSON.stringify(isFunction(value) ? value(item) : value);
+      localStorage.setItem(storageKey, newValue);
+      window.dispatchEvent(new StorageEvent("storage", { key, newValue }));
+    } catch (error) {
+      console.warn(`Error setting localStorage key “${storageKey}”:`, error);
+    }
+  };
+
+  return [item, setItem, { isUnset: serializedItem == null }];
```
dani-mp: Can we have just [item, setItem] and assume that it's unset if it's null/undefined?

Otherwise I rather have `{ item, setItem, isUnset }` in the return.

packages/slice-machine/src/hooks/useLocalStorageItem.ts
```diff
@@ -0,0 +1,62 @@
...
+      return defaultValue;
+    } catch (error) {
+      console.warn(`Error reading localStorage key “${storageKey}”:`, error);
+      return defaultValue;
+    }
+  }, [serializedItem, defaultValue, storageKey]);
+
+  const setItem = (value: SetStateAction<T | undefined>) => {
+    try {
+      const newValue = JSON.stringify(isFunction(value) ? value(item) : value);
```
dani-mp: What's the use case for this `isFunction`? I think we could simplify this implementation by not allowing to pass a function. If the caller needs this, they can do it by themselves.

packages/slice-machine/src/hooks/useLocalStorageItem.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  const subscribe = (listener: () => void) => {
+    window.addEventListener("storage", listener);
+    return () => void window.removeEventListener("storage", listener);
+  };
+
+  const serializedItem = useSyncExternalStore(subscribe, getSnapshot);
+  const item = useMemo(() => {
+    try {
+      if (serializedItem != null) return JSON.parse(serializedItem) as T;
```
dani-mp: Rather than casting here, I would pass a zod schema to parse the value.

This will help us in the future because if the requirements change and you have a different schema, the zod version will fail and provide the default value, while the casting version will just lie about it.

packages/slice-machine/src/hooks/useLocalStorageItem.ts
```diff
@@ -0,0 +1,62 @@
...
+      return defaultValue;
+    } catch (error) {
+      console.warn(`Error reading localStorage key “${storageKey}”:`, error);
+      return defaultValue;
+    }
+  }, [serializedItem, defaultValue, storageKey]);
```
dani-mp: This memo doesn't feel like the best solution here, and there seem to be some issues (unless I'm being mistaken).

- If the `storageKey` changes, `getSnapshot` is not called again, so an item will be provided for the old key.
- `defaultValue` can be a non-primitive, so this memo will be re-evaluated in every render.

I think this logic must be inside `getSnapshot`, and a new subscription must happen when there's a new key, but not for every render.

packages/slice-machine/src/features/onboarding/helpers/useOnboardingProgress.ts
```diff
@@ -0,0 +1,70 @@
...
+  useEffect(() => {
+    // populate onboarding status if not defined
+    if (isStatusUnset) setStatus(initialState);
+
+    /* eslint-disable-next-line react-hooks/exhaustive-deps -- 
+    We don't care if the values of isStatusUnset, setStatus or initialState change here. */
+  }, [status]);
+
+  return [status, setStatus];
+};
```
dani-mp: I suggest we never disable the exhaustive deps rule. They're there for a reason and most of the time there's a better way of doing what we need to do if we feel the need of bypassing them. Also, in the future we'll have the react compiler that will do this for us and it won't skip them, so our code will behave differently.

For this concrete use case, the recommended approach is to use an inline set state: https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

packages/slice-machine/src/features/onboarding/content.tsx
```diff
@@ -0,0 +1,77 @@
...
+    videoUrl:
+      "https://res.cloudinary.com/dmtf1daqp/video/upload/v1700213517/IN-APP-GUIDE-SM/PREVIEW.mp4",
+    content:
+      "Ipsum aliqua do consequat eiusmod id. Reprehenderit consectetur sit officia consequat velit non officia aliquip laboris incididunt cillum proident incididunt. Ad quis laborum tempor dolor duis ea cillum aliqua occaecat. Sunt reprehenderit fugiat et ullamco proident pariatur deserunt minim. Irure enim nulla et ad ut id anim elit. Voluptate culpa esse qui et reprehenderit aute est.",
+  },
+  writeContent: {
+    videoUrl:
+      "https://res.cloudinary.com/dmtf1daqp/video/upload/v1700213517/IN-APP-GUIDE-SM/WRITE.mp4",
+    content: () => (
+      <span>
```
dani-mp: Let's use Text here.

dani-mp (commented):
Great job, I've left some comments.

Apart from the comments in the code, something that caught my attention was that the folder structure was too nested and difficult to follow.

I'm not sure yet how the team prefers to do this, but my personal preference is to have:
- A clear entry point with an index in the top level
- Every folder must also have an index to avoid confusion about the implementation details
- Avoid extracting to a new folder/module things that are used only once, prefer to inline them in the module they're used (colocation)
- Declare the types together with the data structures, to avoid indirection

And a final note, in the future we need to create smaller PRs that are easier/faster to review. For instance, this PR could have been split into 3:
1. `useLocalStoreItem`, because it's a generic hook and we need to get the API right.
2. The dropdown
3. The dialog

packages/slice-machine/src/features/onboarding/helpers/useOnboardingProgress.ts
```diff
@@ -0,0 +1,70 @@
...
+  return Object.fromEntries(
+    steps.map((step) => [step.id, startComplete]),
+  ) as OnboardingStepStatuses;
+};
+
+const useOnboardingStepStatus = (): [
+  OnboardingStepStatuses,
+  Dispatch<SetStateAction<OnboardingStepStatuses>>,
+] => {
+  const initialState = useRef(getInitialState()).current;
```
dani-mp: Forgot this one, it's not recommended to read (or set) refs during the render of a component: https://react.dev/reference/react/useRef#caveats

packages/slice-machine/src/features/onboarding/helpers/useOnboardingProgress.ts
```diff
@@ -0,0 +1,70 @@
...
+  const [status, setStatus, { isUnset: isStatusUnset }] = useLocalStorageItem(
+    "onboardingSteps",
+    initialState,
```
dani-mp: An initial state is not the same as a default value, so either the API for useLocalStorageItem is not the best for this or the usage is not accorded to the API.

dani-mp (commented):
Left two more comments.

packages/slice-machine/src/features/onboarding/OnboardingGuide.tsx
jomifepe: I could do that, but since `Text` renders a span, it will appear inline by default. I could create a class to apply `display: block` to the parent Text. Not sure I prefer having that, but I also don't like unnecessary nesting just for styling, so what do you prefer?

dani-mp: not important, you can leave it like this

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
jomifepe: Yeah that's fair, I'll change to two `useState`s, as with only one I would have to spread the current value in at least one call, and IMO that's doesn't look great.

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
jomifepe: Sounds good to me 👍

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
dani-mp: But why two? I mean you can keep the same state in a use state and use the spread to merge with the previous state. Same as before but a bit more explicit.

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.module.css
jomifepe: I get the confusion. I don't think I'll use a header, because in really it's not really a header, but I changed it to apply the bottom margin directly to the element that need it instead of having a parent. 

https://github.com/prismicio/slice-machine/pull/1399/commits/cbe7c4b684f85548a82fbce83bfdeb82fa3cac9a#diff-b6bf9a3e2b4ced38967ebbb7c67a7a4d0baa20e5ea3758c09170bd4005791b6aR1

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
```diff
@@ -0,0 +1,75 @@
...
+    toggleStepComplete(step.id);
+    onClose();
+  };
+
+  return (
+    <Dialog open={isOpen} onOpenChange={onClose}>
+      <DialogHeader title="Learn" />
+      <DialogContent>
+        <ScrollArea>
+          <article className={styles.scrollableContent}>
```
xrutayisire: 💡 #idea

This article is making one more wrapper, we could just use `ScrollArea` no? You can see the usage of `ScrollArea` in `editor` project with unlined style

One goal we try to have in to prevent using CSS modules in features folder, something we didn't really do before, but that can be a good indication that something is wrong.

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
```diff
@@ -0,0 +1,75 @@
...
+            <section>
+              <Text
+                className={styles.stepNumberText}
```
xrutayisire: 💡 #idea

I don't think you need this classname, see my message above about prevent using CSS modules in `features` folder.

I think you could just use `SX` prop from `Text` right?

packages/slice-machine/src/hooks/useLocalStorageItem.ts
jomifepe: Yeah, completely missed that when I was moving the code around, nice catch 😅

packages/slice-machine/src/hooks/useLocalStorageItem.ts
jomifepe: The idea here was to have the hook be usable like a `useState`, and be able to pass a value or a value setter, I believe it's useful to have that. The `isFunction` was to make TS happy, as `typeof value === "function"` was not enough, but I could also use `value instanceof Function`.

I'm making use of this here: https://github.com/prismicio/slice-machine/pull/1399/files#diff-98f74169fc5172f72a4d695241328bab2abeed3b11a4e5e559cdd3270e231337R47

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
```diff
@@ -0,0 +1,72 @@
...
+        onClose={() => setDialogOpen(false)}
+      />
+      <DropdownMenu>
+        <DropdownMenuTrigger>
+          <Button color="grey" sx={{ width: "100%" }}>
+            <Box alignItems="center" gap={4}>
+              <Text variant="bold" color="grey12">
+                {completedStepCount > 0 ? "Continue" : "Start"}
+              </Text>
+              <Icon name="playCircle" size="small" color="grey11" />
```
xrutayisire: ⚠️ #issue

Button have props for icons, and then you don't need Box, it's all handled with the SD.
You also don't need the wrapper for `Text`.

packages/slice-machine/src/hooks/useLocalStorageItem.ts
jomifepe: So this was mostly to avoid having extra null check around the `status` value. By passing it the initial value, it will always have a value and you can use it freely, when spreading the current value and so on.

jomifepe: I also thought about, since this is used to manage a local storage item, having the `defaultValue` be written to the storage if it's unset there (maybe having a prop to tell it to do so), that way we'll always have something to work with. This would also solve https://github.com/prismicio/slice-machine/pull/1399#discussion_r1680779137

packages/slice-machine/src/hooks/useLocalStorageItem.ts
jomifepe: I initially went that route (handling it inside `getSnapshot`), but my issue was that it kept yelling at me with: `useLocalStorageItem.ts:43 Warning: The result of getSnapshot should be cached to avoid an infinite loop`, even with a try catch inside of it.

jomifepe: You're right about those pitfalls, but I believe they can be solved by simply restricting how the hook can be used. Both `key` and `defaultValue` should be hardcoded or only the initial value is considered, just like `useState`/`useRef` initializer.

packages/slice-machine/src/features/onboarding/content.tsx
jomifepe: Sounds good 👍

packages/slice-machine/src/features/onboarding/helpers/useOnboardingProgress.ts
jomifepe: I would usually never agree with `never disable the exhaustive deps rule`, but you made a valid point about adjusting the state during render time, so that sounds like a good alternative to me 👀

packages/slice-machine/src/features/onboarding/helpers/useOnboardingProgress.ts
jomifepe: That's a nice catch, thanks 🙂 I ended up removing this line entirely while refactoring.

packages/slice-machine/src/features/onboarding/helpers/useOnboardingProgress.ts
jomifepe: In this case, it was a default state (also the name of the hook prop), I was writing it as an initial state outside. I ended up changing the API and treating it as an initial state, and now the hook writes it to the storage if it's not defined already.

packages/slice-machine/src/hooks/useLocalStorageItem.ts
jomifepe: I agree with that, but I don't see `zod` being used anywhere in the slice-machine codebase, would it be worth to introduce it for this use case?

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
jomifepe: Since the radix dialog uses an `open` prop to manage itself, and have a open/close animation, I have one for that, and another for the `step` to pass to the dialog. I avoid setting the step to `undefined` to avoid unnecessary null check and still allow the whole dialog to be rendered and the closing animation to happen.

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
jomifepe: Do you mean not allowing strings and always defining a component?

packages/slice-machine/src/hooks/useLocalStorageItem.ts
dani-mp: Let's ask @xrutayisire, we use zod a lot across prismic for everything that is validation, from API responses to make sure environment variables are correct, etc.

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
dani-mp: yes

packages/slice-machine/src/hooks/useLocalStorageItem.ts
xrutayisire: Yes, Zod is the new way to go globally at Prismic. I don't see problem to introduce it now 🙂

BUT haha, my question is actually more with the approach, we already have a hook `usePersistedState` that allow us to sync a state with the storage. 

Why not go with this approach? What are the pros / cons of this one vs yours? I just see this one as a bit more complex, no?

Note: Even `usePersistedState` would benefit from a Zod schema, so not really related.

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
jomifepe: Sure that also works, changed!

packages/slice-machine/src/hooks/useLocalStorageItem.ts
jomifepe: I've been meaning to talk about this. I wanted to have the state react to store changes, but I could also use the `usePersistedState` hook and use a context above the onboarding components, for reactivity purposes. 

Would you prefer that?

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
jomifepe: The reason to use the inner container here is that by styling the `ScrollArea` with a property like padding, the scrollbar will get pushed inwards. 

What would be the correct approach to apply styles to these components? Create them in the components folder?

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
jomifepe: I don't particularly enjoy applying inline styles 😅 It's fine for one or two properties, but doesn't seem like a approach. Can you clarify why we shouldn't create CSS modules in `features`? 

There will always be the need for specific components, which don't make sense to have under a `components` directory because they are not used anywhere else. So for those ones, we will often need to apply some styles.

I applied that for the `Text`, as it's only a `marginBottom`, but what should I do with the `section` without creating a CSS module?

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
jomifepe: https://github.com/prismicio/slice-machine/pull/1399/commits/f73d86f900b01bd2c53ebc4672b8e0bb8e9fcc67
Much cleaner, thanks for the suggestion 🔥

packages/slice-machine/src/hooks/useLocalStorageItem.ts
dani-mp: I think that approach would be better for at least three reasons:
- One is the separation of concerns.
- The other one is that here we're dispatching a broad event. If you use this new hook in several places, they will all get notified even though their key hasn't changed.
- Finally, I think the performance of reading from React Context should be better than reading from local storage + parsing every time.

packages/slice-machine/src/features/onboarding/helpers/useOnboardingStepsContent.ts
```diff
@@ -0,0 +1,21 @@
...
+): OnboardingStepContentDefinition;
+export function useOnboardingStepsContent(
+  stepId?: OnboardingStepType,
+): OnboardingStepContent | OnboardingStepContentDefinition {
+  const { repositoryUrl } = useRepositoryInformation();
+  const content = getOnboardingStepsContent({ repositoryUrl });
+
+  if (stepId) return content[stepId];
+  return content;
+}
```
dani-mp: So now that we've refactored this, it's clear that this is no longer a hook, but a component: receives some props, renders something.

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
xrutayisire: That's the question, do you need CSS modules at all ever in features? One goal is to prevent specific components with specific style. If it's only Layout, then you `Box`. 
It's something PB introduced and prevent usage of styling directly in their wroom app. We could discuss this topic with them, I think they would give us better perspective on this.

packages/slice-machine/src/hooks/useLocalStorageItem.ts
jomifepe: Yeah sure, I'm fine with both. Moved into a context based approach. Adjusted the `usePersistedState` a bit to suit simplify storing the initial state @xrutayisire.

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
jomifepe: I personally like to leave styles to CSS, mixing CSS with a styled system can get messy. But if that's how we're doing things now, I can refactor.

jomifepe: Refactored 🧼

packages/slice-machine/src/features/onboarding/helpers/useOnboardingStepsContent.ts
jomifepe: Yeah sounds better, refactored!

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
jomifepe: Solved: Refactored to a Box

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
jomifepe: I'm looking into this one. It's doable using a `setTimeout` to match the time of the animation, or maybe by leveraging the `onAnimationStart` prop from radix, which is not being exposed right now by the `Dialog` component from editor-ui.

packages/slice-machine/src/hooks/useLocalStorageItem.ts
jomifepe: Not applicable anymore, moved to `usePersistedState` hook

packages/slice-machine/src/hooks/useLocalStorageItem.ts
jomifepe: No longer applicable, but I applied the same logic mention in the last comment to the existing `usePersistedState`

packages/slice-machine/src/features/onboarding/helpers/useOnboardingProgress.ts
jomifepe: Refactored to `usePersistedState`, this might still be relevant there

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
dani-mp: Too complicated, I think, not sure if it's worth it. I think we can keep it like this.

packages/slice-machine/src/features/onboarding/content.tsx
```diff
@@ -0,0 +1,115 @@
...
+  repositoryUrl,
+}: GetOnboardingStepsContentProps): Record<
+  OnboardingStepType,
+  OnboardingStepContentDefinition
+> => ({
+  createPage: {
+    videoUrl:
+      "https://res.cloudinary.com/dmtf1daqp/video/upload/v1700213517/IN-APP-GUIDE-SM/ADD_SLICE.mp4",
+    content: () => (
+      <>
```
dani-mp: Shouldn't these be <Text>?

jomifepe: Be Text? Yeah, I didn't add it because it was just placeholder text

dani-mp: yeah, i promise you i wrote `<Text>` without the code formatting and when i posted github editor must have drop it - I'm not talking about the content of the component, but the component itself, not sure what you mean with placeholder here, i understand that the text won't be final, but i guess we want to use our own font and styling, right?

jomifepe: Not sure I'm following 😕 Can you give a code example?

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
xrutayisire: I could explain a bit more the vision next time we talk, and you can tell me what you think.

packages/slice-machine/src/features/onboarding/content.tsx
dani-mp: it's ok, i saw that you already fixed it https://github.com/prismicio/slice-machine/pull/1399/commits/76f217b8fd65a4fb8605bacbecda9cc58814dcd3

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
dani-mp: yeah, i meant that before (where you were using useReducer) you had the state in one object, but you split it into two states when you moved to useState. But that's ok, both work 👍

packages/slice-machine/src/features/onboarding/content.tsx
jomifepe: Oh ok, so I understood it correctly 😄

packages/slice-machine/src/features/onboarding/OnboardingProvider.tsx
```diff
@@ -0,0 +1,85 @@
...
+  isComplete: boolean;
+};
+
+export const OnboardingContext = createContext<OnboardingContext | undefined>(
+  undefined,
+);
+
+const getInitialState = (): OnboardingStepStatuses => {
+  // if the old guide was dismissed, all steps start as complete
+  const startComplete =
```
BohdanOne: 💡 #idea: tiny thing, but maybe this could be renamed to a `isOnboardingCompleted`, or `isCompleted`

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
jomifepe: Yeah I used two to avoid doing stuff like `setDialog((prev) => ({ ...prev: open: false }))` just to open the dialog, this way you only change the `setDialogOpen` in that case.

packages/slice-machine/src/features/onboarding/OnboardingProvider.tsx
jomifepe: I prefer to mention that it's regarding the previous guide, but I renamed it to be more clear (`wasOldGuideDismissed`)

BohdanOne: actually, we shouldn't really care if it was new or the old guide that was completed - if it was completed then we don't show it.
Which brings me idea that if all steps are completed in the new guide we also should persist that information and use it here, WDYT?

jomifepe: Regarding the old guide, we discussed not displaying the new one if the user had previously dismissed it before, that's why I'm starting with all completed. This value is also only used if there isn't already something stored. Not sure if that's the issue you're describing 😄

BohdanOne: I was thinking about keeping one more piece of persistent state -> that whole onboarding was completed, then we can exit early, without computing each step etc.
On the other hand there is not much things to process anyway so the way you did is perfectly fine I guess.

packages/slice-machine/src/hooks/useLocalStorageItem.ts
jomifepe: Agreed to implement in a different PR. Resolving.

packages/slice-machine/src/hooks/useLocalStorageItem.ts
jomifepe: No longer applicable. Resolving.

packages/slice-machine/src/features/onboarding/OnboardingProvider.tsx
jomifepe: I don't think it's necessary, it's not too much to compute and having an extra value means having multiple sources of truth.

dani-mp (approved):
🚀 

Let's work on the feature flag in a different PR before merging this one so we have that code available to hide it!

jomifepe:
Almost 100 individual comments 😄 
Thank you for the reviews @BohdanOne @dani-mp @xrutayisire 🚀

PR #1400, opened 2024-07-17
feat: add field dropdown

xrutayisire:
**Resolves**: DT-2224

### Description

- Rework to have a clean way to manage fields in SM, we will have the two ways to work with fields until we totally finish migrating fields to new design / typescript.
- The goal is now to use `domain/fields.ts` file instead of the old one. The old widgets should start to get less and less use until totally removed.
- Create a new functional component `AddFieldDropdown` with no link at all with old code (that was the goal here).
- Let template management for next PR.
- Align old icons with new icons from new design.
- Update `editor-ui` to have new `DropdownMenu` features.
- Reorder widgets to ensure it's displayed in the order we would like now.

## TODO

- Confirm and validate the copy for all fields
- Update the tooltip preview images

### Checklist

### Preview

https://github.com/user-attachments/assets/b539cfe6-6482-4df9-9d24-117cdf1e31f0

### How to QA [^1]

- Install `slice-machine-ui` alpha `2.2.2-alpha.xru-add-field-dropdown.1`
- Anywhere you would like to add a filed you should not see the modal anymore but the add field dropdown.

[^1]:


packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
```diff
@@ -171,15 +170,35 @@ export const CustomListItem = ({
...
-            onClick={() => setSelectMode(true)}
-          >
-            Add Field
-          </Button>,
+          <AddFieldDropdown
+            key="add-field-dropdown"
+            disabled={newFieldData !== null}
+            onSelectField={onSelectFieldType}
+            fields={
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
```
jomifepe: Since we have a `eslint-disable` on top, I guess you can avoid disabled the rules here (an below), no? 👀

💡 Maybe we could disable all TS eslint rules for all JS files or the one under `legacy/`

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
```diff
@@ -171,15 +170,35 @@ export const CustomListItem = ({
...
+              widgetsArray
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+                .filter((e) => e)
```
jomifepe: Tiny improvement suggestion 🧼

```suggestion
                .filter(Boolean)
```

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
xrutayisire: Didn't even see the global disabled rule, I copy-paste the old code. Good catch.

Hum, it could be a good thing to do for sure, even if we would need to understand what would the consequence for us. Sometimes we are working on JS files still and we may want to have some protection. WDYT?

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
xrutayisire: Yeah better good catch, copy-paste this old code, when working on JS file, not trying to really improve things 😅 
I'm mostly focusing on extracting a little by little code from legacy to new folders.

xrutayisire: Done 🙂

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
xrutayisire: Done 🙂

packages/slice-machine/src/domain/fields.ts
```diff
@@ -0,0 +1,202 @@
...
+  GroupFieldType,
+  ImageFieldType,
+  LinkFieldType,
+  NestableFieldTypes,
+  NumberFieldType,
+  RichTextFieldType,
+  SelectFieldType,
+  TextFieldType,
+  TimestampFieldType,
+  UIDFieldType,
```
dani-mp: 💡 In the past we've discussed a lot with other folks about removing the usage of these constants and use string literals instead. This is because for TypeScript they're the same, but for the developer is fewer things to maintain and less indirection.

packages/slice-machine/src/domain/fields.ts
```diff
@@ -0,0 +1,202 @@
...
+  description: string;
+  icon: IconName;
+  name: string;
+  type: FieldType;
+  variant?: FieldVariants;
+}
+
+export type NestableFieldVariants = "ContentRelationship" | "LinkToMedia";
+export type GroupFieldVariants = "NestedGroup";
+export type FieldVariants = NestableFieldVariants | GroupFieldVariants;
```
dani-mp: 💡 
```suggestion
export type NestableFieldVariant = "ContentRelationship" | "LinkToMedia";
export type GroupFieldVariant = "NestedGroup";
export type FieldVariant = NestableFieldVariants | GroupFieldVariants;
```

packages/slice-machine/src/domain/fields.ts
```diff
@@ -0,0 +1,202 @@
...
+  TimestampFieldType,
+  UIDFieldType,
+} from "@prismicio/types-internal/lib/customtypes";
+
+export interface Field {
+  description: string;
+  icon: IconName;
+  name: string;
+  type: FieldType;
+  variant?: FieldVariants;
```
dani-mp: ❓ does it make sense to export this base type?

packages/slice-machine/src/features/builder/AddFieldDropdown.tsx
```diff
@@ -0,0 +1,106 @@
...
+  return (
+    <DropdownMenu>
+      <DropdownMenuTrigger disabled={disabled}>
+        <Button startIcon="add" color="grey" data-testid={triggerDataTestId}>
+          Add a field
+        </Button>
+      </DropdownMenuTrigger>
+      <DropdownMenuContent align="end" maxHeight={400} collisionPadding={8}>
+        <DropdownMenuLabel>Single fields</DropdownMenuLabel>
+        {singleFieldsToRender.map((field) => renderDropdownItem(field))}
```
dani-mp: 💡 I'd use a `<DropdownItem field={field} />` component here.

packages/slice-machine/src/features/builder/AddFieldDropdown.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+        {groupFieldToRender && (
+          <>
+            <DropdownMenuLabel>Sets of fields</DropdownMenuLabel>
+            {renderDropdownItem(groupFieldToRender)}
```
dani-mp: 💡 Same as the previous one.

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
```diff
@@ -68,22 +67,38 @@ const Zone = ({
...
-    setSelectModalData({ isOpen: false });
   };
 
   const onCancelNewField = () => setNewFieldData(null);
+  const addFieldDropdown = (
+    <AddFieldDropdown
+      disabled={newFieldData !== null}
+      onSelectField={onSelectFieldType}
+      fields={
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
```
dani-mp, line 86: ❓ What's the problem here? Can't it be solved with an assert?

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
```diff
@@ -68,22 +67,38 @@ const Zone = ({
...
+          const { TYPE_NAME, CUSTOM_NAME } = widget;
+          return allFields.find(
+            (f) =>
+              f.type === TYPE_NAME &&
+              (CUSTOM_NAME === undefined || f.variant === CUSTOM_NAME),
+          );
+        })
+      }
+      triggerDataTestId={
+        isRepeatable === true ? "add-field-in-items" : undefined
```
dani-mp: 💡 
```suggestion
        isRepeatable ? "add-field-in-items" : undefined
```

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
```diff
@@ -171,15 +170,28 @@ export const CustomListItem = ({
...
-            variant="buttons.darkSmall"
-            onClick={() => setSelectMode(true)}
-          >
-            Add Field
-          </Button>,
+          <AddFieldDropdown
+            key="add-field-dropdown"
+            disabled={newFieldData !== null}
+            onSelectField={onSelectFieldType}
+            fields={widgetsArray.filter(Boolean).map((widget) => {
```
dani-mp, line 177: ❓ is this a big list? does it make sense to memoize it?

dani-mp (approved):
Left some comments, but nothing important that should block the PR.

packages/slice-machine/src/domain/fields.ts
xrutayisire: Nope you're right, mistake 👍

packages/slice-machine/src/domain/fields.ts
xrutayisire: I wanted to be the good dev here haha, I started without, If we believe we can slowly stop using them I'm up for it. Just need to be careful with it in PRs.

packages/slice-machine/src/features/builder/AddFieldDropdown.tsx
xrutayisire: Ah, you mean just creating another component in this file instead of a function? Good idea

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
xrutayisire: I don't even know nor try x)
It's old JS code from this file that I try not to touch, all JS files are so ugly with all of that. I'm focusing myself really on extracting code a little by little and prevent touching anything there.

I could try to prevent using all of this also, of course. WDYT?

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
xrutayisire: In SM we have an eslint rule that prevent that... --"

`An explicit comparison or type cast is required.eslint[@typescript-eslint/strict-boolean-expressions](https://typescript-eslint.io/rules/strict-boolean-expressions)`

xrutayisire: Maybe we could just remove it haha
We will need to review our setup in SM, lot and lot of things to improve there!

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
xrutayisire: It's not really a big list, like 15 simple objects. I don't think so, WDYT?

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
dani-mp: If it's legacy that's ok! we don't have to waste time on it 👍

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
dani-mp: Ah, I see. Maybe you can do just `Boolean(isRepeatable)`, then?

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
dani-mp: yeah, no need

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
jomifepe: I agree, we could keep the general rules but disable the TS ones for JS files to reduce the noise of all the disabled comments in these legacy components.

packages/slice-machine/src/domain/fields.ts
xrutayisire: Done

packages/slice-machine/src/domain/fields.ts
xrutayisire: Done

packages/slice-machine/src/domain/fields.ts
xrutayisire: At the end I updated it, the BaseField type no but Field type yes, used in AddFieldDropdown

packages/slice-machine/src/features/builder/AddFieldDropdown.tsx
xrutayisire: Done, I did spread {...field}, let me know what you think

packages/slice-machine/src/features/builder/AddFieldDropdown.tsx
xrutayisire: Done

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
xrutayisire: Done, yeah a bit better

packages/slice-machine/src/features/builder/AddFieldDropdown.tsx
dani-mp: It's ok. Usually, I don't spread because if tomorrow you need to add another prop, it can conflict with the name of one of the field properties, and for the component prop types you have to merge the two.

But if that's a problem in the future we can always put the field back as a prop, so no biggie.

dani-mp:
@xrutayisire LGTM!

packages/slice-machine/src/features/builder/AddFieldDropdown.tsx
xrutayisire: Oh good point also! I think I will put field then, it will be easy to read for the props interface also

PR #1401, opened 2024-07-18
feat(onboarding): allow to validate usePersistedDate value with schema

jomifepe:
**Resolves**: DT-2218

### Description

Allow to pass a `schema` prop to the `usePersistedState` to validate the persisted value when reading from local storage.

### Checklist

### How to QA [^1]

Define a `schema` object and pass it to `usePersistedState`'s options (third parameter).

[^1]:


packages/slice-machine/src/hooks/usePersistedState.ts
```diff
@@ -1,27 +1,40 @@
 import { Dispatch, SetStateAction, useEffect, useState } from "react";
+import { AnyZodObject } from "zod";
 
 type PersistedState<T> = [T, Dispatch<SetStateAction<T>>];
 
 const SLICE_MACHINE_STORAGE_PREFIX = "slice-machine";
 
+type UsePersistedStateOptions = {
+  schema?: AnyZodObject;
```
dani-mp: This must match T, you can use something like `ZodType<T, z.ZodTypeDef, unknown>` (maybe there's a better way, but we've been using this one in other codebases). See https://github.com/prismicio/editor/blob/df9589d286ec095ae29f17115074b5241b1721e0/packages/editor-support/src/Http.ts#L109.

packages/slice-machine/src/hooks/usePersistedState.ts
```diff
@@ -1,27 +1,40 @@
...
     try {
       const value = localStorage.getItem(computedKey);
 
-      return value !== null ? (JSON.parse(value) as T) : defaultValue;
+      if (value == null) return defaultValue;
+      if (!schema) return JSON.parse(value) as T;
+
+      const result = schema.safeParse(JSON.parse(value));
+      if (!result.success) throw result.error;
+      return result.data as T;
```
dani-mp: If we're going to throw anyway, let's use `schema.parse`. Also, we shouldn't cast here, as the schema is supposed to help us with that. But I guess with the previous comment this will be solved.

packages/slice-machine/src/features/onboarding/types.ts
```diff
@@ -17,9 +21,17 @@ export type OnboardingStepContentDefinition = {
   videoUrl: string;
 };
 
+export const onboardingStepStatusesSchema = z.object(
+  Object.fromEntries(
+    onboardingSteps.map((step) => [step.id, z.boolean()]),
+  ) as Record<OnboardingStepType, z.ZodBoolean>,
```
dani-mp: No need to cast here. Create an empty z.object({}), then iterate over each step and extend the initial schema with the id and the boolean. Something like:
```
schema = schema.extend({
              [key]: z.boolean(),
            })
```

packages/slice-machine/src/hooks/usePersistedState.ts
jomifepe: Makes sense, thanks! Made that change.

packages/slice-machine/src/hooks/usePersistedState.ts
jomifepe: Yeah makes total sense, done!

packages/slice-machine/src/features/onboarding/types.ts
jomifepe: Solved after a quick chat to understand the suggestion better ✅

dani-mp (approved):

PR #1402, opened 2024-07-18
feat(onboarding): add feature flag to onboarding guide component

jomifepe:
**Resolves**: DT-2218

### Description

Feature flag for the new revamped onboarding guide.

### Checklist

### How to QA [^1]

Toggle the feature flag and see if it shows up ¯\_(ツ)_/¯

[^1]:

### DIFF
--- a/packages/slice-machine/src/features/onboarding/useOnboardingExperiment.ts
+++ b/packages/slice-machine/src/features/onboarding/useOnboardingExperiment.ts
@@ -1,0 +1,6 @@
+import { useExperimentVariant } from "@/hooks/useExperimentVariant";
+
+export const useOnboardingExperiment = () => {
+  const variant = useExperimentVariant("slice-machine-onboarding");
+  return { eligible: variant?.value === "on" };
+};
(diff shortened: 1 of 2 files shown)


packages/slice-machine/src/features/onboarding/useOnboardingExperiment.ts
dani-mp, line 1: no need for this extra file, right?

xrutayisire: That's how we did in the past btw. It allowed us to share a feature flag everywhere. We could not to it when we know it's super scoped.

jomifepe: Renamed to follow the same pattern as the other experiment hooks

dani-mp: sounds good, i have a strong preference for co-location and reducing as much files to look at and maintain - i only look for opportunities to move things to a different place when i need to reuse them

but I'm ok with this too

jomifepe: I get that. I do think that, in this case, since we're not type checking the key, it's preferable to extract the hook.

dani-mp (approved):

PR #1403, opened 2024-07-19
chore(packages): upgrade typescript from v4 to v5

jomifepe:
**Resolves**: DT-2228

### Description

`5.5` was already release, but it requires some additional code changes. To keep it simple, `5.4.5` is the latest version that didn't need any major change. This upgrade also required newer `@typescript-eslint/*` and `eslint` versions, which only needed the following changes to keep most of the code unchanged:
- TSConfig: Remove monorepo root node_modules from type roots: https://devblogs.microsoft.com/typescript/announcing-typescript-5-1/#explicit-typeroots-disables-upward-walks-for-node_modules-types
- Suppress new stricter `no-unsafe-enum-comparison` rule in some cases, to keep code as is
- Convert some logical operators to nullish coalescing, to meet the stricter `prefer-nullish-coalescing` rule
- Suppress some more non-critical `no-unsafe-*` rules in legacy code files which already have a lot of these.
- 🐞 Fixed `restore-cache` ci action, which was not caching playwright's `node_modules`

### Checklist

### How to QA [^1]

[^1]:


packages/slice-machine/src/features/sync/AutoSyncProvider.tsx
```diff
@@ -110,7 +111,7 @@ export const AutoSyncProvider: FC<PropsWithChildren> = (props) => {
 
           // Hard limit reached is the only limit that can be reached as we set
           // confirmDeleteDocuments to true.
-          if (response && response.type === "HARD") {
+          if (response && response.type === PushChangesLimitType.HARD) {
```
dani-mp: ⚠️ Let's use string literals instead of referencing the constants. For TypeScript there's no difference, and for us it's less code to maintain (you don't have to import and reference and can use just the value directly).

jomifepe: Yeah I agree, I was in the process of changing that because exporting the enum from the manager module was causing issues, so I suppressed the rule in existing lines.

dani-mp: Alright, I guess the real fix is to stop using enums and use unions instead. But we can do this incrementally when we see the opportunity. For legacy enums, I prefer to reference them instead of disabling the rule (just because it's safer), I didn't realize this one was an enum. 

But you can also leave it like this if exporting the enum is causing problems.

dani-mp (approved):

PR #1406, opened 2024-07-23
feat(onboarding): update onboarding step modals to match design

jomifepe:
**Resolves**: DT-2219

### Description

- [X] Make the onboarding modals similar to the design
- [X] Avoid displaying "Undo step" labels right after marking the step as complete, while the dialog is closing
- [x] Update modals with the latest content (defined by product and design)

### Checklist

### How to QA [^1]

[^1]:


BohdanOne (approved):
🎉

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
```diff
@@ -21,26 +22,42 @@ export const OnboardingStepDialog = ({
...
   const { toggleStepComplete, isStepComplete } = useOnboardingContext();
+  const [ctaOkText, updateCtaOkText] = useUpdatableState(() => {
+    return isStepComplete(step.id) ? "Undo step" : "Mark as done";
+  });
 
-  const markAsDone = () => {
+  const execIfOpen = (fn: () => void) => () => {
     if (!isOpen) return;
+    fn();
+  };
```
dani-mp: ⚠️ I think we could make all this code explicit, and remove the `useUpdatableState` and `execIfOpen`. See the comments below.

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
```diff
@@ -21,26 +22,42 @@ export const OnboardingStepDialog = ({
...
   };
 
   return (
-    <Dialog open={isOpen} onOpenChange={onClose}>
-      <DialogHeader title="Learn" />
+    <Dialog
+      open={isOpen}
+      onOpenChange={onClose}
+      onAnimationEnd={execIfOpen(updateCtaOkText)}
+      onAnimationStart={execIfOpen(updateCtaOkText)}
```
dani-mp: ⚠️ These two can be passed the same function without having to rely on the double callback.

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
```diff
@@ -21,26 +22,42 @@ export const OnboardingStepDialog = ({
...
+    >
+      <DialogHeader title="Onboarding" />
       <DialogContent>
         <OnboardingStepDialogContent step={step} />
         <DialogActions
-          ok={{
-            text: isStepComplete(step.id) ? "Undo step" : "Mark as done",
-            onClick: markAsDone,
-          }}
+          ok={{ text: ctaOkText, onClick: execIfOpen(markAsDone) }}
```
dani-mp: ⚠️ This one doesn't make sense to be wrapped into `execIfOpen`, because you can't click the button if the dialog is not open.

packages/slice-machine/src/features/onboarding/content.tsx
```diff
@@ -6,94 +6,118 @@ import type {
...
-  repositoryUrl,
-}: GetOnboardingStepsContentProps): Record<
+export const stepIndexNumeral: Record<number, string> = {
+  0: "one",
+  1: "two",
+  2: "three",
+  3: "four",
+  4: "five",
+  5: "six",
+} satisfies OnboardingStepIndexNumeralMap;
```
dani-mp: ⚠️ Why not a list here? Also, I'd call it step name or something like that. Furthermore, there's one step name per step, so I think it makes sense that they live together within the same data structure `onboardingSteps`.

dani-mp (commented):
Left some comments.

packages/slice-machine/src/features/onboarding/content.tsx
jomifepe: This isn't really supposed to be related to the steps, the type is only there to prevent you from adding a step and forgetting to add a label. Think of it as just a map of numbers to it's numeral word, but I only need 6 for this case, so I can go with a simple map. 

If the numeral label for the index lives next to the step, you have to update it if you change the order of the steps.

Also, I didn't go for a tuple because they are harder to type check.

jomifepe: But this method has to go, unfortunately, `satisfies` doesn't work with next 12 and I can't do any of this type checking 😔

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
jomifepe: Yeah I can simplify it, changed, tell me what you think.

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
jomifepe: You do when the dialog is closing 😅 So this is to solve this issue:
https://github.com/prismicio/slice-machine/pull/1399#discussion_r1680750473

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
jomifepe: Refactored: https://github.com/prismicio/slice-machine/pull/1406#discussion_r1691438981

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
```diff
@@ -21,24 +22,38 @@ export const OnboardingStepDialog = ({
   onClose,
 }: OnboardingStepDialogProps) => {
   const { toggleStepComplete, isStepComplete } = useOnboardingContext();
+  const [ctaOkText, setCtaOkText] = useState(getCtaOkText);
```
dani-mp, line 25: ⚠️ I think you meant to call the fn here?
```suggestion
  const [ctaOkText, setCtaOkText] = useState(getCtaOkText());
```

packages/slice-machine/src/features/onboarding/content.tsx
dani-mp: How is not related to the steps? It's literraly called step index. Maybe I'm missing something, but there are 6 steps, and these are 6 labels. I still think the label (which I like better as name than stepIndexNumeral) should live close to the rest of the step data. It's true that you have to update the label if you move the step, but I don't think that's a big deal. I would even ask design if we can use Step 1, step 2, etc. and use the index + 1 instead.

jomifepe: Since I cannot type check it correctly right now, changed to a simple unchecked array.

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
jomifepe: Actually no, `useState` accepts an initializer function and calls it internally.

packages/slice-machine/src/features/onboarding/content.tsx
jomifepe: What I meant by "not related" was that they are simply a map of number (index) to label. The only reason I added a `satisfies` was to make sure we have enough labels for every step, so that I don't have to create labels up to 20 (arbitrary number) just to be safe.

Anyway, I changed it 😅

packages/slice-machine/src/features/onboarding/OnboardingStepDialog/OnboardingStepDialog.tsx
dani-mp: Ah, right 👍

dani-mp (approved):

packages/slice-machine/src/features/onboarding/content.tsx
```diff
@@ -8,92 +8,106 @@ import type {
...
+  {
+    id: "reviewAndPush",
+    title: "Review and Push Changes",
+    description: "Enable editors to create content.",
   },
   {
     id: "createContent",
-    title: "Create content",
-    description: "Create engaging content.",
+    title: "Create Content in the Page Builder",
```
BohdanOne: I think there were some more recent updates to the copy

jomifepe: Yeah I'm on it

dani-mp (approved):

packages/slice-machine/src/features/onboarding/content.tsx
jomifepe: Final update: Asked benjamin if we could display a number. Changed it to a number.

PR #1409, opened 2024-07-25
Bi/DT-2221 update onboarding progress UI

BohdanOne:
**Resolves**: [DT-2221](https://linear.app/prismic/issue/DT-2221/create-onboarding-progress-listdropdown-component)

### Description

Replaced temporary UI elements in `OnboardingGuide` MVP with elements matching designs. 

### Checklist

### Preview

https://github.com/user-attachments/assets/91ba495e-ac47-47ef-a207-b20eb24836a0

### How to QA [^1]

[^1]:


packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
```diff
@@ -33,37 +35,47 @@ export const OnboardingProgressStepper = () => {
...
           </Button>
         </DropdownMenuTrigger>
-        <DropdownMenuContent align="start">
-          <DropdownMenuLabel>Progress</DropdownMenuLabel>
-          {steps.map((step) => {
-            const isCompleted = isStepComplete(step.id);
+        <DropdownMenuContent align="center" sideOffset={-56} minWidth={256}>
+          <div
+            onMouseLeave={() => {
+              if (!isListOpen) return;
```
dani-mp: 💡 i think you can avoid this as setting the state to the value it is already is a no-op.

dani-mp (approved):
Nice!

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
BohdanOne: thanks, will remove that line!

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
```diff
@@ -33,37 +35,46 @@ export const OnboardingProgressStepper = () => {
...
         </DropdownMenuTrigger>
-        <DropdownMenuContent align="start">
-          <DropdownMenuLabel>Progress</DropdownMenuLabel>
-          {steps.map((step) => {
-            const isCompleted = isStepComplete(step.id);
+        <DropdownMenuContent align="center" sideOffset={-56} minWidth={256}>
+          <div
+            onMouseLeave={() => {
+              setListOpen(false);
+            }}
```
jomifepe:
```suggestion
            onMouseLeave={() => setListOpen(false)}
```

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
```diff
@@ -33,37 +35,46 @@ export const OnboardingProgressStepper = () => {
...
+            onMouseEnter={() => setListOpen(true)}
           >
             {completedStepCount > 0 ? "Continue" : "Start"}
           </Button>
         </DropdownMenuTrigger>
-        <DropdownMenuContent align="start">
-          <DropdownMenuLabel>Progress</DropdownMenuLabel>
-          {steps.map((step) => {
-            const isCompleted = isStepComplete(step.id);
+        <DropdownMenuContent align="center" sideOffset={-56} minWidth={256}>
```
jomifepe: I think the menu can be better aligned with the card. I see that this uses arbitrary numbers to position itself, which doesn't seem like a good way.

jomifepe: I believe that doing it with `asPortal={false}` and CSS would not be pretty, so I think we can maybe position a hidden trigger, so that the menu opens from the correct place.

Something like:

```js
<DropdownMenuTrigger>
  <div className={styles.hiddenTrigger} />
</DropdownMenuTrigger>
```

```css
.hiddenTrigger {
  position: absolute;
  right: 50%;
  bottom: 0;
  left: 50%;
  transform: translateX(-50%);
  visibility: none;
  opacity: 0;
}
```

dani-mp: yeah, i agree, i missed this 👍

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
```diff
@@ -33,37 +35,46 @@ export const OnboardingProgressStepper = () => {
...
+          <div
+            onMouseLeave={() => {
+              setListOpen(false);
+            }}
+          >
+            <DropdownMenuLabel>
+              <Text variant="small" color="grey11">
+                Progress
+              </Text>
```
jomifepe: This label needs some margin/padding to be aligned with the content

Screenshot 2024-07-25 at 16 14 06

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
BohdanOne: Honestly I think it's just fine like that: the dropdown content is positioned against the dropdown trigger. Now, calculating distance to the card (which is outside of this component) would add some complexity that for me is not necessary here. Should the design change, some change in code would be necessary anyways and then updating the offset value is no effort at all.

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
BohdanOne: Yes, I'm aware of that, this problem was present already in storybook of DropdownMenu. 
I showed this to Benjamin and he is fine with present version.

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
jomifepe: We should avoid arbitrary numbers though, `56` doesn't mean anything to me, and can easily be wrong if the button changes size because of the content (unlikely but possible). With a trigger at the edge of the card you can have something that makes more sense as an offset.

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
jomifepe: You can easily apply it to this `Text`, why not make it pixel perfect? To me it looks misaligned 😅 
But if the plan is to fix is in the editor-ui component, I'm okay with it.

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
dani-mp: For me it's ok if we leave it like this. I don't think the complexity is worth it. I would leave a comment there explaining why it's 56 and what to check if it changes in the future.

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
BohdanOne: let me just double-check if I can do that with `Text`, maybe I'm missing something here

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
BohdanOne: @jomifepe I understand that using magic numbers is not a good practice, my bad that I didn't add any comment on that! 
Still I think it is just better to leave it (but with added comment) than adding more complexity (which probably also would need some explanation in the comments)

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
BohdanOne: Ok, fixed :)

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
jomifepe: Still think that it was worth trying out my suggestion. Your comment didn't clarify anything regarding the number 😅 `50` is what? Most likely `36px (button height) + 16px (padding) + 2px (default gap for radix dropdown menu)`.

PR #1412, opened 2024-07-26
feat(onboarding): fade out card and fire 🎉 on complete

jomifepe:
**Resolves**: DT-2237

### Description

Small UX improvements for the onboarding guide component:
- Fire confetti when all steps are complete. `dom-confetti` is the one used in out [blog](https://prismic.io/blog/headless-cms) and it's [pretty small](https://bundlephobia.com/package/dom-confetti@0.2.2)
- Fade out the card component before unmounting it.

### Checklist

### Preview

https://github.com/user-attachments/assets/c99cf716-6df7-459c-9222-db20ec91f20a

### How to QA [^1]

[^1]:


BohdanOne (approved):
🎉 😍 confetti!

packages/slice-machine/src/features/onboarding/OnboardingProvider.tsx
```diff
@@ -93,3 +93,16 @@ export const useOnboardingContext = () => {
...
+  const { isComplete } = useOnboardingContext();
+  const previousIsCompleteRef = useRef<boolean>();
+
+  useEffect(() => {
+    previousIsCompleteRef.current = isComplete;
+  });
+
+  if (previousIsCompleteRef.current !== isComplete && isComplete) {
+    handler();
+  }
```
dani-mp: we shouldn't read a ref in render, and we shouldn't perform a side effect in render

what are we trying to accomplish here? there's probably a more idiomatic way of doing it

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.module.css
```diff
@@ -0,0 +1,23 @@
...
+
+.confettiCannon {
+  position: absolute;
+  bottom: 0;
+  left: 0;
+}
+
+.invisible {
+  opacity: 0;
+  visibility: hidden;
```
dani-mp: Do we need both opacity and visibility?

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.module.css
```diff
@@ -0,0 +1,23 @@
+.container {
+  overflow: hidden;
+  position: relative;
+  opacity: 1;
+  visibility: visible;
```
dani-mp: i guess this is the default?

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.module.css
```diff
@@ -0,0 +1,23 @@
...
+.confettiCannon {
+  position: absolute;
+  bottom: 0;
+  left: 0;
+}
+
+.invisible {
+  opacity: 0;
+  visibility: hidden;
+  transition-delay: 2000ms;
```
dani-mp: confetti duration is 3s and this delay is 2s, not sure if they must go hand in hand

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.tsx
```diff
@@ -0,0 +1,83 @@
...
+  duration: 3000,
+};
+
+const OnboardingGuideContent = () => {
+  const { steps, completedStepCount, isComplete } = useOnboardingContext();
+  const [isVisible, setVisible] = useState(true);
+  const isMediaQueryVisible = useMediaQuery({ min: "medium" });
+  const confettiCannonRef = useRef<HTMLDivElement>(null);
+
+  useOnOnboardingComplete(() => {
```
dani-mp: I would pass a callback to the onboarding context so we can avoid this hook (see the comment about its implementation).

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.tsx
```diff
@@ -0,0 +1,83 @@
...
+    const { current: confettiCannon } = confettiCannonRef;
+    if (confettiCannon) fireConfetti(confettiCannon, confettiConfig);
+    setTimeout(() => setVisible(false), confettiConfig.duration);
+  });
+
+  if (!isVisible || !isMediaQueryVisible) return null;
```
dani-mp: If we make it invisible via code, why do we need the css?

dani-mp (commented):
Left some comments, I'll give it another pass later after the changes.

packages/slice-machine/src/features/onboarding/OnboardingProvider.tsx
jomifepe: > we shouldn't read a ref in render

Why not?

This expression is to make sure if the handler is only called if the value of `isComplete` changes to `true`, and no subsequent calls are made if a re-render happens. I can use a `useEffect`, I just don't like how eslint always yells at you when I'm sure I don't want it to be triggered when the `handler` changes.

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.module.css
jomifepe: Yeah I guess the opacity is sufficient here because the component is unmounted after 😅

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.module.css
jomifepe: Correct

packages/slice-machine/src/features/onboarding/OnboardingProvider.tsx
dani-mp: I think I sent you the docs last week: https://react.dev/reference/react/useRef#caveats

>Do not write or read ref.current during rendering, except for [initialization.](https://react.dev/#avoiding-recreating-the-ref-contents) This makes your component’s behavior unpredictable.

You can do this in a use effect by passing a stable handler, but the idiomatic way is to execute the callback in an event handler. That's why I suggested passing it to the context in the other comment.

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.module.css
jomifepe: 2s for the fade out to start looked like the best timing (as seen in the video), even if the confetti is still being rendered. If I increase this, the fadeout happens too late.

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.tsx
jomifepe: I did that the first time, just thought this looked cleaner, but I will do that, I liked both methods

jomifepe: Changed, tell me what you think

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.tsx
jomifepe: The CSS is only used to the fade out effect. After the transition happens, this boolean will make sure the component is unmounted.

packages/slice-machine/src/features/onboarding/OnboardingProvider.tsx
jomifepe: Yeah I remember that, but it was regarding accessing the current right after the definition. I don't think referencing refs in render time is a bad thing if you're sure you don't want it to change or trigger a re-render. Quite useful I would say.

I changed it though 👀 I agree and prefer to handle these using a callback.

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.tsx
```diff
@@ -0,0 +1,84 @@
...
+  function onComplete() {
+    const { current: confettiCannon } = confettiCannonRef;
+    if (confettiCannon) fireConfetti(confettiCannon, confettiConfig);
+    setTimeout(() => setVisible(false), confettiConfig.duration);
+  }
+
+  if (!isVisible || !isMediaQueryVisible) return null;
+
+  return (
+    <div className={clsx(styles.container, isComplete && styles.invisible)}>
```
dani-mp: Btw, I think this works by chance. I think we should have the container, the visible and the invisible styles, otherwise, the styles in invisible don't necessarily override the ones set in the container.

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.tsx
jomifepe: I thought introducing `framer-motion`, because it's an awesome tool to achieve great UX with little effort. But it's a bit big at 2.19 MB 🫠

packages/slice-machine/src/features/onboarding/OnboardingProvider.tsx
```diff
@@ -91,5 +104,9 @@ export const useOnboardingContext = () => {
     );
   }
 
+  if (onComplete) {
+    context.setOnCompleteCallback(onComplete);
```
dani-mp: There are some problems with this approach. We're performing a side effect on render, and also different useOnboardingContext usages could override the onComplete callback from others.

A better approach is to move this to the provider directly (pass `onComplete` to `OnboardingProvider`), this is easy to do if you move the conffeti implementation to `OnboardingGuide`.

dani-mp (commented):
Almost there!

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.module.css
dani-mp: Ah, i see, do you mind leaving a comment there in the CSS about this? I think it's important for the reader to understand that the value is not random.

jomifepe: Yeah for sure, done!

packages/slice-machine/src/features/onboarding/OnboardingProvider.tsx
jomifepe: Does it matter if the only thing we're doing it assigning a callback to a ref?

dani-mp: yes because we shouldn't read or set refs in render, haha (don't kill the messenger, i don't make the rules 🙈)

dani-mp: but the other problem is also important, imagine having two components setting the callback in every render, overriding one another

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.tsx
jomifepe: I think it would always work because the order of the rules is correct, but I agree with that approach. Done!

packages/slice-machine/src/features/onboarding/OnboardingProvider.tsx
jomifepe: Man, react and it's never ending pitfalls. They say it makes the `"component’s behavior difficult to predict"` but I don't even know how or why 😅

dani-mp: react is nice but it's definitely not perfect

i asked dan abramov in the past what that "unpredictable behavior" means in practice but he never replied

i understand it has to do with the concurrent mode and the ability for the components to pause/cancel rendering and start back again later

i also agree with you that the rules of hooks are annoying and that they shouldn't have been just eslint rules, as they're too important; hopefully, that'll change with the new react compiler (which is also a reason to respect the rules now, so we get an easier path to migrate to it in the future)

i think the best we can do is to learn these rules and caveats together and help each other find ways of implementing things the "react way", even if sometimes we get annoyed by it

in the end, react adds a lot of value and it's a nice way of building UIs

jomifepe: Changed, I avoided passing the callback directly to the provider because I wanted to container everything in the "Content" component, but I extract an intermediate one and I'm happy with it. Tell me what you think.

jomifepe: Yeah I'm used to these rules and caveats, but the ref one I wasn't, to be honest 😄

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.tsx
```diff
@@ -0,0 +1,96 @@
...
+    setTimeout(() => setVisible(false), confettiConfig.duration);
+  };
+
+  if (!isVisible) return null;
+
+  return (
+    <OnboardingProvider onComplete={onComplete}>
+      <OnboardingGuideContent>
+        <div ref={confettiCannonRef} className={styles.confettiCannon} />
+      </OnboardingGuideContent>
```
dani-mp:
```suggestion
<div className={clsx(
        styles.container,
        isComplete ? styles.invisible : styles.visible,
      )} >
      <OnboardingGuideContent />
      <div ref={confettiCannonRef} className={styles.confettiCannon} />
</div>
```

jomifepe: Done

dani-mp (approved):
🎊

PR #1413, opened 2024-07-26
feat(onboarding): change onboarding flag and hide old guide when active

jomifepe:
**Resolves**: DT-2234

### Description

If the new onboarding is not enabled, use the old one.

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/features/inAppGuide/InAppGuideContext.tsx
+++ b/packages/slice-machine/src/features/inAppGuide/InAppGuideContext.tsx
@@ -8,6 +8,7 @@ import {
   useMemo,
 } from "react";
 
+import { useOnboardingExperiment } from "@/features/onboarding/useOnboardingExperiment";
 import { usePersistedState } from "@/hooks/usePersistedState";
 
 type InAppGuideContextValue = {
@@ -21,15 +22,18 @@ export const InAppGuideContext = createContext<
 
 export const InAppGuideProvider: FC<PropsWithChildren> = (props) => {
   const { children } = props;
+  const { eligible: isNewOnboardingEnabled } = useOnboardingExperiment();
   const [isInAppGuideOpen, setIsInAppGuideOpen] = usePersistedState(
     "isInAppGuideOpen",
     true,
   );
 
-  const memoizedValue = useMemo(
-    () => ({ isInAppGuideOpen, setIsInAppGuideOpen }),
-    [isInAppGuideOpen, setIsInAppGuideOpen],
-  );
+  const memoizedValue = useMemo(() => {
+    if (isNewOnboardingEnabled) {
+      return { isInAppGuideOpen: false, setIsInAppGuideOpen: () => undefined };
+    }
+    return { isInAppGuideOpen, setIsInAppGuideOpen };
+  }, [isInAppGuideOpen, setIsInAppGuideOpen, isNewOnboardingEnabled]);
 
   return (
     <InAppGuideContext.Provider value={memoizedValue}>
--- a/packages/slice-machine/src/features/onboarding/useOnboardingExperiment.ts
+++ b/packages/slice-machine/src/features/onboarding/useOnboardingExperiment.ts
@@ -1,6 +1,6 @@
 import { useExperimentVariant } from "@/hooks/useExperimentVariant";
 
 export const useOnboardingExperiment = () => {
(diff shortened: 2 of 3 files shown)


BohdanOne (approved):
🎉

packages/slice-machine/src/features/inAppGuide/InAppGuideContext.tsx
```diff
@@ -21,14 +22,18 @@ export const InAppGuideContext = createContext<
...
   );
 
   const memoizedValue = useMemo(
-    () => ({ isInAppGuideOpen, setIsInAppGuideOpen }),
-    [isInAppGuideOpen, setIsInAppGuideOpen],
+    () => ({
+      isInAppGuideOpen: isNewOnboardingEnabled ? false : isInAppGuideOpen,
+      setIsInAppGuideOpen,
+    }),
+    [isInAppGuideOpen, setIsInAppGuideOpen, isNewOnboardingEnabled],
```
dani-mp: do we need this code? if we don't show the guide this won't be able to be executed, right?

jomifepe: We kinda do, because we're using `useInAppGuide` in [another place](https://github.com/prismicio/slice-machine/blob/main/packages/slice-machine/src/legacy/components/ReviewModal/ReviewModal.tsx#L14), so the provider is above most of the component tree. This way the context still works but shows the value of `isInAppGuideOpen` is always `false` if the new onboarding is enabled.

dani-mp: in that case, i guess what we don't need is the other code in this PR, as the value will always be false, right?

jomifepe: If we disabled the new onboarding flag, we want everything to work as it did before, so we should keep that code, no?

dani-mp: yes, i mean that i believe we don't need to do the check twice

if we do it in the provider, i think that's sufficient and the other part where you check the flag is not needed, because that dialog is driven by this provider anyway (less code to touch and maintain)

jomifepe: But what if a user has an open inAppGuide dialog? If we turn on the new onboarding flag, the `useInAppGuide` will return the value of the `isInAppGuideOpen` storage value, and we want to always have it be `false` in that case.

dani-mp: sorry, i'm confused, don't undertand what you're saying

my suggestion: leave this code, remove the flag check in the component, as it's anyway driven by the provider

jomifepe: Clarified via DM. Done ✅

dani-mp (approved):

packages/slice-machine/src/features/onboarding/useOnboardingExperiment.ts
```diff
@@ -1,6 +1,6 @@
 import { useExperimentVariant } from "@/hooks/useExperimentVariant";
 
 export const useOnboardingExperiment = () => {
-  const variant = useExperimentVariant("slice-machine-onboarding");
+  const variant = useExperimentVariant("slicemachine-onboarding");
```
xrutayisire, line 4: ❓ #question

Why this rename? It was because of the non-consistent naming with `-`?
I still see `slice-machine-onboarding` in Amplitude, should it be removed, deactivated?

jomifepe: Yeah I've been meaning to discuss this. It was a typo, Levi pointed out that all the other slice machine ones didn't have a dash, but since that version was released, I'm not sure it was wise to do so. I was planning on archive it after releasing.

xrutayisire: Oh I see, you can test in stage anyway the thing you want to do, to ensure it doesn't cause a problem while testing the prod slice machine version that target it.

xrutayisire: To test in stage you need `SM_ENV=staging npm run slicemachine`

PR #1419, opened 2024-07-31
fix(adapter-next): improve usability of snippets

angeloashmore:
**Resolves**: A request by @a-trost

### Description

This PR updates `@slicemachine/adapter-next`'s snippets. The updated snippets better reflect how developers render and use the data.

#### Group snippet

The previous snippet included a fragment (`<></>`) and did not use an implicit return (`() => ()`), requiring users to modify the snippet before using it. It was written that way for Prettier compatibility, but a snippet that works immediately is better than one that requires editing by a user.

> **Old**:
> 
> ```tsx
> <>
>   {slice.primary.my_group.map((item) => {
>     // Render the item
>   })}
> </>
> ```
> 
> **New**:
> 
> ```tsx
> {slice.primary.my_group.map((item) => (
>   // Render the item
> ))}
> ```

#### Other snippets

The wrapping `<></>` fragments were removed. They were originally added to support Prettier formatting, but the snippets are simple enough to not need automatic formatting.

> **Old**:
> 
> ```tsx
> <>{slice.primary.my_key_text}</>
> ```
> 
> **New**:
> 
> ```tsx
> {slice.primary.my_key_text}
> ```

### Checklist

### Preview

image

### How to QA [^1]

1. Start a new playground with `yarn dev` + `yarn play --new`.
2. In Slice Machine, add a group to an existing custom type.
3. View the code snippet.

🐤

[^1]:

### DIFF
--- a/packages/adapter-next/src/hooks/snippet-read.ts
+++ b/packages/adapter-next/src/hooks/snippet-read.ts
@@ -85,16 +85,13 @@ export const snippetRead: SnippetReadHook<PluginOptions> = async (
 		}
 
 		case "Group": {
-			const code = await format(
-				stripIndent`
-					<>
-						{${dotPath(fieldPath)}.map((${itemName}) => {
-							// Render the ${itemName}
-						})}
-					</>
-				`,
-				helpers,
-			);
+			// We cannot use `format` since this snippet contains invalid syntax.
+			// Please ensure this snippet is manually formatted correctly.
+			const code = stripIndent`
+				{${dotPath(fieldPath)}.map((${itemName}) => (
+				  // Render the ${itemName}
+				))}
+			`;
 
 			return {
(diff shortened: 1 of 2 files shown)


a-trost:
Thanks so much for changing this! Definitely more immediately usable this way.

dani-mp (approved):

xrutayisire (commented):
It's the same for all the others snippets that include a fragment, right?

It could be improved to prevent having a fragment so the copy-paste works without user having to remove the fragment. Or for you, it's not better for the others snippets?

angeloashmore:
@xrutayisire Good point! We can remove fragments from all snippets that have them using the same reasoning.

xrutayisire (approved):
Nice work!

PR #1421, opened 2024-08-05
[DT-2242] feat: rework uid field

BohdanOne:
**Resolves**: [DT-2242](https://linear.app/prismic/issue/DT-2242/rework-uid-field)

[FIGMA](https://www.figma.com/design/IzGRpY6Gp5wDr7XCTwGl6G/Onboarding?node-id=460-3693&t=FOfLAxaUTkWa25UB-4)

### Description

- for reusable page uid field is now moved to the top of the editor
- for single pages uid is moved to the last position on the dropdown field when adding a field

### Checklist

### Preview

https://github.com/user-attachments/assets/932c9d0a-4882-4411-9354-e726fee3a663

### How to QA [^1]

[^1]:


packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/TabZone/index.tsx
```diff
@@ -74,8 +74,13 @@ const TabZone: FC<TabZoneProps> = ({ tabId }) => {
   const customTypeSM = CustomTypes.toSM(customType);
   const sliceZone = customTypeSM.tabs.find((tab) => tab.key === tabId)
     ?.sliceZone;
-  const fields: TabFields =
+  const allFields: TabFields =
     customTypeSM.tabs.find((tab) => tab.key === tabId)?.value ?? [];
+  // on repeatable pages uid field is movet to the top of the editor
```
dani-mp:
```suggestion
  // on repeatable pages, the uid field is moved to the top of the editor
```

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
```diff
@@ -41,11 +41,20 @@ const Zone = ({
...
+    const widgets = widgetsArray.filter(({ TYPE_NAME }) => TYPE_NAME !== "UID");
+
     // eslint-disable-next-line @typescript-eslint/no-unsafe-return
-    return hasUid
-      ? // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
-        widgetsArray.filter(({ TYPE_NAME }) => TYPE_NAME !== "UID")
-      : widgetsArray;
+    if (hasUid) return widgets;
+
+    // Move UID widget to the end of the array
```
dani-mp: ⚠️ This comment explains what the code does (which in this case it's self-explanatory) but not why we do it, could you add a bit of context here?

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+    });
+
+    setCustomType(newCustomType);
+  };
+
+  return (
+    <Dialog
+      open={isOpen}
+      onOpenChange={setOpen}
+      size={{ width: 448, height: 248 }}
```
dani-mp: ❓ Can't we use one of the default sizes?

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+      trigger={
+        <Button
+          color="grey"
+          textColor="placeholder"
+          startIcon="language"
+          sx={{ marginInline: "auto" }}
+        >
+          :uid
```
dani-mp: ❓ do you know why we use this `:uid` label? I mean lowercase and with the colon.

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+        </Button>
+      }
+    >
+      <DialogHeader title="Update UID label" />
```
dani-mp:
```suggestion
      <DialogHeader title="Update the UID label" />
```

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+              errors.label = "This field is required";
+            }
+            return errors;
+          }}
+          onSubmit={(values) => {
+            setSaving(true);
+
+            saveUIDLabel(values.label);
+
+            setSaving(false);
```
dani-mp: ⚠️ Is this synchronous? if so, do we need to toggle the saving true/false? Otherwise, are we missing an await?

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+                  type="text"
+                  label="Label *"
+                  placeholder="UID"
+                  value={formik.values.label}
+                  onValueChange={(value) =>
+                    void formik.setFieldValue("label", value)
+                  }
+                  error={typeof formik.errors.label === "string"}
+                />
+                <Box alignItems="center" gap={4} padding={{ top: 4 }}>
```
dani-mp: ⚠️ Let's use gap instead of padding top.

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+  FormInput,
+  Icon,
+  Text,
+} from "@prismicio/editor-ui";
+import {
+  Group,
+  NestableWidget,
+  UID,
+} from "@prismicio/types-internal/lib/customtypes";
+import { Formik } from "formik";
```
dani-mp: ❓ Using formik for this little form seems overkill. I don't know if we want to sue formik for everything, or just for complex forms, or not use it at all because it might be a legacy thing. Worth asking @xrutayisire.

dani-mp (commented):
Left some comments!

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
BohdanOne: I've added saving state early assuming it will be an async operation, but if it's not in the end, let me double-check that

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
BohdanOne: that's copy from the design, but I guess your version is the correct one, will consult that with Benjamin

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
BohdanOne: that's how it is in the design, I didn't really think about why 😅

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
BohdanOne: the smallest from the defaults was still much bigger than that, but will ask Benjamin what he thinks about that

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
BohdanOne: Ok, will try to see if I can remove that padding

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
dani-mp: haha, sounds good, let me know if you find out!

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
dani-mp: thanks! i think it's important if we can get it to work with one of the default ones

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
BohdanOne: added reason

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
BohdanOne: anyways, now it will be dynamic, displaying the actual set value of the label

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
BohdanOne: It's working synchronously, removed `isSaving`

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
BohdanOne: done

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
BohdanOne: changing to default "small"

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
BohdanOne: changed

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
xrutayisire: Good point @dani-mp!

We need to define what do we want for the future of our forms. Do we want a form management library or not at all? Depending on that, define if we want to apply the choice for all forms or only "complex" one, but then which one are the complex ones? When are we deciding that a form need a form library?

In the case, it seems Formik may be overkill, I also agree, but research needs to be done for our forms so we can have a clear shared strategy. 
@BohdanOne, for me, it's a necessary study for this ticket to ensure we are going to the right direction with this new form. 
The study also come with form validation library question, Yup is used for legacy reasons, but maybe we would like to use Zod as it's the new standard at Prismic.

------

Following that comment, it's opening another problem, this file is new but in a legacy folder. We should not create fully new components like that in "legacy" folder. We choose this name to make it clear that it was legacy and not the direction to go. Here, we will never finish removing files in legacy if we keep adding new ones.

It's related to the previous topic, if we have this file in our new folder, it pushes us to think more about best practices and tech strategies as it's not in a "legacy" folder where we are less picky about things.

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+          sx={{ marginInline: "auto" }}
+        >
+          {field?.value?.config?.label ?? "UID"}
+        </Button>
+      }
+    >
+      <DialogHeader title="Update the UID label" />
+      <DialogContent>
+        <Formik
+          initialValues={{ label: field?.value?.config?.label ?? "" }}
```
xrutayisire: ⚠️ #issue

Let's not mix anymore `domain` related things with product features. Here we tried to regroup what we are doing manipulating custom types and slices in the `src/domain` folder. This ensures when in Slice Machine, we try to access a label, it will always be the same.

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+  const { customType, setCustomType } = useCustomTypeState();
+  const customTypeSM = CustomTypes.toSM(customType);
+
+  const sectionId =
+    customTypeSM.tabs.find((tab) => {
+      return tab.value.find((f) => f.key === field.key);
+    })?.key ?? "";
+
+  const saveUIDLabel = (label: string) => {
```
xrutayisire: ⚠️ #issue

Here it seems we are also doing `domain` related manipulation, right? This then should be in `src/domain` if it's the case. The goal is to prevent manipulating directly custom types everywhere in our product features. Much more because we are using "legacy" things directly and in non `legacy` folder we don't want any usage of "legacy" things.

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+  field: { key: string; value: TabField };
+}
+
+export const UIDEditor: FC<UIDEditorProps> = ({ field }) => {
+  const [isOpen, setOpen] = useState(false);
+
+  const { customType, setCustomType } = useCustomTypeState();
+  const customTypeSM = CustomTypes.toSM(customType);
+
+  const sectionId =
```
xrutayisire: ⚠️ #issue

Here it seems we are also doing domain related manipulation, right? This then should be in src/domain if it's the case. The goal is to prevent manipulating directly custom types everywhere in our product features. Much more because we are using "legacy" things directly and in non legacy folder we don't want any usage of "legacy" things.

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+  TabFieldsModel,
+} from "@/legacy/lib/models/common/CustomType";
+import { Widgets } from "@/legacy/lib/models/common/widgets";
+import { type Widget } from "@/legacy/lib/models/common/widgets/Widget";
+
+interface UIDEditorProps {
+  field: { key: string; value: TabField };
+}
+
+export const UIDEditor: FC<UIDEditorProps> = ({ field }) => {
```
xrutayisire: 💡#nitpick

We are trying to merge our best practices with editor, you have here some things that don't follow, them, take a look here: https://github.com/prismicio/editor/blob/main/CONTRIBUTING.md#coding-guidelines

Note: we need to improve the discoverability of the best practices, but they are referenced in our Notion doc with a link to the contributing guide: https://www.notion.so/prismic/Slice-Machine-UI-s-Coding-Guidelines-d377c3e33faa44b3a5e0ad84a942616d

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+                  type="text"
+                  label="Label *"
+                  placeholder="UID"
+                  value={formik.values.label}
+                  onValueChange={(value) =>
+                    void formik.setFieldValue("label", value)
+                  }
+                  error={typeof formik.errors.label === "string"}
+                />
+                <Box alignItems="center" gap={4}>
```
xrutayisire: ❓#question

This will be replaced by that: https://github.com/prismicio/editor/pull/1146?

playwright/pages/shared/TypeBuilderPage.ts
```diff
@@ -129,6 +130,9 @@ export class TypeBuilderPage extends BuilderPage {
       name: "Remove slice",
       exact: true,
     });
+    this.uidEditorDialog = page
```
xrutayisire: ⚠️ #issue

We are using the POM pattern in our playwright tests. Here you didn't follow how it's done for other Dialog and everything is directly in `TypeBuilderPage`. It should be separated to respect our current pattern.

playwright/pages/shared/TypeBuilderPage.ts
```diff
@@ -285,6 +293,13 @@ export class TypeBuilderPage extends BuilderPage {
     ).toBeVisible();
   }
 
+  async saveUidLabel(label: string) {
+    await this.getUidEditorButton("UID").click();
+    await expect(this.uidEditorDialog).toBeVisible();
+    await this.page.getByRole("textbox", { name: "Label *" }).fill(label);
```
xrutayisire: ⚠️ #issue

Please follow best practices: https://github.com/prismicio/slice-machine/tree/main/playwright#always-try-to-do-an-exact-matching-with-locators

Here for `Always try to do an exact matching with locators`

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/index.tsx
```diff
@@ -42,10 +43,22 @@ export const CustomTypeBuilder = () => {
...
+  );
+  const uidField = tabWithUid?.value.find((field) => field.key === "uid");
+
   return (
     <>
       <Window sx={sliceZoneEmpty ? { flexGrow: 1 } : undefined}>
-        {customType.format === "page" ? <WindowFrame /> : undefined}
+        {customType.format === "page" ? (
+          <WindowFrame
+            title={
```
xrutayisire: ❓ #question

Here we may give `false` value no?

xrutayisire (changes requested):
Leave some comments that should help you to move forward 💪

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
BohdanOne: most probably yes

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/index.tsx
BohdanOne: If the value is `false` then the title is not displayed at all.
Is it a problem?

playwright/pages/shared/TypeBuilderPage.ts
BohdanOne: done

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+            const errors: Record<string, string> = {};
+            if (!values.label) {
+              errors.label = "This field is required";
+            }
+            return errors;
+          }}
+          onSubmit={(values) => {
+            saveUIDLabel(values.label);
+            setOpen(false);
+          }}
```
jomifepe: 💡 #nitpick

I believe the component would be more readable if these handlers were extracted from the render logic, especially multiline ones. WDYT?

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
BohdanOne: moved that logic to domain/customType

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
BohdanOne: moved that logic to domain/customType

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/UIDEditor/index.tsx
BohdanOne: 100% agree, will move handlers outside the template

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/index.tsx
xrutayisire: I'm not a fan of passing `false` as a prop value just because ReactNode accept everything. I always like to be more clear and pass undefined. No strong opinion here.

packages/slice-machine/src/domain/customType.ts
```diff
@@ -129,6 +129,28 @@ export function getSectionSliceZoneConfig(
   return maybeSliceZone?.config ?? undefined;
 }
 
+// TODO add tests
+// We block user from deleteing section containing UID field,
+// so we can expect uid to be in the main section
+export function getUidField(customType: CustomType): UID | undefined {
```
xrutayisire: 💡 #idea

```suggestion
export function getMainSectionUidField(customType: CustomType): UID | undefined {
```

packages/slice-machine/src/domain/customType.ts
```diff
@@ -129,6 +129,28 @@ export function getSectionSliceZoneConfig(
...
+  const mainSectionEntry = getMainSectionEntry(customType);
+  const [_, mainSection] = mainSectionEntry ?? [];
+
+  return Object.values(mainSection ?? {}).find(
+    (field): field is UID => field.type === "UID",
+  );
+}
+
+// TODO add tests
+export function findSectionWithUidField(customType: CustomType): string | null {
```
xrutayisire: ❓#question

I'm lost here, you're saying that we expect UID to be in the main section but here you have another function to find the section with the UID field. One of the two think is strange then. Should we not use that management in the previous one?

Also maybe if we sure it's always the main section, we don't care to look for the UID field in other section. Can it be other sections?

packages/slice-machine/src/domain/customType.ts
```diff
@@ -129,6 +129,28 @@ export function getSectionSliceZoneConfig(
...
+  const sectionEntries = getSectionEntries(customType);
+  const sectionWithUid = sectionEntries.find(([_, section]) =>
```
xrutayisire: ❓#question

Can we have more than one section with UID? You are checking all sections but actually only get the first one anyway after with [0].

packages/slice-machine/src/domain/customType.ts
```diff
@@ -425,6 +447,41 @@ export function reorderField(args: ReorderFieldArgs): CustomType {
...
+// TODO add tests
+export function updateUidField(
+  label: string,
+  customType: CustomType,
+): CustomType {
+  const field = getUidField(customType);
+  if (!field) return customType;
+
+  const section = findSectionWithUidField(customType);
+  const sectionId = section ?? "Main";
```
xrutayisire: ⚠️ #issue

2 things for me.

1 - I don't understand the management, the function is name `updateUidField`, and if you don't find a section with a field in it, you still want to update it, but then it's creating it. Is that what we want? If yes maybe the name should be clearer.

2 - We cannot rely on "Main" directly for me here. I would say that you want to create also and that's why you added the "Main" If you want to "add" you need to get the main section id, because it can NOT be "Main", it's technically possible. Then with this code you would create on the fly another section for the user.

packages/slice-machine/src/domain/customType.ts
```diff
@@ -425,6 +447,41 @@ export function reorderField(args: ReorderFieldArgs): CustomType {
...
+    ...field,
+    config: {
+      ...field.config,
+      label,
+    },
+  };
+
+  const updatedSection = updateFields({
+    fields: customType.json[sectionId],
+    previousFieldId: "uid",
```
xrutayisire: ❓#question

But here then I don't understand, we already saw from your previous code that there was no UID existing, how can we have a previous uid field in that case?

packages/slice-machine/src/domain/customType.ts
```diff
@@ -425,6 +447,41 @@ export function reorderField(args: ReorderFieldArgs): CustomType {
...
+    newFieldId: "uid",
```
xrutayisire: ⚠️ #issue

Today technically you CAN have a UID that is not named uid. If you create a Page Type then open the JSON file and manually change the API ID, you will see that it's possible. 
You can see that prismic-types-internal that control what can be pushed or not never enforce that it should be name UID: https://github.com/prismicio/prismic-types-internal/pull/27

The goal would be use the field id instead of hardcoding it. Since you're finding it with the type, you should have it after.

packages/slice-machine/src/domain/customType.ts
xrutayisire: 💡#nitpick

I see that sometimes you are doing `Uid` or `UID` for example for `UIDEditor`.
I would be consistent and do `getMainSectionUIDField `.

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { type KeyboardEvent, useState } from "react";
+
+import { getUidField, updateUidField } from "@/domain/customType";
+import { useCustomTypeState } from "@/features/customTypes/customTypesBuilder/CustomTypeProvider";
+
+export function UIDEditor() {
+  const [isOpen, setOpen] = useState(false);
+  const { customType, setCustomType } = useCustomTypeState();
+  const field = getUidField(customType);
+  const [label, setLabel] = useState(field?.config?.label ?? "");
```
xrutayisire: ⚠️ #issue

You still are getting the label direction without using domain function here

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/index.tsx
```diff
@@ -42,10 +43,17 @@ export const CustomTypeBuilder = () => {
     customTypeSM.tabs.find((tab) => tab.key === tabValue)?.sliceZone?.value
       .length === 0;
 
+  // block removing tab with UID field
```
xrutayisire: 💡 #idea

I was lost to understand the comment, was reading block as a block of text or something like that 😅
Maybe we could move that line directly above where this is done, here you're not blocking the tab with UID field, this code is only to get the `tabWithUID`.

Also improving naming to ensure nobody face the same problem as me haha.

```
                   <DropdownMenuItem
                        color="tomato"
                        // Blocking the ability to remove a tab if it's the last one or it contains an UID.
                        disabled={
                          customTypeSM.tabs.length <= 1 ||
                          tabWithUid?.key === tab.key
                        }
```

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/index.tsx
```diff
@@ -73,7 +81,10 @@ export const CustomTypeBuilder = () => {
                       </DropdownMenuItem>
                       <DropdownMenuItem
                         color="tomato"
-                        disabled={customTypeSM.tabs.length <= 1}
+                        disabled={
+                          customTypeSM.tabs.length <= 1 ||
+                          tabWithUid?.key === tab.key
```
xrutayisire: ❓ #question

I'm not sure of the product impact of this. Are you sure you want to block also all custom type tabs that contains a UID? It seems bad to do that. Is it not only for page type that you want to do that?

xrutayisire (changes requested):

packages/slice-machine/src/domain/customType.ts
BohdanOne: yes, you're right, since the Main section has an uid field, and we will prevent user form deleting that section, we shouldn't need to look for uid in other ones

packages/slice-machine/src/domain/customType.ts
BohdanOne: need to re-think that, but clearly it needs some fixes, basicly the uid is already attached to main section so we shouldn't need to create the field or the section

packages/slice-machine/src/domain/customType.ts
BohdanOne: didn't know that, will fix it

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/index.tsx
BohdanOne: didn't think about that, will narrow down the condition to prevent deleting section with uid only on reusable pages

packages/slice-machine/src/domain/customType.ts
xrutayisire: As discussed in private, old version of SM may not have the UID in the main section

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/index.tsx
xrutayisire: It's maybe a product question if you're not sure 🙂

BohdanOne: that was already discussed, the decision was : prevent main tab from being deleted, and it was said in the context of the main tab containing uid be default: https://prismic-team.slack.com/archives/C02L3FN3AJK/p1722852039737899

packages/slice-machine/src/domain/customType.ts
BohdanOne: changed the logic to find the UID wherever it is, 
also after chat with Benjamin, added option to add uid via the new button if for any reason it's missing in all sections

packages/slice-machine/src/domain/customType.ts
BohdanOne: done

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
BohdanOne: added domain function to get the label

playwright/pages/shared/TypeBuilderPage.ts
BohdanOne: done

playwright/tests/pageTypes/pageTypeBuilderFields.spec.ts
```diff
@@ -491,24 +491,31 @@ test("I can see and copy the code snippets for nested groups and their sub field
...
+  await pageTypesBuilderPage.goto(singlePageType.name);
 
   await expect(
-    pageTypesBuilderPage.getFieldMenuButton("uid"),
+    pageTypesBuilderPage.getListItemFieldId("uid"),
   ).not.toBeVisible();
 });
 
-test("I cannot see default UID field for single page type", async ({
+test("I can edit the UID field", async ({
```
xrutayisire: 💡 #idea

Should we add a `I cannot edit the UID field for single page type`? To prevent having a regression where the UID editor button would be visible.

xrutayisire: 💡 #idea

```suggestion
test("I can edit the UID field for repeatable page type", async ({
```

packages/slice-machine/src/domain/customType.ts
```diff
@@ -90,13 +90,41 @@ export function getMainSectionEntry(
   return sections[0];
 }
 
+export function getMainSectionId(customType: CustomType): string | undefined {
```
xrutayisire: 💡 #idea

We have some code in this file that don't use that management:

```
const mainSectionEntry = getMainSectionEntry(customType);
  if (mainSectionEntry) {
    const [mainSectionKey] = mainSectionEntry;
```

I would say that we need to stay consistent, like either using this new function where we get only the key, or just do: `const [mainSectionKey] = mainSectionEntry;` Is it not enough?

---

Note: if we have to use `getMainSectionId` I would keep using `getMainSectionEntry`, just to ensure we always have one central way to get the main section, and then you can just return the key if it exists.


WDYT?

packages/slice-machine/src/domain/customType.ts
```diff
@@ -425,6 +453,64 @@ export function reorderField(args: ReorderFieldArgs): CustomType {
...
+  const sectionId = getMainSectionId(customType);
+  if (typeof sectionId === "undefined") {
+    return customType;
+  }
+  const newFieldId = "uid";
+  const newField: UID = {
+    config: {
+      label,
+    },
+    type: "UID",
```
xrutayisire: 💡 #idea

I would move the `type` above to keep current management with the type is first.

packages/slice-machine/src/domain/customType.ts
```diff
@@ -425,6 +453,64 @@ export function reorderField(args: ReorderFieldArgs): CustomType {
...
 
+// if the UID is not existing in any section, it should be added to the main section
+export function addUIDField(label: string, customType: CustomType): CustomType {
+  const sectionId = getMainSectionId(customType);
+  if (typeof sectionId === "undefined") {
+    return customType;
+  }
+  const newFieldId = "uid";
+  const newField: UID = {
+    config: {
```
xrutayisire, line 464: ⚠️ #issue

Potential issue, did you discuss the placeholder? I don't see it here at all, but how would a dev configure the placeholder visible for the editor?

packages/slice-machine/src/domain/customType.ts
```diff
@@ -425,6 +453,64 @@ export function reorderField(args: ReorderFieldArgs): CustomType {
...
+  return addField({ customType, sectionId, newField, newFieldId });
+}
+
+export function updateUIDField(
+  label: string,
+  customType: CustomType,
+): CustomType {
+  const [sectionId, section] = getSectionWithUIDField(customType) ?? [];
+
+  const fieldId = Object.keys(section ?? {})?.find(
```
xrutayisire: 💡 #idea

I don't understand why you don't get the field ID from the next line as you have the field. It's really strange because you have a manual code here for the `fieldId` but for the `fieldValue` you have a function, it's not consistent.

It's coming back to the reason why we have for example `getMainSectionEntry`, by getting the entry you always ensure you get either the ID, the value or both.

What about having the same here?

packages/slice-machine/src/domain/customType.ts
```diff
@@ -425,6 +453,64 @@ export function reorderField(args: ReorderFieldArgs): CustomType {
...
+    (key) => section?.[key]?.type === "UID",
+  );
+
+  const field = getUIDField(customType);
+
+  if (
+    !field ||
+    typeof fieldId === "undefined" ||
```
xrutayisire: 💡 #idea

Why `typeof fieldId === "undefined"` and not as for field `!field`?

packages/slice-machine/src/domain/customType.ts
```diff
@@ -90,13 +90,41 @@ export function getMainSectionEntry(
...
+export function getUIDField(customType: CustomType): UID | undefined {
+  const sectionEntry = getSectionWithUIDField(customType);
+  const [_, section] = sectionEntry ?? [];
+
+  return Object.values(section ?? {}).find(
+    (field): field is UID => field.type === "UID",
+  );
+}
+
+export function getUIDFieldLabel(customType: CustomType): string | undefined {
```
xrutayisire: 💡 #idea

Let's be more generic here, we just want a way to get the label from a field as it's always the same way to get it.

```suggestion
export function getFieldLabel(field: NestableWidget | UID | Group): string | undefined {
```

packages/slice-machine/src/domain/customType.ts
```diff
@@ -90,13 +90,41 @@ export function getMainSectionEntry(
...
+}
+
 export function getSection(
   customType: CustomType,
   sectionId: string,
 ): DynamicSection | undefined {
   return customType.json[sectionId];
 }
 
+export function getSectionWithUIDField(
```
xrutayisire: 💡 #idea

You can see in the file that there is a pattern to name with entry when it's an array return so it's clear that it's not direction the section value itself

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
```diff
@@ -0,0 +1,83 @@
...
+          color="grey"
+          textColor="placeholder"
+          startIcon={field ? "language" : "add"}
+          sx={{ marginInline: "auto" }}
+        >
+          {field ? uidFieldLabel : "Add an UID"}
+        </Button>
+      }
+    >
+      <DialogHeader title="Edit the UID label" />
```
xrutayisire, line 74: 💡 #idea

Should we also change the title as you change the button text like:

```suggestion
      <DialogHeader title="Add an UID" />
```

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
```diff
@@ -0,0 +1,83 @@
...
+          <Box flexDirection="column" padding={16} gap={4}>
+            <FormInput
+              type="text"
+              label="Label *"
+              placeholder="UID"
+              value={label}
+              onValueChange={setLabel}
+              error={!label && "This field is required"}
+            />
+            {label && <Text color="grey11">A label for the UID</Text>}
```
xrutayisire: ⚠️ #issue

TODO just to remember to use latest editor package when ready 🙂

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
```diff
@@ -0,0 +1,83 @@
...
+          </Box>
+          <DialogActions>
+            <DialogCancelButton size="medium" />
```
xrutayisire, line 92: ⚠️ #issue

Potential issue, when you click on "Cancel" or just close the Modal with the icon or ESC it will keep the value inside the field, looks strange as you want to cancel your action and I would have expected to have back the original value when opening the modal again no? Is this desired?

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
```diff
@@ -0,0 +1,83 @@
...
+      <DialogContent>
+        <Form onSubmit={handleSubmit}>
+          <Box flexDirection="column" padding={16} gap={4}>
+            <FormInput
+              type="text"
+              label="Label *"
+              placeholder="UID"
+              value={label}
+              onValueChange={setLabel}
+              error={!label && "This field is required"}
```
xrutayisire: ⚠️ #issue

You didn't add the error management that is today on the UID field for max length.

xrutayisire (changes requested):

playwright/tests/pageTypes/pageTypeBuilderFields.spec.ts
BohdanOne: Was considering that, will add!

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
BohdanOne: I discussed that with Benjamin, and the final decision was 'Edit the UID label" for both use cases (as this title kind of cover both cases)

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
BohdanOne: You mean that this text would be passed directly to FormInput with new package, right?
I'm not 100% sure if this change will be applied in the FormInput, as far as I know there is still some discussion about that. 
But I can add comment to update the code if the change is made in editor-ui.

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
BohdanOne: Wasn't 100% sure what would be the best option. 
But you're right, reseting the field on exiting modal is more expected behaviour. 
Will change that.

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
BohdanOne: right, missed that, will add!

packages/slice-machine/src/domain/customType.ts
BohdanOne: yes, you're right that function is not needed `getMainSectionEntry` can be used instead. Removing it.

packages/slice-machine/src/domain/customType.ts
BohdanOne: I'm not sure if I understand 😅

packages/slice-machine/src/domain/customType.ts
BohdanOne: That's good observation, current design doesn't provide any option to configure placeholder - will ask Benjamin!

packages/slice-machine/src/domain/customType.ts
BohdanOne: yes, that's good idea, will change that

packages/slice-machine/src/domain/customType.ts
BohdanOne: typescript didn't like the other option: `Unexpected nullable string value in conditional`

packages/slice-machine/src/domain/customType.ts
BohdanOne: good idea, will change that

packages/slice-machine/src/domain/customType.ts
BohdanOne: true, will rename it

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
xrutayisire: Let's ask José, I think something is still in progress.

packages/slice-machine/src/domain/customType.ts
xrutayisire: Today in the json, "type" is the first property on top of config, I would keep it like this. No?

BohdanOne: ah ok, then I understood correctly, will change that in `DEFAULT_MAIN_WITH_UID` as well then

packages/slice-machine/src/domain/customType.ts
BohdanOne: talked with Benjamin, he is ok with not adding placeholder here
also talk to Come, he is also ok with not having option to configure placeholder, but asked to default it to ""Enter a unique identifier"

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
BohdanOne: yeah, it is still not unclear, will leave comment like that:
`{/* TODO: refactor if change proposed in: https://github.com/prismicio/editor/pull/1151 is released */}`

packages/slice-machine/src/domain/fields.ts
```diff
@@ -196,6 +197,18 @@ export const UIDField: UIDField = {
...
+export const UIDFieldCustomErrorMap: z.ZodErrorMap = (issue, ctx) => {
+  if (issue.code === z.ZodIssueCode.too_big) {
+    return { message: `String is too long. Max: ${issue.maximum}` };
+  }
+  if (issue.code === z.ZodIssueCode.too_small) {
+    return { message: `This field is required` };
+  }
+  return { message: ctx.defaultError };
+};
+
```
BohdanOne: I wasn't sure where to put this piece of config, this place felt like the best but I'm happy to move it elsewhere

packages/slice-machine/src/domain/customType.ts
xrutayisire: Screenshot 2024-08-08 at 12 16 32

Hum I don't have that 🧐

playwright/tests/pageTypes/pageTypeBuilderFields.spec.ts
xrutayisire: Did you add it? I don't see the check to verify that the UID edit button is not here on a single page type?
I only see the check that the field is not visible in the static fields.

BohdanOne: Gosh, I misunderstood your comment.
Will add that
anyway, will need to push fixes to e2e tests - two are failing on github :( which is strange because locally they work, need to investigate that

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
```diff
@@ -0,0 +1,108 @@
...
+            <FormInput
+              type="text"
+              label="Label *"
+              placeholder="UID"
+              value={label}
+              onValueChange={handleValueChange}
+              error={error}
+            />
+            {/* TODO: refactor if change proposed in: https://github.com/prismicio/editor/pull/1151 is released */}
+            {typeof error === "undefined" && (
```
xrutayisire: 💡 #idea

I think you have a problem with your config, we never did that kind of things to check for undefined or not:

Screenshot 2024-08-08 at 12 35 34

xrutayisire (changes requested):

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
BohdanOne: ok, will need to check that, you think I should ignore those warnings for now?

xrutayisire: I think you should make sure VSCode is not telling you something different than us, could be hard for you to work :/

playwright/pages/components/UIDEditor.ts
```diff
@@ -29,11 +29,13 @@ export class UIDEditor {
   /**
    * Dynamic locators
    */
-
   getDialogTrigger(label: string): Locator {
     return this.page.getByRole("button", { name: label, exact: true });
   }
 
+  getErrorMessage(message: string): Locator {
+    return this.dialog.getByText(message);
```
xrutayisire: 💡 #idea

`exact` could be used here, right?

In best practice, we anyway said:
> Use `false` when you deliberately don't want an exact matching.

So it clearly shows if you thought about the best practice or not.

xrutayisire (changes requested):

playwright/pages/components/UIDEditor.ts
BohdanOne: fixed

packages/slice-machine/src/domain/fields.ts
```diff
@@ -196,6 +197,18 @@ export const UIDField: UIDField = {
...
+export const UIDFieldCustomErrorMap: z.ZodErrorMap = (issue, ctx) => {
+  if (issue.code === z.ZodIssueCode.too_big) {
+    return { message: `String is too long. Max: ${issue.maximum}` };
+  }
+  if (issue.code === z.ZodIssueCode.too_small) {
+    return { message: "This field is required" };
+  }
+  return { message: ctx.defaultError };
+};
+
```
dani-mp: I'd move this to the form module, as a private function - i would also:
- encapsulate the schema parsing in a function so you don't have to call parse and pass the map
- use switch for the error map implementation
- provide a default error message that is controlled by us, because we don't know what zod spits out in the default case
- use more consistent punctuation in the sentences, for example, the first can be "The value can't be longer than x characters`

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const uidFieldLabel = field ? getFieldLabel(field) : "";
+  const [label, setLabel] = useState(uidFieldLabel ?? "");
+  const [error, setError] = useState<string | undefined>();
+
+  function handleOpenChange(open: boolean) {
+    setOpen(open);
+    if (!open) {
+      setLabel(uidFieldLabel ?? "");
+      setError(undefined);
+    }
```
dani-mp: `handleOpenChange` is not called when you pass `false` to the `open` property of the dialog. That means that in those cases you won't be reseting the form.

To fix this, in concurrent React is ok to do something like this in the render function:
```
if (!open) { setLabel(...), etc }
```
when a set state happens in the render, react interrupts the rendering and starts over with the new values

this ensures that no matter who changes the open prop, you'll have an opportunity to reset it.

playwright/utils/generateRandomStringOfLength.ts
```diff
@@ -0,0 +1,7 @@
+export function generateRandomStringOfLength(length: number): string {
```
dani-mp:
```suggestion
export function getRandomString({ length }: { length: number }): string {
```

dani-mp (commented):
left 3 small comments

playwright/utils/generateRandomStringOfLength.ts
BohdanOne: Good idea with shortening the name, it's still clear what is it for.
But I don't understand what is the benefit of wrapping length in an object as a parameter?

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
BohdanOne: nice catch, will add that

playwright/utils/generateRandomStringOfLength.ts
dani-mp: not important, but two benefits:
- the caller side will be `getRandomString({ length: 32 })` instead of `getRandomString(32)`.
- if tomorow you need to add another param, you add it to the object instead of as a positional argument

BohdanOne: Ok, make a lot of sense!

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
BohdanOne: actually, as mentioned in private message, I'll need a bit more help with that

jomifepe (approved):
Nothing to point out, LGTM 🎉🚀

xrutayisire (approved):
Didn't QA back from your last changes, trusting you on that 🙂

LGTM!

PR #1424, opened 2024-08-07
[DT-2245] feat: display static and slice zone right away

BohdanOne:
**Resolves**: [DT-2245](https://linear.app/prismic/issue/DT-2245/replace-the-initial-empty-page-type-screen)

### Description

Simplify the onboarding by removing the difference in page type builder page that occurs on adding new page.
This is not completed feature! In next steps placeholder for empty Static Zone will be removed.

### Checklist

### Preview
**BEFORE**
new-page-before
**AFTER**
new-page-after

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/legacy/components/Forms/CreateCustomTypeModal/CreateCustomTypeModal.tsx
+++ b/packages/slice-machine/src/legacy/components/Forms/CreateCustomTypeModal/CreateCustomTypeModal.tsx
@@ -81,12 +81,6 @@ export const CreateCustomTypeModal: React.FC<CreateCustomTypeModalProps> = ({
 
         await router.push({
           pathname: customTypesConfig.getBuilderPagePathname(id),
-          query:
-            newCustomType.format === "page"
-              ? {
-                  newPageType: true,
-                }
-              : undefined,
         });
 
         syncChanges();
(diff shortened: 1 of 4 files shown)


levimykel (dismissed):
Looks good to me 😄

levimykel (commented):
In `SliceZone`, is `redirectToEditMode` still needed? And thus, is `useRouter` still needed?

BohdanOne:
> In `SliceZone`, is `redirectToEditMode` still needed? And thus, is `useRouter` still needed?

not sure, was focused only on UI changes, but it's possible you're right, will double-check that!

xrutayisire (changes requested):
⚠️ #issue

I don't understand this implementation here. Are we not supposed to remove the placeholder to push devs to add slices and not see the huge placeholder to create fields?

Screenshot 2024-08-08 at 09 34 34

BohdanOne:
> ⚠️ #issue
> 
> I don't understand this implementation here. Are we not supposed to remove the placeholder to push devs to add slices and not see the huge placeholder to create fields?
> 
@xrutayisire 
Yes, that is next step, in order to do that we need to move 'add fields' button. 
So this is just the first step, more to follow today (my bad, I could state that in pr description).

xrutayisire (commented):
Ok nice 👌 No problem I didn't know, that's also the goal of PR 💪

xrutayisire:
I let Levi approve as I didn't check the code on my side 🙂

levimykel (approved):
Looks good to me, nice work 😄

PR #1425, opened 2024-08-08
[DT-2249] Create new fields using the settings dialog directly

jomifepe:
**Resolves**: DT-2249

### Description

We got some feedback from our users stating that the process of creating a new field is annoying. Mainly due to the fact that every time the users wants to create one, they have to choose it, specify a `label` and `id`, click on `Add`, and open the edit modal to fill the rest of the fields, even though the fields for `label` and `id` are already in the edit modal. These extra clicks add up if the user adds many fields. This middle step of filling just the label and id before being able to fill the rest could be removed and the modal could be used to specify everything right away.

These changes preserve the experience we currently have with the new field form:
- Generate a slug of the label and pre-fill the `id` while typing (if the id was not manually set)
- Automatically focus the `label` input when adding a field

#### Tasks
- [X] Path: Add new field to a page type
- [X] Path: Add new field to a custom type
- [X] Path: Add new field to a slice
- [X] Adjust E2E tests
- [x] Remove Group placeholder field. This field was not supposed to be in the Group edit form, as this property is not defined in the [Group widget config](https://github.com/prismicio/prismic-types-internal/blob/master/src/customtypes/widgets/Group.ts#L11), and it's currently not being saved. This field was also not present in the currently used `NewField` component (inline form to create a new field), so having it in the edit modal to create a new group will probably confuse the user.

#### Legacy concerns

These code changes mostly touch on legacy code, so the following things were considered:
- _Minor_ changes to the existing legacy files. Only the necessary for the basic functions (create & edit) and have auto focus
- `SlugifyLabelObserver` is self-managed, doesn't have any dependency to the existing legacy code, and only needs to tap into the formik context

#### ❌ Old flow

https://github.com/user-attachments/assets/d4be6698-eb13-465a-9677-8f1e83862a22

#### ✅ New flow

https://github.com/user-attachments/assets/7b04e2b3-ae88-4bef-8735-79fca21cb8ae

### How to QA [^1]

```bash
npx create-next-app@latest add-field-modal && cd add-field-modal && npx @slicemachine/init@2.10.4-alpha.DT-2249-add-field-open-modal.15
```

[^1]:


packages/slice-machine/src/legacy/components/FormFields/Input.tsx
```diff
@@ -62,6 +62,7 @@ interface FormFieldInputProps {
   initialValues?: Record<string, string>;
   isDisabled?: boolean;
   variant?: string;
+  autoFocus?: boolean;
```
levimykel, line 65: ⚠️ It seems that this isn't needed. It looks like it isn't used anywhere.

packages/slice-machine/src/legacy/lib/forms/defaults.ts
```diff
@@ -29,6 +29,7 @@ export const DefaultFields: Record<string, InputType> = {
     undefined,
     undefined,
     "Label for content creators (defaults to field type)",
+    true,
```
levimykel, line 33: ❓ Do we want this to always be true, to always focus on the Label field when opened? We definitely do when the field is created, but this causes it to focus on this field even when editing after creation. This wasn't the behavior before and I personally think it shouldn't autofocus after creation.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
```diff
@@ -7,7 +7,10 @@ import {
   DialogContent,
   DialogHeader,
 } from "@prismicio/editor-ui";
-import { GroupFieldType } from "@prismicio/types-internal/lib/customtypes";
+import {
+  GroupFieldType,
+  SlicePrimaryWidget,
```
levimykel: ⚠️ Since you're only using this as a type, import as such: `type SlicePrimaryWidget`.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
```diff
@@ -93,79 +102,73 @@ const FieldZones: FC = () => {
...
+    widgetArea: WidgetsArea,
+    { apiId: previousKey, newKey, value }: OnSaveFieldProps,
+  ) => {
+    const newSlice = updateField({
+      slice,
+      variationId: variation.id,
+      widgetArea,
+      previousFieldId: previousKey,
+      newFieldId: newKey,
+      newField: value as SlicePrimaryWidget,
```
levimykel, line 115: 💡 Instead of forcing the type like this would it be better to update `updateField` to allow `newField` to be `SlicePrimaryWidget` or `SlicePrimaryFieldSM`?

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
```diff
@@ -93,79 +102,73 @@ const FieldZones: FC = () => {
...
-      );
-      if (!widget) {
-        throw new Error(`Unsupported Field Type: ${widgetTypeName}`);
-      }
+  const _onSaveNewField = (
+    widgetArea: WidgetsArea,
+    { apiId: id, value: newField }: OnSaveFieldProps,
+  ) => {
+    const { type: widgetTypeName, config } = newField;
+    const label: string = config?.label ?? "";
```
levimykel: 💡 I don't think that you need to explicitly type this.

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
```diff
@@ -82,13 +80,16 @@ const Zone = ({
     setEditModalData({ isOpen: false });
   };
 
+  /** @param {keyof typeof Widgets} widgetTypeName */
```
levimykel, line 83: 🎉 Nice addition

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
```diff
@@ -82,13 +80,16 @@ const Zone = ({
...
   const onSelectFieldType = (widgetTypeName) => {
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-assignment
-    setNewFieldData({ widgetTypeName, fields });
+    /** `widgetTypeName` might have less keys than `Widgets`, but we lost track 
+    of the types because the `widgetsArray` is not typed and is also filtered into 
+    `widgetsArrayWithCondUid`. Although it's safe to use it to index the `Widgets`
```
levimykel: Minor grammar & spelling fixes in italics:
"_Although,_ it's safe to use it to index the `Widgets` as _long_ as..."

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
```diff
@@ -198,22 +198,6 @@ const Zone = ({
               testId={testId}
               // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
               isRepeatableCustomType={isRepeatableCustomType}
-              newField={
```
levimykel, line 201: ❓ Does this prop need to be removed from `Card` now?

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
```diff
@@ -40,26 +39,23 @@ export const CustomListItem = ({
...
   const onSelectFieldType = (widgetTypeName) => {
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-    setNewFieldData({ widgetTypeName });
-  };
-
-  const onCancelNewField = () => {
-    setNewFieldData(null);
+    /** `widgetTypeName` might have less keys than `Widgets`, but we lost track 
+    of the types because the `widgetsArray` is not typed and is also filtered into 
+    `widgetsArrayWithCondUid`. Although it's safe to use it to index the `Widgets`
```
levimykel: Minor grammar & spelling fixes in italics:
"_Although,_ it's safe to use it to index the `Widgets` as _long_ as..."

levimykel (changes requested):
I left a few comments & questions for you. Nice work, this is a great change.

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
jomifepe: Corrected ✅

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
jomifepe: Corrected ✅

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
jomifepe: Correct, fixed ✅

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
jomifepe: Added ✅

packages/slice-machine/src/legacy/lib/forms/defaults.ts
jomifepe: Correct, I initially though I needed to pass it as a separate prop. Fixed ✅

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
jomifepe: Sounds safe to me, they are the same type.

jomifepe: But I would be accepting a legacy type in a new domain util

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
jomifepe: Not needed, removed ✅

packages/slice-machine/src/legacy/components/FormFields/Input.tsx
jomifepe: ~Correct, I initially though I needed to pass it as a separate prop. Fixed ✅~

Now it is, because of https://github.com/prismicio/slice-machine/pull/1425#discussion_r1726842687 😅

packages/slice-machine/src/legacy/lib/forms/defaults.ts
jomifepe: Makes sense, adjusted to only autofocus when creating a new field.

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
levimykel: Missed one last grammar fix:
"Although,"

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
levimykel: Missed one last grammar fix:
"Although,"

levimykel (approved):
Approved with two quick grammar fixes that were missed. Great job!

packages/slice-machine/src/legacy/components/FormFields/Input.tsx
levimykel: Okay

packages/slice-machine/src/legacy/lib/forms/defaults.ts
levimykel: Looks good 👍

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
levimykel: That's a good point, your original solution might be better. I'll leave this to you.

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
jomifepe: Fixed ✅

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
jomifepe: Fixed ✅

PR #1426, opened 2024-08-09
feat(onboarding): simplify and onboard to static and slice zone

BohdanOne:
**Resolves**: [DT-2246](https://linear.app/prismic/issue/DT-2246/adapt-static-and-slice-zone-ui)
**Resolves**: [DT-2247](https://linear.app/prismic/issue/DT-2247/display-about-static-fields-dialog-when-adding-a-slice-to-a-page)
### Description

- Adjustments to the Static Zone and Slice Zone UI: Name changes, buttons look. 
- Added `StaticFieldsInfoDialog` which is displayed on first click of button adding static field.

### Checklist

- [X] Final modal content

### Preview

Screenshot 2024-08-13 at 10 17 33

Screenshot 2024-08-13 at 10 17 54

### How to QA [^1]

[^1]:


packages/slice-machine/src/features/builder/AddStaticFieldButton.tsx
```diff
@@ -0,0 +1,92 @@
...
+  );
+
+  return (
+    <StaticFieldsInfoContext.Provider
+      value={{ isInfoDialogSeen, setInfoDialogSeen }}
+    >
+      {children}
+    </StaticFieldsInfoContext.Provider>
+  );
+}
```
jomifepe: 💡 

I don't really see a reason to use context here, do we plan on opening the modal from somewhere else? Seems like having the `usePersistedState` be use directly in the `AddStaticFieldButtonComponent` would suffice.

BohdanOne: I had some issue with that, but will look into it again!

BohdanOne: removed!

packages/slice-machine/src/features/builder/AddFieldDropdown.tsx
```diff
@@ -20,19 +21,25 @@ import {
...
 
 export function AddFieldDropdown(props: AddFieldDropdownProps) {
   const {
     disabled,
     onSelectField,
     fields,
-    triggerDataTestId = "add-field",
+    triggerDataTestId,
+    trigger,
+    ...implicitProps
```
levimykel: ❓ Why is `implicitProps` added here?

levimykel: I guess it's to pass `open` and `onOpenChange` to `DropdownMenu`. Why not explicitly pass those to make that more clear?

packages/slice-machine/src/features/builder/AddStaticFieldButton.tsx
```diff
@@ -0,0 +1,54 @@
...
+        onClose={dismissDialog}
+        trigger={!isInfoDialogSeen ? trigger : null}
+      />
+      <AddFieldDropdown
+        open={isAddFieldDropdownOpen}
+        onOpenChange={setAddFieldDropdownOpen}
+        disabled={disabled}
+        onSelectField={onSelectField}
+        fields={fields}
+        trigger={isInfoDialogSeen ? trigger : null}
```
xrutayisire: 💡 #best-practice

Here we have a best practice about not using null if undefined is possible. Same for others places where null is used in this PR.

```suggestion
        trigger={isInfoDialogSeen ? trigger : undefined}
```

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
```diff
@@ -0,0 +1,78 @@
...
+      size="small"
+    >
+      <DialogHeader title="Static zone" />
+      <DialogContent>
+        <ScrollArea>
+          <Box as="article" flexDirection="column" padding={16} gap={4}>
+            <Text sx={{ marginBottom: 4 }} color="purple9" variant="smallBold">
+              About
+            </Text>
+            <Text variant="h3">What is the Static zone</Text>
```
xrutayisire: 💡 #wording

Here there is "Static zone" but later on "static zone", we should have consitency, no?

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
```diff
@@ -0,0 +1,78 @@
...
+            />
+          </Box>
+        </ScrollArea>
+      </DialogContent>
+      <DialogActions>
+        <DialogActionButton
+          size="medium"
+          sx={{ flexGrow: 1 }}
+          onClick={dismissDialog}
+          data-testid="static-fields-info-dialog-cta"
```
xrutayisire: 💡 #idea

Can we not use just "Got it" text? We prefer to use text over test id when it can prevent us to update code. + it makes us test more like the user would, by first seeing the text and after clicking on it.

Edit: Much more because I see we query it like that already: 
```
await this.page
      .getByRole("button", { name: "Got it", exact: true })
      .click();
```

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
```diff
@@ -55,6 +55,13 @@ export function UIDEditor() {
     setOpen(false);
   }
 
+  const uidIcon = useCallback(
```
xrutayisire, line 58: ❓ #question

Why do we need that change?

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
```diff
@@ -82,11 +90,8 @@ export function UIDEditor() {
               value={label}
               onValueChange={handleValueChange}
               error={error}
+              description="A label for the UID"
```
xrutayisire, line 93: 🎉 #nice

So much better like that 🙂

packages/slice-machine/src/legacy/components/Navigation/index.tsx
```diff
@@ -155,15 +144,6 @@ const Navigation: FC = () => {
           </SideNavListItem>
         )}
 
-        <ErrorBoundary>
-          <Suspense>
-            <VideoItem
```
xrutayisire: ❓ #question

What is this change related to? What is the product desire related to that?

packages/slice-machine/src/legacy/components/Navigation/index.test.tsx
```diff
@@ -140,7 +140,7 @@ describe("Side Navigation", () => {
     expect(within(changesItem).getByText("1")).toBeVisible();
   });
 
-  test("Video Item with next", async (ctx) => {
+  test.skip("Video Item with next", async (ctx) => {
```
xrutayisire: 💡 #idea

Is it a permanent removal of the code related to that? Why not deleting related tests then?

packages/slice-machine/src/legacy/components/Navigation/index.tsx
```diff
@@ -25,10 +24,6 @@ import { FolderIcon } from "@/icons/FolderIcon";
 import { LightningIcon } from "@/icons/Lightning";
 import { MasterSliceLibraryIcon } from "@/icons/MasterSliceLibraryIcon";
 import { SettingsIcon } from "@/icons/SettingsIcon";
-import VideoItem from "@/legacy/components/Navigation/VideoItem";
```
xrutayisire: ❓ #question

Should we remove related code?

packages/slice-machine/src/features/builder/AddStaticFieldButton.tsx
```diff
@@ -0,0 +1,54 @@
...
+import {
+  AddFieldDropdown,
+  type AddFieldDropdownProps,
+} from "@/features/builder/AddFieldDropdown";
+import { StaticFieldsInfoDialog } from "@/features/builder/StaticFieldsInfoDialog";
+import { usePersistedState } from "@/hooks/usePersistedState";
+
+const LOCAL_STORAGE_KEY = "staticFieldsInfoDialogDismissed";
+
+export function AddStaticFieldButton(props: AddFieldDropdownProps) {
```
xrutayisire: 💡 #idea

I was a bit lost with the name. Here it's still a `AddFieldDropdown`, why not something like `AddStaticFieldDropdown`? We understand it's just a wrapper around `AddFieldDropdown` directly.

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
```diff
@@ -87,63 +88,76 @@ const Zone = ({
...
+                    // TODO(DT-1710): add the missing `flexShrink: 0` property to the Editor's Switch component.
+                    style={{ flexShrink: 0 }}
+                    data-testid="code-snippets-switch"
+                  />
+                </>
+              )
+            }
+            {
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+              zoneType === "slice" && fields.length > 0 ? (
```
xrutayisire: 💡 #idea

I would like if we can prevent using double ternary in our code, here it's old code, so I don't care much, but it's more a guideline for us.

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
```diff
@@ -0,0 +1,78 @@
...
+        <ScrollArea>
+          <Box as="article" flexDirection="column" padding={16} gap={4}>
+            <Text sx={{ marginBottom: 4 }} color="purple9" variant="smallBold">
+              About
+            </Text>
+            <Text variant="h3">What is the Static zone</Text>
+            <Text color="grey11">
+              The static zone in Prismic page types contain fields that are
+              always present on the page. Use it for essential information like
+              a page's title, page design settings or metadata. Unlike slices,
```
levimykel: FYI, we use the oxford comma as part of our  [writing style guide](https://www.notion.so/prismic/Style-Guide-b1e1865ef5a54edeaf0d9ce6f4734d66?pvs=4#7face3f6c8a74615ad4582ce44559077).

"...a page's title, page design settings, or metadata."

playwright/tests/common/sideNav.spec.ts
```diff
@@ -138,7 +138,7 @@ test.describe("Tutorial tooltip", () => {
     },
   });
 
-  test("I can close the tutorial video tooltip and it stays close", async ({
+  test.skip("I can close the tutorial video tooltip and it stays close", async ({
```
xrutayisire: 💡 #idea

Is it a permanent removal of the code related to that? Why not deleting related tests then?

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
levimykel: Since we aren't referring to this as the Static zone anymore, I would change this title to match what we call it in the UI. For example, "What are static fields?"

playwright/pages/shared/TypeBuilderPage.ts
```diff
@@ -226,6 +223,15 @@ export class TypeBuilderPage extends BuilderPage {
       .click();
   }
 
+  async dismissStaticFieldsInfoDialog() {
+    await this.page.getByTestId("add-field").click();
+    await this.page
+      .getByRole("button", { name: "Got it", exact: true })
+      .click();
+    await expect(this.addFieldDropdown.menu).toBeVisible();
+    await this.staticZoneAddFieldButton.click();
```
xrutayisire: 💡 #idea

Finishing with button seems strange, much more how it's used later on. The function name is to dismiss static fields info dialog, but we are doing a side effect to click on the add field button at the end. I guess we could improve here and prevent side effect without assertion after.

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  return (
+    <Dialog
+      open={isOpen}
+      onOpenChange={handleOpenChange}
+      trigger={trigger}
+      size="small"
+    >
+      <DialogHeader title="Static zone" />
```
levimykel, line 29: Since we aren't referring to this as the Static zone anymore, I would change this title to match what we call it in the UI, "Static fields".

packages/slice-machine/src/features/builder/AddStaticFieldButton.tsx
```diff
@@ -0,0 +1,54 @@
...
+    />
+  );
+
+  return (
+    <>
+      <StaticFieldsInfoDialog
+        onClose={dismissDialog}
+        trigger={!isInfoDialogSeen ? trigger : null}
+      />
+      <AddFieldDropdown
```
levimykel: One of the requirements in the [linear ticket](https://linear.app/prismic/issue/DT-2247/display-about-static-fields-dialog-when-adding-a-slice-to-a-page) is to add a ? button near the CTA to re-open the info dialog. Was that requirement removed, or does this still need to be added?

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
```diff
@@ -87,63 +88,76 @@ const Zone = ({
...
+                <AddFieldDropdown {...addFieldDropdownProps} />
+              ) : zoneType === "customType" ? (
```
xrutayisire: ⚠️ #issue

For me there is an issue about when I'm seeing the new static zone info dialog, for custom types.

Even the message talk about page types, did product ask us to do both custom types AND page types?

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
```diff
@@ -0,0 +1,78 @@
...
+        </ScrollArea>
+      </DialogContent>
+      <DialogActions>
+        <DialogActionButton
+          size="medium"
+          sx={{ flexGrow: 1 }}
+          onClick={dismissDialog}
+          data-testid="static-fields-info-dialog-cta"
+        >
+          Got it
```
xrutayisire, line 58: ❓#question

Even if you never click on "Got it" you will not see it again after, is that wanted?

xrutayisire (changes requested):
Nice work! Left some simple comments, didn't see anything big 🙂

packages/slice-machine/src/features/builder/AddStaticFieldButton.tsx
jomifepe: I'm actually passing `null` to hide it. I could also create a prop to do this, but the idea is that `trigger` overrides the default one, but I also want to hide the default one and show nothing 😅

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
jomifepe: Yeah, changed 👍

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
jomifepe: Yeah it's not being used, but IMO test ids for E2E are better, that way we don't keep breaking the tests if we want to update a simple label, only the full flow matters.

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
jomifepe: To update the icon size, unfortunately the `renderStartIcon` accepts a component definition, which IMO is wrong, it should accept a `ReactNode`, but for that reason, I define one and memoize it.

packages/slice-machine/src/legacy/components/Navigation/index.tsx
jomifepe: This was a request from Côme, but I extracted a separate PR for it: https://github.com/prismicio/slice-machine/pull/1428

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
xrutayisire: I don't agree to that, I prefer to have text when text is part of your product. It's our case on our side, with exact, it allows us to prevent someone to make a change of text without wanting to. I remember one time when I added a "s" because I wanted to do CTRL + s to save and e2e tests broke to warn me about it. 
When your text is part of your product, I would always favor text.

packages/slice-machine/src/legacy/components/Navigation/index.test.tsx
jomifepe: Same as https://github.com/prismicio/slice-machine/pull/1426#discussion_r1715061859, but yeah I agree that they can be removed ✅

packages/slice-machine/src/legacy/components/Navigation/index.tsx
jomifepe: Can do that, I didn't do it right away because I see that that it was quite a few lines of code, and we could want to bring it back in the future 🤷‍♂️

Will do it here: https://github.com/prismicio/slice-machine/pull/1428

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
jomifepe: Not sure I understood, if you dismiss the modal it won't be opened again on add static field.

packages/slice-machine/src/features/builder/AddStaticFieldButton.tsx
xrutayisire: Oh ok, got it! 🙌

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
xrutayisire: Ahhh ok, I don't like that we changed the default, like icon size small if for button size small normally 😬
But btw, I'm not a fan of use useCallback just for the sake of using it without a reason.

Do we have a performance problem? React don't encourage using it if not for performance optimization.
See: https://react.dev/reference/react/useCallback

Screenshot 2024-08-13 at 13 35 11

`useCallback` cost perf to do the memoize, and sometimes it's costing more than what you saved.

Also: [Should you add useCallback everywhere? ](https://react.dev/reference/react/useCallback#should-you-add-usecallback-everywhere)

I let you read this part and let you tell me what you think for this use case, if you think we should still wrap it or not necessary. It something I always like to learn more and more about when its good / bad to do.

Screenshot 2024-08-13 at 13 37 08

packages/slice-machine/src/legacy/components/Navigation/index.test.tsx
xrutayisire: Prevent redoing the same comment there but yeah If product wants us to remove the menu item totally, let's remove ALL code related to that, components, tests, everything

packages/slice-machine/src/legacy/components/Navigation/index.tsx
xrutayisire: Same:
> Prevent redoing the same comment there but yeah If product wants us to remove the menu item totally, let's remove ALL code related to that, components, tests, everything


I guess it depends on what Côme said, if they want to remove it totally and don't know about a usage again, let's remove it totally. We can find it after anyway with git history

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
xrutayisire: What I mean is when you don't click on "Got it" but close the modal with another way like the close icon, it will still be taken as "Got it". Just wanted to be sure it's ok for product.

packages/slice-machine/src/features/builder/AddStaticFieldButton.tsx
jomifepe: I guess that was the name chosen because the only visible element without interaction is a button

jomifepe: But I'm okay with that, renamed

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
jomifepe: Done

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
jomifepe: I'm aware of the premature optimization issue with `useMemo` and `useCallback`, but I believe defining components at render time is a worse pitfall than that.

xrutayisire: Good then, wanted to be sure you considered that 🙂

xrutayisire: Btw it will still be defined at render time, that didn't avoid that, just the number of time you're doing it

jomifepe: Yeah, I would prefer to fix this issue in the editor side, accepting a definition is not justifiable for an icon 😅

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
jomifepe: Ah, will check that

packages/slice-machine/src/features/customTypes/customTypesBuilder/UIDEditor.tsx
xrutayisire: It as discussed a lot in the past, you get the Icon in param, that's why it's like this. But I don't remember everything. Agree could be good to refresh on the reasons haha. Anyway good for here

packages/slice-machine/src/legacy/lib/builders/common/Zone/index.jsx
jomifepe: Confirmed this with product, made the change to only display the modal for page types.

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
```diff
@@ -0,0 +1,77 @@
...
+      size="small"
+    >
+      <DialogHeader title="Static zone" />
+      <DialogContent>
+        <ScrollArea>
+          <Box as="article" flexDirection="column" padding={16} gap={4}>
+            <Text sx={{ marginBottom: 4 }} color="purple9" variant="smallBold">
+              About
+            </Text>
+            <Text variant="h3">What is the static zone?</Text>
```
levimykel, line 36: Since we aren't referring to this as the Static zone anymore, I would change this to match what we call it in the UI, "What are static fields?".

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
```diff
@@ -0,0 +1,77 @@
...
+            <Text color="grey11">
+              The static zone in Prismic page types contain fields that are
```
levimykel, line 38: Since we aren't referring to this as the Static zone anymore, I would change this to match what we call it in the UI, "The static fields in Prismic...".

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
```diff
@@ -0,0 +1,77 @@
...
+              always present on the page. Use it for essential information like
+              a page's title, page design settings or metadata. Unlike slices,
+              which can be added or removed, static zone fields remain fixed.
```
levimykel, line 41: Since we aren't referring to this as the Static zone anymore, I would change this to match what we call it in the UI, "..., static fields remain fixed."

packages/slice-machine/src/features/builder/AddStaticFieldDropdown.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  const trigger = (
+    <Button
+      {...commonTriggerProps}
+      {...(triggerDataTestId != null && { "data-testid": triggerDataTestId })}
+    />
+  );
+
+  return (
+    <>
```
levimykel: ⚠️ There's no need to wrap this in fragments, I think.

levimykel (commented):
Looks good, I added some questions and comments.

packages/slice-machine/src/features/builder/AddFieldDropdown.tsx
jomifepe: Unsure of the reason, but I'm okay with that change 😄

jomifepe: Done

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
jomifepe: Ah, makes sense, changed

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
jomifepe: Make sense to me, will check with Côme

jomifepe: Missed the final correction from Côme, and we're not changing anything to Static fields after all, so these labels make sense, will change the only `Static fields` one back to `Static zone`.

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
jomifepe: https://github.com/prismicio/slice-machine/pull/1426#discussion_r1715315531

packages/slice-machine/src/features/builder/AddStaticFieldButton.tsx
jomifepe: Still checking this with Benjamin, there's no design for it yet.

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
jomifepe: This clears this up: https://github.com/prismicio/slice-machine/pull/1426#discussion_r1715315531

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
jomifepe: This clears this up: https://github.com/prismicio/slice-machine/pull/1426#discussion_r1715315531

packages/slice-machine/src/features/builder/AddStaticFieldDropdown.tsx
jomifepe: Yeah fixed 😄

packages/slice-machine/src/features/builder/StaticFieldsInfoDialog.tsx
jomifepe: Checked with Benj & Côme and now closing the dialog doesn't prevent it from showing up the next time

playwright/pages/shared/TypeBuilderPage.ts
jomifepe: Refactored, tell me what you think

xrutayisire: This `getStaticZoneInfoDialogConfirmCta` can be a simple static field as you don't pass param, but looks good otherwise

xrutayisire: I let you do the change and you can resolve the comment yourself

jomifepe: Won't that fail? Since it will look for it right away

levimykel (approved):
Nice work 😄

playwright/pages/shared/TypeBuilderPage.ts
xrutayisire: No it doesn't, that's the beauty of it 🤩

xrutayisire: Dynamic field like you did is only when you want, for example an item with a name that you don't have before the test. But everything in our classes will only be executed when you call them in a test.

jomifepe: Awesome then, refactored, thanks for the suggestion 🚀

xrutayisire (approved):
LGTM! Nice work to both you!

🎉

PR #1431, opened 2024-08-20
feat: add support for display text in links

levimykel:
**Resolves**: [DT-2256](https://linear.app/prismic/issue/DT-2256/aadev-slice-machine-supports-text-in-link-fields)

### Description

Adds support for the display text property in link fields. 

### Checklist

### How to QA [^1]

[^1]:


packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
dani-mp, line 1: this is a whole new component, do we want to add it to the legacy folder and use legacy stuff to implement it?

dani-mp: we talked and the decision is to keep it because we can't implement legacy design with new code

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
```diff
@@ -0,0 +1,67 @@
...
+type DisplayTextModel = { type: string };
+
+interface DisplayTextCheckboxProps {
+  text?: DisplayTextModel;
+  height?: 130 | 127;
+  setFieldValue: (
+    a: string,
+    b?: DisplayTextModel,
+    // eslint-disable-next-line @typescript-eslint/no-redundant-type-constituents, @typescript-eslint/no-explicit-any
+  ) => void | Promise<void | any>;
```
jomifepe: 💡 Small abbreviation

```suggestion
  ) => void | Promise<any>;
```

jomifepe: 💡 Maybe even like this, but I'm not sure if you want to enforce `void` if it's not a promise

```suggestion
  ) => any | Promise<any>;
```

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
```diff
@@ -0,0 +1,67 @@
...
+          <Box
+            display="flex"
+            justifyContent="space-between"
+            alignItems="center"
+            width="100%"
+          >
+            <Flex>
+              <Checkbox
+                checked={allowDisplayText}
+                onChange={() => setAllowDisplayText(!allowDisplayText)}
```
jomifepe: 💡 We could use the setter here

```suggestion
                onChange={(current) => setAllowDisplayText(!current)}
```

jomifepe (approved):
LGTM 🚀 Left only some minor suggestions

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
```diff
@@ -32,6 +41,38 @@ const Form = (props) => {
           </Col>
         ))
       }
+      <Col />
+      <Col key="allowTargetBlank">
+        <Box sx={{ mt: 2 }}>
```
dani-mp, line 46: two things about this:
- could we avoid the extra element? maybe placing the styling in the parent instead
- if this just adds some spacing, maybe this should be placed in the grid itself with `gap`

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
```diff
@@ -32,6 +41,38 @@ const Form = (props) => {
...
+          <Label
+            htmlFor="allowTargetBlank"
+            variant="label.primary"
+            sx={{
+              // mt: 1,
```
dani-mp:
```suggestion
```

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
```diff
@@ -0,0 +1,67 @@
+import { Box, Icon, Tooltip } from "@prismicio/editor-ui";
+import React, { useEffect, useState } from "react";
+import { Checkbox, Flex, Label } from "theme-ui";
+
+import { Col } from "@/legacy/components/Flex";
+
+type DisplayTextModel = { type: string };
```
dani-mp:
```suggestion
type DisplayTextModel = { type: "Text" };
```

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+interface DisplayTextCheckboxProps {
+  text?: DisplayTextModel;
+  height?: 130 | 127;
```
dani-mp, line 8: I wonder if we can avoid this, I don't really understand why we have to do it but it seems odd that we have these two values and then we use them as a percentage `height: `${height}%``

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  const [allowDisplayText, setAllowDisplayText] = useState(Boolean(text));
+
+  useEffect(() => {
+    void setFieldValue(
+      "config.text",
+      allowDisplayText ? { type: "Text" } : undefined,
+    );
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [allowDisplayText]);
```
dani-mp: Instead of doing this, let's call `setFieldValue` in the checkbox's on change, and derive the boolean for the checkbox. That will remove the effect, the extra state and the extra render.

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  return (
+    <Col>
+      <Flex
+        sx={{
+          mt: 2,
```
dani-mp, line 23: Same as before, i wonder if we can remove this extra wrapping element and if the spacing should be applied in the parent instead with `gap`.

dani-mp (commented):
left some comments

xrutayisire (commented):
Just tested the alpha and saw that the default value for "Allow display text" is false where it should be true by default.

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
levimykel: Good call, thanks. Fixed ✅

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
levimykel: Yeah, that's better. Updated ✅

levimykel:
> Just tested the alpha and saw that the default value for "Allow display text" is false where it should be true by default.

Done ✅

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
levimykel: @dani-mp It's there to create this gap in the grid:
Screenshot 2024-08-27 at 08 41 17

I could achieve the same thing by expanding the 3rd item to 100% width, but I'd have to add another grid to keep the field at 50%. Like this:

```
Object.entries(DefaultFields).map(([key, field], index) => (
  <Col key={key} cols={index === 2 ? 1 : 2}>
    <FlexGrid>
      <Col key={key} cols={index === 2 ? 2 : 1}>
        <WidgetFormField
          ...fieldProps
        />
      </Col>
    </FlexGrid>
  </Col>
))
```

But I like that less than I like the empty `<Col />`. What do you think? Or do you have another idea of how I can do this?

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
levimykel: @dani-mp I would like to avoid this, I don't like it. It's there to get the "Allow display text" field to align. The "130%" value aligns the checkbox for content relationships and link to media fields:

Screenshot 2024-08-27 at 08 55 29

Then the "127%" option is needed for the link field. Without it, the fields are slightly misaligned: 

Screenshot 2024-08-27 at 08 55 42

I think it's just enough of a misalignment that it needs to be adjusted for. I can keep trying to find another solution for alignment. What do you think?

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
xrutayisire: Side note: Remember that we are talking about legacy code here, we want it to be understandable but it's ok if it's not our best code as the goal is to crash it all with the redesign.

levimykel: Oh, I think I misunderstood you, sorry. 

**With a gap**
Yes, the `<Box sx={{ mt: 2 }}>` can be avoided by adding a gap to the grid. But it will mean having to update every form for every field.

**Placing the styling on the parent**
This can also be accomplished but it would mean adding a prop to the `WidgetFormField` component so that it ignores the default "130%" height for a checkbox field. That or create a whole new component for the "Allow target blank" checkbox.

None of these options seem worth it to me as I believe that we're going to rework all the field forms soon anyway. What do you think?

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
levimykel: Same comment as my response above:

> Yes, the <Box sx={{ mt: 2 }}> can be avoided by adding a gap to the grid. But it will mean having to update every form for every field.
> ...
> This doesn't seem worth it to me as I believe that we're going to rework all the field forms soon anyway. What do you think?

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
levimykel: Removed ✅

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
levimykel: Good catch. It's fixed ✅

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
levimykel: That's better, yeah. Done ✅

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
dani-mp: by placing the margin in the parent i meant moving it to the above Col and remove the extra box

but i understand this is legacy code, most changes will be required and we don't want to invest on that atm, so you can discard this comment

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
dani-mp: not sure why this is happening, i guess it's a problem with the original layout and the intrinsic heights of those elements

i'm ok if we want to leave it like this as this is legacy code

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
```diff
@@ -0,0 +1,61 @@
...
+          >
+            <Flex>
+              <Checkbox
+                checked={Boolean(text)}
+                onChange={(event) => {
+                  void setFieldValue(
+                    "config.text",
+                    event.target.checked ? { type: "Text" } : undefined,
+                  );
+                }}
```
dani-mp: nice 👍

dani-mp (approved):

PR #1432, opened 2024-08-21
feat: add hook to manage ab-test

BohdanOne:
**Resolves**: [DT-2287](https://linear.app/prismic/issue/DT-2287/set-up-ab-test)

### Description
By providing prompts we can improve conversion to a first document published.
This PR adds a hook allowing to serve users new prompt to create content in PB

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/hooks/usePromptToCreateContentExperiment.ts
+++ b/packages/slice-machine/src/hooks/usePromptToCreateContentExperiment.ts
@@ -1,0 +1,8 @@
+import { useExperimentVariant } from "@/hooks/useExperimentVariant";
+
+type UsePromptToCreateContentExperimentReturnType = { eligible: boolean };
+
+export function usePromptToCreateContentExperiment(): UsePromptToCreateContentExperimentReturnType {
+  const variant = useExperimentVariant("slicemachine-prompt-to-create-content");
+  return { eligible: variant?.value === "on" };
+}


levimykel (approved):
Looks good to me 🎉

jomifepe (approved):
LGTM!

packages/slice-machine/src/hooks/usePromptToCreateContentExperiment.ts
```diff
@@ -0,0 +1,8 @@
+import { useExperimentVariant } from "@/hooks/useExperimentVariant";
+
+type UsePromptToCreateContentExperimentReturnType = { eligible: boolean };
+
+export function usePromptToCreateContentExperiment(): UsePromptToCreateContentExperimentReturnType {
+  const variant = useExperimentVariant("slicemachine-prompt-to-create-content");
+  return { eligible: variant?.value === "on" };
```
xrutayisire, line 7: What happen if a user deactivated telemetry? It will not be eligible with that, right? I thought it was something we wanted, as written in the ticket.

BohdanOne: yes, I guess user without telemetry will not see that feature.
@xrutayisire which ticket are you referring to? 
I fas as I understand this feature will be subject to ab-test, is there any other way to do ab-testing that doesn't require telemetry?

xrutayisire: I'm referring to ticket DT-2280, so it's not about the A/B testing directly here. It's about what if someone deactivated Telemetry? So of course they will not be part of the experiment, but we want a fallback variant so. The question is, what is the default variant we want when user is not part of the A/B testing experiment because he deactivated telemetry? The answer from product is in the ticket and we want the new version. I guess the equivalent to "on". With "on" being the new version (can be confirmed with Mabel of course)

BohdanOne: ah, ok
sorry, I haven't seen that ticket before
will add the fallback then!

PR #1434, opened 2024-08-22
feat: update change page blank slate content

BohdanOne:
**Resolves**: [DT-2285](https://linear.app/prismic/issue/DT-2285/update-change-page-blank-slate)

### Description

Update the blank state content depending for pre- and post-push state

### Checklist

### Preview

BEFORE
Zrzut ekranu 2024-08-22 o 11 11 07

AFTER
no new changes to push
Zrzut ekranu 2024-08-22 o 11 10 58

changes pushed
Zrzut ekranu 2024-08-22 o 11 10 42

### How to QA [^1]

[^1]:


packages/slice-machine/src/hooks/usePromptToCreateContentExperiment.ts
```diff
@@ -4,5 +4,7 @@ type UsePromptToCreateContentExperimentReturnType = { eligible: boolean };
 
 export function usePromptToCreateContentExperiment(): UsePromptToCreateContentExperimentReturnType {
   const variant = useExperimentVariant("slicemachine-prompt-to-create-content");
-  return { eligible: variant?.value === "on" };
+
+  // serve the experimental version to user with Telemetry deactivated
```
dani-mp, line 8: If we do this, then we won't be able to understand how well it performs, because part of the experiment will come from a different, non controlled scenario, right?

packages/slice-machine/src/pages/changes.tsx
```diff
@@ -111,7 +113,7 @@ const Changes: React.FunctionComponent = () => {
       return <AuthErrorPage authStatus={authStatus} />;
     }
     if (numberOfChanges === 0) {
-      return <NoChangesBlankSlate />;
+      return <NoChangesBlankSlate isPostPush={isPushed} />;
```
dani-mp, line 116: How does this behave when I reload the page? I wonder if there's a way of knowing if we have ever pushed changes before. Otherwise, we only have local state. When we reload the page, we'll have 0 changes but no post push, and we'll get a different blank slate content.

packages/slice-machine/src/features/changes/BlankSlates/getBlankSlateContent.ts
dani-mp, line 1: i would move this function as a private one in the main module, avoiding the indirection.

packages/slice-machine/src/pages/changes.tsx
BohdanOne: when we reload we see the content confirming there are no new changes to pushed (it's named prePush, but i just realised this names make no sense, will change it).
the postPush should be visible only right after the successful push.

packages/slice-machine/src/hooks/usePromptToCreateContentExperiment.ts
BohdanOne: I'm kind of not sure how to tackle this one. 
For me control version of ab-test is the old version, and the experiment is the new version.
But here product decision was to serve new version by default, also to users that are not in the ab-test at all (the case of disabled telemetry). That's why I added this fallback.
But would that influence the experiment metrics? We don't track actions of users with telemetry disabled, right? Or am I missing something here?

dani-mp: i guess you're right (if we assume that no variant means that segment events are not sent either)

but in that case, i would have modeled this the other way around: i would create a flag to opt in to the version that is not the default one

BohdanOne: I thought about that, but I'm not 100% sure how Amplitude will treat that (as in their docs the control version is the old version)

packages/slice-machine/src/features/changes/BlankSlates/getBlankSlateContent.ts
BohdanOne: done

dani-mp (approved):

PR #1435, opened 2024-08-22
feat: add success toast when push ready

BohdanOne:
**Resolves**: [DT-2277](https://linear.app/prismic/issue/DT-2277/add-success-toast-when-push-ready)

### Description

We saw users feeling confused after the push not knowing where they should go next.
In order to explain users that they need to go to Page Bulder and create content, we will display a toast with success message and link to PB.

### Checklist

### Preview
Zrzut ekranu 2024-08-22 o 13 24 45

### How to QA [^1]

[^1]:


packages/slice-machine/src/pages/changes.tsx
```diff
@@ -50,6 +55,13 @@ const Changes: React.FunctionComponent = () => {
   >(undefined);
   const { autoSyncStatus } = useAutoSync();
   const router = useRouter();
+  const [isToastOpen, setIsToastOpen] = useState(false);
+  const { repoName } = useSelector((state: SliceMachineStoreType) => ({
+    repoName: getRepoName(state),
+  }));
```
dani-mp:
```suggestion
  const repoName = useSelector(getRepoName);
```

packages/slice-machine/src/pages/changes.tsx
```diff
@@ -155,6 +171,28 @@ const Changes: React.FunctionComponent = () => {
                 void onPush(false); // not deleting documents by default
               }}
             />
+            <Toast
+              anchor={<Box position="fixed" top={30} right={32} />}
```
dani-mp: Position feels weird here, imo, because top and right paddings don't match (personal opinion, btw)

image

packages/slice-machine/src/pages/changes.tsx
```diff
@@ -84,7 +96,11 @@ const Changes: React.FunctionComponent = () => {
         // Update last sync value in local storage
         pushChangesSuccess();
 
-        toast.success("All slices and types have been pushed");
+        // TODO: hide this toast for users in experiment eligible version (DT-2287)
+        setIsToastOpen(true);
```
BohdanOne: I'll refactor this to use useToast hook when [PR with some fixes in `editor-ui`](https://github.com/prismicio/editor/pull/1193) is merged,
This will make this component and managing toast cleaner.

packages/slice-machine/src/pages/changes.tsx
```diff
@@ -155,6 +171,28 @@ const Changes: React.FunctionComponent = () => {
...
               }}
             />
+            <Toast
+              anchor={<Box position="fixed" top={30} right={32} />}
+              open={isToastOpen}
+              onOpenChange={setIsToastOpen}
+              variant="card"
+              seconds={20}
+              title={
+                <Box flexDirection="column" alignItems="flex-start" as="span">
```
dani-mp: why `span`?

packages/slice-machine/src/pages/changes.tsx
```diff
@@ -84,7 +96,11 @@ const Changes: React.FunctionComponent = () => {
         // Update last sync value in local storage
         pushChangesSuccess();
 
-        toast.success("All slices and types have been pushed");
+        // TODO: hide this toast for users in experiment eligible version (DT-2287)
+        setIsToastOpen(true);
+
+        // TODO: display this toast for users in experiment control version (DT-2287)
+        // toast.success("All slices and types have been pushed");
```
dani-mp: I'm confused about these two todos, is actually the same todo, no? why not add the flag right away? (or add the flag first, then add the toast second)

that way we keep the project in a shippable state

packages/slice-machine/src/pages/changes.tsx
BohdanOne: you're right, I just blindly copied it - anyway this is repeated inside `NoChangesBlankSlate` component, I plan to refactor it and leave it in only one place here, when both my PR's are merged

packages/slice-machine/src/pages/changes.tsx
BohdanOne: this goes inside `Text` component in the Toast which is rendered as `p` and this caused some warnings showing up in console

packages/slice-machine/src/pages/changes.tsx
BohdanOne: that's the paddings from Figma, will ask on daily if we can make paddings even :)

packages/slice-machine/src/pages/changes.tsx
BohdanOne: there is still discussion about the experiment management open in the second PR, that's why I didn't include it here yet
since we want the new Toast to be a default one I thought is ok like that

BohdanOne: I see you approved the other PR, will merge it and get rid of that todo right away then :)

packages/slice-machine/src/pages/changes.tsx
dani-mp: in that case, i would talk with design to align on this

we're workarounding the _title_ (that originally seemed to mean for a paragraph, so we could argue that `title` here is misleading) to show a header and a body

maybe those can be two different props

BohdanOne: Originally Toast accepted a ReactNode as a title, i didn't want to add too many new props too it, so I used that. 
But you're right, this might be confusing. There is [open PR](https://github.com/prismicio/editor/pull/1193) for few more tweaks to the Toast, so I can still change that

packages/slice-machine/src/pages/changes.tsx
BohdanOne: changed (actually there was a hook that returns repoName, so used it instead)

packages/slice-machine/src/pages/changes.tsx
BohdanOne: todos removed

packages/slice-machine/src/pages/changes.tsx
dani-mp: yeah, so i would first align with design on this, and then post a message in #frontend-engineering about the best way to tackle this:
- have an unopinionated react node so the consumer can set whatever
- have a heading and a body (both optional, some optional...?)
- something else?

packages/slice-machine/src/pages/changes.tsx
```diff
@@ -158,6 +186,28 @@ const Changes: React.FunctionComponent = () => {
                 void onPush(false); // not deleting documents by default
               }}
             />
+            <Toast
```
jomifepe, line 187: 💡 Maybe we could start using `ToastProvider` and the `useToast` hook.

packages/slice-machine/src/pages/changes.tsx
jomifepe: ❓ According to the design, shouldn't the toast be a bit wider?

packages/slice-machine/src/pages/changes.tsx
BohdanOne: yes, I want to add that, but missed one prop when adjusting ToastProvider, so this [PR](https://github.com/prismicio/editor/pull/1193) is needed first

packages/slice-machine/src/pages/changes.tsx
```diff
@@ -86,7 +101,10 @@ const Changes: React.FunctionComponent = () => {
         pushChangesSuccess();
 
         setIsPushed(true);
-        toast.success("All slices and types have been pushed");
+
+        isPromptToCreateContentExperimentEligible
+          ? setIsToastOpen(true)
+          : toast.success("All slices and types have been pushed");
```
jomifepe: 💡 Ternaries are not very readable to perform actions, only for assignments IMO. A regular if statement would look better here. WDYT?

```suggestion
        if (isPromptToCreateContentExperimentEligible) {
          setIsToastOpen(true);
        } else {
          toast.success("All slices and types have been pushed");
        }
```

BohdanOne: Don't have strong opinions here, will change it to use if statement

packages/slice-machine/src/pages/changes.tsx
BohdanOne: good observation, it was more narrow, but designs changed a little bit during the week, but this change will have to be included in Toast component itself

packages/slice-machine/src/pages/changes.tsx
BohdanOne: actually, in will leave it like that, 
when toast is launched with useToast, it stays on page when navigating away from `/changes` page, which in general is good but not in this use-case

jomifepe: Sure, but sounds like a feature that should be added to the `useToast` then 👀

jomifepe (approved):
LGTM, great job 🚀

packages/slice-machine/src/pages/changes.tsx
BohdanOne: yes probably, in general I have impression that Toast would need some proper revision

dani-mp (approved):

PR #1436, opened 2024-08-23
[DT-2249] feat: display toast on field create

jomifepe:
**Resolves**: DT-2249

### Description

~Previously, when adding a new item, the view scroll into the form, and after saving, the user was looking at the added item. After changing to the modal, that stopped happening, so to have the same experience, this PR scroll to the last added item.~ 

This PR displays a toast after creating an item.

ℹ️ The toast is only displayed once the auto-sync succeeds

### Preview

https://github.com/user-attachments/assets/fb11f734-081d-40dc-9618-20b636bd4220

### How to QA [^1]

[^1]:


packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/index.jsx
```diff
@@ -129,4 +139,28 @@ const FieldZone = ({
...
+
+  useEffect(() => {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    lastFieldsRef.current = fields;
+    if (!isReadyRef.current) isReadyRef.current = true;
+  }, [fields]);
+
+  return { lastItemRef };
+}
+
```
jomifepe: Not sure if this would be better under the features folder, instead of more code under legacy. WDYT @dani-mp @xrutayisire? 👀

xrutayisire: Arf, this one is tricky, because a hook like that could be nice and simple in our hooks folder. Generic enough, so you pass an array of items and it scrolls to last one. WDYT? Is it really more complex?

jomifepe: It's a bit dependent on the `fields` list, and we might be able to this in a more procedural way, instead of handling it as a side-effect no?

xrutayisire: Personally fine for me, it's today dependent to this specific design that will be changed, so I don't mind much.

jomifepe: Got it, it's currently being changed anyway, so this hook doesn't exist anymore 😅

dani-mp (dismissed):

BohdanOne (approved):
🎉 LGTM!

jomifepe:
@dani-mp @BohdanOne 

As discussed in the daily, I changed the approach to only display the new field success toast after syncing.

In this PR, I use a side-effect approach, but I also tried a callback-based one here https://github.com/prismicio/slice-machine/pull/1441. 

Tell me what you think!

dani-mp:
@jomifepe shared via slack earlier, but i'll also post here so other folks can follow along:

the new hook introduced in this PR doesn’t respect the rules of hooks

if we had to do all that it’s probably because the use-effect approach is not the best for this

i’ve also seen that we have created an alternative approach, let’s focus on that one, but could we:
- apply those changes (the callback approach) against main instead of the other PR, so it’s easier to understand each alternative in isolation (without one depending on the other)
- use a promise instead of a callback

jomifepe:
Discussed a couple of issue with @xrutayisire and we went with a callback approach and skipped two failing E2E test for a feature that is not active.

This will get merged hopefully tomorrow, when approved by @xrutayisire 

cc @dani-mp

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/TabZone/index.tsx
```diff
@@ -132,7 +135,7 @@ const TabZone: FC<TabZoneProps> = ({ tabId }) => {
       sectionId: tabId,
     });
 
-    setCustomType(newCustomType);
+    setCustomType(newCustomType, () => toast.success("Field added"));
```
xrutayisire, line 138: 💡 #nit-pick

For consistency, I would use the same everywhere with or with the `{}` for the arrow function

xrutayisire: 💡 #nit-pick

For consistency, I would use the same everywhere, with or without the `{}` for the arrow function.

Same for other code that is different.

playwright/tests/common/autoSync.spec.ts
```diff
@@ -44,7 +44,11 @@ test("I can see the auto-sync succeed when making a change", async ({
   await expect(pageTypesBuilderPage.menu.autoSyncSynced).toBeVisible();
 });
```
xrutayisire, line 46: ⚠️ #coverage

I would quickly update the test when adding a field not in a group + in a group (both case) with the check of the success toaster, as it's a product requirement so we ensure we don't miss it one day.

xrutayisire (changes requested):
Code LGTM, but I would just add the update of the tests to protect ourselves 🙂

Concerning your code, we can always globally improve this management, but it fixes a bug where the error of the auto-save would not be visible! And it's good improvement, well done!

playwright/tests/common/autoSync.spec.ts
jomifepe: Done!

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/TabZone/index.tsx
jomifepe: Changed!

xrutayisire (approved):
Let's gooo!! Nice work!

PR #1441, opened 2024-08-26
refactor(🔄): use sync queue callback approach to display success toast

jomifepe:
### Description

Alternative approach to https://github.com/prismicio/slice-machine/pull/1436

### How to QA [^1]

[^1]:


packages/slice-machine/src/features/sync/AutoSyncProvider.tsx
```diff
@@ -72,79 +73,85 @@ export const AutoSyncProvider: FC<PropsWithChildren> = (props) => {
...
+              // know that they need to login again.
+              // This can easily happen if the the token expires when the user is
+              // offline.
+              return;
+            }
+
+            // If the sync failed, we want to display the error message with
+            // the retry button.
+            reject?.(error);
+            throw error;
```
dani-mp: Two things here:
- above if `isUnauthenticatedError(error)`, we're not resolving/rejecting the promise, what should we do here?
- here, i don't think we want to both reject and throw, it's like bubbling the error twice

packages/slice-machine/src/features/sync/AutoSyncProvider.tsx
```diff
@@ -42,14 +42,15 @@ export type AutoSyncStatus =
...
-  syncChanges: (args?: SyncChangesArgs) => void;
+  syncChanges: (args?: SyncChangesArgs) => void | Promise<void>;
 };
 
 type SyncChangesArgs = {
   environment?: Environment;
   loggedIn?: boolean;
   changedCustomTypes?: ChangedCustomType[];
   changedSlices?: ChangedSlice[];
+  callback?: () => void;
```
dani-mp: we don't need this, right?

packages/slice-machine/src/features/sync/AutoSyncProvider.tsx
```diff
@@ -42,14 +42,15 @@ export type AutoSyncStatus =
 
 type AutoSyncContext = {
   autoSyncStatus: AutoSyncStatus;
-  syncChanges: (args?: SyncChangesArgs) => void;
+  syncChanges: (args?: SyncChangesArgs) => void | Promise<void>;
```
dani-mp: why not?
```suggestion
  syncChanges: (args?: SyncChangesArgs) => Promise<void>;
```

jomifepe: Forgot to remove it 😅

packages/slice-machine/src/features/sync/AutoSyncProvider.tsx
jomifepe: Yeah makes sense, fixed!

PR #1442, opened 2024-08-27
feat: track experiment exposure event

BohdanOne:
**Resolves**: [DT-2276](https://linear.app/prismic/issue/DT-2287/set-up-ab-test)

### Description

Added experience exposure event needed for handling ab-test data in Amplitude.

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/manager/src/managers/telemetry/types.ts
+++ b/packages/manager/src/managers/telemetry/types.ts
@@ -366,7 +366,7 @@ type SliceMachineExperimentExposure = SegmentEvent<
 	typeof SegmentEventType.experiment_exposure,
 	{
 		flag_key: string;
-		variant?: string;
+		variant: string;
 	}
 >;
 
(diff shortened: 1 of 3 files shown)


packages/manager/src/managers/telemetry/types.ts
```diff
@@ -360,6 +362,14 @@ type SliceMachinePostPushToastCtaClicked = SegmentEvent<
 	typeof SegmentEventType.postPush_toastCtaClicked
 >;
 
+type SliceMachineExperimentExposure = SegmentEvent<
+	typeof SegmentEventType.experiment_exposure,
+	{
+		flag_key: string;
+		variant?: string;
```
xrutayisire: 💡 #suggestion

I think we can remove the `?` as variant should always be passed in our case, right?

xrutayisire (approved):
LGTM!

Just added a simple comment, you can tackle 🙂

packages/manager/src/managers/telemetry/types.ts
BohdanOne: added it because amplitude `Variant` has it also with `?` and it made TypeScript not happy with line 310 in `TelemetryManager` (even though we check that `value` is thruthy so can't be `undefined` few lines up 🤷‍♂️ )

xrutayisire: I think it's because you don't define the value before.
You could do: `const variantValue = ...` And then using variantValue should work

BohdanOne: yes, that worked, fixed

jomifepe (approved):
🚀

PR #1448, opened 2024-08-29
feat(onboarding): improve onboarding card visibility

BohdanOne:
**Resolves**: [DT-2294](https://linear.app/prismic/issue/DT-2294/improve-onboardingcard-visibility)

### Description

In order to increase `OnboardingGuide` visibility, we introduce experiment to test 2 more visible variants with animated card.

### Checklist

### How to QA [^1]

[^1]:


packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
```diff
@@ -17,7 +17,13 @@ import type { OnboardingStep } from "@/features/onboarding/types";
 
 const EndCtaIcon = () => <Icon name="playCircle" size="small" color="grey11" />;
 
-export const OnboardingProgressStepper = () => {
+interface OnboardingProgressStepperProps {
+  hasLargeButton: boolean;
```
dani-mp: 💡 if we make it `buttonSize: "large" | "medium"` it's both nicer in terms of api and more flexible

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.tsx
```diff
@@ -60,36 +119,51 @@ const confettiConfig: ConfettiConfig = {
...
+  descriptionColor?: "grey11" | "purple6";
+  descriptionVariant?: "small";
+  progressColor?: "white";
+  progressBackgroundColor?: "purple8";
+  progressLabelColor?: "purple3";
+}
 
-  if (!eligible || sliceMachineUpdateAvailable || adapterUpdateAvailable) {
-    return null;
+const getCardPropsFromExperimentVariant = (variant?: string): CardProps => {
```
dani-mp: 💡 if we do this, we avoid having to replace `{number}` later
```suggestion
const getCardPropsFromExperimentVariant = (props: { variant?: string, steps: number }): CardProps => {
```

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.tsx
```diff
@@ -60,36 +119,51 @@ const confettiConfig: ConfettiConfig = {
...
+  switch (variant) {
+    case "light":
+      return {
+        cardDescription:
+          "Render a live page with content coming from Prismic in 5 mins",
+        cardTitle: "Build your first Prismic Page in {number} simple steps",
```
dani-mp: 💡 i prefer the solution i shared above but another option could be to make this a function that receive the steps instead of a string literal, then in the caller side you execute the function to get the value rather than use the value directly

jomifepe (approved):
LGTM 🚀

packages/slice-machine/src/features/onboarding/OnboardingProgressStepper.tsx
BohdanOne: changed, also added this prop to function returning rest of the ui configuration

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.tsx
BohdanOne: good idea, done

PR #1449, opened 2024-08-29
Use link text in adapters

dani-mp:
https://linear.app/prismic/issue/DT-2271/aadev-slice-templates-use-text-in-link-fields


packages/adapter-next/src/sliceTemplates/CallToAction/index.ts
```diff
@@ -42,13 +42,9 @@ export const mocks: SharedSliceContent[] = [
 				value: {
 					__TYPE__: "ExternalLink",
 					url: "https://twitter.com/prismicio",
+					text: "Learn more",
 				},
 			},
-			buttonlabel: {
```
xrutayisire, line 47: 😅 That was a bug here 😬 Nice we removed it

xrutayisire (approved):
To be able to properly test, you also need to update prismic-types-internal for all adapters.

It's important since today using your alpha will strip the link text from both custom types and mock.

Anyway, we need to update all versions of prismic-types-internal + prismic-mock from SM to be safe.

But code LGTM! 🚀

packages/adapter-next/public/CallToAction/javascript.jsx
```diff
@@ -41,9 +41,7 @@ const PascalNameToReplace = ({ slice }) => {
 					<PrismicNextLink
 						className="es-call-to-action__button"
 						field={slice.primary.buttonLink}
-					>
-						{slice.primary.buttonLabel || "Learn more…"}
-					</PrismicNextLink>
+					/>
```
angeloashmore: ❓ #question: We lost the fallback values ("Learn more…"). Is that intentional? I think that's okay for a slice template, just want to make sure it's a conscious decision.

dani-mp: it was conscious for two reasons, but maybe we do not want this and want something else:
- they were not being shown, as every field had a value
- with them, we lose the ability to show how "powerful" the new link is

wdyt, should we use the fallback somewhere? maybe in one of the usages for each adapter? i'd say no because this is about the optimal way, not the "edge cases", but i don't have a strong opinion

xrutayisire: I did the same choice in our examples, really for the second reason:

> with them, we lose the ability to show how "powerful" the new link is

It's so cool to let them see that you don't need to do anything right now

angeloashmore: Works for me! 🙂 We can always explain how to do a fallback in the documentation.

PR #1452, opened 2024-09-02
feat(tracking): pascal case error tracking

BohdanOne:
**Resolves**: [DT-2322](https://linear.app/prismic/issue/DT-2322/tracking-add-tracking-to-understand-the-reach-of-the-error-pascalcase)

### Description

We saw a common error when creating or updating a Slice and would like to know the reach of user encountering this issue before fixing it.
This code will add tracking occurrences of `Value has to be PascalCased` error, adding `errorType` information on which type of action error happened.

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/manager/src/managers/telemetry/types.ts
+++ b/packages/manager/src/managers/telemetry/types.ts
@@ -374,7 +374,7 @@ type SliceMachineExperimentExposure = SegmentEvent<
 >;
 
 type SliceMachineSliceNamePascalCase = SegmentEvent<
-	typeof SegmentEventType.sliceName_pascalCase,
+	typeof SegmentEventType.sliceName_pascalCaseError,
 	{ errorType: "create" | "update" }
 >;
 
--- a/packages/manager/src/managers/telemetry/types.ts
+++ b/packages/manager/src/managers/telemetry/types.ts
@@ -41,7 +41,7 @@ export const SegmentEventType = {
 	postPush_emptyStateCtaClicked: "post-push:empty-state-cta-clicked",
 	postPush_toastCtaClicked: "post-push:toast-cta-clicked",
 	experiment_exposure: "experiment:exposure",
-	sliceName_pascalCase: "slice-name:pascal-case",
+	sliceName_pascalCaseError: "slice-name-error:pascal-case",
 } as const;
 type SegmentEventTypes =
 	(typeof SegmentEventType)[keyof typeof SegmentEventType];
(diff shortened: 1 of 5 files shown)


packages/manager/src/managers/telemetry/types.ts
```diff
@@ -41,6 +41,7 @@ export const SegmentEventType = {
 	postPush_emptyStateCtaClicked: "post-push:empty-state-cta-clicked",
 	postPush_toastCtaClicked: "post-push:toast-cta-clicked",
 	experiment_exposure: "experiment:exposure",
+	sliceName_pascalCase: "slice-name:pascal-case",
```
xrutayisire: 💡 #nit-pick

Should we have the info that it's about the error? 

```suggestion
	sliceNameError_pascalCase: "slice-name-error:pascal-case",
```

Or something like this?

packages/slice-machine/src/legacy/components/Forms/formsValidator.ts
```diff
@@ -17,6 +18,7 @@ export function validateSliceModalValues(
   { sliceName }: SliceModalValues,
   localLibs: ReadonlyArray<LibraryUI>,
   remoteLibs: ReadonlyArray<SliceSM>,
+  event: string,
```
xrutayisire: ⚠️ #issue

As you're passing a fixed string, it would be better to type it to prevent any wrong string:

```suggestion
  event: "create" | "update",
```

packages/manager/src/managers/telemetry/types.ts
```diff
@@ -370,6 +373,11 @@ type SliceMachineExperimentExposure = SegmentEvent<
 	}
 >;
 
+type SliceMachineSliceNamePascalCase = SegmentEvent<
+	typeof SegmentEventType.sliceName_pascalCase,
+	{ errorType: string }
```
xrutayisire: ⚠️ #issue

As you're passing a fixed string, it would be better to type it to prevent any wrong string:

```suggestion
	{ errorType: "create" | "update" }
```

xrutayisire (approved):
Renaud will love this event! 🎉

I left 3 small comments, tagged two as issues, just because it's always better to have typescript protection even if it's a small temporary experiment.
Anyway, as it's super small I also approve as you can easily take into account what you think should be done 👍

Of course, there is Segment setup for stage / prod that I didn't check to do.

PR #1455, opened 2024-09-11
fix: add field modal does not close on first close button click

BohdanOne:
**Resolves**: [DT-2345](https://linear.app/prismic/issue/DT-2345/add-field-modal-dont-close-one-the-first-click)

### Description
When adding a new field in a a page or custom type in `slice-machine` the edit modal won't close on the first X button click.
This is happening because Label input is being auto-focused.

onMouseDown added to the close button will execute close function before the focus is being hijacked by input.

### Checklist

### Preview

BEFORE

https://github.com/user-attachments/assets/d9e3e0d1-aeb9-428e-85c0-6fffc24f97a3

AFTER

https://github.com/user-attachments/assets/843c8f9a-29dc-4e77-bf97-637eec21b7da

### How to QA [^1]

[^1]:


packages/slice-machine/src/legacy/lib/builders/common/EditModal/index.jsx
```diff
@@ -234,6 +234,8 @@ const EditModal = ({ close, data, fields, onSave, zoneType }) => {
                   />
                   <Close // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-assignment
                     onClick={close}
+                    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
dani-mp: Let's add this comment at the top of the file so we don't have to add it in each line. We said we were going to do this but if we never start we'll never do it!

packages/slice-machine/src/legacy/lib/builders/common/EditModal/index.jsx
```diff
@@ -234,6 +234,8 @@ const EditModal = ({ close, data, fields, onSave, zoneType }) => {
...
+                    onMouseDown={close} // will execute the close function before Label input hijacks the focus
```
dani-mp: I wonder if this is the best way to solve this, it looks like a hack. This is a common scenario on the web (clicking a button while another element has the focus). We rarely have to duplicate the on-click handler to fix this, and with this, we're executing the close function twice (it might be ok for this particular scenario, but it is still not ideal).

jomifepe:
Small comment on the description: This doesn't seem to be caused by the label autofocus, the same happens if you open the modal, focus an input and try to close the modal with the X button. This is may be related to the re-render that happens on input blur.

BohdanOne:
> Small comment on the description: This doesn't seem to be caused by the label autofocus, the same happens if you open the modal, focus an input and try to close the modal with the X button. This is may be related to the re-render that happens on input blur, that validates the form.
I'm not 100% sure what is exactly the problem, and yes I also think is related to the re-render. 
What I'm sure about is that the issue is not present when there is no autofocus at all.

jomifepe:
> > Small comment on the description: This doesn't seem to be caused by the label autofocus, the same happens if you open the modal, focus an input and try to close the modal with the X button. This is may be related to the re-render that happens on input blur, that validates the form.
> > I'm not 100% sure what is exactly the problem, and yes I also think is related to the re-render.
> > What I'm sure about is that the issue is not present when there is no autofocus at all.

That's not the result I'm getting when I disabled the autofocus 😅

https://github.com/user-attachments/assets/2d9a8a09-2210-4887-b414-8ee0faab6ac7

BohdanOne:
> > > Small comment on the description: This doesn't seem to be caused by the label autofocus, the same happens if you open the modal, focus an input and try to close the modal with the X button. This is may be related to the re-render that happens on input blur, that validates the form.
> > > I'm not 100% sure what is exactly the problem, and yes I also think is related to the re-render.
> > > What I'm sure about is that the issue is not present when there is no autofocus at all.
> 
> That's not the result I'm getting when I disabled the autofocus 😅
> 
>  Kapture.2024-09-11.at.12.37.32.mp4

~~How do you disabled it? (I just removed the passed autofocus prop)~~ Ok, I see it, I didn't try to click on the input when autofocus was disabled 🤦

packages/slice-machine/src/legacy/lib/builders/common/EditModal/index.jsx
BohdanOne: I'm not 100% sure about what is exact root of the problem, it's most likely related with re-render happening when input is validated, and my idea was that after the re-render focus is brought back to input (if I disable autofocus, button works without any problem). 
I agree the solution looks like a hack, but couldn't find any other working (might try to spend some more time on it).

packages/slice-machine/src/legacy/lib/builders/common/EditModal/index.jsx
BohdanOne: forgot to comment that it was done, sorry

packages/slice-machine/src/legacy/lib/builders/common/EditModal/index.jsx
BohdanOne: I didn't find solution that would not require two handlers, but prevented firing close function twice

BohdanOne: @dani-mp I manage to get rid of those 2 event listeners by preventing the default field validation on blur (but not 100% sure if this is what we want)

PR #1460, opened 2024-09-30
feat: Make allow target blank true by default

dani-mp:
**Resolves**: 
https://linear.app/prismic/issue/DT-2359/slice-machine-link-field-have-target-blank-by-default-allowed

### Description

Make allow target blank true by default.

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
+++ b/packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
@@ -11,7 +11,7 @@ import { DisplayTextCheckbox } from "./components";
 // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 const FormFields = {
   ...DefaultFields,
-  allowTargetBlank: CheckBox("Allow target blank", false, false),
+  allowTargetBlank: CheckBox("Allow target blank", false, true),
 };
 
 const Form = (props) => {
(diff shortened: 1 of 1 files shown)


packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
```diff
@@ -11,7 +11,7 @@ import { DisplayTextCheckbox } from "./components";
 // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 const FormFields = {
   ...DefaultFields,
-  allowTargetBlank: CheckBox("Allow target blank", false, false),
+  allowTargetBlank: CheckBox("Allow target blank", false, true),
```
xrutayisire, line 14: 💡 #idea

Even if not really used, I would change the `defaultValue` param in Checkbox function `packages/slice-machine/src/legacy/lib/forms/fields.ts`. 

So it's clear everywhere what default we want.

xrutayisire (approved):
Small comment but LGTM!

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
xrutayisire: Or even directly used the fallback

dani-mp: why? that'd assume that we want all checkboxes checked by default, no?

xrutayisire: Hum, it's because I saw now we always pass true, but not really required of course.

xrutayisire: You're right that it's 2 different things

PR #1461, opened 2024-10-03
feat(onboarding): add shared onboarding guide

BohdanOne:
**Resolves**: [DT-2380](https://linear.app/prismic/issue/DT-2380/shared-onboarding-in-slice-machine)

### Description

In order to improve onboarding process of new user `OnboardingGuide` will be now shared between Slice Machine and Page Builder.
It will use new component from `editor-fields` package. ([Related PR in `editor` repository](https://github.com/prismicio/editor/pull/1260#discussion_r1785900604))
The state will be persisted and shared between two projects using new `repository` service created in `Obelix`.
New guide is subject to an ab-test experiment.

### Checklist

### Preview

https://github.com/user-attachments/assets/8342c317-6fd3-4612-8a19-e418046b365c

### How to QA [^1]

[^1]:


packages/manager/src/managers/prismicRepository/types.ts
```diff
@@ -167,3 +167,32 @@ export const Environment = t.type({
...
+	t.literal("chooseLocale"),
+	t.literal("createProject"),
+	t.literal("setupSliceMachine"),
+	t.literal("createPageType"),
+	t.literal("codePage"),
+	t.literal("createSlice"),
+	t.literal("reviewAndPush"),
+	t.literal("createContent"),
+	t.literal("renderPage"),
+]);
```
jomifepe: ⚠️ I wouldn't validate this, since the plan is to allow them to change, and be dynamic, we shouldn't maybe avoid having these in the client (same for PB)

BohdanOne: I'm not sure what is the best option:
- on one hand, if the content is updated we have to update the types (I could add a comment about that next to the content to remind of that)
- on the other, the steps will be toggled in other parts of the application using step id, having those types will help prevent any typos etc.

I'm more in favour of the first version since updating content would require changes in code anyway. And hopefully the content wouldn't be updated too often (if at all).

jomifepe (approved):
LGTM 🚀

packages/manager/src/constants/API_ENDPOINTS.ts
```diff
@@ -33,6 +34,7 @@ export const API_ENDPOINTS: APIEndpoints = (() => {
 					process.env.slice_machine_v1_endpoint ??
 						"https://mc5qopc07a.execute-api.us-east-1.amazonaws.com/v1/",
 				),
+				RepositoryService: "https://repository.internal.wroom.io/",
```
xrutayisire: ⚠️ You should manage the environment variables setup as other endpoint here. It should come from the .env of the dev env directly here right? Is `RepositoryService` fully functional is the dev env already as it's part of Obelix?

packages/manager/src/managers/prismicRepository/types.ts
xrutayisire: ⚠️ Here for me, it's also a product question. In the case we are receiving something unknown, how do we want to react? Hiding the onboarding card? Only displaying the known steps? etc

packages/manager/src/managers/prismicRepository/types.ts
```diff
@@ -167,3 +167,32 @@ export const Environment = t.type({
...
+
+export const OnboardingState = t.type({
+	completedSteps: t.array(OnboardingStepId),
+	isDismissed: t.boolean,
+	context: t.type({
+		framework: t.union([
+			t.literal("next"),
+			t.literal("nuxt"),
+			t.literal("sveltekit"),
+			t.literal("other"),
```
xrutayisire: ❓ How can it be `other`?

packages/manager/src/managers/prismicRepository/types.ts
```diff
@@ -167,3 +167,32 @@ export const Environment = t.type({
...
+	t.literal("createPageType"),
+	t.literal("codePage"),
+	t.literal("createSlice"),
+	t.literal("reviewAndPush"),
+	t.literal("createContent"),
+	t.literal("renderPage"),
+]);
+export type OnboardingStepId = t.TypeOf<typeof OnboardingStepId>;
+
+export const OnboardingState = t.type({
```
xrutayisire: ⚠️ For everything relating to types validation, and for new code, we want to use Zod and not io-ts. Only exception i prismic-types-internal as it's whole project. But now for new code where we want to validate a simple thing like that, Zod should be used.

packages/manager/src/managers/telemetry/types.ts
```diff
@@ -35,13 +35,13 @@ export const SegmentEventType = {
 	sliceMachine_start: "slice-machine:start",
 	sliceLibrary_beta_modalOpened: "slice-library:beta:modal-opened",
 	sliceLibrary_beta_codeOpened: "slice-library:beta:code-opened",
-	onboarding_step_opened: "onboarding:step-opened",
```
xrutayisire, line 38: ❓How can we removed the past event if we keep it for some use cases as a fallback?

packages/manager/src/managers/telemetry/types.ts
```diff
@@ -35,13 +35,13 @@ export const SegmentEventType = {
...
-	onboarding_step_completed: "onboarding:step-completed",
-	onboarding_completed: "onboarding:completed",
 	postPush_emptyStateCtaClicked: "post-push:empty-state-cta-clicked",
 	postPush_toastCtaClicked: "post-push:toast-cta-clicked",
 	experiment_exposure: "experiment:exposure",
 	sliceName_pascalCaseError: "slice-name-error:pascal-case",
+	sharedOnboardingGuide_stepCompleted: "shared-onboarding-guide:step-completed",
```
xrutayisire: ❓ Actually, why we don't re-use existing event? It's a demand from Product?

packages/slice-machine/src/features/onboarding/SharedOnboardingGuide.tsx
```diff
@@ -0,0 +1,26 @@
+import { OnboardingGuide } from "@prismicio/editor-fields";
+import { useMediaQuery } from "@prismicio/editor-ui";
+
+import { telemetry } from "@/apiClient";
+
+import { useOnboarding } from "./useOnboarding";
+
+export function SharedOnboardingGuide() {
+  const isOnboardingVisible = useMediaQuery({ min: "medium" });
```
xrutayisire: 💡 Why not having all conditions of the onboarding card display in the parent `packages/slice-machine/src/features/onboarding/OnboardingGuide.tsx`? So it's easier to know what conditions handle the display?

Here it's duplicated in both onboarding guide card, shared or not.

packages/slice-machine/src/features/onboarding/SliceMachineOnboardingGuide/OnboardingProgressStepper.tsx
```diff
@@ -35,9 +36,10 @@ export function OnboardingProgressStepper(
     setActiveStep(step);
     setDialogOpen(true);
     void telemetry.track({
-      event: "onboarding:step-opened",
+      event: "shared-onboarding-guide:step-open",
```
xrutayisire: ⚠️ That's wrong no? If it's the new guide it's good but if the old guide is displayed here it will sent the wrong event and mislead product analysis, right?

packages/slice-machine/src/features/onboarding/useOnboarding.ts
```diff
@@ -0,0 +1,70 @@
...
+  async function toggleStep(stepId: OnboardingStepId) {
+    if (!onboarding) return [];
+
+    const previousSteps = onboarding.completedSteps;
+
+    const newSteps = previousSteps.includes(stepId)
+      ? previousSteps.filter((id) => id !== stepId)
+      : [...previousSteps, stepId];
+
+    updateData(getOnboarding, [], {
```
xrutayisire: 💡Shouldn't you update the data only after it's a success in the back-end?

It would prevent to revert the completed steps after and risk a flash change.
This would even prevent to have the `newSteps` definition, I guess that is error-prone, no?

packages/slice-machine/src/features/onboarding/useOnboarding.ts
```diff
@@ -0,0 +1,70 @@
...
+        await managerClient.prismicRepository.toggleOnboardingStep(stepId);
+
+      updateData(getOnboarding, [], {
+        ...onboarding,
+        completedSteps,
+      });
+
+      return completedSteps;
+    } catch (err) {
+      console.error("Error toggling step:", err);
```
xrutayisire: ⚠️ What happens on the UI if you cannot complete a step? Here I don't see anything, but the user would not understand? Is that something you discussed with Design / Product? Should we have a toaster or something? It's part of the onboarding and I guess it's important to have the best flow even in case of errors.

packages/slice-machine/src/features/onboarding/SharedOnboardingGuide.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+  const { onboarding, toggleStep, toggleGuide } = useOnboarding();
+
+  if (!onboarding || !isOnboardingVisible) return null;
```
xrutayisire: 💡 Typically something we can propose to product is to have an event when the onboarding was not display due to an error. This could allow us to track how much onboarding guide are not visible because of network error, problem in our code etc.

xrutayisire (changes requested):
Some comments that I think will ensure the feature is more robust 🚀

packages/manager/src/constants/API_ENDPOINTS.ts
BohdanOne: ok, I'll check that

packages/manager/src/managers/telemetry/types.ts
BohdanOne: yes, I was discussing that with Come, now the same event will be also triggered in PB so we want a new one with changed name and one property added

packages/manager/src/managers/telemetry/types.ts
BohdanOne: ok, I need to check that, it shouldn't be used anymore - maybe I missed some place

packages/slice-machine/src/features/onboarding/SliceMachineOnboardingGuide/OnboardingProgressStepper.tsx
BohdanOne: That was discussed with Come, I think we need  the same events so we can compare experiment variants

packages/slice-machine/src/features/onboarding/useOnboarding.ts
BohdanOne: that was initial version, but thought that this optimistic update is better, since it take some time to get the response from the service, and then there is quite a delay between marking the step as done and having this reflected in ui.

packages/slice-machine/src/features/onboarding/useOnboarding.ts
BohdanOne: Asked Come about that and he decided to keep it super simple like it is now.

packages/slice-machine/src/features/onboarding/SharedOnboardingGuide.tsx
BohdanOne: good idea

packages/manager/src/managers/prismicRepository/types.ts
jomifepe: For me, with something unknown, we just want to ignore it and keep going, only considering the valid ones, because of deprecation purposes.

packages/manager/src/managers/prismicRepository/types.ts
jomifepe: When the user selects other? Or is that just `null`?

packages/manager/src/managers/prismicRepository/types.ts
jomifepe: The only issue, as of now, is that this will get passed to utils like `decode`, which accept a io-ts type

packages/manager/src/managers/prismicRepository/types.ts
xrutayisire: Same as my comment on editor PR, SM is something today only for the 3 frameworks above, we don't support anything else than that. So I don't understand all this other management.

packages/slice-machine/src/features/onboarding/SliceMachineOnboardingGuide/OnboardingProgressStepper.tsx
xrutayisire: I don't like the "I think" because we need to be sure 😛

packages/slice-machine/src/features/onboarding/useOnboarding.ts
xrutayisire: Ok, for me it's not about "simple" but bug. A user that would get an error would not have something "simple" but  would experience a bug. The error is not a bug, not having feedback is for me. I would like we push product to ensure error feedback as for us it can be 2 min dev. 
This can even help engineering in case of problem because the user will report I saw an alert with this message.

I guess here it's "less" a problem than editor because dev often open their console but still.

packages/manager/src/managers/prismicRepository/types.ts
xrutayisire: So it should just not fail if a step is not known, just not displaying it, right?

packages/manager/src/managers/prismicRepository/types.ts
xrutayisire: I would take the time to just do the decode with io-ts, as we use zod everywhere else for new code at Prismic, it's not very long to update only for here, no?

packages/manager/src/constants/API_ENDPOINTS.ts
jomifepe: Changed to check `process.env` first and use this as a fallback ✅

packages/manager/src/managers/prismicRepository/types.ts
jomifepe: Correct

packages/manager/src/managers/prismicRepository/types.ts
jomifepe: Honestly, here is doesn't really matter no? These are the possible values for this field, if it's `other`, the onboarding guide won't be show, but this can still return `other`.

xrutayisire: But for me it's like writing "toto" and saying it's not a problem. It's not, but why writing it? Why not removing it if it's not possible?

packages/manager/src/managers/prismicRepository/types.ts
jomifepe: Removed this ✅

packages/manager/src/managers/prismicRepository/types.ts
jomifepe: Yeah sure, I don't have a strong opinion here, but we should still validate it somewhere no?

Removed this ✅

packages/manager/src/managers/telemetry/types.ts
jomifepe: Ok so I don't have all the context, but as far as I understood by talking with Côme, we will keep the old events and also have new ones, as we will have an experiment hook controlling whether the new shared onboarding is visible.

Refactored to preserve the old events and add new ones ✅

packages/slice-machine/src/features/onboarding/SharedOnboardingGuide.tsx
jomifepe: Changed ✅

packages/slice-machine/src/features/onboarding/SliceMachineOnboardingGuide/OnboardingProgressStepper.tsx
jomifepe: Changed to use old/current events ✅

packages/manager/src/managers/prismicRepository/types.ts
xrutayisire: Hum, it will never happen so it's good without so. I mean you know the code better. But for me if other never happen no need to add it.

packages/slice-machine/src/features/onboarding/useOnboarding.ts
jomifepe: Change to show a small toast on error ✅

packages/slice-machine/src/features/onboarding/useOnboarding.ts
jomifepe: Did the same as in editor, since in this particular case, we want to avoid having a delay before throwing the confetti after completing the last step, this mutation doesn't have an optimistic update anymore ✅

packages/slice-machine/src/features/onboarding/SharedOnboardingGuide.tsx
jomifepe: Shouldn't we be using sentry or something similar for that?

xrutayisire: We could, I was thinking of Amplitude if product wants to track that as part of their data. Côme is not here so let's skip that and see later on :)

packages/manager/src/constants/API_ENDPOINTS.ts
```diff
@@ -33,6 +34,10 @@ export const API_ENDPOINTS: APIEndpoints = (() => {
 					process.env.slice_machine_v1_endpoint ??
 						"https://mc5qopc07a.execute-api.us-east-1.amazonaws.com/v1/",
 				),
+				RepositoryService: addTrailingSlash(
+					process.env.repository_endpoint ??
```
xrutayisire: ⚠️ I see that you're defining `repository_api` in this [PR](https://github.com/prismicio/prismicdev-cli/pull/59/files), this will be used here:

```suggestion
					process.env.repository_api ??
```

packages/manager/src/lib/DecodeError.ts
```diff
@@ -1,18 +1,37 @@
...
 	input: TInput;
-	errors: t.Errors;
+	errors: t.Errors | ZodIssue[];
 };
 
+function isZodIssueArray(errors: object[]): errors is ZodIssue[] {
+	return "path" in errors[0];
+}
+
+function formatZodErrors(errors: ZodIssue[]): string[] {
```
xrutayisire, line 14: 🎉 Nice to have managed this with Zod

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
```diff
@@ -506,6 +508,112 @@ export class PrismicRepositoryManager extends BaseManager {
...
+	async toggleOnboarding(): Promise<{ isDismissed: boolean }> {
+		const repositoryName = await this.project.getRepositoryName();
+
+		const url = new URL("/onboarding/toggle", API_ENDPOINTS.RepositoryService);
+		url.searchParams.set("repository", repositoryName);
+		const res = await this._fetch({ url, method: "PATCH" });
+
+		if (res.ok) {
+			const json = await res.json();
+			const { value, error } = decode(t.type({ isDismissed: t.boolean }), json);
```
xrutayisire: ⚠️ Could be with Zod now, right?

packages/manager/src/managers/prismicRepository/types.ts
```diff
@@ -167,3 +168,40 @@ export const Environment = t.type({
...
+	z.enum([...supportedSliceMachineFrameworks, "other"]),
+);
+
+export type RepositoryFramework = z.TypeOf<typeof repositoryFramework>;
+
+export const OnboardingState = z.object({
+	completedSteps: z.array(z.string()),
+	isDismissed: z.boolean(),
+	context: z.object({
+		framework: repositoryFramework.optional().default("other"),
```
xrutayisire: 💡 Is the default here necessary? As you already handle a default within the preprocess, no? Or when there is no value it's jumping the preprocess?

packages/manager/src/managers/telemetry/types.ts
```diff
@@ -338,26 +348,34 @@ type SliceLibraryBetaCodeOpened = SegmentEvent<
 	typeof SegmentEventType.sliceLibrary_beta_codeOpened
 >;
 
+type OnboardingCommonPayload = { stepId: string; stepTitle: string };
+type SharedOnboardingProperties<T = {}> = T & { source: "SliceMachine" }; // eslint-disable-line @typescript-eslint/ban-types
```
xrutayisire: ⚠️ Hum don't like this ban 😛
In TS `{}` should almost never be used, are you sure you don't want to do something like that: 

```suggestion
type SharedOnboardingProperties<T = Record<string, string>> = T & { source: "SliceMachine" };
```

?

packages/slice-machine/src/components/SideNav/SideNav.module.css
```diff
@@ -180,6 +180,10 @@
   }
 }
 
+button.link {
```
xrutayisire: ❓Why the button in front? Why just `.link` was not enough?

packages/slice-machine/src/components/SideNav/SideNav.tsx
xrutayisire, line 1: ⚠️ We need to rethink this file, I don't understand why we do have this complexity, seems really too much for what we want.

packages/slice-machine/src/features/onboarding/SliceMachineOnboardingGuide/OnboardingProvider.tsx
```diff
@@ -60,14 +60,18 @@ export const OnboardingProvider = ({
 
     if (newCompleteState) {
       void telemetry.track({
-        event: "onboarding:step-completed",
+        event: "shared-onboarding-guide:step-completed",
```
xrutayisire: ❓That's the old guide, no? Should take the old guide tracking key, no?

packages/slice-machine/src/features/onboarding/useOnboarding.ts
```diff
@@ -0,0 +1,62 @@
...
+    console.error("Failed to fetch onboarding", error);
+    return undefined;
+  }
+}
+
+export function useOnboarding() {
+  const onboarding = useRequest(getOnboarding, []);
+  const getOnboardingState = useRefGetter(onboarding);
+
+  function updateCache(onboarding: OnboardingState) {
```
xrutayisire: 💡 
```suggestion
  function updateCache(newOnboardingState: OnboardingState) {
```

A bit hard to read when it has the same name as a parent `onboarding` var.

xrutayisire (changes requested):

packages/slice-machine/src/features/onboarding/SliceMachineOnboardingGuide/OnboardingProvider.tsx
jomifepe: Yeah, you're correct, fixed ✅

packages/manager/src/constants/API_ENDPOINTS.ts
jomifepe: Do you want me to rename it to `repository_endpoint` instead?

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
jomifepe: Missed this one, refactored ✅

packages/manager/src/managers/prismicRepository/types.ts
jomifepe: Yeah looks like it's redundant, changed ✅

packages/manager/src/managers/telemetry/types.ts
jomifepe: Refactored ✅

packages/slice-machine/src/components/SideNav/SideNav.module.css
jomifepe: The button didn't stretch horizontally like the anchor 😕

Screenshot 2024-10-17 at 15 44 51

packages/slice-machine/src/features/onboarding/useOnboarding.ts
jomifepe: Fine by me, renamed ✅

packages/manager/src/managers/prismicRepository/types.ts
jomifepe: Done ✅

packages/manager/src/constants/API_ENDPOINTS.ts
xrutayisire: Honesty, it's ok, I'm not sure which one is the most up-to-date for SRE team. I guess `api` as Michel did that for the Search.

packages/slice-machine/src/components/SideNav/SideNav.tsx
xrutayisire: So decision, we remove all the Get Started new UI management, we keep as Bohdan left the PR. 
Then in a second PR we can have the auto check management. Then in a third PR I will check with Product what to do for the Get Started management.

packages/manager/src/constants/API_ENDPOINTS.ts
jomifepe: Done ✅

packages/slice-machine/src/components/SideNav/SideNav.tsx
jomifepe: Just committed a simpler approach to the side nav button. Avoided adapting the existing one in favor of a dedicated button, WDYT?

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.tsx
```diff
@@ -1,101 +0,0 @@
-import {
-  Card,
-  CardContent,
-  ProgressBar,
-  Text,
-  useMediaQuery,
-} from "@prismicio/editor-ui";
-import { confetti as fireConfetti, ConfettiConfig } from "dom-confetti";
```
xrutayisire, line 8: ⚠️ CI failing because you need to remove this dependency

packages/slice-machine/src/components/SideNav/SideNav.tsx
```diff
@@ -132,6 +132,38 @@ export const SideNavLink: FC<SideNavLinkProps> = ({
...
+  component?: "button" | FC<{ onClick?: (event: MouseEvent) => void }>;
+  onClick?: (event: MouseEvent) => void;
+};
+
+export const SideNavButton: FC<SideNavButtonProps> = ({
+  title,
+  RightElement,
+  Icon,
+  active,
+  component: Comp = "button",
```
xrutayisire: ⚠️ Let's not accept a component, today, if it's only a button and always a button, we don't need it.
Simplify a lot the code.

packages/slice-machine/src/components/SideNav/SideNav.tsx
```diff
@@ -132,6 +132,38 @@ export const SideNavLink: FC<SideNavLinkProps> = ({
...
+  ...otherProps
+}) => {
+  const visible = useMediaQuery({ max: "medium" });
+
+  return (
+    <Tooltip content={title} side="right" visible={visible}>
+      <Comp {...otherProps} className={styles.link} data-active={active}>
```
xrutayisire: ⚠️ Let's have a proper style management, then. 

`.button` instead of link, and same for other, it will prevent the `button.link` in style + will prevent having link used for button by doing.

```
.link,
.button
```

Where needed

xrutayisire (changes requested):
Next time, create another PR as right now it's starting to create a huge PR hard to review when you are adding something else. You work on the code of someone else, so ok to take into account comments about simple things, but for new code, a new PR is always better.

packages/slice-machine/src/features/onboarding/OnboardingGuide/OnboardingGuide.tsx
jomifepe: Done ✅

packages/slice-machine/src/components/SideNav/SideNav.tsx
jomifepe: Makes sense, done ✅

packages/slice-machine/src/components/SideNav/SideNav.tsx
jomifepe: Done ✅

xrutayisire: I just that for that:

```
button.link {
  width: 100%;
}
```

You can do that now:

```
.button
```

jomifepe: Missed that one, done ✅

xrutayisire (approved):
LGTM!

PR #1462, opened 2024-10-03
fix(onboarding): make createProject step optional

BohdanOne:
**Resolves**: no ticket

### Description

After removing the ab-test on OnboardingGuide the new version has one more onboarding step. 
If a user had old steps saved in localStorage and updated SM version it caused to show warning in console.
Screenshot 2024-10-03 at 17 05 18

### Checklist

### How to QA [^1]

[^1]:


packages/slice-machine/src/features/onboarding/OnboardingProvider.tsx
```diff
@@ -54,7 +54,7 @@ export const OnboardingProvider = ({
   );
 
   const toggleStepComplete = (step: OnboardingStep) => {
-    const isComplete = !stepStatus[step.id];
+    const isComplete = !(stepStatus[step.id] ?? false);
```
xrutayisire: ⚠️ 

That form is too weird / complex for little for me, you start to do an opposite of something or false. Can you rethink that? Seems too complex? 

What are the possible values, what are the ouput you desire and check what can we do.

packages/slice-machine/src/features/onboarding/OnboardingProvider.tsx
```diff
@@ -76,13 +76,11 @@ export const OnboardingProvider = ({
   };
 
   const isStepComplete = (step: OnboardingStep) => {
-    return stepStatus[step.id] ?? false;
+    return (stepStatus[step.id] ?? false) || (step.defaultCompleted ?? false);
```
xrutayisire: ⚠️ 

Same here, it feels like 3 conditions inlined for not a lot of gain here.

packages/slice-machine/src/features/onboarding/types.ts
```diff
@@ -1,7 +1,7 @@
 import { z } from "zod";
 
 export const onboardingStepStatusesSchema = z.object({
-  createProject: z.literal(true),
+  createProject: z.boolean().optional(),
```
xrutayisire, line 4: ❓

This fix is globally until we have the new onboarding right?

xrutayisire (changes requested):

packages/slice-machine/src/features/onboarding/types.ts
BohdanOne: yes, this fix is for the 'old' guide which still be one of the experiment versions, so we will still use that for some time

packages/slice-machine/src/features/onboarding/OnboardingProvider.tsx
BohdanOne: yeah, you're right, I'll rethink naming so we don't have to cast opposite boolean values.
stepStatus[step.id] could be boolean, but now - since one of the steps is optional it can be undefined as well, and we want isComplete to have a boolean value

packages/slice-machine/src/features/onboarding/OnboardingProvider.tsx
BohdanOne: ok, will think of way to simplify that line

packages/slice-machine/src/features/onboarding/OnboardingProvider.tsx
xrutayisire: But when you do `!undefined` it's the same as `!(undefined ?? false)`

BohdanOne: For me it is, but TS is not happy with this
Zrzut ekranu 2024-10-4 o 10 48 55

xrutayisire: Shitty rule haha

`!Boolean(<var>)` should make it work

BohdanOne: yeah, it does! 
i thought !!(value) does the same what Boolean(value) and only checked the first option -but I was wrong
changed with your suggestion

packages/slice-machine/src/features/onboarding/OnboardingProvider.tsx
BohdanOne: changed

xrutayisire (approved):
I let you do some QA with the alpha version and good for me 🙂

PR #1463, opened 2024-10-04
chore: update editor-ui with form breaking changes

xrutayisire:
**Resolves**: DT-2382

### Description

- Rework form that were using FormInput from editor-ui to ensure it's working with editor-ui version 0.4.49

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/features/slices/convertLegacySlice/ConvertLegacySliceAsNewSliceDialog.tsx
+++ b/packages/slice-machine/src/features/slices/convertLegacySlice/ConvertLegacySliceAsNewSliceDialog.tsx
@@ -51,7 +51,6 @@ export const ConvertLegacySliceAsNewSliceDialog: FC<DialogProps> = ({
   };
 
   const [values, setValues] = useState<FormValues>(defaultValues);
-
   const [errors, setErrors] = useState<SliceModalValuesValidity>();
 
   useOnChange(isOpen, () => {
--- a/packages/slice-machine/src/features/slices/convertLegacySlice/ConvertLegacySliceAsNewSliceDialog.tsx
+++ b/packages/slice-machine/src/features/slices/convertLegacySlice/ConvertLegacySliceAsNewSliceDialog.tsx
@@ -1,3 +1,4 @@
+import { useOnChange } from "@prismicio/editor-support/React";
 import {
   Box,
   Dialog,
@@ -6,23 +7,31 @@ import {
   DialogCancelButton,
   DialogContent,
   DialogHeader,
+  Form,
   FormInput,
   ScrollArea,
   Select,
   SelectItem,
   Text,
 } from "@prismicio/editor-ui";
-import { Formik } from "formik";
-import type { FC } from "react";
+import { type FC, useState } from "react";
 import { useSelector } from "react-redux";
 
-import { validateSliceModalValues as validateAsNewSliceValues } from "@/legacy/components/Forms/formsValidator";
+import {
+  SliceModalValuesValidity,
+  validateSliceModalValues as validateAsNewSliceValues,
+} from "@/legacy/components/Forms/formsValidator";
 import { pascalize } from "@/legacy/lib/utils/str";
 import { getRemoteSlices } from "@/modules/slices";
 import { SliceMachineStoreType } from "@/redux/type";
 
 import styles from "./ConvertLegacySliceButton.module.css";
 import { DialogProps } from "./types";
+
+type FormValues = {
+  from: string;
+  sliceName: string;
+};
 
 export const ConvertLegacySliceAsNewSliceDialog: FC<DialogProps> = ({
   isOpen,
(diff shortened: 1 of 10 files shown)


packages/slice-machine/src/theme.ts
```diff
@@ -886,10 +886,6 @@ const AppTheme = (): Theme =>
         variant: "textStyles.heading",
         fontSize: 5,
       },
-      h3: {
```
xrutayisire, line 889: Removed that as it was causing a previous change I did bad. The blank slate title was too big because of that style being applied. As we don't want theme-ui styles + we don't have other h3 that we want to style with that, I removed it.

packages/slice-machine/src/legacy/components/Forms/RenameVariationModal/RenameVariationModal.tsx
```diff
@@ -6,20 +7,19 @@ import {
...
 import { type FC, useState } from "react";
+import { z } from "zod";
 
 import { renameVariation } from "@/features/slices/sliceBuilder/actions/renameVariation";
 import { useSliceState } from "@/features/slices/sliceBuilder/SliceBuilderProvider";
 import type { ComponentUI } from "@/legacy/lib/models/common/ComponentUI";
 import type { VariationSM } from "@/legacy/lib/models/common/Slice";
 import useSliceMachineActions from "@/modules/useSliceMachineActions";
 
-import styles from "./RenameVariationModal.module.css";
```
BohdanOne, line 21: ❓ shouldn't the css file be removed as well

packages/slice-machine/src/features/slices/convertLegacySlice/ConvertLegacySliceAsNewSliceDialog.tsx
```diff
@@ -36,6 +45,37 @@ export const ConvertLegacySliceAsNewSliceDialog: FC<DialogProps> = ({
...
   }));
 
+  const defaultValues = {
+    from: libraries[0]?.name,
+    sliceName: pascalize(slice.key),
+  };
+
+  const [values, setValues] = useState<FormValues>(defaultValues);
+
+  const [errors, setErrors] = useState<SliceModalValuesValidity>();
```
dani-mp, line 54: If we keep an empty object here we don't have to check for errors all the time before checking if it has content.

packages/slice-machine/src/features/slices/convertLegacySlice/ConvertLegacySliceMergeWithIdenticalDialog.tsx
```diff
@@ -24,6 +25,31 @@ export const ConvertLegacySliceMergeWithIdenticalDialog: FC<DialogProps> = ({
   isLoading,
   identicalSlices,
 }) => {
+  const [path, setPath] = useState<string>(identicalSlices[0]?.path ?? "");
```
dani-mp:
```suggestion
  const defaultPath = identicalSlices[0]?.path ?? "";
  const [path, setPath] = useState(defaultPath);
```
so you can reuse it below as well

packages/slice-machine/src/legacy/components/Forms/RenameVariationModal/RenameVariationModal.tsx
```diff
@@ -34,9 +34,47 @@ export const RenameVariationModal: FC<RenameVariationModalProps> = ({
...
+  const [variationName, setVariationName] = useState<string>("");
+
+  const [error, setError] = useState<string | undefined>();
+
   const { setSlice } = useSliceState();
   const { saveSliceSuccess } = useSliceMachineActions();
 
+  useOnChange(isOpen, () => {
+    if (isOpen && variation?.name !== variationName) {
+      setVariationName(variation?.name ?? "");
```
dani-mp, line 46: in other forms, we do this reset when the form is closed, not a biggie but maybe we want to use always the same approach?

dani-mp (approved):
left 3 comments but they're not blocking

packages/slice-machine/src/legacy/components/Forms/RenameVariationModal/RenameVariationModal.tsx
xrutayisire: Good one, I forgot to check all css files following my changes, will take a look at all 🙂

xrutayisire: Done, was the only one 🙂

packages/slice-machine/src/features/slices/convertLegacySlice/ConvertLegacySliceAsNewSliceDialog.tsx
xrutayisire: Hum, typescript don't like it, because of the type `SliceModalValuesValidity` that is shared with others form I didn't touch. It include the fact it can be undefined. WDYT?

xrutayisire: I did it for the other form, where form errors type is scoped

packages/slice-machine/src/features/slices/convertLegacySlice/ConvertLegacySliceMergeWithIdenticalDialog.tsx
xrutayisire: Done 👍

packages/slice-machine/src/legacy/components/Forms/RenameVariationModal/RenameVariationModal.tsx
xrutayisire: So this is the only form we do this, it's because the form is a bit different, and we don't always give the variation. So it means with the other way we would init with an empty value because it's not passed.,

dani-mp (approved):

PR #1464, opened 2024-10-04
[DT-2335] Introduce local package linking script

jomifepe:
**Resolves**: DT-2335

### Description

The current package development process required the publishing of a public alpha (of editor packages) on npm and installing it here. In an effort of improving this flow, this PR creates a script to link/unlick the packages locally, using yalc, so that they can be tested before publishing publicly.

Package publishing PR: https://github.com/prismicio/editor/pull/1322

### Preview

https://github.com/user-attachments/assets/0a7cf2c1-a864-4800-9396-0c1f4f30345f

### How to QA [^1]

[^1]:

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "test": "yarn workspaces foreach --parallel --topological-dev --verbose run test",
     "publish": "tsx scripts/publish.ts",
     "play": "tsx scripts/play.ts",
-    "yalc:slice-machine": "tsx scripts/yalc-sm.ts"
+    "yalc:editor-packages": "tsx scripts/yalc-sm.ts"
   },
   "dependenciesMeta": {
     "@sentry/cli": {
--- a/package.json
+++ b/package.json
@@ -27,7 +27,8 @@
     "prettier:check": "prettier --check .",
     "test": "yarn workspaces foreach --parallel --topological-dev --verbose run test",
     "publish": "tsx scripts/publish.ts",
-    "play": "tsx scripts/play.ts"
+    "play": "tsx scripts/play.ts",
+    "yalc:slice-machine": "tsx scripts/yalc-sm.ts"
   },
   "dependenciesMeta": {
     "@sentry/cli": {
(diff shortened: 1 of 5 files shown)


scripts/yalc-sm.ts
```diff
@@ -0,0 +1,64 @@
...
+  if (prismicPkgs.length === 0) {
+    fail("No packaged to link");
+  }
+
+  for await (const name of packages) {
+    if (prismicPkgs.includes(name)) {
+      execaSync("yalc", ["link", name, "--dev"], { cwd: smWorkspaceLocation });
+      console.log(`${greenBright("✔")} Linked ${bold(green(name))}`);
+    } else {
+      console.log(`${red('✘')} Package ${bold(name)} not found in yalc`);
```
xrutayisire: 💡 Maybe we could explain what can be done here to if the package is not found.

package.json
```diff
@@ -27,7 +27,8 @@
     "prettier:check": "prettier --check .",
     "test": "yarn workspaces foreach --parallel --topological-dev --verbose run test",
     "publish": "tsx scripts/publish.ts",
-    "play": "tsx scripts/play.ts"
+    "play": "tsx scripts/play.ts",
+    "yalc:slice-machine": "tsx scripts/yalc-sm.ts"
```
xrutayisire: 💡 Is it not more `yalc:editor-packages`?

scripts/yalc-sm.ts
```diff
@@ -0,0 +1,64 @@
+import { existsSync, readdirSync } from "node:fs";
```
xrutayisire, line 1: ⚠️ For me it would be perfect to a small and simple bloc in the CONTRIBUTING.md guide about that. 
Like "How to work locally with editor packages?" something like that.

xrutayisire (changes requested):
Super early review, didn't test it yet, but the review can be shared with both PRs I guess for some comments.

scripts/yalc-sm.ts
jomifepe: Done!

package.json
jomifepe: Sounds good, didn't really know what to name it. Changed ✅

scripts/yalc-sm.ts
jomifepe: Added ✅

xrutayisire (approved):
LGTM, super work for our team!

I just let you handle the prettier issue 🙂

PR #1465, opened 2024-10-21
fix: add field modal does not close on the first close-button click

BohdanOne:
**Resolves**: [DT-2345](https://linear.app/prismic/issue/DT-2345/add-field-modal-dont-close-one-the-first-click)

### Description
When adding a new field in a a page or custom type in `slice-machine` the edit modal won't close on the first X button click.

The root cause is passing a Header to the Card as a function which gets re-defined after each state update.
This PR is a temporary solution to remove the bug: in order to remove Header as a function radius (that should be passed down to the Header) value was hard-coded.

Tech Debt ticket was created to remove the root issue: [DT-2384](https://linear.app/prismic/issue/DT-2384/tech-debt-aadev-i-can-use-card-component-in-a-predictable-way).

### Checklist

### Preview

BEFORE

https://github.com/user-attachments/assets/d9e3e0d1-aeb9-428e-85c0-6fffc24f97a3

AFTER

https://github.com/user-attachments/assets/843c8f9a-29dc-4e77-bf97-637eec21b7da

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/legacy/components/Card/index.tsx
+++ b/packages/slice-machine/src/legacy/components/Card/index.tsx
@@ -3,8 +3,13 @@ import { Card as ThemeCard, ThemeUIStyleObject } from "theme-ui";
 
 import { CardBox, CardBoxProps } from "./CardBox";
 
+// passing Header as a function caused a bug in EditModal
+// where the function was re-defined after each state update
+// which led to a bug where context was lost and a modal button was not working as expected
+// TODO: tech debt issue DT-2384
+
 interface CardProps extends Omit<CardBoxProps, "withRadius"> {
-  Header?: React.FC<{ radius: string }> | null;
+  Header?: React.FC<{ radius: string }> | JSX.Element | null;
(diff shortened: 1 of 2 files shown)


packages/slice-machine/src/legacy/components/Card/index.tsx
```diff
@@ -3,8 +3,13 @@ import { Card as ThemeCard, ThemeUIStyleObject } from "theme-ui";
 
 import { CardBox, CardBoxProps } from "./CardBox";
 
+// passing Header as a function caused a bug in EditModal
+// where the function was re-defined after each state update
+// which led to a bug where context was lost and a modal button was not working as expected
+// TODO: tech debt issue DT-2384
```
dani-mp: Let's remove this comment from here, as technically there's nothing wrong with this API. The problem is in the caller side (where we also left a comment), where they were defining the component in render.

dani-mp (approved):

PR #1467, opened 2024-10-24
[DT-2385] chore: rebuild navigation with design system components

BohdanOne:
**Resolves**: [DT-2385](https://linear.app/prismic/issue/DT-2385/aadev-i-want-to-have-a-sidenav-component-without-custom-styles)

### Description

This PR recreates `SideNav` with design system based components from `editor-ui` package. 

### Checklist

### Preview
Before
old
After
new

### How to QA [^1]

[^1]:


packages/slice-machine/src/features/navigation/Navigation.tsx
```diff
@@ -0,0 +1,151 @@
...
+  const gitIntegrationExperiment = useGitIntegrationExperiment();
+  const [isSliceLibraryDialogOpen, setIsSliceLibraryDialogOpen] =
+    useState(false);
+  const { masterSliceLibrary } = useMarketingContent();
+
+  return (
+    <Box
+      as="nav"
+      flexDirection="column"
+      padding={{ top: 16, right: 16, left: 16, bottom: 0 }}
```
dani-mp: curious why we have margin everywhere but the bottom, making the component not symmetrical

packages/slice-machine/src/features/navigation/Navigation.tsx
```diff
@@ -0,0 +1,151 @@
...
+      gap={16}
+      minWidth={0}
```
dani-mp, line 58: is this needed?

packages/slice-machine/src/features/navigation/Navigation.tsx
```diff
@@ -0,0 +1,151 @@
...
+    >
+      <ErrorBoundary>
+        <Suspense>
```
dani-mp: no fallback here means layout shifts all over the place

does it make sense to show the navigation content at all before this _header_ is loaded? if the answer is not, then maybe we should move this suspense boundary to the root of the component so it's all or nothing

packages/slice-machine/src/features/navigation/Navigation.tsx
```diff
@@ -0,0 +1,151 @@
...
+        </Suspense>
+      </ErrorBoundary>
+
+      <Separator style="dashed" />
+
+      <RepositoryInfo />
+
+      <ChangesItem />
+
+      <Box flexDirection="column" padding={{ top: 8 }} flexGrow={1}>
```
dani-mp: here another instance where padding is not symmetrical 

imo, it's important to keep components symmetrical and use the parent to add spacing, preferably via gap

that way, the components are truly independent and can be moved around when needed without having to reach out to its internal implementation

packages/slice-machine/src/features/navigation/Navigation.tsx
```diff
@@ -0,0 +1,151 @@
...
+            title={CUSTOM_TYPES_MESSAGES["custom"].name({
+              start: true,
+              plural: true,
+            })}
+            href={CUSTOM_TYPES_CONFIG["custom"].tablePagePathname}
+            active={CUSTOM_TYPES_CONFIG["custom"].matchesTablePagePathname(
+              router.asPath,
+            )}
+            Icon={CUSTOM_TYPES_CONFIG.custom.Icon}
+          />
```
dani-mp: these two are identical, maybe it's a good idea to create a small, private component that takes the key of the config and renders the thing

packages/slice-machine/src/features/navigation/Navigation.tsx
```diff
@@ -0,0 +1,151 @@
...
+          <NavigationItem
+            title="Slices"
+            href="/slices"
+            Icon={FolderIcon}
+            active={router.asPath.startsWith("/slices")}
+          />
+        </ActionList>
+
+        <ErrorBoundary>
+          <Suspense>
```
dani-mp, line 89: Same as before, this can potentially produce a layout shift.

packages/slice-machine/src/features/navigation/Navigation.tsx
```diff
@@ -0,0 +1,151 @@
...
+            <UpdateInfo />
+          </Suspense>
+        </ErrorBoundary>
+      </Box>
+
+      <Box flexDirection="column" padding={{ top: 48, bottom: 16 }}>
```
dani-mp: same comment about the component not being symmetrical

packages/slice-machine/src/features/navigation/Navigation.tsx
```diff
@@ -0,0 +1,151 @@
...
+        <ActionList variant="compact">
+          <ErrorBoundary>
+            <Suspense>
```
dani-mp, line 98: potential layout shift

packages/slice-machine/src/features/navigation/Navigation.tsx
```diff
@@ -0,0 +1,151 @@
...
+              <OnboardingGuide />
+            </Suspense>
+          </ErrorBoundary>
+          {masterSliceLibrary !== undefined && (
```
dani-mp: does this work?

```suggestion
          {masterSliceLibrary && (
```

packages/slice-machine/src/features/navigation/Navigation.tsx
```diff
@@ -0,0 +1,151 @@
...
+                    event: "slice-library:beta:modal-opened",
+                  });
+
+                  setIsSliceLibraryDialogOpen(true);
+                }}
+              />
+            </>
+          )}
+
+          {gitIntegrationExperiment.eligible ? (
```
dani-mp:
```suggestion
          {gitIntegrationExperiment.eligible && (
```

packages/slice-machine/src/features/navigation/Navigation.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+          <NavigationItem
+            title="Changelog"
+            href="/changelog"
+            Icon={LightningIcon}
+            active={router.asPath.startsWith("/changelog")}
+            RightElement={
+              <ErrorBoundary>
+                <Suspense>
+                  <Box>
```
dani-mp: why the box here?

packages/slice-machine/src/features/navigation/NavigationItem.tsx
```diff
@@ -0,0 +1,63 @@
...
+    props;
+
+  const isCollapsed = useMediaQuery({ max: "medium" });
+
+  const Content = (
+    <ActionListItem
+      textVariant="normal"
+      backgroundColor="transparent"
+      renderStartIcon={() => <Icon width={isCollapsed ? 28 : 32} />}
+      renderEndIcon={() => RightElement}
```
dani-mp: this is a bit confusing because we don't end up rendering an icon, but something completely different

also, IMO these should be memoized because otherwise we're mounting and unmounting these components all the time

packages/slice-machine/src/features/navigation/NavigationItem.tsx
```diff
@@ -0,0 +1,63 @@
...
+      selected={active}
+    >
+      {isCollapsed ? null : title}
```
dani-mp, line 49: just to be sure, we do want to show the action list item when is collapsed, but just without a title, right?

packages/slice-machine/src/features/navigation/NavigationItem.tsx
```diff
@@ -0,0 +1,63 @@
...
+    </ActionListItem>
+  );
+
+  return (
+    <Tooltip content={title} side="right" visible={isCollapsed}>
+      {href !== undefined ? (
```
dani-mp, line 55:
```suggestion
      {href ? (
```

packages/slice-machine/src/features/navigation/NavigationItem.tsx
```diff
@@ -0,0 +1,63 @@
...
+        <Link href={href} style={{ textDecoration: "none" }} {...otherProps}>
```
dani-mp: can we be explicit about other props here?

packages/slice-machine/src/features/navigation/RepositoryInfo.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export function RepositoryInfo() {
+  const { repositoryName, repositoryDomain, repositoryUrl } =
+    useRepositoryInformation();
+  const isCollapsed = useMediaQuery({ max: "medium" });
+
+  return (
+    <Box
+      justifyContent={isCollapsed ? "center" : "space-between"}
+      padding={{ bottom: isCollapsed ? 16 : 0 }}
```
dani-mp: asymmetrical component

packages/slice-machine/src/features/navigation/UpdateInfo.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  if (
+    (!sliceMachineUpdateAvailable && !adapterUpdateAvailable) ||
+    isCollapsed
+  ) {
+    return null;
+  }
+
+  return (
+    <Card color="grey3" padding={16} sx={{ marginTop: 32 }}>
```
dani-mp: asymmetrical component here, in this case it's even worse i'd say because we're pushing down this component 32px no matter where it's rendered - the parent should be responsible for this

dani-mp (commented):
left some comments

packages/slice-machine/src/features/navigation/Navigation.tsx
BohdanOne: which part do you mean exactly?

packages/slice-machine/src/features/navigation/Navigation.tsx
BohdanOne: I just moved the values from css file in here, but let me see, I think we add padding or margin to the last element so it could be added here instead

packages/slice-machine/src/features/navigation/NavigationItem.tsx
BohdanOne: yes, when it's collapsed only the icon is visible, and when it's hovered or focused the tooltip with the title shows up

packages/slice-machine/src/features/navigation/RepositoryInfo.tsx
BohdanOne: yes, it became asymmetrical when collapsed, but thanks to it the height of that element is maintained, so the rest of the sidebar below that doesn't jump up and down (in the collapsed view, there is only icon visible)

packages/slice-machine/src/features/navigation/Navigation.tsx
BohdanOne: fixed

packages/slice-machine/src/features/navigation/Navigation.tsx
BohdanOne: fixed

packages/slice-machine/src/features/navigation/Navigation.tsx
BohdanOne: fixed

packages/slice-machine/src/features/navigation/Navigation.tsx
BohdanOne: honestly, I don't remember why I left here, and it seems it's not needed, removing

packages/slice-machine/src/features/navigation/Navigation.tsx
dani-mp: minwidth

packages/slice-machine/src/features/navigation/NavigationItem.tsx
BohdanOne: it's because of that
Zrzut ekranu 2024-10-25 o 15 59 48

packages/slice-machine/src/features/navigation/RepositoryInfo.tsx
dani-mp: maybe we should use height instead then? or rendering an icon that takes the same amount of space so when the text is not there there's no layout shift

packages/slice-machine/src/features/navigation/Navigation.tsx
BohdanOne: yes it's needed, ActionListItem (when collapsed) are bit wider than the available space causing the nav to overflow, with this the nav can shrink to the width of the parent

packages/slice-machine/src/features/navigation/Navigation.tsx
BohdanOne: removed

packages/slice-machine/src/features/navigation/Navigation.tsx
BohdanOne: done

packages/slice-machine/src/features/navigation/Navigation.tsx
BohdanOne: yes, it does :) changed

packages/slice-machine/src/features/navigation/Navigation.tsx
BohdanOne: changed

packages/slice-machine/src/features/navigation/RepositoryInfo.tsx
BohdanOne: Changed to fixed height.
Also thanks to it spotted the difference that I overlooked: in old version height was fixed, in my (before this fix) if repo name would be long it would wrap and cause to increase height of that component.
Should be good now.

packages/slice-machine/src/features/navigation/NavigationItem.tsx
BohdanOne: done

packages/slice-machine/src/features/navigation/NavigationItem.tsx
BohdanOne: good call, changed

packages/slice-machine/src/features/navigation/Navigation.tsx
BohdanOne: Here the problem is that fallback might not make much sense since we display this element only when update is available, so not that often - not sure if it's good idea to delay loading of whole navbar until we have information should we display that information or not. **update**: confirmed with Come, no fallback here

packages/slice-machine/src/features/navigation/Navigation.tsx
BohdanOne: also here, fallback will be tricky as we can end up with 3 different things on the screen when data is loaded:
- display nothing - depending on the feature flag
- display the guide
- display only button to open the guide
**update**: confirmed with Come, no fallback here

packages/slice-machine/src/features/navigation/Navigation.tsx
BohdanOne: added skeleton loader as a fallback for this element (confirmed with Come)

dani-mp (approved):
Thanks for changes!

jomifepe (approved):
Aside from what Dani pointed out, looks good to me ✅

packages/slice-machine/src/features/navigation/NavigationItem.tsx
```diff
@@ -0,0 +1,80 @@
...
+type NavigationItemProps = NavigationLinkItemProps | NavigationButtonItemProps;
+
+export function NavigationItem(props: NavigationItemProps) {
+  const { title, href, target, active, Icon, RightElement, onClick } = props;
+
+  const isCollapsed = useMediaQuery({ max: "medium" });
+
+  const ItemIcon = useMemo(() => {
+    return () => <Icon width={isCollapsed ? 28 : 32} />;
+  }, [Icon, isCollapsed]);
```
jomifepe: We really need to suggest fixing this on editor-ui 😅 

Nit: Usually functions/components defined at render time use a `useCallback`.

```suggestion
  const ItemIcon = useCallback(() => {
    return () => <Icon width={isCollapsed ? 28 : 32} />;
  }, [Icon, isCollapsed]);
```

BohdanOne: changed, thanks!

PR #1470, opened 2024-10-28
[DT-2333] chore: remove pascal case error tracking event

BohdanOne:
**Resolves**: [DT-2333](https://linear.app/prismic/issue/DT-2333/remove-pascalcase-error-tracking)

### Description

The tracking event was added to gain knowledge about the scale of a problem with a PascalCase Error. 
All necessary data was gathered so the event is no longer needed.

### Checklist

### How to QA [^1]

[^1]:


dani-mp (approved):
Let's make sure to remove it from the segment schema as well for both dev and prod.

PR #1471, opened 2024-10-29
fix(adapter-next): async access to `params` and `searchParams` in Next.js 15

angeloashmore:
**Resolves**: https://community.prismic.io/t/nextjs-15-and-draftmode/17855

### Description

This PR updates `@slicemachine/adapter-next` to support async access to a page's `params` and `searchParams` props. The props are now wrapped in a `Promise` [as of Next.js 15](https://nextjs.org/blog/next-15#async-request-apis-breaking-change).

This change works in Next.js 13-15 with the App Router. The `Promise` wrappers are technically incorrect in Next.js 13 and 14, but it shouldn't break projects.

Projects using the App Router should be updating to the latest Next.js version as regular maintenance, reducing the effect of a slightly incorrect type.

### Checklist

### Preview

N/A

### How to QA [^1]

1. Set up:

   ```sh
   npx create-next-app@latest
   cd my-app
   npx @slicemachine/init@alpha.aa-nextjs-15
   ```

1. Add a slice.

1. Add a page and its code.

1. Confirm slice simulator works in the repo (testing the changes to `searchParams`).

1. Confirm pages render correctly (testing changes to `params`).

1. Confirm page previews work (testing previews don't break any of the above).

[^1]:

### DIFF
--- a/packages/adapter-next/src/hooks/documentation-read.ts
+++ b/packages/adapter-next/src/hooks/documentation-read.ts
@@ -41,10 +41,10 @@ export const documentationRead: DocumentationReadHook<PluginOptions> = async (
 
 					type Params = { uid: string };
 
-					export default async function Page({ params }: { params: Params }) {
+					export default async function Page({ params }: { params: Promise<Params> }) {
 						const client = createClient();
 						const page = await client
-							.getByUID("${model.id}", params.uid)
+							.getByUID("${model.id}", (await params).uid)
 							.catch(() => notFound());
 
 						return <SliceZone slices={page.data.slices} components={components} />;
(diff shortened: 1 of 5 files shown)


levimykel (approved):
Looks good to me. I QA'd it as well, and everything worked as expected. Nice work 🚀

dani-mp (commented):
I'd move all the awaits to their own lines, to make things less cluttered.

No need to block the PR because of this, as it fixes things for our users, we can merge and apply this change in a follow-up.

jomifepe (approved):
🔥

angeloashmore:
@dani-mp I moved the `await`s to individual lines, as seen in Next.js 15's upgrade guide:

```tsx
// Source: https://nextjs.org/docs/app/api-reference/file-conventions/page#displaying-content-based-on-params

export default async function Page({
  params,
}: {
  params: Promise<{ slug: string }>
}) {
  const { slug } = await params
  return <h1>Blog Post: {slug}</h1>
}
```

The docs don't seem to have a preferred way of `await`ing the props. The following examples are used throughout the docs:

```ts
const { foo } = await searchParams
const foo = (await searchParams).foo
(await searchParams).foo // Used inline
function Page(props) {
  const searchParams = await props.searchParams
}
```

The first one seems the cleanest and most likely to match users' existing code.

PR #1472, opened 2024-10-29
chore: prettier ignore playgrounds

jomifepe:
**Resolves**: 

### Description

Playgrounds tend to grow, and having prettier running over them takes a "long" time

### How to QA [^1]

[^1]:

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -20,3 +20,4 @@ e2e-projects/**
 dist
 packages/init/playground
 coverage/
+playgrounds


dani-mp (changes requested):
via slack, we agreed that we only want to continue with this change if we confirm that these files are never meant to be touched manually by a dev

if a dev has to read/touch them, they should be prettified

if these are just machine-generated artifacts opaque to the dev, we can ignore them

jomifepe:
Closing this, not critical and after some discussion, some people may want to edit playground and have them formatted.

PR #1473, opened 2024-10-30
[DT-2405] Support repeatable links

jomifepe:
Resolves: DT-2405

- [Repeatable links project](https://www.notion.so/prismic/Repeatable-Links-Work-1213bee086698021a7f9d5f20e20a47b)

### Description

Includes the necessary UI to mark a `Link` and `LinkToMedia` as repeatable.

### Preview

Screenshot 2024-10-30 at 12 42 56

Screenshot 2024-10-30 at 17 14 22

### How to QA [^1]

[^1]:


packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
```diff
@@ -6,12 +6,11 @@ import { createFieldNameFromKey } from "@/legacy/lib/forms";
 import { DefaultFields } from "@/legacy/lib/forms/defaults";
 import { CheckBox } from "@/legacy/lib/forms/fields";
 
-import { DisplayTextCheckbox } from "./components";
+import { DisplayTextCheckbox, RepeatableCheckbox } from "./components";
 
 // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 const FormFields = {
   ...DefaultFields,
-  allowTargetBlank: CheckBox("Allow target blank", false, true),
```
jomifepe, line 14: @levimykel Correct me if I'm wrong, but this is not needed right? Since we're not mapping over `FormFields` below to render `allowTargetBlank`, we can just set the default [here](https://github.com/prismicio/slice-machine/blob/ccffe983f91b1b66cddb12952bf1b372bbd64fe6/packages/slice-machine/src/legacy/lib/models/common/widgets/Link/index.ts#L109) and remove this.

package.json
```diff
@@ -69,7 +69,7 @@
   "resolutions": {
     "connected-next-router/react-redux": "8.0.7",
     "react-beautiful-dnd/react-redux": "8.0.7",
-    "@prismicio/types-internal": "2.8.0",
+    "@prismicio/types-internal": "2.9.0-alpha.0",
```
jomifepe: To be applied when package is released

```suggestion
    "@prismicio/types-internal": "2.9.0",
```

packages/slice-machine/package.json
```diff
@@ -49,7 +49,7 @@
     "@prismicio/mock": "0.3.3",
     "@prismicio/mocks": "2.4.0",
     "@prismicio/simulator": "0.1.4",
-    "@prismicio/types-internal": "2.8.0",
+    "@prismicio/types-internal": "2.9.0-alpha.0",
```
jomifepe: To be applied when package is released

```suggestion
    "@prismicio/types-internal": "2.9.0",
```

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
```diff
@@ -20,27 +18,24 @@ const Form = (props) => {
...
+            fieldName={createFieldNameFromKey(key)}
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+            formField={field}
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+            fields={fields}
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+            initialValues={initialValues}
+          />
+        </Col>
+      ))}
```
jomifepe: Diff context: Just removed a `eslint-disable`

packages/slice-machine/src/legacy/lib/models/common/widgets/LinkToMedia/Form.jsx
```diff
@@ -16,33 +15,36 @@ const Form = (props) => {
...
+          <WidgetFormField
+            fieldName={createFieldNameFromKey(key)}
+            formField={field}
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+            fields={fields}
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+            initialValues={initialValues}
+          />
+        </Col>
+      ))}
```
jomifepe: Diff context: Just removed a `eslint-disable`

dani-mp (approved):

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
levimykel: @jomifepe I don't remember there being a specific need for this to be there. As long as everything is working as expected, then it seems that this isn't needed.

packages/slice-machine/src/legacy/lib/models/common/widgets/ContentRelationship/index.ts
```diff
@@ -52,6 +52,7 @@ export const ContentRelationshipWidget: Widget<Link, typeof schema> = {
     config: {
       label,
       select: "document",
+      repeat: false,
```
jomifepe, line 55: Still trying to understand why this is needed to make [this test](https://github.com/prismicio/slice-machine/blob/main/packages/slice-machine/test/lib/models/common/widgets/index.test.ts) pass

dani-mp: which of those tests is failing? what does the error message say? can you reproduce it locally?

jomifepe: Screenshot 2024-11-04 at 10 18 46

jomifepe: I see now that it's because of `withFallback` [here](https://github.com/prismicio/prismic-types-internal/pull/89/files#diff-b7285001e1fb0c3aeffcc649ccae5c113c6c0755100b038b91e2c9937dd04231R82)

jomifepe: Ok so since the `Link` has a fallback for the `repeat` property, and the `ContentRelationship` widget is just a `Link`, I will keep this default value in the config

PR #1474, opened 2024-11-04
[DT-2939] fix: reordering fields on repeatable pages

BohdanOne:
**Resolves**: [DT-2939](https://linear.app/prismic/issue/DT-2393/bug-aauser-i-can-reorder-fields-in-slice-machine)

### Description

Users are encountering [issue](https://community.prismic.io/t/drag-drop-for-ordering-fields-is-broken-slicemachine/17848) when trying to drag'n'drop fields in repeatable page.

The root cause of the bug was UID field that was moved away from the list of the fields (and now is displayed at the top of the TypeBuilder), but is still present in the array of fields.

The fix was to omit UID field while reordering fields and keep it at the first position.

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/domain/customType.ts
+++ b/packages/slice-machine/src/domain/customType.ts
@@ -430,8 +430,21 @@ export function reorderField(args: ReorderFieldArgs): CustomType {
     ),
   );
 
+  // On repetable pages UID field shouldn't be reordered
+  const isRepeatablePage =
+    customType.format === "page" && customType.repeatable;
+
+  // Remove UID field from the fields to be reordered
+  const sectionFieldsWithoutUid = Object.fromEntries(
+    Object.entries(sectionFields).filter(([_, value]) => value.type !== "UID"),
+  );
+
+  const fieldsToReorder = isRepeatablePage
+    ? sectionFieldsWithoutUid
+    : sectionFields;
+
   const updatedSection = reorderFields({
-    fields: sectionFields,
+    fields: fieldsToReorder,
     sourceIndex,
     destinationIndex,
   });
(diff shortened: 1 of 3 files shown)


packages/slice-machine/src/domain/customType.ts
```diff
@@ -442,10 +455,20 @@ export function reorderField(args: ReorderFieldArgs): CustomType {
     updatedSection[sliceZoneKey] = sliceZoneField;
   }
 
+  // Put the UID field back at the begginnig of the reordered fields
```
xrutayisire: 💡 
```suggestion
  // Put the UID field back at the beginning of the reordered fields
```

packages/slice-machine/src/domain/customType.ts
```diff
@@ -430,8 +430,21 @@ export function reorderField(args: ReorderFieldArgs): CustomType {
     ),
   );
 
+  // On repetable pages UID field shouldn't be reordered
```
xrutayisire: 💡
```suggestion
  // On repeatable pages UID field shouldn't be reordered
```

packages/slice-machine/src/domain/customType.ts
```diff
@@ -537,8 +560,8 @@ export function reorderFields<T>(args: ReorderFieldsArgs<T>) {
   const fieldEntries = Object.entries(fields);
   const [removedEntry] = fieldEntries.splice(sourceIndex, 1);
   fieldEntries.splice(destinationIndex, 0, removedEntry);
-  const reorderedFields = Object.fromEntries(fieldEntries);
 
+  const reorderedFields = Object.fromEntries(fieldEntries);
```
xrutayisire, line 564: 💡 Most of the time we recommend having a blank line before return, for example: https://eslint.org/docs/latest/rules/newline-before-return

xrutayisire (approved):
Some minor comments, LGTM!

packages/slice-machine/src/domain/customType.ts
BohdanOne: fixed

packages/slice-machine/src/domain/customType.ts
BohdanOne: fixed

PR #1475, opened 2024-11-04
Refetch field snippet when field config changes

dani-mp:
**Resolves**: 
https://linear.app/prismic/issue/DT-2433/aauser-when-i-edit-a-link-field-i-can-see-the-code-snippet-change

### Description

Use a dynamic key based on the config of some fields to make useSWR refetch the field snippet to show. More as comments in the PR.

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/components/Hints/index.tsx
+++ b/packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/components/Hints/index.tsx
@@ -20,7 +20,7 @@ const Hint: React.FC<HintProps> = ({
 }) => {
   const fieldPathString = renderHintBase({ item });
 
-  const snippetCacheKey = [fieldPathString];
+  const snippetCacheKey = [item.value.type];
   if (item.value.type === "Link") {
     if (item.value.config?.allowText ?? false)
       snippetCacheKey.push("allowText");
--- a/packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/components/Hints/index.tsx
+++ b/packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/components/Hints/index.tsx
@@ -20,7 +20,7 @@ const Hint: React.FC<HintProps> = ({
 }) => {
   const fieldPathString = renderHintBase({ item });
 
-  const snippetCacheKey = [item.value.type];
+  const snippetCacheKey: string[] = [item.value.type];
   if (item.value.type === "Link") {
     if (item.value.config?.allowText ?? false)
       snippetCacheKey.push("allowText");
--- a/packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/components/Hints/index.tsx
+++ b/packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/components/Hints/index.tsx
@@ -20,7 +20,7 @@ const Hint: React.FC<HintProps> = ({
 }) => {
   const fieldPathString = renderHintBase({ item });
 
-  const snippetCacheKey: string[] = [item.value.type];
+  const snippetCacheKey = [fieldPathString];
   if (item.value.type === "Link") {
     if (item.value.config?.allowText ?? false)
       snippetCacheKey.push("allowText");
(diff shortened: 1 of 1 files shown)


packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/components/Hints/index.tsx
```diff
@@ -20,9 +20,15 @@ const Hint: React.FC<HintProps> = ({
 }) => {
   const fieldPathString = renderHintBase({ item });
 
-  // TODO: Call `swr`'s global `mutate` function when something changes to clear the cache.
```
dani-mp, line 23: I also tried this solution, but I didn't like it. We have at least two places where we'd have to make this call, making sure the key is the same in both places, and it wouldn't be declarative.

packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/components/Hints/index.tsx
```diff
@@ -20,9 +20,15 @@ const Hint: React.FC<HintProps> = ({
...
+  const snippetCacheKey = [fieldPathString];
+  if (item.value.type === "Link") {
+    if (item.value.config?.allowText ?? false)
+      snippetCacheKey.push("allowText");
+    if (item.value.config?.repeat ?? false) snippetCacheKey.push("repeat");
```
dani-mp:
```suggestion
```

packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/components/Hints/index.tsx
```diff
@@ -20,9 +20,14 @@ const Hint: React.FC<HintProps> = ({
 }) => {
   const fieldPathString = renderHintBase({ item });
 
-  // TODO: Call `swr`'s global `mutate` function when something changes to clear the cache.
+  const snippetCacheKey = [fieldPathString];
+  if (item.value.type === "Link") {
+    if (item.value.config?.allowText ?? false)
+      snippetCacheKey.push("allowText");
```
dani-mp, line 26: Here we'll add also the new `repeat` config so it also reacts to it.

packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/components/Hints/index.tsx
```diff
@@ -20,9 +20,14 @@ const Hint: React.FC<HintProps> = ({
 }) => {
   const fieldPathString = renderHintBase({ item });
 
-  // TODO: Call `swr`'s global `mutate` function when something changes to clear the cache.
+  const snippetCacheKey = [fieldPathString];
```
dani-mp, line 23:
```suggestion
  const snippetCacheKey = [item.value.type];
```

packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/components/Hints/index.tsx
```diff
@@ -20,9 +20,14 @@ const Hint: React.FC<HintProps> = ({
 }) => {
   const fieldPathString = renderHintBase({ item });
 
-  // TODO: Call `swr`'s global `mutate` function when something changes to clear the cache.
+  const snippetCacheKey = [item.value.type];
```
dani-mp:
```suggestion
  const snippetCacheKey: string[] = [item.value.type];
```

packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/components/Hints/index.tsx
```diff
@@ -20,9 +20,14 @@ const Hint: React.FC<HintProps> = ({
 }) => {
   const fieldPathString = renderHintBase({ item });
 
-  // TODO: Call `swr`'s global `mutate` function when something changes to clear the cache.
+  const snippetCacheKey: string[] = [item.value.type];
```
dani-mp:
```suggestion
  const snippetCacheKey = [fieldPathString];
```

jomifepe (approved):
LGTM ✅

packages/slice-machine/src/legacy/lib/builders/common/Zone/Card/components/Hints/index.tsx
jomifepe, line 26: Nit 😇

```suggestion
    if (item.value.config?.allowText ?? false) {
      snippetCacheKey.push("allowText");
    }
```

PR #1476, opened 2024-11-05
fix: stop using internal domain

jomifepe:
https://prismic-team.slack.com/archives/C0441S8SKQR/p1730800413994889


xrutayisire:
For me, we don't need to revert our change, as discussed in Slack

PR #1486, opened 2024-11-20
feat: add method to set default master locale

BohdanOne:
**Resolves**: [DT-2481](https://linear.app/prismic/issue/DT-2481/set-master-language-for-a-repository-via-manager-package)

### Description

When a user creates a repository through the terminal (without starter), it should set a default master lang (en-us) for the repository.
In order to do that there should be a method in manager package sending request to an obelix locale service.

### Checklist

### How to QA [^1]

[^1]:


packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
```diff
@@ -673,6 +673,29 @@ export class PrismicRepositoryManager extends BaseManager {
 		}
 	}
 
+	async setDefaultMasterLocale(): Promise<void> {
+		const repositoryName = await this.project.getRepositoryName();
+
+		const localeServiceBaseUrl = new URL("locale/", API_ENDPOINTS.PrismicWroom);
```
xrutayisire: ⚠️ Following this message from Michel, could you ensure you're using the right domain?

https://prismic-team.slack.com/archives/C039VR18ZDY/p1731667211291229

If we contact an internal service, as you see, there is a specific format instead of directly contacting wroom base url.

```
Format: [api.internal.prismic.io/](http://api.internal.prismic.io/)<service>
```

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
```diff
@@ -673,6 +673,29 @@ export class PrismicRepositoryManager extends BaseManager {
...
+		const url = new URL("repository/locales", localeServiceBaseUrl);
+		url.searchParams.set("repository", repositoryName);
+
+		const res = await this._fetch({
+			url,
+			method: "POST",
+			body: {
+				id: "en-us",
+				isMaster: true,
+				label: "English - United States",
```
xrutayisire: ❓ I don't find where did you see a `label` as body param of this call?

xrutayisire (changes requested):

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
BohdanOne: Yes, you're right, it's not needed.
It's removed.

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
BohdanOne: url updated

packages/manager/src/constants/API_ENDPOINTS.ts
```diff
@@ -37,6 +38,7 @@ export const API_ENDPOINTS: APIEndpoints = (() => {
 				RepositoryService: addTrailingSlash(
 					process.env.repository_api ?? "https://repository.internal.wroom.io/",
 				),
+				LocaleService: addTrailingSlash("https://locale.internal.wroom.io/"),
```
xrutayisire: ⚠️ Here it's the dev mode, and it should take env variable, you can have a fallback as done just above for repository service, but when running in dev mode with all other env variables it would not work with wroom.io.

And so at the same time you need to ensure that locale service env variable is available in the dev env: you can ask a question in Obelix why it was never done. 
See an example how it was done for repository service: https://github.com/prismicio/prismicdev-cli/pull/59

packages/manager/src/constants/API_ENDPOINTS.ts
```diff
@@ -102,6 +105,7 @@ If you didn't intend to run Slice Machine this way, stop it immediately and unse
 				PrismicUnsplash: "https://unsplash.prismic.io/",
 				SliceMachineV1: "https://sm-api.prismic.io/v1/",
 				RepositoryService: "https://repository.prismic.io/",
+				LocaleService: "https://locale.prismic.io/",
```
xrutayisire: ⚠️ Did you test that? I don't understand this related to Michel communication. It should be `"https://locale.internal.prismic.io/"`, correct?

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
```diff
@@ -673,6 +673,26 @@ export class PrismicRepositoryManager extends BaseManager {
...
+		const res = await this._fetch({
+			url,
+			method: "POST",
+			body: {
+				id: "en-us",
+				isMaster: true,
+			},
+		});
+
+		if (!res.ok) {
```
xrutayisire, line 691: ⚠️ You should take care of all potential status to make it clear when debugging.

```
                         case 400:
			case 401:
				throw new UnauthenticatedError();
			case 403:
				throw new UnauthorizedError();
```

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
```diff
@@ -673,6 +673,26 @@ export class PrismicRepositoryManager extends BaseManager {
...
+			throw new Error("Failed to set default master locale.");
```
xrutayisire: 💡 I think we should return also the cause:

```suggestion
		const text = await res.text();
		throw new Error("Failed to set default master locale.", { cause: text });
```

Not done for a lot of calls in the project I see, but it's better.

xrutayisire (changes requested):

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
xrutayisire: 💡 We should not return `master locale` to our users:

```suggestion
			throw new Error("Failed to set main content language.");
```

xrutayisire (changes requested):

packages/manager/src/constants/API_ENDPOINTS.ts
BohdanOne: I got a little bit confused with the repository url one line up that don't have the internal. added.
Just tested and it seems that both options (with and without internal) work at the moment.
Will add that (should I also add it for repository service?)

packages/manager/src/constants/API_ENDPOINTS.ts
BohdanOne: fixed, 
PR to add endpoint to dev env here: https://github.com/prismicio/prismicdev-cli/pull/60

packages/manager/src/constants/API_ENDPOINTS.ts
BohdanOne: fixed

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
BohdanOne: done

packages/manager/src/managers/prismicRepository/PrismicRepositoryManager.ts
BohdanOne: done

BohdanOne: done

packages/slice-machine/test/src/modules/__fixtures__/serverState.ts
```diff
@@ -25,6 +25,7 @@ export const dummyServerState: Pick<
       PrismicUnsplash: "https://unsplash.prismic.io/",
       SliceMachineV1: "https://sm-api.prismic.io/v1/",
       RepositoryService: "https://repository.prismic.io/",
+      LocaleService: "https://locale.prismic.io/",
```
jomifepe: ⚠️ Should it be `locale.internal.prismic.io` here also?

```suggestion
      LocaleService: "https://locale.internal.prismic.io/",
```

BohdanOne: yes, I think so, thanks!

BohdanOne: updated

packages/manager/src/constants/API_ENDPOINTS.ts
BohdanOne: urls updated

xrutayisire (approved):
LGTM! 🚀

packages/init/src/SliceMachineInitProcess.ts
```diff
@@ -965,6 +966,42 @@ ${chalk.cyan("?")} Your Prismic repository name`.replace("\n", ""),
...
+					task.title = `Main content language set to ${chalk.cyan(
+						"English - United States",
+					)} 🇺🇸. You can change it anytime in your project settings.`;
+
+					try {
+						const { value: onboardingExperimentVariant } =
+							(await this.manager.telemetry.getExperimentVariant(
+								"shared-onboarding",
+							)) ?? {};
+						if (onboardingExperimentVariant === "with-shared-onboarding") {
```
jomifepe: 💡 Nit: Can you extract this to a `private` function, since we already have it for `createNewRepository`, something like:

```ts
const isSharedOnboardingEnabled = await checkSharedOnboardingExperimentEnabled();
if (isSharedOnboardingEnabled) {
```

jomifepe (approved):

packages/init/src/SliceMachineInitProcess.ts
BohdanOne: good idea! will do

BohdanOne: done

PR #1487, opened 2024-11-20
feat: set up default master locale

BohdanOne:
**Resolves**: [DT-2482](https://linear.app/prismic/issue/DT-2482/set-default-master-language-when-creating-new-repo-via-cli)

### Description

When a user creates a repository through the terminal (without starter), it should set a default master lang (en-us) for the repository.

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/init/src/SliceMachineInitProcess.ts
+++ b/packages/init/src/SliceMachineInitProcess.ts
@@ -184,6 +184,9 @@ export class SliceMachineInitProcess {
 			}
 			if (!this.context.repository.exists) {
 				await this.createNewRepository();
+				if (!this.options.starter) {
+					await this.setDefaultMasterLocale();
+				}
 			}
 
 			await this.syncDataWithPrismic();
@@ -960,6 +963,20 @@ ${chalk.cyan("?")} Your Prismic repository name`.replace("\n", ""),
 					task.title = `Created new repository ${chalk.cyan(
 						this.context.repository.domain,
 					)}`;
+				},
+			},
+		]);
+	}
+
+	protected setDefaultMasterLocale(): Promise<void> {
+		return listrRun([
+			{
+				title: `Setting default master language...`,
+				task: async (_, task) => {
+					await this.manager.prismicRepository.setDefaultMasterLocale();
+					task.title = `Default master language set to ${chalk.cyan(
+						"en-US",
+					)} 🇺🇸. You can change it anytime in your project settings.`;
 				},
 			},
 		]);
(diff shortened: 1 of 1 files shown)


packages/init/src/SliceMachineInitProcess.ts
```diff
@@ -184,6 +184,9 @@ export class SliceMachineInitProcess {
 			}
 			if (!this.context.repository.exists) {
 				await this.createNewRepository();
+				if (!this.options.starter) {
```
xrutayisire: ⚠️ Technically, this is wrong, and we can and will still have the bug. You assume here you start with a starter with the command the UI gave you, but the bug happens actually when you don't start with the UI. 

Imagine you start for here: https://github.com/prismicio-community/nextjs-starter-prismic-minimal/tree/master
Or from here: https://prismic.io/docs/nextjs-example-projects

As you see, there are no starter option. The condition to check if you need to setup the master locale is if there are documents to push. If you have documents to push, it's what we consider a starter today.

packages/init/src/SliceMachineInitProcess.ts
```diff
@@ -965,6 +968,20 @@ ${chalk.cyan("?")} Your Prismic repository name`.replace("\n", ""),
 		]);
 	}
 
+	protected setDefaultMasterLocale(): Promise<void> {
+		return listrRun([
+			{
+				title: `Setting default master language...`,
```
xrutayisire: 💡 I don't think we are talking about `master` language in the product, no?

See the title of the PB onboarding step: `What's your main content language?` 

I would do something like:

```suggestion
				title: `Setting main content language...`,
```

And also:

```
task.title = `Main content language set to ${chalk.cyan(
  "English - United States",
)} 🇺🇸. You can change it anytime in your project settings.`;
```

I would also use the label, as in the select in PB.

xrutayisire (changes requested):

packages/init/src/SliceMachineInitProcess.ts
BohdanOne: Ok, will confirm the change of en-Us to the label with Come

packages/init/src/SliceMachineInitProcess.ts
BohdanOne: Ok, thanks for pointing that out!
Changed

packages/init/src/SliceMachineInitProcess.ts
BohdanOne: message change confirmed with Come, updated

packages/init/src/SliceMachineInitProcess.ts
xrutayisire: Hum so you decide to not check if there are documents to push?

BohdanOne: ah, ok. I misunderstood your suggestion. Going back to this.

xrutayisire (approved):
LGTM!

PR #1488, opened 2024-11-21
feat: onboarding step of master locale setting done when creating new repo with CLI

BohdanOne:
**Resolves**: [DT-2483](https://linear.app/prismic/issue/DT-2483/onboarding-step-of-master-locale-setting-done-when-creating-new-repo)

### Description
When a user creates a repository through the terminal, the master locale is already set, so the onboarding guide step of choosing master locale should be marked as done.

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/init/src/SliceMachineInitProcess.ts
+++ b/packages/init/src/SliceMachineInitProcess.ts
@@ -977,6 +977,20 @@ ${chalk.cyan("?")} Your Prismic repository name`.replace("\n", ""),
 					task.title = `Default master language set to ${chalk.cyan(
 						"en-US",
 					)} 🇺🇸. You can change it anytime in your project settings.`;
+
+					try {
+						const { value: onboardingExperimentVariant } =
+							(await this.manager.telemetry.getExperimentVariant(
+								"shared-onboarding",
+							)) ?? {};
+						if (onboardingExperimentVariant === "with-shared-onboarding") {
+							this.manager.prismicRepository.completeOnboardingStep(
+								"chooseLocale",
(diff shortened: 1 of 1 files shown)


packages/init/src/SliceMachineInitProcess.ts
```diff
@@ -977,6 +977,20 @@ ${chalk.cyan("?")} Your Prismic repository name`.replace("\n", ""),
 					task.title = `Default master language set to ${chalk.cyan(
 						"en-US",
 					)} 🇺🇸. You can change it anytime in your project settings.`;
+
+					try {
```
xrutayisire, line 979: ⚠️ This code is at the wrong place, right? You're really focusing on your flow but for starter we also want to set that this as been done right?

Best place would be with the other `completeOnboardingStep` call of this file, no? So you directly set all steps as done directly together.

xrutayisire (changes requested):

packages/init/src/SliceMachineInitProcess.ts
BohdanOne: isn't [that change](https://github.com/prismicio/slice-machine/pull/1487/commits/d2cc45f76ef85eb887d233e9382921c51570d079) making this change not needed anymore?
now we will set the language for every repo created with CLI, right?
Then in my opinion it's better to complete the step when it's actually being complete, so we don't complete it if `setDefaultMasterLocale` fail

xrutayisire (approved):
LGTM!

PR #1492, opened 2024-11-27
feat(tracking): track repeat property on link field added

BohdanOne:
**Resolves**: [DT-2429](https://linear.app/prismic/issue/DT-2429/aapm-i-can-see-a-property-to-the-tracking-event-slice-machine-field)

### Description

Added tracking for a `repeat` property of a new link field added.
Also moved that tracking into separate method to avoid duplications.

### Checklist

### How to QA [^1]

[^1]:


packages/slice-machine/src/utils/tracking/trackFieldAdded.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { telemetry } from "@/apiClient";
+import { SlicePrimaryFieldSM } from "@/legacy/lib/models/common/Slice";
+
+import { getContentTypeForTracking } from "./getContentTypeForTracking";
+
+export function trackFieldAdded(
+  id: string,
+  field: SlicePrimaryFieldSM | NestableWidget | UID | Group,
+) {
```
dani-mp: let's use one param only with an object, rather than positional arguments

packages/slice-machine/src/utils/tracking/trackFieldAdded.ts
```diff
@@ -0,0 +1,28 @@
...
+  void telemetry.track({
+    event: "field:added",
+    id,
+    name: field.config?.label ?? "",
+    type: field.type,
+    isInAGroup: false,
```
xrutayisire: ⚠️ Hum, here you're just removing a tracking feature, it should be true when calling from inside a group.

xrutayisire (changes requested):
I saw only one problem, rest is fine for me.

packages/slice-machine/src/utils/tracking/trackFieldAdded.ts
BohdanOne: yeah, I wasn't sure about that, in all 3 places that I replace with new method it was hard coded to false, thought there is some reason for that 😅 will fix that then

dani-mp: i guess it was never relly implemented? lol

curious why product never complained about this, then

packages/slice-machine/src/utils/tracking/trackFieldAdded.ts
BohdanOne: changed

packages/slice-machine/src/utils/tracking/trackFieldAdded.ts
xrutayisire: It was not:

Old code:
```
void telemetry.track({
      event: "field:added",
      id,
      name: label,
      type: newField.type,
      isInAGroup: true,
      contentType: getContentTypeForTracking(window.location.pathname),
      ...(newField.type === "Link" && {
        allowText: newField.config?.allowText,
      }),
    });
```

BohdanOne: fixed

jomifepe (approved):

packages/slice-machine/src/utils/tracking/trackFieldAdded.ts
```diff
@@ -0,0 +1,33 @@
...
+  NestableWidget,
+  UID,
+} from "@prismicio/types-internal/lib/customtypes";
+
+import { telemetry } from "@/apiClient";
+import { SlicePrimaryFieldSM } from "@/legacy/lib/models/common/Slice";
+
+import { getContentTypeForTracking } from "./getContentTypeForTracking";
+
+type TrackFieldAddedParam = {
```
xrutayisire: 💡 Nit: `TrackFieldAddedArgs`

xrutayisire (approved):
Small nit for you but LGTM

packages/slice-machine/src/utils/tracking/trackFieldAdded.ts
BohdanOne: true, changed

PR #1493, opened 2024-11-29
feat: track groups pushed

BohdanOne:
**Resolves**: [DT-2439](https://linear.app/prismic/issue/DT-2439/aapm-i-can-track-when-a-user-pushed-a-group)

### Description

In order to gain more knowledge about the usage of the groups there is tracking added to get the data of the groups in new custom types or slices.

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/features/sync/actions/trackPushChangesSuccess.ts
+++ b/packages/slice-machine/src/features/sync/actions/trackPushChangesSuccess.ts
@@ -1,7 +1,12 @@
-import { FieldType } from "@prismicio/types-internal/lib/customtypes";
+import {
+  FieldType,
+  NestableWidget,
+  UID,
+} from "@prismicio/types-internal/lib/customtypes";
 
 import { telemetry } from "@/apiClient";
 import { countMissingScreenshots } from "@/domain/slice";
+import { GroupSM } from "@/legacy/lib/models/common/Group";
 import {
   ChangedCustomType,
   ChangedSlice,
@@ -120,20 +125,8 @@ function trackPushedGroups(args: TrackPushedGroupsArgs) {
       if (customType.status !== ModelStatus.New) return acc;
 
       customType.customType.tabs.forEach((tab) => {
-        tab.value.forEach((field) => {
-          if (field.value.type === "Group" && field.value.config?.fields) {
-            const fieldsCount: FieldsCount = {};
-            field.value.config.fields.forEach(({ value: fieldValue }) => {
-              const value = fieldsCount[fieldValue.type] ?? 0;
-              fieldsCount[fieldValue.type] = value + 1;
-            });
-            acc.push({
-              isInStaticZone: true,
-              isInSlice: false,
-              ...fieldsCount,
-            });
-          }
-        });
+        const fieldsCount = countGroupFields(acc, tab.value, "custom-type");
+        acc.push(...fieldsCount);
       });
 
       return acc;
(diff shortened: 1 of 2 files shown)


packages/slice-machine/src/features/sync/actions/trackPushChangesSuccess.ts
```diff
@@ -94,4 +96,82 @@ export function trackPushChangesSuccess(args: TrackPushChangesSuccessArgs) {
...
+            };
+            field.value.config.fields.forEach((field) => {
+              if (typeof fieldStats[field.value.type] !== "number") {
+                fieldStats[field.value.type] = 1;
+              } else {
+                fieldStats[field.value.type]! += 1;
+              }
+            });
+            acc.push(fieldStats);
+          }
```
jomifepe: 💡 This logic seems to be the same as the one below, so it could be extracted to a function.

jomifepe: 💡 It could also be simplified to something like this, to avoid non-null assertions

```suggestion
          if (field.value.type === "Group" && field.value.config?.fields) {
            const fieldCount: { [key in FieldType]?: number } = {};
            field.value.config.fields.forEach(({ value: fieldValue }) => {
              const value = fieldCount[fieldValue.type] ?? 0;
              fieldCount[fieldValue.type] = value + 1;
            });
            acc.push({ isInStaticZone: false, isInSlice: true, ...fieldCount });
          }
```

BohdanOne: Nice, I'll try that, was already trying to extract that duplicated part but it got super confusing with types very quickly, but will try another attempt.

BohdanOne: I added your suggestion, but didn't extract repeated logic to another function as correct typing for that function would be very convoluted.

jomifepe (approved):

packages/slice-machine/src/features/sync/actions/trackPushChangesSuccess.ts
jomifepe: Would be something like this:

```ts
customType.customType.tabs.forEach((tab) => {
  const fieldsCount = countGroupFields(acc, tab.value);
  acc.push(...fieldsCount);
});

// ...

slice.slice.model.variations.forEach((variation) => {
  const fieldsCount = countGroupFields(acc, variation.primary);
  acc.push(...fieldsCount);
});


// ...

interface Field {
  key: string;
  value: GroupSM | UID | NestableWidget;
}

function countGroupFields<TField extends Field>(
  fieldStats: FieldStats[],
  fields: TField[] | undefined,
) {
  if (!fields) return fieldStats;

  const newFieldState = fieldStats.slice();
  fields.forEach((field) => {
    if (field.value.type === "Group" && field.value.config?.fields) {
      const fieldsCount: FieldsCount = {};
      field.value.config.fields.forEach(({ value: fieldValue }) => {
        const value = fieldsCount[fieldValue.type] ?? 0;
        fieldsCount[fieldValue.type] = value + 1;
      });
      newFieldState.push({
        isInStaticZone: false,
        isInSlice: true,
        ...fieldsCount,
      });
    }
  });

  return newFieldState;
}
```

BohdanOne: nice! applied your suggestion!

jomifepe (approved):

PR #1502, opened 2024-12-12
Add link variant box

dani-mp:
**Resolves**: 
Part of https://linear.app/prismic/issue/DT-2474/aadev-i-can-create-a-link-field-with-the-new-option-to-enable-and.

### Description

Adds the box introducing the new feature, without any functionality.

### Checklist

### Preview

image

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
+++ b/packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
@@ -1,4 +1,4 @@
-import { Box, Icon, Text, Tooltip } from "@prismicio/editor-ui";
+import { Box, Icon, Switch, Text, Tooltip } from "@prismicio/editor-ui";
 import { Checkbox, Flex, Label } from "theme-ui";
 
 import { Col } from "@/legacy/components/Flex";
@@ -57,13 +57,7 @@ export function RepeatableCheckbox(props: CommonCheckboxProps) {
 
   return (
     <Box flexDirection="column">
-      <Label
-        htmlFor="repeat"
-        variant="label.primary"
-        sx={{
-          mt: 2,
-        }}
-      >
+      <Label htmlFor="repeat" variant="label.primary">
         Repeatable
       </Label>
       <Label variant="label.border" sx={{ display: "flex" }}>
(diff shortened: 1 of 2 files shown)


packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
```diff
@@ -95,3 +89,41 @@ export function RepeatableCheckbox(props: CommonCheckboxProps) {
...
+        <Text variant="h4" color="grey12">
+          Variants
+        </Text>
+        <Text color="grey11">
+          Add variants like "Primary" and "Secondary" to allow editors to choose
+          the link's style by selecting one of them.
+        </Text>
+        <Box gap={8}>
+          <Switch />
+          <Text color="grey11">Disabled</Text>
```
xrutayisire, line 111: ⚠️ If I get the Figma, I think the text should change depending on whether it's checked or not.

If your plan was to do it later on, forget my comment.

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
```diff
@@ -95,3 +89,41 @@ export function RepeatableCheckbox(props: CommonCheckboxProps) {
...
+        border={{ bottom: true }}
+        padding={12}
+        flexDirection="column"
+        gap={8}
+      >
+        <Text variant="h4" color="grey12">
+          Variants
+        </Text>
+        <Text color="grey11">
+          Add variants like "Primary" and "Secondary" to allow editors to choose
```
xrutayisire, line 106: 💡 I also prefer your new copy, I let you confirm it with Côme for both sentences and all good for me.

xrutayisire, line 106: ⚠️ For me I don't think it's the correct color corresponding to the Figma.

Note: At the same time, I couldn't find the correct grey value from Figma, seems weird and not coming from the DS. Maybe we could just took the darker version of this as visible in the design.

xrutayisire (approved):
Two ⚠️ comments, even if I don't think I need to revalidate after, as they are straightforward.

LGTM after my comments 🚀

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
dani-mp: yup, this was just putting the stuff there with no behavior

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
dani-mp: yeah, the color doesn't exist in the new design system, so i picked the closest one that looked right

>Maybe we could just took the darker version of this as visible in the design

not sure i understood this message, could you clarify?

xrutayisire: From the design, I feel the color should be darker. Maybe we could use one a darker to be closer to the design?
Since this one is the same as the others texts it's too close and don't match design.

dani-mp: i see what you mean now, i've used the darker color now, let mem know what you think!

PR #1504, opened 2024-12-13
Link variants switch behavior

dani-mp:
**Resolves**: 
Part of https://linear.app/prismic/issue/DT-2474/aadev-i-can-create-a-link-field-with-the-new-option-to-enable-and

### Description

Checking/unchecking the link variants toggle will add two variants ("Primary" and "Secondary") to the model, and update the toggle label.

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
+++ b/packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
@@ -90,7 +90,21 @@ export function RepeatableCheckbox(props: CommonCheckboxProps) {
   );
 }
 
-export function Variant() {
+export function Variants({
+  variants,
+  onVariantsChange,
+}: {
+  variants?: string[];
+  onVariantsChange: (variants?: string[]) => void;
+}) {
+  const enabled = Boolean(variants);
+
+  const onCheckedChange = (checked: boolean) => {
+    onVariantsChange(checked ? ["Primary", "Secondary"] : undefined);
+  };
+
+  const switchLabel = enabled ? "Enabled" : "Disabled";
+
   return (
     <Box overflow="hidden" flexDirection="column" border borderRadius={6}>
       <Box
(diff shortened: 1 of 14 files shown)


packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
```diff
@@ -90,7 +90,21 @@ export function RepeatableCheckbox(props: CommonCheckboxProps) {
...
+  variants,
+  onVariantsChange,
+}: {
+  variants?: string[];
+  onVariantsChange: (variants?: string[]) => void;
+}) {
+  const enabled = Boolean(variants);
+
+  const onCheckedChange = (checked: boolean) => {
+    onVariantsChange(checked ? ["Primary", "Secondary"] : undefined);
```
xrutayisire, line 103: 💡 Potential product problem, are we sure we want to reset the variants every time you check / uncheck?

I'm wondering because, let say you're a dev with 5 variants, click accidentally or "just to see" what happens, you will lose existing data. 

I would have said that `["Primary", "Secondary"]` would only be the default if you don't have already, no?

dani-mp: my opinion is that resetting is better, especially because adding them is not costly if you accidentally disable it

otherwise, you sometimes will see the default values, and sometimes you won't, and you won't necessarily know why

for instance, see this scenario:
1. enable (you see the default options)
2. change an option
3. disable
4. enable (you see the changed option)
5. disable
6. close the modal without clicking "done"
7. enable (you see the default options)

if we always reset, the behavior in those two scenarios would be the same (default options)

xrutayisire: I see, I like the consistency argument. Let's just mention that during our daily so Côme knows about it but should be good!

xrutayisire (approved):
LGTM!

PR #1506, opened 2024-12-13
Link variant options (1)

dani-mp:
**Resolves**: 
Part of https://linear.app/prismic/issue/DT-2474/aadev-i-can-create-a-link-field-with-the-new-option-to-enable-and

### Description

Shows the list of options and allows editing them, handling empty values errors.

### Checklist

### Preview

image

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
+++ b/packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
@@ -94,7 +94,7 @@ export default Form;
 function VariantsWrapper({ variants, setFieldValue }) {
   const fieldKey = "config.variants";
 
-  const onVariantsChange = (variants) => setFieldValue(fieldKey, variants);
+  const onVariantsChange = (newVariants) => setFieldValue(fieldKey, newVariants);
 
   const [_, meta] = useField(fieldKey);
 
--- a/packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
+++ b/packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
@@ -3,6 +3,7 @@
 /* eslint-disable @typescript-eslint/no-unsafe-return */
 
 import { Box } from "@prismicio/editor-ui";
+import { useField } from "formik";
 import { Flex } from "theme-ui";
 
 import { Col, Flex as FlexGrid } from "@/legacy/components/Flex";
@@ -81,12 +82,7 @@ const Form = (props) => {
           // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
           setFieldValue={setFieldValue}
         />
-        <Variants
-          variants={variants}
-          onVariantsChange={(variants) =>
-            setFieldValue("config.variants", variants)
-          }
-        />
+        <VariantsWrapper variants={variants} setFieldValue={setFieldValue} />
       </Box>
     </>
   );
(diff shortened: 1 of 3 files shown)


packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
```diff
@@ -81,16 +82,29 @@ const Form = (props) => {
...
   );
 };
 
 export { FormFields };
 export default Form;
+
+function VariantsWrapper({ variants, setFieldValue }) {
+  const fieldKey = "config.variants";
+
+  const onVariantsChange = (variants) => setFieldValue(fieldKey, variants);
```
dani-mp:
```suggestion
  const onVariantsChange = (newVariants) => setFieldValue(fieldKey, newVariants);
```

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/index.ts
```diff
@@ -78,7 +78,9 @@ export const linkConfigSchema = yup
     allowTargetBlank: yup.boolean().strict().optional(),
     allowText: yup.boolean().strict().optional(),
     repeat: yup.boolean().strict().optional(),
-    variants: yup.array(yup.string()).optional(),
+    variants: yup
+      .array(yup.string().required("values can't be empty"))
```
jomifepe, line 82: 💡
```suggestion
      .array(yup.string().required("Values can't be empty"))
```

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
```diff
@@ -125,6 +136,44 @@ export function Variants({
...
+              key={position}
+              backgroundColor="white"
+              padding={{ inline: 12, block: 8 }}
+              flexDirection="column"
+              gap={8}
+              border
+              borderRadius={4}
+            >
+              <TextInput
+                placeholder="Variation option (e.g. Primary)"
```
jomifepe, line 163: I'm not sure if this was changed intentionally, but in the the design, the placeholder is different.

```suggestion
                placeholder="Enter your variant name (eg. Primary)"
```

jomifepe (approved):

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
```diff
@@ -3,6 +3,7 @@
 /* eslint-disable @typescript-eslint/no-unsafe-return */
 
 import { Box } from "@prismicio/editor-ui";
+import { useField } from "formik";
```
xrutayisire, line 6: 💡 What made you use Formik, knowing that we don't want to continue with it? Could it be simple also without?

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/index.ts
dani-mp: other errors are not capitalized

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/components.tsx
dani-mp: i changed it intentionally, i'll check with come all copies before releasing

packages/slice-machine/src/legacy/lib/models/common/widgets/Link/Form.jsx
dani-mp: i tried but no... to handle errors in a centralized place to avoid closing the modal with the wrong values, the whole thing is already using formik

that's why at least i tried to separate all the logic from formik with the silly variants wrapper

xrutayisire: Got it, thanks!

xrutayisire (approved):
LGTM!

PR #1507, opened 2024-12-13
feat: prevent creating slices and custom types with reserved name

BohdanOne:
**Resolves**: [DT-2239](https://linear.app/prismic/issue/DT-2239/bug-aauser-i-cant-create-a-slice-or-a-ct-called-update-or-insert-as)

### Description

Added validation to prevent creating or renaming slices and custom types with reserved endpoints name : update, insert.
 

### Checklist

### Preview
Zrzut ekranu 2024-12-13 o 16 00 36

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/legacy/components/Forms/CreateCustomTypeModal/CreateCustomTypeModal.tsx
+++ b/packages/slice-machine/src/legacy/components/Forms/CreateCustomTypeModal/CreateCustomTypeModal.tsx
@@ -173,6 +173,10 @@ export const CreateCustomTypeModal: React.FC<CreateCustomTypeModalProps> = ({
           errors.id = "ID cannot be empty.";
         }
 
+        if (["update", "insert"].includes(id.toLowerCase())) {
+          errors.id = `Id "${id}" is reserved for Slice Machine use.`;
+        }
+
         // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
         if (!errors.id && id && !API_ID_REGEX.exec(id)) {
           errors.id = "Invalid id: No special characters allowed.";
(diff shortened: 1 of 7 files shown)


packages/slice-machine/src/legacy/components/Forms/CreateCustomTypeModal/CreateCustomTypeModal.tsx
```diff
@@ -165,10 +165,18 @@ export const CreateCustomTypeModal: React.FC<CreateCustomTypeModalProps> = ({
           })} name is already taken.`;
         }
 
+        if (["update", "insert"].includes(label.toLowerCase())) {
```
jomifepe, line 168: ~⚠️ Isn't this a bit overly strict? Do we want to prevent customtypes/slices containing these? What would be the issue of having a label with the words `Inserted`, `Insertable`, `Insertion`, `Updated`, `Updater`, `Self-update`, `Updateable`, etc?~

~IMO we should be careful not to restrict this field even more, as we already want to remove the PascalCase rule.~

~💡 Some alternatives:~
~1. **Less strict**: Only block if it's just `insert` or `update`, and not a word containing it.~
~2. Block if the word is by itself, even on a sentence (e.g. `insert words` or `word update`), by using a regex like `/\b(insert|update)\b/gi`.~

BohdanOne: @jomifepe I'm not sure if I understand the issue, here we check only if we match whole word  to be either 'update' or 'insert' not a part of it, so we still allow all of examples you provided

jomifepe: Nevermind, I misread the code 🤦‍♂️

jomifepe, line 168: 💡 This could also be extracted, like you did for the slice name

```suggestion
        if (RESERVED_CUSTOM_TYPE_NAMES.includes(label.toLowerCase())) {
```

jomifepe (approved):
LGTM

PR #1517, opened 2024-12-19
Track link variants property when field is added

dani-mp:
Depends on https://github.com/prismicio/slice-machine/pull/1516. Review that one first.

**Resolves**: 
https://linear.app/prismic/issue/DT-2473/aapm-i-can-see-a-property-to-the-event-slice-machine-field-added-type

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/manager/src/managers/telemetry/types.ts
+++ b/packages/manager/src/managers/telemetry/types.ts
@@ -210,6 +210,7 @@ type FieldAddedSegmentEvent = SegmentEvent<
 		contentType: "page type" | "custom type" | "slice";
 		allowText?: boolean;
 		repeat?: boolean;
+		variants?: string[];
 	}
 >;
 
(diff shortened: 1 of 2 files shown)


packages/manager/src/managers/telemetry/types.ts
```diff
@@ -210,6 +210,7 @@ type FieldAddedSegmentEvent = SegmentEvent<
 		contentType: "page type" | "custom type" | "slice";
 		allowText?: boolean;
 		repeat?: boolean;
+		variants?: string[];
```
xrutayisire, line 213: ❓ Interesting, so we can send an array directly to Segment / Amplitude? How does it's translated in Amplitude, as a string?

dani-mp: not sure how it gets in amplitude, but it reaches segment as it should (see debugger screenshot)

image

xrutayisire: Actually that's so coooool, I just checked on Amplitude your event!
It allows Côme to say: I want to check all that contain Primary, or those you have exactly Primary and Secondary, etc etc.

Screenshot 2024-12-19 at 19 45 48

xrutayisire (approved):
LGTM!

packages/manager/src/managers/telemetry/types.ts
dani-mp: irrc, the page builder team has some weird types for tracking events, only allowing strings

maybe it's because of the proxy we use, idk

but both segment and amplitude allow strings, numbers, arrays, booleans... which could improve a lot our analytics, imo

PR #1522, opened 2024-12-23
Add variants to link to media

dani-mp:
**Resolves**: 
https://linear.app/prismic/issue/DT-2527/aadev-i-can-enable-link-variants-for-link-to-media-field

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/legacy/lib/models/common/widgets/LinkToMedia/index.ts
+++ b/packages/slice-machine/src/legacy/lib/models/common/widgets/LinkToMedia/index.ts
@@ -41,7 +41,6 @@ export const LinkToMediaWidget: Widget<Link, typeof schema> = {
       placeholder: "",
       select: "media",
       allowText: true,
-      repeat: false,
     },
   }),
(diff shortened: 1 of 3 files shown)


packages/slice-machine/src/legacy/lib/models/common/widgets/LinkToMedia/index.ts
```diff
@@ -41,7 +41,6 @@ export const LinkToMediaWidget: Widget<Link, typeof schema> = {
       placeholder: "",
       select: "media",
       allowText: true,
-      repeat: false,
```
dani-mp, line 44: not sure why this was here but IMO it's not needed and it's misleading because there's no repeat in link to media

xrutayisire: I think it was because of the first version of types internal, when it was required to set `repeat`

xrutayisire (approved):
LGTM

PR #1536, opened 2025-01-21
feat: add table field

levimykel:
**Resolves**: [DT-2547](https://linear.app/prismic/issue/DT-2547/%5Bm%5D-aadev-i-can-add-a-table-field)

### Description

Adds the ability to add the table field to your models in Slice Machine. 

### Checklist

### Preview

Screenshot 2025-01-21 at 14 52 39

### How to QA [^1]

[^1]:


packages/slice-machine/src/domain/slice.ts
```diff
@@ -71,7 +71,9 @@ export function getNonSharedSliceLabel(
     slice.config?.label ??
     (slice.type === "Group" ||
     slice.type === "Slice" ||
-    (slice.type !== "Boolean" && slice.type !== "Separator")
+    (slice.type !== "Boolean" &&
+      slice.type !== "Separator" &&
+      slice.type !== "Table")
```
dani-mp, line 76: if you added `fieldset` to the model as you mentioned in slack, we can remove this now

btw, what does `fieldset` represent in our fields? i thought this was a legacy thing, should we add it to new fields?

packages/slice-machine/src/legacy/lib/models/common/widgets/Table/index.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/** : {
+  "type" : "Table",
+  "config" : {
+    "label" : "Gable's Fabled Table"
+  }
+} */
+
+const FormFields = {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
```
dani-mp: are all of these needed or were just copy pasted? if they're still needed, let's move the rules that we repeat several times to the top of the file so we have to declare them only once

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/index.tsx
```diff
@@ -54,9 +56,17 @@ const hintItemName = "item";
...
+  let maybeFilteredWidgetsArray = nestedGroupExperiment.eligible
     ? widgetsArray
     : widgetsArray.filter((widget) => widget.CUSTOM_NAME !== "NestedGroup");
+
+  const tableFieldExperiment = useTableFieldExperiment();
+  maybeFilteredWidgetsArray = tableFieldExperiment.eligible
+    ? maybeFilteredWidgetsArray
+    : maybeFilteredWidgetsArray.filter(
+        (widget) => widget.TYPE_NAME !== "Table",
+      );
```
dani-mp: we can do all this in a single map over the original list:
`const widgets = widgetsArray.map(widget => ...)`

same for the rest of the places where we do this

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
```diff
@@ -79,6 +81,11 @@ const FieldZones: FC = () => {
     setIsDeleteRepeatableZoneDialogOpen,
   ] = useState(false);
 
+  const tableFieldExperiment = useTableFieldExperiment();
+  const maybeFilteredItemsWidgetsArray = tableFieldExperiment.eligible
```
dani-mp, line 83: sholud we also take into account this for `primaryWidgetsArray`? probably a good idea to do the filtering directly when we declare the array, to avoid any issues

packages/slice-machine/src/domain/slice.ts
levimykel: Yeah, I should have looked into this before adding it. It is deprecated and isn't needed for new fields. In fact, it isn't used for Boolean as seen with your code addition above. We should leave this code and remove the property from the `types-internal` type (which I'll do).

packages/slice-machine/src/legacy/lib/models/common/widgets/Table/index.ts
levimykel: They were copy and pasted. I should have checked them instead of blindly trusting them. I cleaned them up as you suggested.

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/index.tsx
levimykel: Good call, I combined them into a single filter. This is the only place where we had 2 filters like this.

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
levimykel: Good catch, I fixed it.

dani-mp (approved):
i'll leave it approved, we need to update types internal to remove the `fieldset`

PR #1548, opened 2025-01-31
fix(adapter-nuxt): add removed wrappers

lihbr:
**Resolves**: https://github.com/nuxt-modules/prismic/issues/231

### Description

With @prismicio/vue v5, default wrappers were removed from `<PrismicRichText>` components: https://prismic.io/docs/technical-reference/prismicio-vue/v5#wrap-prismicrichtext-output

This creates an issue with an existing slice templates, this PR therefore adds them back using the `wrapper` prop :)

### Checklist

### Preview

N/A

### How to QA [^1]

1. Create a Nuxt project with Slice Machine
2. Create a slice from the alternative grid template
3. No Vue warning should be emitted after bootstrapping it

[^1]:

### DIFF
--- a/packages/adapter-nuxt/public/AlternateGrid/javascript.vue
+++ b/packages/adapter-nuxt/public/AlternateGrid/javascript.vue
@@ -40,11 +40,13 @@ defineProps(getSliceComponentProps(["slice", "index", "slices", "context"]));
 						v-if="isFilled.richText(slice.primary.title)"
 						:field="slice.primary.title"
 						class="es-alternate-grid__primary-content__intro__headline"
+						wrapper="div"
 					/>
 					<PrismicRichText
 						v-if="isFilled.richText(slice.primary.description)"
 						:field="slice.primary.description"
 						class="es-alternate-grid__primary-content__intro__description"
+						wrapper="div"
 					/>
 				</div>
 				<div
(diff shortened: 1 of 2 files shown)


dani-mp (approved):

packages/adapter-nuxt/public/AlternateGrid/javascript.vue
```diff
@@ -40,11 +40,13 @@ defineProps(getSliceComponentProps(["slice", "index", "slices", "context"]));
 						v-if="isFilled.richText(slice.primary.title)"
 						:field="slice.primary.title"
 						class="es-alternate-grid__primary-content__intro__headline"
+						wrapper="div"
 					/>
 					<PrismicRichText
 						v-if="isFilled.richText(slice.primary.description)"
 						:field="slice.primary.description"
 						class="es-alternate-grid__primary-content__intro__description"
+						wrapper="div"
```
angeloashmore, line 49: 💡 #idea: What are your thoughts on using a more modular version instead?

```html
<div
  v-if="isFilled.richText(slice.primary.description)"
  class="es-alternate-grid__primary-content__intro__description"
>
  <PrismicRichText :field="slice.primary.description" />
</div>
```

With this version, we teach users that the wrapper can be any component and accept any props; it's just another element. It also removes complexity around how to treat extra props (like `class`) when `wrapper` isn't used.

lihbr: Discussed: this feels more like a natural pattern for Vue components, we'll monitor usage to see if we transition to React-style or not at a future time.

PR #1549, opened 2025-02-17
fix: fix team env login

xrutayisire:
**Resolves**: [DT-2603: Fix slice machine login with team env](https://linear.app/prismic/issue/DT-2603/fix-slice-machine-login-with-team-env)

### Description

- Team env are not working when you try to login as it's pointing to prismic.io

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/legacy/lib/utils/preferWroomBase.ts
+++ b/packages/slice-machine/src/legacy/lib/utils/preferWroomBase.ts
@@ -1,13 +1,23 @@
 export default function preferWroomBase(smApiUrl: string): string {
   try {
-    const urlFromSmJson = new URL(smApiUrl);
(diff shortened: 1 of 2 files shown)


packages/slice-machine/src/legacy/lib/utils/preferWroomBase.ts
```diff
@@ -1,13 +1,23 @@
 export default function preferWroomBase(smApiUrl: string): string {
```
xrutayisire, line 1: I kept this function as its legacy and a hotfix, but this should be reworked totally, and we should ask the manager for the correct login URL as the manager already handles that.

dani-mp (approved):

PR #1559, opened 2025-03-18
feat: add AI slice generation option and empty modal

levimykel:
**Resolves**: [DT-2606](https://linear.app/prismic/issue/DT-2606/%5Bm%5D-aadev-in-slice-machine-i-can-add-slices-screenshots-from-page-type)

### Description

Design: https://www.figma.com/design/RGAejB5IuJtUsplHMFxN9H/AI?node-id=95-6171&t=07Z8aeiuQcSxSUQa-0

- Adds the "slicemachine-ai-slice-generation" feature flag hook
- Adds the "Generate with AI" option to slice zone
- Adds a blank Generate with AI modal

### Checklist

### Preview

https://github.com/user-attachments/assets/e265e212-fc06-4007-88ab-14ee64fd3c71

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/components/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
+++ b/packages/slice-machine/src/components/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
@@ -1,0 +1,29 @@
+import {
+  Box,
+  Dialog,
+  DialogActions,
+  DialogCancelButton,
+  DialogContent,
+  DialogHeader,
+} from "@prismicio/editor-ui";
+
+interface GenerateSliceWithAiModalProps {
+  onClose: () => void;
+}
+
+export function GenerateSliceWithAiModal(props: GenerateSliceWithAiModalProps) {
+  const { onClose } = props;
+
+  return (
+    <Dialog open onOpenChange={onClose}>
+      <DialogHeader title="Generate with AI" />
+      <DialogContent gap={0}>
+        <Box height="100%" />
+
+        <DialogActions>
+          <DialogCancelButton />
+        </DialogActions>
+      </DialogContent>
+    </Dialog>
+  );
+}
--- a/packages/slice-machine/src/components/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
+++ b/packages/slice-machine/src/components/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
@@ -16,7 +16,12 @@ export function GenerateSliceWithAiModal(props: GenerateSliceWithAiModalProps) {
   const { open, onClose } = props;
 
   return (
-    <Dialog open={open} onOpenChange={onClose}>
+    <Dialog
+      open={open}
+      onOpenChange={(open) => {
+        if (!open) onClose();
+      }}
+    >
       <DialogHeader title="Generate with AI" />
       <DialogContent gap={0}>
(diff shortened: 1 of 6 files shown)


jomifepe (approved):
LGTM!

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
```diff
@@ -366,6 +381,10 @@ const SliceZone: React.FC<SliceZoneProps> = ({
           onClose={closeCreateSliceModal}
         />
       )}
+      {aiSliceGenerationExperiment.eligible &&
+        isGenerateSliceWithAiModalOpen && (
```
dani-mp: if you do this, the modal doesn't properly animate out, it's just removed from the dom suddenly instead 

this is why modals have the `open` prop when they're controlled, so you can switch it off, and they do the animation

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
```diff
@@ -366,6 +381,10 @@ const SliceZone: React.FC<SliceZoneProps> = ({
           onClose={closeCreateSliceModal}
         />
       )}
+      {aiSliceGenerationExperiment.eligible &&
```
dani-mp: i don't think we need this here, just having the flag in the option list is enough as an entry point (without that entry point, nothing can flip the boolean to open the modal)

dani-mp (changes requested):

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
levimykel: Removed ✅

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
levimykel: Fixed ✅

packages/slice-machine/src/components/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
```diff
@@ -0,0 +1,29 @@
...
+interface GenerateSliceWithAiModalProps {
+  open: boolean;
+  onClose: () => void;
+}
+
+export function GenerateSliceWithAiModal(props: GenerateSliceWithAiModalProps) {
+  const { open, onClose } = props;
+
+  return (
+    <Dialog open={open} onOpenChange={onClose}>
```
dani-mp: here we're calling `onClose` every time `onOpenChange` is called, but they're not the same, `onOpenChange` has an `open` param that could potentially be `true`, and we only want to call `onClose` when it is `false`

dani-mp (approved):

packages/slice-machine/src/components/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
levimykel: Updated ✅

PR #1561, opened 2025-03-19
feat(ai-slice): add file drop zone and blank state

levimykel:
**Resolves**: N/A

### Description

AI slice generation modal updates:
- Adds the file drop zone
- Adds the blank state
- Adds a disabled "Add to page" button  

### Checklist

### Preview

https://github.com/user-attachments/assets/ab2822d7-b1f4-4faf-8fc0-fbd90344a38c

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/components/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
+++ b/packages/slice-machine/src/components/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
@@ -23,8 +23,8 @@ interface GenerateSliceWithAiModalProps {
 export function GenerateSliceWithAiModal(props: GenerateSliceWithAiModalProps) {
   const { open, onClose } = props;
 
-  const onFilesSelected = (files: File[]) => {
-    console.log(files);
+  const onImagesSelected = (images: File[]) => {
+    console.log(images);
   };
 
   return (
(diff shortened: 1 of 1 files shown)


packages/slice-machine/src/components/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
```diff
@@ -24,11 +36,68 @@ export function GenerateSliceWithAiModal(props: GenerateSliceWithAiModalProps) {
...
       <DialogContent gap={0}>
-        <Box height="100%" />
+        <Box padding={16} height="100%">
+          <FileDropZone
+            onFilesSelected={onFilesSelected}
+            assetType="image"
+            overlay={
+              <UploadBlankSlate
+                onFilesSelected={onFilesSelected}
+                droppingFiles
```
dani-mp, line 46: this feels a bit weird, having the same component rendered twice, one as a child and another as an "overlay", with different configurations, rather than having a render prop coming from the file drop zone, is this how they implemented it in the page builder?

levimykel: I agree that it feels weird. In the page builder, the overlay is different from the blank state, so they don't have this issue.

dani-mp: i've taken a look at the implementation and yeah, it's super weird, also because the "overlay" is only rendered while dropping files, but at the same time, nothing in its name indicates this

dani-mp (approved):

PR #1563, opened 2025-03-20
feat(ai-slice): add slice card

levimykel:
**Resolves**: N/A

### Description

- Adds a slice card with a loading state.
- Fixes the "Missing description" console error

Note in the next PR, I will relocate the `GenerateSliceWithAiModal` component.

### Checklist

### Preview

https://github.com/user-attachments/assets/5f41fda3-c357-422d-b02d-aca24c1883d1

### How to QA [^1]

[^1]:


packages/slice-machine/src/components/Card/Card.tsx
```diff
@@ -136,18 +136,29 @@ export const CardActions: FC<PropsWithChildren> = (props) => (
 );
 
 type CardFooterProps = {
+  startIcon?: IconName | "loading";
```
dani-mp: i'd add a separate boolean `loading` optinoal prop instead, that would allow us to use an icon named `loading`, for instance

packages/slice-machine/src/components/Card/Card.module.css
```diff
@@ -200,3 +204,13 @@
   background-color: var(--purple4);
   color: var(--purple11);
 }
+
+.startIconBox {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 32px;
+  height: 32px;
```
dani-mp: this box will contain elements with fixed sizes defined elsewhere, rather than fixing its size and hoping that they both will work fine together, using padding will make the parent box adapt to the children size even if it changes underneath

packages/slice-machine/src/components/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
```diff
@@ -22,38 +27,84 @@ interface GenerateSliceWithAiModalProps {
...
+            thumbnailUrl:
+              "https://images.unsplash.com/photo-1588315029754-2dd089d39a1a?w=512",
+            status: "success",
+          })),
+        );
+      }, 2000);
+
+      return () => clearTimeout(timeout);
+    }
+  }, [slices]);
```
dani-mp: use effect is not recommended for this, it's better to fire the action in the handler https://react.dev/learn/you-might-not-need-an-effect

packages/slice-machine/src/components/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
```diff
@@ -22,38 +27,84 @@ interface GenerateSliceWithAiModalProps {
...
+              }
+            >
+              <UploadBlankSlate onFilesSelected={onImagesSelected} />
+            </FileDropZone>
+          </Box>
+        ) : (
+          <ScrollArea>
+            <Box padding={16} height="100%" gap={16}>
+              {slices.map((slice) => (
+                <SliceCard slice={slice} />
```
dani-mp: probably need a key here

packages/slice-machine/src/components/GenerateSliceWithAiModal/SliceCard.tsx
```diff
@@ -0,0 +1,48 @@
+import { Card, CardFooter, CardMedia } from "@/components/Card";
+
+import { Slice } from "./GenerateSliceWithAiModal";
```
dani-mp: we import something from the modal here, and then we import this card from the modal, creating a circular reference 

in this case is not a problem because one is a type, but it'd feel better if you declare the slice type within the slice card, and then import both the slice type and the slice card to build the more complex slice modal

packages/slice-machine/src/components/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
```diff
@@ -62,6 +113,13 @@ export function GenerateSliceWithAiModal(props: GenerateSliceWithAiModalProps) {
   );
 }
 
+export interface Slice {
```
dani-mp: we need discriminated union here, where each status will have some properties; for instance, you don't have a url before the image is submitted or if it failed, but you will for sure have it when it has been uploaded

packages/slice-machine/src/components/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
```diff
@@ -62,6 +113,13 @@ export function GenerateSliceWithAiModal(props: GenerateSliceWithAiModalProps) {
...
+  status: "loading" | "success" | "error";
+  displayName: string;
```
dani-mp: you can use the file name in the beginning, then we can swap it with the model name when it has been generated

dani-mp (commented):
left some comments, there's another thing about the how to store the slices in the state and the upload promises but we can see that in the next pr(s)

packages/slice-machine/src/components/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
levimykel: That's not really true, yeah? Once the image has been uploaded, the thumbnail url will exist. The AI endpoint could still be processing (the "loading" state) or could fail, but the thumbnail url would still be there. Or am I misunderstanding how this will work?

packages/slice-machine/src/components/Card/Card.tsx
levimykel: Done ✅

packages/slice-machine/src/components/Card/Card.module.css
levimykel: Done ✅

packages/slice-machine/src/components/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
levimykel: I removed `useEffect` and created a function to mock an API call. Let me know what you think.

packages/slice-machine/src/components/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
levimykel: Yes, of course.

packages/slice-machine/src/components/GenerateSliceWithAiModal/SliceCard.tsx
levimykel: Done ✅

packages/slice-machine/src/components/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
levimykel: Done ✅

packages/slice-machine/src/components/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
dani-mp: when we introduce the inferring part, that'll be a different "loading" state, with different properties: as you said, at that point, you'll have a url, but before you don't

so it's not about the properties that you keep, which you can forward from one state to the next; it's about the properties that you don't have

the idea is to create a state machine, where in each step you don't have a bunch of optional properties, but a discriminated union with the exact state each time

levimykel: Done as per our discussion ✅

packages/slice-machine/src/components/Card/Card.tsx
```diff
@@ -136,18 +136,28 @@ export const CardActions: FC<PropsWithChildren> = (props) => (
...
+  startIcon,
   subtitle,
   title,
   ...otherProps
 }) => (
   <div {...otherProps} className={styles.footer}>
+    {(startIcon || loading) && (
+      <div className={styles.startIconBox}>
+        {loading && <ProgressCircle color="grey11" />}
+        {startIcon && <Icon name={startIcon} size="small" />}
```
dani-mp: this can potentially show two elements on the screen, right? shouldn't it be
```suggestion
        {!loading && startIcon && <Icon name={startIcon} size="small" />}
```

dani-mp (approved):

PR #1573, opened 2025-03-25
infer slice - connect the pieces

dani-mp:
**Resolves**: 

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/manager/src/managers/slices/SlicesManager.ts
+++ b/packages/manager/src/managers/slices/SlicesManager.ts
@@ -364,6 +364,7 @@ export class SlicesManager extends BaseManager {
 			([ids, names], { model }) => {
 				ids.add(model.id);
 				names.add(model.name.toLowerCase());
+
 				return [ids, names];
 			},
 			[new Set(), new Set()],
--- a/packages/manager/src/managers/slices/SlicesManager.ts
+++ b/packages/manager/src/managers/slices/SlicesManager.ts
@@ -344,40 +344,55 @@ export class SlicesManager extends BaseManager {
 		};
 	}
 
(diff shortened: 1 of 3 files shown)


levimykel (approved):
Everything looks good to me 💯

packages/manager/src/managers/slices/SlicesManager.ts
```diff
@@ -344,41 +344,57 @@ export class SlicesManager extends BaseManager {
...
+				name = `${model.name}${counter}`;
+				counter++;
+			}
+			existingNames.add(name.toLowerCase());
 
 			return {
 				...model,
-				id: newId,
+				id,
+				name,
```
jomifepe, line 394: ~💡 I found this code a bit hard to read, mainly because we're mapping over models and mutating data outside of it, which is unnexpected.~

~I would go to a more traditional approach with a `for...of`, that way it's clear we're just iterating over data to mutate other variables, and not only the return value of the map matters.~

~WDYT?~

~Something like:~

```ts
const updatedModels: SharedSlice[] = [];

for (const model of models) {
    // fix id
    let id = model.id;
    let counter = 2;
    while (existingIds.has(id)) {
      id = `${model.id}${counter}`;
      counter++;
    }
    existingIds.add(id);
    
    // fix name
    let name = model.name;
    counter = 2;
    while (existingNames.has(name.toLowerCase())) {
      name = `${model.name}${counter}`;
      counter++;
    }
    existingNames.add(name.toLowerCase());

    updatedModels.push({
        ...model,
        id,
        name,
    });
}
```


EDIT: We discussed that it's not that bad here because `existingIds` and `existingNames` are not used elsewhere, so this is okay.

jomifepe (approved):
Looks great 💯

PR #1576, opened 2025-03-26
feat(ai-slices): add onSuccess on submit

levimykel:
**Resolves**: [DT-2606](https://linear.app/prismic/issue/DT-2606/%5Bm%5D-aadev-in-slice-machine-i-can-add-slices-screenshots-from-page-type)

### Description

This PR:
- Adds `onSuccess` function to run on successful submit
- Adds the inferred slices to the page type
- Adds toast message on success or failure
- Adds the standard "slice:created" tracking (to be improved in a following PR)
- Removes the scroll area padding-right for the scrollbar
- Updates the displayed slice name after receiving the inferred slice model
- Adds a 10-image limit to the file drop zone
- Fixes a random typo in a comment

### Checklist

### Preview

https://github.com/user-attachments/assets/42ddd221-ecfd-4bfa-9fcb-7a53bd1120a6

### How to QA [^1]

[^1]:


packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
```diff
@@ -146,17 +148,19 @@ export function GenerateSliceWithAiModal(props: GenerateSliceWithAiModalProps) {
...
-      });
+
+    try {
+      const slicesAdded = await addSlices(newSlices);
+      if (currentId !== id.current) return;
+      await onSuccess(slicesAdded);
+    } catch (e) {
+      if (currentId !== id.current) return;
+      const errorMessage = "An unexpected error happened while adding slices.";
+      console.error(errorMessage, e);
```
dani-mp: should we keep this error log?

packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
```diff
@@ -146,17 +148,19 @@ export function GenerateSliceWithAiModal(props: GenerateSliceWithAiModalProps) {
...
-      .catch(() => {
-        if (currentId !== id.current) return;
-        setIsCreatingSlices(false);
-        // TODO: Show error?
-      });
+
+    try {
+      const slicesAdded = await addSlices(newSlices);
+      if (currentId !== id.current) return;
+      await onSuccess(slicesAdded);
```
dani-mp: we shouldn't await here - if for some reason `onSuccess` fails, the catch would continue but in theory the modal has finished its job already

packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
```diff
@@ -146,17 +148,19 @@ export function GenerateSliceWithAiModal(props: GenerateSliceWithAiModalProps) {
...
-      .then(() => {
-        if (currentId !== id.current) return;
-        setIsCreatingSlices(false);
-        // TODO: Execute modal callback.
-      })
-      .catch(() => {
-        if (currentId !== id.current) return;
-        setIsCreatingSlices(false);
-        // TODO: Show error?
-      });
```
dani-mp, line 159: i prefer to keep the .then and .catch to make it consistent with the other async ops within the modal, if you want to `setIsCreatingSlices(false)` in a single place you can use `.finally`, but imo it's a good idea to separate both paths because it makes things easier to reason about, even if there's a small duplication in some of the steps

for instance, we forgot to reset the slices and the modal id in the happy path

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
```diff
@@ -429,6 +433,38 @@ const SliceZone: React.FC<SliceZoneProps> = ({
       )}
       <GenerateSliceWithAiModal
         open={isGenerateSliceWithAiModalOpen}
+        onSuccess={async (slices: SharedSlice[]) => {
```
dani-mp: the `CreateSliceModal` also has a `onSuccess` callback and does the same thing, but we're doing here more things than we're doing there, probably because `CreateSliceModal` partially does some of those things

ideally, we should be able to remove somethings from the `CreateSliceModal` and bring them here, and have a single function that can be called from both `onSuccess` implementations here to add the slices to the slice zone, that way we avoid complex duplication

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
```diff
@@ -429,6 +433,38 @@ const SliceZone: React.FC<SliceZoneProps> = ({
...
+          const serverState = await getState();
+          // Update Redux store
```
dani-mp:
```suggestion
```

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
```diff
@@ -429,6 +433,38 @@ const SliceZone: React.FC<SliceZoneProps> = ({
...
+          const newCustomType = addSlicesToSliceZone({
+            customType,
+            tabId,
+            slices,
+          });
+          setCustomType(CustomTypes.fromSM(newCustomType), () => {
+            toast.success(
+              <ToastMessageWithPath
+                message="New slices added to slice zone and created at: "
+                path={`${localLibraries[0].name}/`}
```
dani-mp: this and line 461 is true only in practice (becuase it happens that we use the first/default library to create the slices to within the modal, but these two parts of the code are completely disconnected

we should either pass to the modal the library we want to use, or the on success callback should pass here the library it used, that way a change in one will be connected with the other

packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
levimykel: This is what's done for all the other slice event errors (creation, deletion, etc.)

https://github.com/prismicio/slice-machine/blob/d8d3abb77feb78b479f1c1b71143e5c6a06c1d39/packages/slice-machine/src/features/slices/actions/createSlice.ts#L37-L41

Not that we have to follow what is done elsewhere. It does provide the error, so it could possibly give additional information not shown in the toast. I say we keep it for those two reasons, but I'm fine removing it if you think we should start removing this sort of thing.

packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
levimykel: I'm good with that. I changed it back and added the slice reset to the happy path ☑️

packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
levimykel: Done ☑️

packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
dani-mp: personally, i would remove it

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
levimykel: Alright, the library has one source of truth now ☑️

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
levimykel: Done ☑️

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
levimykel: I've given this a go. Let me know if this is what you had in mind.

packages/manager/src/managers/telemetry/types.ts
```diff
@@ -271,7 +271,14 @@ type CustomTypeSavedSegmentEvent = SegmentEvent<
...
 type SliceCreatedSegmentEvent = SegmentEvent<
 	typeof SegmentEventType.slice_created,
-	{ id: string; name: string; library: string; sliceTemplate?: string }
+	{
+		id: string;
+		name: string;
+		library: string;
+		sliceTemplate?: string;
+		location: "custom_type" | "page_type" | "slices";
+		mode: "ai" | "manual" | "template";
```
dani-mp: i guess it doesn't make sense to pass a slice template if the mode is not template

so this can be enforced via typescript

you can do:

{ ... } & ({ mode: "ai" } | { mode: "manual" } | { mode: "template", sliceTemplate: string })

if in the future we start adding other props related to other modes, they can be added safely too

packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
```diff
@@ -26,7 +26,7 @@ import { Slice, SliceCard } from "./SliceCard";
 
 interface GenerateSliceWithAiModalProps {
   open: boolean;
-  onSuccess: (models: SharedSlice[]) => Promise<void>;
+  onSuccess: (slices: SharedSlice[], libraryName: string) => void;
```
dani-mp: let's avoid positional arguments

packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
```diff
@@ -147,15 +149,19 @@ export function GenerateSliceWithAiModal(props: GenerateSliceWithAiModalProps) {
     const currentId = id.current;
     setIsCreatingSlices(true);
     addSlices(newSlices)
-      .then(() => {
+      .then(({ slices, libraryName }) => {
         if (currentId !== id.current) return;
         setIsCreatingSlices(false);
-        // TODO: Execute modal callback.
+        onSuccess(slices, libraryName);
```
dani-mp: we need to reset the modal id here as well
```suggestion
        id.current = crypto.randomUUID();
        onSuccess(slices, libraryName);
```

maybe it's a good idea to have a reset fn that puts back the common things

packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
```diff
@@ -147,15 +149,19 @@ export function GenerateSliceWithAiModal(props: GenerateSliceWithAiModalProps) {
...
+        setSlices([]);
       })
-      .catch(() => {
+      .catch((e) => {
         if (currentId !== id.current) return;
         setIsCreatingSlices(false);
-        // TODO: Show error?
+        const errorMessage =
+          "An unexpected error happened while adding slices.";
+        console.error(errorMessage, e);
```
dani-mp: still unsure about the console.error - i think we should remove it, as we don't want to populate the user's console with our own errors

but you can also ask the team to see if there's a discussed convention about this already

packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
```diff
@@ -307,5 +314,5 @@ async function addSlices(newSlices: NewSlice[]) {
     ),
   );
 
-  return slices;
+  return { slices, libraryName: library };
```
dani-mp: why not just library (it's just a string)
```suggestion
  return { slices, library };
```

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/SlicesTemplatesModal.tsx
```diff
@@ -18,7 +18,7 @@ import UpdateSliceZoneModalList from "./UpdateSliceZoneModalList";
 interface UpdateSliceModalProps {
   formId: string;
   close: () => void;
-  onSuccess: (slices: SharedSlice[]) => void;
+  onSuccess: (slices: SharedSlice[], libraryName: string) => void;
```
dani-mp: i'd use `library` here, and also an object to avoid positional arguments

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/SlicesTemplatesModal.tsx
```diff
@@ -55,11 +55,13 @@ export const SlicesTemplatesModal: FC<UpdateSliceModalProps> = ({
             // Update Redux store
             createSliceSuccess(serverState.libraries);
 
+            const libraryName = localLibraries[0].name;
```
dani-mp: same here, `library`, sometimes we use library name, sometimes we use library id... i think when it's a string we can just call it library
```suggestion
            const library = localLibraries[0].name;
```

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
```diff
@@ -203,6 +207,45 @@ const SliceZone: React.FC<SliceZoneProps> = ({
     setIsSlicesTemplatesModalOpen(false);
   };
 
+  const addSlices = async (args: {
+    slices: SharedSlice[];
+    libraryName: string;
```
dani-mp:
```suggestion
    library: string;
```

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
```diff
@@ -203,6 +207,45 @@ const SliceZone: React.FC<SliceZoneProps> = ({
...
+        event: "slice:created",
+        id: slice.id,
+        name: slice.name,
+        library: libraryName,
+        location: `${customType.format}_type`,
+        mode,
+      });
+    }
+
+    syncChanges();
```
dani-mp: we can sync before tracking, as it's a more important thing

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
```diff
@@ -203,6 +207,45 @@ const SliceZone: React.FC<SliceZoneProps> = ({
...
+    void completeStep("createSlice");
+
+    for (const slice of slices) {
+      void telemetry.track({
+        event: "slice:created",
+        id: slice.id,
+        name: slice.name,
+        library: libraryName,
+        location: `${customType.format}_type`,
+        mode,
```
dani-mp: we're not passing the slice template here when the mode is template (not sure if we should, but i remember having seen that param in the tracking event type

packages/slice-machine/src/pages/slices.tsx
```diff
@@ -233,6 +234,16 @@ const SlicesIndex: React.FunctionComponent = () => {
                   sliceName: newSlice.name,
                   variationId,
                 });
+
+                void telemetry.track({
```
dani-mp, line 262: this was missing?

dani-mp (commented):
i think the refactor was very good, we merged 3 places where we had duplicated code into one

i've left some comments, most of them quite small

packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
levimykel: I added the snippet you mentioned, but I'm not sure what you mean by adding a reset fn. Can you elaborate on that?

packages/slice-machine/src/pages/slices.tsx
levimykel: No, it wasn't. This was originally handled in the `createSlice` action function. Because we removed the tracking from there, I needed to add a tracking event here to make sure that this is still tracked.

packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
dani-mp: we have at least a couple of places (when we're about to close the modal and when we're about to call `onSuccess`) where we have to repeat the same logic, to _reset_ the modal to its original state, preparing it to be open again with a fresh instance - i thought it was a good idea to encapsulate this logic in a function named `reset` that you can call from both places, that way, if tomorrow we need to add some logic there, we'll do it in a single place and we won't forget to add it to the other branches

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
levimykel: Fixed, good catch 🙏

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
levimykel: Done ☑️

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
levimykel: Done ☑️

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/SlicesTemplatesModal.tsx
levimykel: Done ☑️

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/SlicesTemplatesModal.tsx
levimykel: Done ☑️

packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
levimykel: Done ☑️

packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
levimykel: I removed it ☑️

packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
levimykel: Done ☑️

packages/manager/src/managers/telemetry/types.ts
levimykel: Done ☑️

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
```diff
@@ -203,6 +207,61 @@ const SliceZone: React.FC<SliceZoneProps> = ({
...
+    for (const [index, slice] of slices.entries()) {
+      if (mode === "template") {
+        void telemetry.track({
+          event: "slice:created",
+          id: slice.id,
+          name: slice.name,
+          library,
+          location: `${customType.format}_type`,
+          mode,
+          sliceTemplate: args.sliceTemplates[index],
```
dani-mp: maybe you can do:

`...(mode === "template" ? {mode, sliceTemplate: ...} : {mode}`

to avoid all the duplication? otherwise, you can try to create an object with all the common properties once, then on each branch you finalize the object depending on the mode

finally, if the first option doesn't work and you have to go with the if, it's always better to use a switch with no default, especially in this case with such a small number of options, so the compiler helps you not forget any branch, and also will help you with future branches

dani-mp (approved):
just left one more comment to avoid some repetition but it's not super important, good job

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
levimykel: I was trying something similar but couldn't get it to work. This is exactly what I needed, thanks 🙏

levimykel:
This was closed because this work was redone in smaller PRs.

PR #1586, opened 2025-04-10
AI feedback

dani-mp:
Resolves: 
https://linear.app/prismic/issue/DT-2643/[m]-implement-ai-slice-feedback-flow-thumbs-updown

### Description

adds a feedback component to gather feedback from the user for ai generated slices, and send the data to analytics

### Checklist

### How to QA [^1]

[^1]:


packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
dani-mp, line 1: review this file hiding whitespaces, as most of the changes were due to indenting the existing component

packages/slice-machine/src/legacy/lib/builders/SliceBuilder/FieldZones/index.tsx
```diff
@@ -209,104 +210,121 @@ const FieldZones: FC = () => {
...
-          `slice.primary${transformKeyAccessor(key)}`
-        }
-        testId="static-zone-content"
-        isRepeatableCustomType={undefined}
-        emptyStateHeading={undefined}
-      />
-      {hasItems ? (
+    <Box
+      flexDirection="column"
+      // backgroundColor="amber10"
```
dani-mp:
```suggestion
```

packages/slice-machine/src/features/aiFeedback/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+  library,
+  sliceId,
+  variationId,
+}: {
+  type: "model";
+  library: string;
+  sliceId: string;
+  variationId: string;
+}) {
+  return ["ai-feedback", type, library, sliceId, variationId].join("#");
```
xrutayisire: 💡I don't think adding the type is necessary. Slices are shared by design between types. library + sliceId + variationId is enough to ensure a unique key.

packages/slice-machine/src/features/aiFeedback/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+  sliceId,
+  variationId,
+}: {
+  type: "model";
+  library: string;
+  sliceId: string;
+  variationId: string;
+}) {
+  const key = getKey({ type, library, sliceId, variationId });
+  const [value, setValue] = usePersistedState(key, undefined, {
```
xrutayisire: ❓I don't fully get the usage of this custom hook, like we don't use it to set the initial value.
I don't really mind, but it feels weird the fact we directly use `localStorage.setItem` but then to get and remove we use this hook.

packages/slice-machine/src/features/aiFeedback/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+    schema: z.object({
+      langSmithUrl: z.string().url().optional(),
+    }),
+  });
+  return {
+    key,
+    value,
+    remove: () => setValue(undefined),
```
xrutayisire: ❓ This will keep the key in the local storage, no? Just without a langSmithUrl.
I find it a bit strange that we rely on the value to decide to display or not, I would use the key for that. And ensure to clean the key after. langSmith is a detail, but is used as the source of truth.

packages/slice-machine/src/features/builder/AddStaticFieldDropdown.tsx
```diff
@@ -9,7 +9,7 @@ import {
 import { StaticFieldsInfoDialog } from "@/features/builder/StaticFieldsInfoDialog";
 import { usePersistedState } from "@/hooks/usePersistedState";
 
-const LOCAL_STORAGE_KEY = "staticFieldsInfoDialogDismissed";
+const LOCAL_STORAGE_KEY = "slice-machine_staticFieldsInfoDialogDismissed";
```
xrutayisire: ⚠️ Changing the key here will make the info dialog visible again for users that already saw it.

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
```diff
@@ -466,15 +465,23 @@ const SliceZone: React.FC<SliceZoneProps> = ({
           void completeStep("createSlice");
           syncChanges();
 
-          for (const slice of slices) {
+          for (const { model, langSmithUrl } of slices) {
             void telemetry.track({
```
xrutayisire, line 469: ❓  I wonder if Benj would like the langSmith url here as well now that we have it?

xrutayisire (changes requested):
Quick review, nothing big, mostly questions 🙂

packages/slice-machine/src/features/aiFeedback/index.tsx
dani-mp: Type here is "model" for now, in case we want to add others in the future like "code" or "mocks". You mentioned that maybe we wanted to ask feedback in different places depending on the feature.

packages/slice-machine/src/features/aiFeedback/index.tsx
dani-mp: When we add it, it's imperative code. We just need to set a bunch of values in local storage. When we read it, it's declarative, we need to re-render.

packages/slice-machine/src/features/aiFeedback/index.tsx
dani-mp: When the value is undefined, the whole item is removed from local storage. I changed the implementation of the hook in the previous PR to account for this.

packages/slice-machine/src/features/builder/AddStaticFieldDropdown.tsx
dani-mp: This is another change I made in the hook, because we had only one usage. In my opinion, it was a mistake that the hook would prepend the key passed with extra content, because then you can't add/remove that items without using the hook. In the previous implementation, I tried exporting the key prefix together with the hook but I think being 100% explicit about the key you want to use is better.

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
dani-mp: What do you mean? We're sending the lang smith url as part of the analytics event.

packages/slice-machine/src/features/aiFeedback/index.tsx
xrutayisire: Oh sorry my bad, I saw it was like the id of the custom type!

packages/slice-machine/src/features/aiFeedback/index.tsx
xrutayisire: Dam I should have check the other PR, perfect then!

packages/slice-machine/src/legacy/lib/builders/CustomTypeBuilder/SliceZone/index.tsx
xrutayisire: I mean, we could propose to send it also on the track of slice created from AI. Not related to this PR of course, but was wondering if it could be interesting for him

packages/slice-machine/src/features/builder/AddStaticFieldDropdown.tsx
xrutayisire: Sorry to not have read the other PR. I agree with being 100% explicit is better.

packages/slice-machine/src/features/aiFeedback/index.tsx
xrutayisire: Ok I see, I was a bit lost with the usage, the hook doesn't feel super right for that as it expects a default value, but you don't really want to set it.

packages/slice-machine/src/features/aiFeedback/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+}: {
+  type: "model";
+  library: string;
+  sliceId: string;
+  variationId: string;
+  langSmithUrl?: string;
+}) {
+  const key = getKey({ type, library, sliceId, variationId });
+  const feedback = JSON.stringify({ langSmithUrl });
+  localStorage.setItem(key, feedback);
```
xrutayisire, line 32: 💡 Actually, the goal with the prefix in the hook was to share a common prefix for all storage keys. Maybe you will tell me it's really not interesting. But if yes, we could keep that and still prefix all storage key with `slice-machine_`?

xrutayisire (approved):
Super clear answers, I left two questions open and let you handle them, LGTM otherwise!

packages/slice-machine/src/features/aiFeedback/index.tsx
dani-mp: Yeah, the default value is optional.

PR #1587, opened 2025-04-10
Fix usePersistedState hook

dani-mp:
Resolves: 

### Description

fixes the hook using the event handler rather than use effect to sync the values, which was causing that changing the key would re-save the content of the previous key using the new one

### Checklist

### How to QA [^1]

[^1]:


packages/slice-machine/src/legacy/lib/builders/SliceBuilder/VariationsList/AddVariationModal.tsx
```diff
@@ -197,7 +197,7 @@ const AddVariationModal: React.FunctionComponent<{
                   autoComplete="off"
                   id="id"
                   name="id"
-                  placeholder="e.g. GridWithIcon"
+                  placeholder="e.g. gridWithIcon"
```
dani-mp, line 200: not related, saw it while testing stuff

levimykel (approved):
Looks good to me 🚀

PR #1588, opened 2025-04-10
[DT-1813] test: setup and teardown test repository

jomifepe:
Resolves: [DT-1813](https://linear.app/prismic/issue/DT-1813)

### Description

Create repository on test setup and delete in on teardown.

### Checklist

- [X] If my changes affect backward compatibility, it has been discussed.
- [X] If my changes require an update to the CONTRIBUTING.md guide, I updated it.

### How to QA [^1]

[^1]:

### DIFF
--- a/playwright/globalSetup.ts
+++ b/playwright/globalSetup.ts
@@ -1,0 +1,44 @@
+import {
+  createRepositoriesManager,
+  type RepositoryConfig,
+} from "@prismicio/e2e-tests-utils";
+
+import {
+  auth,
+  baseUrl,
+  manageV2Config,
+  prismicCluster,
+} from "./playwright.config";
+
+import { setRepositoryEnvVar } from "./fixtures";
+
+async function globalSetup() {
+  const config: RepositoryConfig = {
+    locales: ["en-us", "en-pl", "it-it"],
+    defaultLocale: "en-us",
+    preview: {
+      name: "Preview",
+      websiteURL: "https://prismic.io",
+      resolverPath: "/preview",
+    },
+    customTypes: [],
+    slices: [],
+  };
+
+  const testUtils = createRepositoriesManager({
+    urlConfig: baseUrl,
+    authConfig: { email: auth.username, password: auth.password },
+    manageV2Config,
+    cluster: prismicCluster,
+  });
+
+  const repository = await testUtils.createRepository(config);
+
+  // Inject the name of the newly created repo into the Env, so we can run the app pointing to it.
+  setRepositoryEnvVar(repository.name);
+
+  // Changing plan to be able to add user
+  await repository.changePlan("platinum");
+}
+
+export default globalSetup;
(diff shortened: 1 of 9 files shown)


playwright/globalSetup.ts
```diff
@@ -0,0 +1,44 @@
...
+import {
+  auth,
+  baseUrl,
+  manageV2Config,
+  prismicCluster,
+} from "./playwright.config";
+
+import { setRepositoryEnvVar } from "./fixtures";
+
+async function globalSetup() {
```
jomifepe, line 15: This was taken/inspired from https://github.com/prismicio/editor/blob/main/packages/wroom/e2e/globalSetup.ts

playwright/playwright.config.ts
```diff
@@ -1,7 +1,46 @@
...
+    BASE_URL: string;
+    E2E_REPOSITORY: string | undefined;
+  };
+};
+
+export const baseUrl = process.env.BASE_URL;
+
+export const prismicCluster = process.env.PRISMIC_CLUSTER;
+
+export const auth = {
```
jomifepe, line 24: Same users as in https://github.com/prismicio/editor/blob/main/packages/wroom/e2e/globalSetup.ts

playwright/globalSetup.ts
```diff
@@ -0,0 +1,44 @@
...
+  const config: RepositoryConfig = {
+    locales: ["en-us", "en-pl", "it-it"],
```
xrutayisire: 💡 I would keep the config minimal and iterate on it depending on our tests.

Like just `en-us` locale and `preview` don't sure it's required.

playwright/globalSetup.ts
```diff
@@ -0,0 +1,44 @@
...
+    cluster: prismicCluster,
+  });
+
+  const repository = await testUtils.createRepository(config);
+
+  // Inject the name of the newly created repo into the Env, so we can run the app pointing to it.
+  setRepositoryEnvVar(repository.name);
+
+  // Changing plan to be able to add user
+  await repository.changePlan("platinum");
```
xrutayisire: 💡Actually, didn't ask you before, but I don't yet see why in SM we would like to test multiple users?
Because here you setup something change plan and add users, but we don't do it, no? Do you think it should be part of the initial setup?

playwright/globalSetup.ts
jomifepe: Yeah, will keep it simple here

playwright/globalSetup.ts
jomifepe: You're right, this is taken from editor and I removed part of the code and this part should also go.

jomifepe: I don't think we need it. If we do in the future, we'll add it.

jomifepe: Removed ✅

playwright/globalSetup.ts
jomifepe: Changed ✅

xrutayisire (approved):
LGTM!

PR #1592, opened 2025-04-11
[DT-1813.rev1] setup and teardown test repository

jomifepe:
Resolves: [DT-1813](https://linear.app/prismic/issue/DT-1813), [DT-2660](https://linear.app/prismic/issue/DT-2660)

### Description

Adding things missed in the first PR:
- Run E2E against production in `main` and the rest against `dev-tools`
- Set `baseUrl` depending on `SM_ENV`
- Allow to set `E2E_REPOSITORY` to skip temporary repo creation
- Default to `dev-tools` for E2E
- Remove unnecessary `manageV2` config
- Remove unnecessary `PRISMIC_CLUSTER` env var and set `cluster` depending on `SM_ENV`
- Add `.env.test.example`

### Checklist

- [X] If my changes affect backward compatibility, it has been discussed.
- [X] If my changes require an update to the CONTRIBUTING.md guide, I updated it.

### How to QA [^1]

[^1]:

### DIFF
--- a/playwright/.env.test.example
+++ b/playwright/.env.test.example
@@ -1,0 +1,6 @@
+SM_ENV=dev-tools # dev-tools, marketing-tools, platform, staging, production
+PLAYWRIGHT_ADMIN_USERNAME=first.last@prismic.io
+PLAYWRIGHT_ADMIN_PASSWORD=secret
+MANAGEV2_SECRET=foo
+MANAGEV2_AUDIENCE=bar
+PRISMIC_CLUSTER=baz
--- a/playwright/.env.test.example
+++ b/playwright/.env.test.example
@@ -2,5 +2,4 @@ PLAYWRIGHT_ADMIN_USERNAME=first.last@prismic.io
 PLAYWRIGHT_ADMIN_PASSWORD=secret
 
 SM_ENV=dev-tools # optional (default: dev-tools) - production, staging, dev-tools, marketing-tools, platform
-PRISMIC_CLUSTER=baz # optional
 E2E_REPOSITORY= # optional - if you want to use an existing repo instead of having a temporary one created for the test run
--- a/playwright/.env.test.example
+++ b/playwright/.env.test.example
@@ -1,7 +1,5 @@
 PLAYWRIGHT_ADMIN_USERNAME=first.last@prismic.io
 PLAYWRIGHT_ADMIN_PASSWORD=secret
-MANAGEV2_SECRET=foo
-MANAGEV2_AUDIENCE=bar
 
 SM_ENV=dev-tools # optional (default: dev-tools) - production, staging, dev-tools, marketing-tools, platform
 PRISMIC_CLUSTER=baz # optional
--- a/playwright/.env.test.example
+++ b/playwright/.env.test.example
@@ -1,6 +1,8 @@
-SM_ENV=dev-tools # dev-tools, marketing-tools, platform, staging, production
 PLAYWRIGHT_ADMIN_USERNAME=first.last@prismic.io
 PLAYWRIGHT_ADMIN_PASSWORD=secret
 MANAGEV2_SECRET=foo
 MANAGEV2_AUDIENCE=bar
-PRISMIC_CLUSTER=baz
+
+SM_ENV=dev-tools # optional (default: dev-tools) - production, staging, dev-tools, marketing-tools, platform
+PRISMIC_CLUSTER=baz # optional
+E2E_REPOSITORY= # optional - if you want to use an existing repo instead of having a temporary one created for the test run
--- a/playwright/playwright.config.ts
+++ b/playwright/playwright.config.ts
@@ -15,7 +15,6 @@ declare const process: {
     MANAGEV2_AUDIENCE: string;
     E2E_REPOSITORY: string | undefined;
     SM_ENV:
-      | "development"
       | "dev-tools"
       | "marketing-tools"
       | "platform"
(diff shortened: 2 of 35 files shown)


playwright/playwright.config.ts
```diff
@@ -14,10 +14,26 @@ declare const process: {
     MANAGEV2_SECRET: string;
     MANAGEV2_AUDIENCE: string;
     E2E_REPOSITORY: string | undefined;
+    SM_ENV:
+      | "development"
```
xrutayisire: 💡 I think we should remove dev, as it doesn't make sense anymore, it would be wroom.test

playwright/playwright.config.ts
```diff
@@ -14,10 +14,26 @@ declare const process: {
...
+      | "marketing-tools"
+      | "platform"
+      | "staging"
+      | "production";
   };
 };
 
-export const baseUrl = "https://wroom.io";
+export const baseUrl = (() => {
+  switch (process.env.SM_ENV) {
```
xrutayisire, line 24: ⚠️ For me, we should support "production" case. With prismic.io, and run the tests against prod.

I even think that we could make the CI of main run the test against production? WDYT?

xrutayisire (changes requested):

playwright/playwright.config.ts
jomifepe: Done ✅

playwright/playwright.config.ts
jomifepe: Discussed in slack, short answer:

Sound like a good pre-release check to me, changed to `main` E2E against `production` and the remaining against `dev-tools`

playwright/playwright.config.ts
```diff
@@ -14,10 +14,27 @@ declare const process: {
...
+    case "dev-tools":
+    case "marketing-tools":
+    case "platform":
+      return `https://${process.env.SM_ENV}-wroom.com/`;
+    case "production":
+      return "https://prismic.io";
```
xrutayisire: ❓Quick question, squad env use `/` but production and stage use without, no problem for you?

playwright/playwright.config.ts
```diff
@@ -14,10 +14,27 @@ declare const process: {
...
+    default:
```
xrutayisire, line 33: 💡 As our yarn play command defaults to dev-tools, should we make it the default, WDYT? 😬

playwright/playwright.config.ts
```diff
@@ -14,10 +14,27 @@ declare const process: {
...
     MANAGEV2_AUDIENCE: string;
     E2E_REPOSITORY: string | undefined;
+    SM_ENV:
+      | "dev-tools"
+      | "marketing-tools"
+      | "platform"
+      | "staging"
+      | "production";
   };
 };
```
xrutayisire, line 22: Actually I totally forgot on the previous PR, but we want to have a way to NOT create a repo if we want to run test locally on a yarn play repo for example, right?

Something like: https://github.com/prismicio/customtypes-api/blob/master/__tests__/api/.env.sample
As you see there is a REPOSITORY ENV you can define if you want, and it will use it.

Second thing I found I forgot for the first PR is having actually a file like this, `.sample` so we can copy and paste it and it help us to run things locally.

playwright/playwright.config.ts
```diff
@@ -14,10 +14,27 @@ declare const process: {
...
 
-export const baseUrl = "https://wroom.io";
+export const baseUrl = (() => {
```
xrutayisire, line 23: Something else I found out, is I think it may not work now, at least not all features of the e2e-test-utils client.
The production secret key for manage is different than stage.

Is it required to pass manage settings? If not I would suggest removing it, and deal when we need that, because manage is to be able to upgrade repo and we don't now.

xrutayisire (changes requested):

playwright/playwright.config.ts
jomifepe: Yeah that makes sense, changed ✅

playwright/playwright.config.ts
jomifepe: Changed for consistency ✅

playwright/playwright.config.ts
jomifepe: Yeah that makes sense, haven't thought of that.

Just changed to look to `E2E_REPOSITORY` and avoid creating a temporary repo.

playwright/playwright.config.ts
jomifepe: Got it, I added it initially, thinking it was required. I see that it's optional and we probably don't need it. 

One less thing, removed ✅

playwright/.env.test.example
```diff
@@ -0,0 +1,6 @@
+PLAYWRIGHT_ADMIN_USERNAME=first.last@prismic.io
+PLAYWRIGHT_ADMIN_PASSWORD=secret
+
+SM_ENV=dev-tools # optional (default: dev-tools) - production, staging, dev-tools, marketing-tools, platform
+PRISMIC_CLUSTER=baz # optional
```
xrutayisire: You can remove prismic cluster btw, stage only have one now, so should be good, you can hard code "exp" directly when it's staging and calling the lib.
This will actually make one less env var to manage.

xrutayisire (changes requested):
One last comment related to the cluster and we are good! Nice work!

playwright/.env.test.example
jomifepe: Should I leave it unset when it's not staging?

xrutayisire: You know what I think empty cluster will work. If you check one run like that and if yes let's remove it totally, even better for us, simpler.

jomifepe: Yes, confirmed, will leave it out!

jomifepe: Done ✅

xrutayisire (approved):
LGTM!

PR #1596, opened 2025-04-16
feat: add image to slice to slices page

levimykel:
Resolves: [DT-2678](https://linear.app/prismic/issue/DT-2678/aauser-i-can-use-image-to-slice-generation-on-the-slices-page)

### Description

This PR updates the Slices page:
- Replaces the Create button with a dropdown menu to generate from images or create from scratch
- Refactors `GenerateSliceWithAiModal` component to be usable on the Slices page
- Updates tests with new dropdown button
- Replaces old failing unit tests with e2e tests
- Fixes small issue with dropdown menu name in custom type SliceZone builder

### Checklist

### Preview

https://github.com/user-attachments/assets/595c4799-6e29-4e2a-942c-7337e24b7a85

### How to QA [^1]

[^1]:


packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
```diff
@@ -168,9 +177,33 @@ export function GenerateSliceWithAiModal(props: GenerateSliceWithAiModalProps) {
       .then(async ({ slices, library }) => {
         if (currentId !== id.current) return;
         id.current = crypto.randomUUID();
-        await onSuccess({ slices, library });
+        const serverState = await getState();
+        createSliceSuccess(serverState.libraries);
+        onSuccess({ slices, library });
```
jomifepe, line 184: 💡 Nit: I would rename this callback prop to something like `onCreateSliceSuccess` just to try to tell the dev that it does not happen after everything is executed anymore, but rather after the `createSliceSuccess` call.

packages/slice-machine/src/pages/slices.tsx
```diff
@@ -83,17 +114,54 @@ const SlicesIndex: React.FunctionComponent = () => {
...
+                      color="white"
+                    />
+                  )}
+                  onSelect={() => setIsCreateSliceModalOpen(true)}
+                  description="Build a custom Slice your way."
+                >
+                  Start from scratch
+                </DropdownMenuItem>
+              </DropdownMenuContent>
+            </DropdownMenu>
```
jomifepe: 💡 Nit: This is the same code as in the `CustomTypeBuilder`, the flag is also only used for it. Do you think we should extract a component for it to keep it consistent?

jomifepe (approved):
LGTM!

packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
dani-mp: even better we could rename the whole modal to CreateSliceWithAiModal or something like that, to reflect better what it does, and the callback can still be `onSucess`

playwright/tests/slices/slicesList.spec.ts
```diff
@@ -235,3 +242,50 @@ test("I can update the screenshot", async ({ slicesListPage, slice }) => {
...
+  await slicesListPage.addSliceDropdownCreateNewAction.click();
+
+  const sliceName = "Slice" + generateRandomId();
+  await slicesListPage.createSliceDialog.createSlice(sliceName);
+
+  await sliceBuilderPage.checkBreadcrumb(sliceName);
+  expect(trackCallCount).toBe(1);
+});
+
+test("I will not send a tracking event when I fail to create a slice", async ({
```
dani-mp: this test can be merged with the one above

- create a slice normally
- check that the event is send
- mock the create call to fail
- check that the event is not send

what we gain is that the setup for both is the same, so it feels like the same path; also, the first working gives confidence that the logic of the second one works too

packages/slice-machine/src/features/customTypes/customTypesBuilder/GenerateSliceWithAiModal/GenerateSliceWithAiModal.tsx
levimykel: Done ☑️

packages/slice-machine/src/pages/slices.tsx
levimykel: Done ☑️

playwright/tests/slices/slicesList.spec.ts
levimykel: Done ☑️

packages/slice-machine/src/features/slices/AddSliceDropdown.tsx
```diff
@@ -0,0 +1,142 @@
...
+  DropdownMenuTrigger,
+} from "@prismicio/editor-ui";
+
+import { useAiSliceGenerationExperiment } from "@/features/builder/useAiSliceGenerationExperiment";
+import { ComponentUI } from "@/legacy/lib/models/common/ComponentUI";
+
+import { SliceTemplate } from "../slicesTemplates/useSlicesTemplates";
+
+type AddSliceDropdownProps = {
+  onOpenCreateSliceFromImageModal?: () => Promise<void>;
```
dani-mp: not sure why this one is optional - also, it shouldn't return a promise either imo

packages/slice-machine/src/features/slices/AddSliceDropdown.tsx
```diff
@@ -0,0 +1,142 @@
...
+  (
+    | {
+        availableSlicesToAdd?: never;
+        onOpenUpdateSliceZoneModal?: never;
+      }
+    | {
+        availableSlicesToAdd: ComponentUI[];
+        onOpenUpdateSliceZoneModal: () => void;
+      }
+  );
```
dani-mp: same comment about these, i don't understand why we need this complex typing

either the parent pass or not a callback for a certain option to have it available/enabled, or always pass a callback for all options and the dropdown itself decides whether to show them or not based on templates available, feature flags, etc. 

but i don't understand why all these different type variations are needed

maybe it's a good idea to do this refactor in its own pr?

levimykel: Good call, I've reverted the refactor and will do it in another PR.

dani-mp (approved):

PR #1600, opened 2025-04-30
[DT-2679] Update linked content relationships on api id update

jomifepe:
Resolves: DT-2679

#### What was done: 

- On custom type API ID update, go through all the existing custom types and slices and check for content relationships that have an outdated ID, and update them.
- Refactor the signature of some methods and components to have some more context of the update reach the manager updating function.

#### Custom relationship in a page/custom type on the left and in a slice on the right

https://github.com/user-attachments/assets/55215ca0-6f2c-4610-8887-3eb779dbb096

### How to QA [^1]

[^1]:


packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -177,9 +307,77 @@ export class CustomTypesManager extends BaseManager {
 			args,
 		);
 
-		return {
-			errors: hookResult.errors,
-		};
+		const { model, updateMeta } = args;
+
+		if (updateMeta?.fieldIdChanged) {
```
xrutayisire, line 316: ❓ For you, the idea would be to use this `updateMeta` when a field is deleted in a custom type? Like a `fieldIdDeleted` property?

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -82,6 +89,13 @@ type CustomTypesMachineManagerDeleteCustomTypeReturnType = {
 	errors: (DecodeError | HookError)[];
 };
 
+type CustomTypeUpdateMeta = {
```
xrutayisire: ❓⚠️ To see, if I correctly understand, this means that only one field id can be changed for a specific custom type update? Like if we batched saves and then call the manager update custom type endpoint, this would not work, correct?

If it's yes, I think this approach will not work with the auto save. Let's say the manager call from SM UI to SM Managers is taking some time (file system a bit slow), and meanwhile you edit multiple api ids, only the last one will be called. It works today, as the whole custom type is always fully defined. But here I guess you will lose your field id updated meta info.

Technically, we are doing this:

`packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx`
```
const setCustomType = useCallback(
    (customType: CustomType, onSaveCallback?: () => void) => {
      setCustomTypeState(customType);
      setNextAction(async () => {
        const { errors } = await updateCustomType(customType);

        if (errors.length > 0) {
          throw errors;
        }

        // Update available custom types store with new custom type
        stableSaveCustomTypeSuccess(customType);

        syncChanges();
        onSaveCallback?.();
      });
    },
    [setNextAction, stableSaveCustomTypeSuccess, syncChanges],
  );
```

So here we are setting the next action to be sent to the action queue. But the action queue is two states queue (maybe the name is too confusing 🫤). 

`packages/slice-machine/src/hooks/useActionQueue.tsx`
```
type ActionQueueStack = {
  pendingAction: NextAction | undefined;
  nextAction: NextAction | undefined;
};
```

So there is only ONE `nextAction` as we don't care to save the intermediate custom type update, we always PASS the whole thing we want to update. It works the same way for slices.

I think you could first try to reproduce what I'm saying and then see to fix the issue.

xrutayisire: 💡 I think it could be nice to have simple comments explain what is `previousPath` and `newPath`, mostly what is a `path` here

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -177,9 +307,77 @@ export class CustomTypesManager extends BaseManager {
...
+						onField: ({ field }) => {
+							return this.updateFieldContentRelationships({
+								field,
+								previousPath,
+								newPath,
+							});
+						},
+					});
+
+					const hookResult = await this.sliceMachinePluginRunner.callHook(
```
xrutayisire: 💡 I think I would do a promise all and batch all the custom type updates (sane for slices after). So when there are an errors, we don't stop at the first one. 

I feel like it's better to try to update as much as possible all the api ids, as that's it, the custom type is already updated on the file system at this point.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -167,6 +181,122 @@ export class CustomTypesManager extends BaseManager {
...
+				...field.config,
+				customtypes: this.updateCRCustomTypes({
+					...updateMeta,
+					customTypes: field.config.customtypes.slice(),
+				}),
+			},
+		};
+	}
+
 	async updateCustomType(
```
xrutayisire: 💡 I wonder if we should not do a rollback system 😬 But maybe not now, we can just keep that in mind. Like we keep the old custom types / slices we update. And in case anything is failing, we rollback everything and tell the user about the problem. 
Here we can leave the user in an inconsistent state without really knowing what to do next.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -167,6 +181,122 @@ export class CustomTypesManager extends BaseManager {
...
+		const [previousId] = previousPath;
+		const [newId] = newPath;
+
+		if (!previousId || !newId) {
+			return customType;
+		}
+
+		if (typeof customType === "string") {
+			if (customType === previousId && customType !== newId) {
+				return newId; // update to new api id
```
xrutayisire: ❓ I don't understand the update of the newId here? Is this for the custom type id? How can it be changed?

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -167,6 +181,122 @@ export class CustomTypesManager extends BaseManager {
...
+					const previousId = previousPath[1];
+					const newId = newPath[1];
+
+					if (!previousId || !newId) {
+						return field;
+					}
+
+					if (typeof field === "string") {
+						if (field === previousId && field !== newId) {
+							return newId; // update to new api id
```
xrutayisire: ❓Here it's the field id we update, right?

I would add comments a bit everywhere where naming is generic to ensure it's clear. I think adding a bit more explanation would help maintenance later on.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -167,6 +181,122 @@ export class CustomTypesManager extends BaseManager {
...
+					}
+
+					return {
+						...field,
+						customtypes: field.customtypes.map((customType) => {
+							return this.updateCRCustomType({
+								customType,
+								previousPath,
+								newPath,
+								// TODO: Fix types. The second level customtypes are not typed
```
xrutayisire: ⚠️ I would like to have tickets for these kinds of things. So you can reference it in the code here, and explain in the ticket what needs to be done. Present this at daily so we can prioritize it 🙂

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -167,6 +181,122 @@ export class CustomTypesManager extends BaseManager {
 		};
 	}
 
+	private updateCRCustomType(
+		args: { customType: CRCustomType } & CustomTypeUpdateMeta,
+	): CRCustomType {
+		const { customType: customTypeArg, previousPath, newPath } = args;
+
+		let customType = customTypeArg;
+		if (typeof customTypeArg === "object") {
```
xrutayisire: 💡 What you want to do with the `.slice()` + this here is a deep clone? Should you just do a `structuredClone` instead of the `.slice()`?

xrutayisire (changes requested):
Nice work to jump on the topic with us 🙂
Added comments, mainly one or two could be big and the rest more smaller improvements.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: Yes, that makes sense, just added them ✅

I also extracted a shared type to avoid having too much duplication.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: Yeah, that makes total sense, changed ✅

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: Do you mean roll back the whole customtype update? Or just the side-effects that update the CR ids?

xrutayisire: I would say everything, because if not, we are not in synced.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: I don't think we need to use `structuredClone`, we only need to shallow clone as we go. I just changed the code to do that in a better way. WDYT?

Side note: This won't matter too much because the objects we're manipulating were read from disk, so if they get directly mutated, we won't have any unexpected side effects.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: Yeah, I still planned to fix this in this PR, but didn't think it would be a blocker for the reviews, so I didn't create a ticket for it.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: We can, for consistency, but we don't really need to, because delete is something we can track by looking at the changed model.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: You're right. Even though this doesn't affect anything, we don't need to do it, because we're not able to change API IDs in the UI. Changed ✅

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: Just added some comments to make it clearer ✅

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -167,6 +180,247 @@ export class CustomTypesManager extends BaseManager {
...
+						return newId;
+					}
+
+					return field;
+				}
+
+				if (field.id === previousId && field.id !== newId) {
+					// We have reached a field id that matches the id that was renamed,
+					// so we update it new one.
+					// Since field is not a string, we don't exit, as we might have
```
jomifepe, line 232: I'm a bit uncertain about this statement. Even though it won't hurt to keep going, we might be ok by returning here, as I don't think we can have more than one rename in the same call by using the UI.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: Makes sense, I created a [small subtask](https://linear.app/prismic/issue/DT-2684/implement-rollback-system-for-when-the-content-relationships-fail-to) and will tackle it there!

packages/manager/src/managers/customTypes/CustomTypesManager.ts
xrutayisire: Ok, if you don't do it in this PR you can create a ticket.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
xrutayisire: Hum but how will you know the diff between a rename or a delete if you use only the diff? You will combine it with the updateMeta object?

packages/manager/src/managers/customTypes/CustomTypesManager.ts
xrutayisire: Actually with what I explained with the queue, I think we can

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -167,6 +180,247 @@ export class CustomTypesManager extends BaseManager {
 		};
 	}
 
+	private updateCRCustomType(
+		args: { customType: CRCustomType } & CustomTypeFieldIdChangedMeta,
+	): { customType: CRCustomType; changed: boolean } {
+		const { previousPath, newPath } = args;
+
+		let changed = false;
```
xrutayisire: I don't understand this new changed things, it was not here before, right?

Does it correspond to a PR review comment? To a bug? Can you explain a bit more.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: Ah, you're right, then I would say we can use this object for the delete.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: It was not there, and it's not in response to a PR comment; It was an optimization I thought about doing initially, because we were unnecessarily calling hooks and writing to the disk without any changes, which could contribute to a delay in the updates.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: I can see the issue and I was able to reproduce. That's definitely something that could happen, since we have optimistic updates, even though the chances are low. I created a [ticket](https://linear.app/prismic/issue/DT-2685) for it as well!

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: I improved it, but, as we discussed, I will remove the recursive call later on in a new PR, as we probably won't need it.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -82,6 +88,21 @@ type CustomTypesMachineManagerDeleteCustomTypeReturnType = {
 	errors: (DecodeError | HookError)[];
 };
 
+type CustomTypeFieldIdChangedMeta = NonNullable<
+	NonNullable<CustomTypeUpdateHookData["updateMeta"]>["fieldIdChanged"]
+>;
+
+type CrCustomTypes = readonly CrCustomType[];
```
xrutayisire, line 95: ❓I think we should have that on types-internal no? If we change types-internal this needs to be aligned and I don't like that we just recreate it manually.

Can you track this to have it there if yes

xrutayisire (approved):
LGTM for the first version without the improvements we talked together, just added a comment about something I just saw that we could improve I believe.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: I agree, I was gonna suggest exporting this from types internal, but by removing the recursive there's less need for these types.

xrutayisire: So I propose you keep that in mind and if at the end of the recursive PR we still need that, we create a ticket for types internal, good for you?

jomifepe: Sounds good

PR #1602, opened 2025-05-08
feat: improve slice create placeholder with mcp server info

xrutayisire:
Resolves: N/A

### Description

- Update the slice code placeholder to have info about our new MCP server

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/adapter-next/src/hooks/slice-create.ts
+++ b/packages/adapter-next/src/hooks/slice-create.ts
@@ -39,6 +39,36 @@ const createComponentFile = async ({
 		options,
 	});
 
+	const placeholder = `
+		Placeholder component for ${data.model.id} (variation: {slice.variation}) slices.
+		<br />
+		<strong>You can edit this slice directly in your code editor.</strong>
+		{/**
+		 * 💡 Use Prismic MCP with your code editor
+		 *
+		 * Get AI-powered help to build your slice components — based on your actual model.
+		 *
+		 * ▶️ Setup:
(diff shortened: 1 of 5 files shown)


packages/adapter-next/src/hooks/slice-create.ts
```diff
@@ -39,6 +39,36 @@ const createComponentFile = async ({
...
+		 * {
+		 *   "mcpServers": {
+		 *     "Prismic MCP": {
+		 *       "command": "npx",
+		 *       "args": ["-y", "@prismicio/mcp-server"]
+		 *     }
+		 *   }
+		 * }
+		 *
+		 * 2. Select Claude 3.7 Sonnet for better output (more for a recommendation)
```
dani-mp: this line will get outdated soon and could be misleading as soon as there's something better out there

dani-mp (commented):
left one comment, applied to all cases

which takes me to the next question: is there a way to put the placeholder in a single place? it seems that all explain the same thing

packages/adapter-next/src/hooks/slice-create.ts
xrutayisire: Yes we will need to keep it up-to-date with our recommendation. But today we cannot let them use whatever, too big a difference.

For the sharing comment, the problem is that it's never fully the same format, how the comment is done, format etc. :/

I tried, and it was more complex than just updating each adapter in the file directly.

dani-mp (approved):

jomifepe (approved):

PR #1605, opened 2025-05-09
[DT-2686] test: content relationship update pure functions

jomifepe:
Resolves: DT-2686

### Description

Add unit tests for the `updateCustomTypeContentRelationships` and `updateSharedSliceContentRelationships` functions.

### How to QA [^1]

[^1]:


packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe, line 1: Diff helper: Since most functions were pure and needed to be tested, they were moved outside the class. The traverse calls were also extracted into two new functions with a callback, so that we could test the result.

packages/manager/test/SliceMachineManager-customTypes-updateCustomType.test.ts
```diff
@@ -40,3 +44,295 @@ it("throws if plugins have not been initialized", async (ctx) => {
...
+		});
+
+		expect(onUpdate).toHaveBeenCalledTimes(3);
+		expect(onUpdate).toHaveBeenCalledWith(
+			getOneLevelCrModel("authorLastName_CHANGED"),
+		); // changed
+		expect(onUpdate).toHaveBeenCalledWith(getOneLevelCrModel("address")); // not changed
+		expect(onUpdate).toHaveBeenCalledWith(
+			getOneLevelCrModel("address", "authorLastName_CHANGED"),
+		); // changed
```
levimykel, line 105: :question: This is showing that the `getOneLevelCrModel` function is being called, but is it possible to check if the ids have actually been changed?

packages/manager/test/SliceMachineManager-customTypes-updateCustomType.test.ts
```diff
@@ -40,3 +44,295 @@ it("throws if plugins have not been initialized", async (ctx) => {
...
+											id: "author",
+											fields: ["authorFirstName", ...ids],
+										},
+									],
+								},
+							},
+						},
+					},
+				],
+			} as const;
```
levimykel: :question: Is this `as const` needed?

packages/manager/test/SliceMachineManager-customTypes-updateCustomType.test.ts
```diff
@@ -40,3 +44,295 @@ it("throws if plugins have not been initialized", async (ctx) => {
...
+												},
+											],
+										},
+									],
+								},
+							},
+						},
+					},
+				],
+			} as const;
```
levimykel: ❓ Is this as const needed?

levimykel (commented):
This looks good. The big question I have with this is if it's possible to check if the ids have actually been changed. As I understand the test is only testing if `onUpdate` has been called with the function and args that you defined.

I left a couple other questions for you as well.

packages/manager/test/SliceMachineManager-customTypes-updateCustomType.test.ts
jomifepe: We're actually testing the id changes here because the functions loop over the models, traverse, and trigger `onUpdate` after updating the ids, so we test if `onUpdate` was called with those updated ids.

packages/manager/test/SliceMachineManager-customTypes-updateCustomType.test.ts
jomifepe: I would actually like to remove it, but for some reason I haven't yet figured out, `types-internal` imports are not being correctly resolved here, so I used `as const` to have the types matching.

jomifepe: Currently testing typing the functions, to see if the CI also sees the import error of it it's just something in my local.

jomifepe: Nope 🥲 I'm not seeing anything in the configuration that might cause this.

Screenshot 2025-05-13 at 10 30 09

jomifepe: Just fixed it, looks like it needs a different import. Should be good!

packages/manager/test/SliceMachineManager-customTypes-updateCustomType.test.ts
jomifepe: Same as above!

packages/manager/test/SliceMachineManager-customTypes-updateCustomType.test.ts
levimykel: Okay, cool, thanks for the explanation.

levimykel (approved):
Looks good 🚀

PR #1606, opened 2025-05-09
[DT-2688] fix: remove CR update recursion and prevent different path updates

jomifepe:
Resolves: DT-2688

### Description

1. Fix: Prevent updating field that have the same field name but different path from the update
2. Refactor: Remove an unnecessary recursive call and deals with the updates inside the map itself. This decreases the type complexity and removes a `@ts-expect-error`.
3. Refactor: Simplify some logic in the `updateCRCustomType` function.

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/manager/src/managers/customTypes/CustomTypesManager.ts
+++ b/packages/manager/src/managers/customTypes/CustomTypesManager.ts
@@ -191,102 +191,94 @@ export class CustomTypesManager extends BaseManager {
 	private updateCRCustomType(
 		args: { customType: CrCustomType } & CustomTypeFieldIdChangedMeta,
 	): CrCustomType {
-		const { previousPath, newPath } = args;
-
-		const customType = shallowClone(args.customType);
-
-		const [previousCustomTypeId, previousFieldId] = previousPath;
-		const [newCustomTypeId, newFieldId] = newPath;
+		const [previousCustomTypeId, previousFieldId] = args.previousPath;
+		const [newCustomTypeId, newFieldId] = args.newPath;
 
 		if (!previousCustomTypeId || !newCustomTypeId) {
 			throw new Error(
-				"Didn't find any customtype id, which should not be possible.",
+				"Could not find a customtype id in previousPath and/or newPath, which should not be possible.",
 			);
 		}
 
 		if (!previousFieldId || !newFieldId) {
 			throw new Error(
-				"Didn't find any field id, which should not be possible.",
+				"Could not find a field id in previousPath and/or newPath, which should not be possible.",
 			);
 		}
 
-		if (typeof customType === "string") {
-			return customType; // we don't support custom type id renaming
+		const customType = shallowCloneIfObject(args.customType);
+
+		if (typeof customType === "string" || !customType.fields) {
+			return customType;
 		}
 
 		const matchedCustomTypeId = customType.id === previousCustomTypeId;
 
-		if (customType.fields) {
-			const newFields = customType.fields.map((fieldArg) => {
-				const nestedField = shallowClone(fieldArg);
+		const newFields = customType.fields.map((fieldArg) => {
+			const nestedField = shallowCloneIfObject(fieldArg);
 
-				if (typeof nestedField === "string") {
-					if (
-						matchedCustomTypeId &&
-						nestedField === previousFieldId &&
-						nestedField !== newFieldId
-					) {
-						// We have reached a field id that matches the id that was renamed,
-						// so we update it new one. The field is a string, so return the new
-						// id.
-						return newFieldId;
-					}
-
-					return nestedField;
+			if (typeof nestedField === "string") {
+				if (
+					matchedCustomTypeId &&
+					nestedField === previousFieldId &&
+					nestedField !== newFieldId
+				) {
+					// We have reached a field id that matches the id that was renamed,
+					// so we update it new one. The field is a string, so return the new
+					// id.
+					return newFieldId;
 				}
 
-				if (nestedField.id === previousFieldId) {
-					if (matchedCustomTypeId && nestedField.id !== newFieldId) {
-						// We have reached a field id that matches the id that was renamed,
-						// so we update it new one.
-						// Since field is not a string, we don't exit, as we might have
-						// something to update further down in customtypes.
-						nestedField.id = newFieldId;
-					}
-				}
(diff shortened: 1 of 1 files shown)


packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -236,24 +234,36 @@ export class CustomTypesManager extends BaseManager {
...
+					customtypes: field.customtypes.map((customTypeArg) => {
+						const customType = shallowClone(customTypeArg);
+						const previousId = previousPath[2];
+						const newId = newPath[2];
+
+						if (!previousId || !newId) {
+							return customType;
+						}
+
+						if (typeof customType === "string") {
```
xrutayisire: ⚠️ I'm maybe lost but I don't get it. This seems to be custom type id here. But the goal is to update the fields right? Not the custom type id

xrutayisire (changes requested):

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: You're right, it's the same case I had previously above, I overlooked it initially. This will not happen because we cannot rename custom type ids, so the code is not doing anything.

jomifepe: Nice catch! Removed ✅

xrutayisire: Hum, I still don't get it. The goal is to go down one more level to rename the fields of the custom type that could have been renamed, no?

Do you have unit test with nested custom types? Could help you understand better all the use cases

jomifepe: After discussing it in slack, there was in fact a missing block of code here, which was handled before with the recursion. It should be fixed now! 

The unit tests were implemented in a separate PR on top of the base branch, so it wasn't yet testing this refactor. I have now changed the base of the unit tests PR so that it's also testing this code.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -236,24 +234,40 @@ export class CustomTypesManager extends BaseManager {
...
 					...field,
-					customtypes: field.customtypes.map((customType) => {
-						return this.updateCRCustomType({
-							customType,
-							previousPath,
-							newPath,
-						});
+					customtypes: field.customtypes.map((customTypeArg) => {
+						const customType = shallowClone(customTypeArg);
+						const previousId = previousPath[0];
```
xrutayisire: ⚠️ I'm not sure to understand how this can work, maybe I'm missing something, but previousPath[0] and [1] was already accessed. So here it's taking the same as the one before, is that normal? Redoing the same conditions?

xrutayisire (changes requested):

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: It can be confusing. We might not have a match on the first level of the relationship, but rather on the second one (nested), and since we entered a new custom type, we have to start checking the path from the beginning. Initially, I implemented it with a variable being incremented, but since we only have two levels, we might as well access the index directly.

For example:
```ts
// updateMeta.fieldIdChanged 
{
  previousPath: ["profession", "name"],
  newPath: ["profession", "name2"]
}

// field
{
  type: "Link",
  config: {
    select: "document",
    customtypes: [
      {
        id: "author", // index: 0
        fields: [
          "firstName", // index: 1
          "lastName",  // index: 1
          {
            id: "professionCr", // index: 1
            customtypes: [
              {
                id: "profession", // index: 0
                fields: [
                  "name",    // index: 1 🎯 Match: name → name2
                  "industry" // index: 1 
                ],
              },
            ],
          },
        ],
      },
    ],
  },
}
```

jomifepe: Resolving: After some private discussion, we improved the structure and reused some variables to improve the readability.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -188,75 +188,97 @@ export class CustomTypesManager extends BaseManager {
...
+					}
 
-			// @ts-expect-error We know that at this level we are returning the
-			// right properties, but TypeScript will not trust it because it might
-			// also have customtypes. This is because the type is not fully
-			// recursive, it just has two levels of depth.
-			return {
-				id: customType.id,
-				fields: newFields,
+					const matchedNestedCustomTypeId =
```
xrutayisire: 💡 No need to even go further if `matchedNestedCustomTypeId` is false no? Because it's second level

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -188,75 +188,97 @@ export class CustomTypesManager extends BaseManager {
...
 
-		if (customType.fields) {
-			const newFields = customType.fields.map((fieldArg) => {
-				const field = shallowClone(fieldArg);
+		const matchedCustomTypeId = customType.id === previousCustomTypeId;
 
-				const previousId = previousPath[1];
-				const newId = newPath[1];
+		const newFields = customType.fields.map((fieldArg) => {
+			const nestedField = shallowCloneIfObject(fieldArg);
```
xrutayisire: 💡 I would have written `customTypeField` here so it matches that it's the custom type field. The word nested here give a bit of confusion and I think it's better to keep it for `nestedCustomType` and `nestedCustomTypeField`, WDYT?

xrutayisire (changes requested):

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: Yeah, that's a good point, we can return early here. Done ✅

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: Makes sense, done ✅

xrutayisire (approved):
NICE! The function is really nicer and simpler to understand

PR #1607, opened 2025-05-09
[DT-2687] perf: only write model if content relationship id was updated

jomifepe:
Resolves: DT-2687

### Description

When updating the content relationship ids of other customtypes/slices after a field API ID rename, we only write the model to disk if something was changed.

### How to QA [^1]

[^1]:


packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -642,18 +634,28 @@ export function updateCustomTypeContentRelationships(
...
+
+				changed ||= update.changed;
+
+				return update.field;
 			},
 		});
 
-		onUpdate(updatedCustomTypeModel);
+		if (changed) {
+			onUpdate(updatedCustomTypeModel);
```
xrutayisire, line 641: ❓ I'm not sure to get all the indirection, when we are saving we are saving an entity. The whole entity. So why not just when we call the hook to save the custom type update we compare there. Not before not multiple times.

jomifepe: Not critical, but the idea behind it was to compare the least amount of properties, but I agree that it's not critical, so I will reduce the indirection.

jomifepe: Tell me what you think now

xrutayisire: I love it! I feel simpler to read as think. Like if it's equal we don't do anything, no need to read more.

xrutayisire (approved):

PR #1609, opened 2025-05-12
test: ensure playwright tests use a real repository

xrutayisire:
Resolves: [DT-2661](https://linear.app/prismic/issue/DT-2661)

### Description

- Use play to create a repository on the fly
- Create a setup before running playwright

### Checklist

### How to QA [^1]

[^1]:


playwright/e2eTestSetup.ts
```diff
@@ -0,0 +1,84 @@
...
+      `[setup] REPOSITORY is set, using existing repo (${repositoryName})`,
+    );
+  } else {
+    repositoryName = `${REPOSITORY_NAME_PREFIX}${randomString()}`;
+
+    console.log(
+      `[setup] REPOSITORY is not set, creating new repo (${repositoryName})`,
+    );
+
+    await exec(
```
jomifepe, line 54: ⚠️ Creating the repo only with `play` doesn't seem to set a master locale, and opening it in page builder throws an error. This is probably not critical, as we won't often look inside the repo in page builder and it will be deleted on teardown. It's just something I noticed changes from creating it with `e2e-tests-utils` vs `play`, and would be solved by creating the repo before and passing the name to `play`.

Screenshot 2025-05-12 at 10 51 30

jomifepe (approved):
Looks and works great! 💯

playwright/e2eTestSetup.ts
xrutayisire: I think it's actually a bug on squad env, the docs from a starter are not pushed, and play use the minimal starter.
We could try to fix the bug OR rework play script. Will check that. As you said not critical for us as of now, but may be annoying in the future.

.github/workflows/ci.yml
```diff
@@ -190,12 +190,8 @@ jobs:
             echo "SM_ENV=dev-tools" >> $GITHUB_ENV
           fi
 
-      - name: Set up Slice Machine project
-        run: |
-          yarn workspaces foreach --include '{@slicemachine/adapter-next,@slicemachine/manager,@slicemachine/plugin-kit,start-slicemachine}' --parallel --topological-dev run build
-          yarn workspace slice-machine-ui pack --out ../../e2e-projects/next/sm.tgz
-          yarn workspace cimsirp add -D ./sm.tgz
-          echo "telemetry=false" > ./e2e-projects/next/.prismicrc
```
dani-mp, line 198: could you explain why we don't need these anymore?

playwright/playwright.config.ts
```diff
@@ -40,15 +35,15 @@ export const cluster = process.env.SM_ENV === "staging" ? "exp" : undefined;
 export const auth = {
   username: process.env.PLAYWRIGHT_ADMIN_USERNAME,
   password: process.env.PLAYWRIGHT_ADMIN_PASSWORD,
-  storageState: ".auth/admin.json",
 };
 
 assert.ok(auth.username, "Missing PLAYWRIGHT_ADMIN_USERNAME env variable.");
 assert.ok(auth.password, "Missing PLAYWRIGHT_ADMIN_PASSWORD env variable.");
 
+export const REPOSITORY_NAME_PREFIX = "e2e-tests-";
```
dani-mp, line 43: what is this for? can't be confusing whether you have to provide a repository with a name like this or without the prefix becuase it's added later?

.github/workflows/ci.yml
xrutayisire: This was made to sync e2e-tests folder with our local code. But the play command is doing that by default, so no need.

BUT I totally forgot that it was also adding telemetry=false, so will add that to the existing setup so e2e tests don't trigger amplitude events.

playwright/playwright.config.ts
xrutayisire: It's not really for "you" meaning dev usage, it's purely for the CI and if you want the system to create a repo for you.
If you provide your own repo name, that's it you already have it, no need to manage that.

Do you think the naming should be different then? You have an idea?

.github/workflows/ci.yml
xrutayisire: Done: https://github.com/prismicio/slice-machine/pull/1609/commits/842ac739a266e25a9efe1e988d2e5d903073c38f

playwright/playwright.config.ts
dani-mp: i think it's unnecessary, i.e. we don't care how the repo is named or if it's named with that particular prefix, unless i'm missing something

the fact that you have to export it from here to use it somewhere else starts smelling a bit

but in the end it's not super important either, i understand now its meaning and that's enough for me, i think the name is ok too - but again, i don't see why it's important that the repositories have this particular prefix, doesn't add anything unless i'm missing something

dani-mp (approved):

playwright/playwright.config.ts
xrutayisire: It's an engineering decision to have all repositories created dynamically for e2e tests to have this pattern.
It helps us to recognize them easily for SRE, debug, maintenance, etc.

PR #1616, opened 2025-05-13
refactor: simplify custom relationship sync data flow

jomifepe:
### Description

Restructure updated paths property to a simple record of paths (`{ [previousPath: string]: string }`)

### How to QA [^1]

[^1]:


packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
```diff
@@ -153,6 +153,13 @@ export const CustomListItem = ({
       apiId: groupItem.key,
       newKey: groupItem.key,
       value: Groups.toSM(newGroupValue),
+      groupId: groupItem.key,
+      updateMeta: {
+        fieldDeletedOrIdChanged: {
+          previousPath: [groupItem.key, key],
+          newPath: null,
+        },
+      },
```
dani-mp: "update meta" sounds like a verb/action

i'd call it just `updates`, to be declarative

and i'd also use an object with key/value, key being the current path and value the new one

it works for deletes too, by just using null as a value

```suggestion
      updates: {
        [[groupItem.key, key].join("#")]: null
      },
```

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
```diff
@@ -153,6 +153,13 @@ export const CustomListItem = ({
       apiId: groupItem.key,
       newKey: groupItem.key,
       value: Groups.toSM(newGroupValue),
+      groupId: groupItem.key,
+      updateMeta: {
+        fieldDeletedOrIdChanged: {
+          previousPath: [groupItem.key, key],
```
xrutayisire: ❓ Actually where I'm lost is the group management, you're managing the group field selection, but it's not yes supported by the back-end no?

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
jomifepe: So I have tried a few things, and I have some concerns about this structure/naming:
1. It does not allow a `null` previous path. Even though we don't currently need details here for added fields, and we can get that information elsewhere, having dedicated properties like `previousPath` and `newPath` allows for more flexibility if we ever need something like that. My idea is to make the `updateMeta` as flexible as possible so that we don't have to add a new property to support another feature in the future.
2. Similar to the one above, removing the nested update name property (e.g., `fieldDeletedOrIdChanged`) makes the `updateMeta` property a bit too specific, and if we need to pass something else there besides path updates, we can no longer do so. I'm thinking about this property as general context about the update made.
3. The name `updates` is a bit too generic, the actual update is in the model passed down, so maybe this should rather be something like `updateDetails`.

That said, I think the action point here is renaming `updateMeta` to something that's clearer (I'm not too happy with "meta" either). WDYT of `updateDetails`?

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
jomifepe: Yeah, it's not supported, I just passed it to be correct, and because I did the same for the rename action.

packages/slice-machine/src/legacy/lib/models/common/widgets/Group/ListItem/index.jsx
dani-mp: 1. why would you need a null as a current path? i don't think we should complicate this more than the strictly necessary today
2. for the same reasons as above, if you already can think of a scenario that we'll need to address within 3 months, i'm happy to give it a thought, but otherwise i'd use the simpler name and data structure possible today, and change it in the future if needed
3. updateDetails is also generic, misleading (what are "details" anyway?) and again looks like an action, not a declarative property as we need

i think `updates` with the structure i shared above (and without `fieldDeletedOrIdChanged`) feels the nicest to me, but feel free to keep the current one or anything else, i'll let you folks decide

jomifepe: I'm not against it, I just had a different idea for the property, something like an update context bag, but I will try your approach.

jomifepe: Updated ✅

PR #1624, opened 2025-05-15
[DT-2684] feat: rollback custom type update on failed CR sync

jomifepe:
Resolves: DT-2684

### Description

When renaming a field API ID in a custom type, we update all existing content relationships with the new id, but if that update fails we want to rollback the API ID rename, otherwise we leave the user in a state that will need manual fixing.

### Preview

Demo with a purposely thrown error in the content relationship update code.

https://github.com/user-attachments/assets/81616c68-275f-4217-bcef-d51d35334e85

### How to QA [^1]

[^1]:


packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -237,49 +255,104 @@ export class CustomTypesManager extends BaseManager {
...
+						rollback: async () => {
+							await Promise.all(crUpdates.map((update) => update.rollback()));
+						},
 					};
 				}
 			}
 		}
 
-		return { errors: [] };
+		return { errors: [], rollback: () => Promise.resolve() };
```
xrutayisire: 💡 Can we not pass rollback at all?

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -237,49 +255,104 @@ export class CustomTypesManager extends BaseManager {
...
+
+		if (args.updateMeta?.fieldIdChanged) {
+			const customTypeRead = await this.readCustomType({ id: model.id });
+
+			if (customTypeRead.errors.length > 0) {
+				return { errors: customTypeRead.errors };
+			}
+			if (!customTypeRead.model) {
+				throw new Error(
+					"Read custom type without errors, but model is undefined.",
```
xrutayisire: 💡 It could benefit from more info on the error like custom type id, WDYT?

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -237,49 +255,104 @@ export class CustomTypesManager extends BaseManager {
...
 		args: CustomTypeUpdateHookData,
-	): Promise<OnlyHookErrors<CallHookReturnType<CustomTypeUpdateHook>>> {
+	): Promise<CustomTypesMachineManagerUpdateCustomTypeReturnType> {
 		assertPluginsInitialized(this.sliceMachinePluginRunner);
+		const { model } = args;
 
-		const hookResult = await this.sliceMachinePluginRunner.callHook(
+		let updateCrPromise: (() => Promise<{ errors: HookError[] }>) | undefined;
+
+		if (args.updateMeta?.fieldIdChanged) {
```
xrutayisire: 💡 I see you also have a `if (updateMeta?.fieldIdChanged) {` in `updateContentRelationships`, then it should be removed no?

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -237,49 +255,104 @@ export class CustomTypesManager extends BaseManager {
...
+					"Read custom type without errors, but model is undefined.",
+				);
+			}
+
+			const previousCustomType = customTypeRead.model;
+
+			updateCrPromise = async () => {
+				const crUpdateResult = await this.updateContentRelationships(args);
+
+				if (crUpdateResult.errors.length > 0) {
```
xrutayisire: I don't get all this management.

I see a global top function that create a promise of update content relationship but still update the custom type passed in args that is then updated anyway from the main function with:

```
const customTypeUpdateResult = await this.sliceMachinePluginRunner.callHook(
			"custom-type:update",
			{ model },
		);
```

I think it could be a bit clearer, don't you think it's too complex here? Like you have some duplication? Something could be removed?

xrutayisire (changes requested):

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: This was done so that all the content relationship code was in the same block, IMO is better for readability. I changed to the previous approach so you can see if you find it better, I'm okay with both!

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: Without changing the types, it could not be removed, but I just changed the function signature so that I don't have to check this again ✅

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: Added a bit more context ✅

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: Done ✅

xrutayisire (approved):

PR #1626, opened 2025-05-15
feat: improve Svelte 5 support

### LINKED ISSUE #28: Svelte 5
Hi!

I've recently started using Prismic together with a Svelte 5 app and noticed that the `prismic-svelte` library hasn't yet been upgraded to use Svelte 5. This is currently causing some type-related issues, especially when using the `PrismicRichText` component and for example use a `Heading1` component that is written i Svelte 5. 

It would be great if you could update the library to support Svelte 5 completly without doing a lot of workarounds. 

Thanks a lot for your great work!

Best regards,
Maja

lihbr:
Resolves: https://github.com/prismicio/prismic-svelte/issues/28

### Description

This pull request improves Svelte 5 support by doing the following:
- The adapter now detects if it's running against a Svelte 4 or 5 project
- Suggested snippets use the modern Svelte 5 syntax on Svelte 5 projects
- Slice template library was updated to propose Svelte 5 versions of component on Svelte 5 projects

### Checklist

> Couldn't test Svelte 5 because it requires mocking `package.json` within tests

### Preview

N/A

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/adapter-sveltekit/src/hooks/documentation-read.ts
+++ b/packages/adapter-sveltekit/src/hooks/documentation-read.ts
@@ -3,6 +3,7 @@ import { source } from "common-tags";
 import type { DocumentationReadHook } from "@slicemachine/plugin-kit";
 
 import { getJSFileExtension } from "../lib/getJSFileExtension";
+import { checkIsSvelte5 } from "../lib/checkIsSvelte5";
 
 import type { PluginOptions } from "../types";
 
@@ -22,6 +23,7 @@ export const documentationRead: DocumentationReadHook<PluginOptions> = async (
 		const { model } = data.data;
 
 		const pageDataExtension = await getJSFileExtension({ helpers, options });
+		const isSvelte5 = await checkIsSvelte5({ helpers });
 
 		const routePath = `src/routes/${model.repeatable ? "[uid]" : model.id}`;
 		const dataFilePath = `${routePath}/+page.server.${pageDataExtension}`;
@@ -72,17 +74,35 @@ export const documentationRead: DocumentationReadHook<PluginOptions> = async (
 			`;
 		}
 
-		let componentFileContent = source`
-			<script>
-				import { SliceZone } from "@prismicio/svelte";
+		let componentFileContent;
 
-				import { components } from "$lib/slices";
+		if (isSvelte5) {
+			// Svelte 5 syntax with runes
+			componentFileContent = source`
+				<script>
+					import { SliceZone } from "@prismicio/svelte";
 
-				export let data;
-			</script>
+					import { components } from "$lib/slices";
 
-			<SliceZone slices={data.page.data.slices} {components} />
-		`;
+					let { data } = $props();
+				</script>
+
+				<SliceZone slices={data.page.data.slices} components={components} />
+			`;
+		} else {
+			// Traditional Svelte 3/4 syntax
+			componentFileContent = source`
+				<script>
+					import { SliceZone } from "@prismicio/svelte";
+
+					import { components } from "$lib/slices";
+
+					export let data;
+				</script>
+
+				<SliceZone slices={data.page.data.slices} {components} />
+			`;
+		}
 
 		if (options.format) {
 			dataFileContent = await helpers.format(
--- a/packages/adapter-sveltekit/src/hooks/documentation-read.ts
+++ b/packages/adapter-sveltekit/src/hooks/documentation-read.ts
@@ -87,7 +87,7 @@ export const documentationRead: DocumentationReadHook<PluginOptions> = async (
 					let { data } = $props();
 				</script>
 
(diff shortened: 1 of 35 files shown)


packages/adapter-sveltekit/src/hooks/documentation-read.ts
```diff
@@ -72,17 +74,35 @@ export const documentationRead: DocumentationReadHook<PluginOptions> = async (
...
-				export let data;
-			</script>
+					import { components } from "$lib/slices";
 
-			<SliceZone slices={data.page.data.slices} {components} />
-		`;
+					let { data } = $props();
+				</script>
+
+				<SliceZone slices={data.page.data.slices} components={components} />
```
angeloashmore: This shortcut should still work in Svelte 5:

```suggestion
				<SliceZone slices={data.page.data.slices} {components} />
```

packages/adapter-sveltekit/src/hooks/documentation-read.ts
```diff
@@ -72,17 +74,35 @@ export const documentationRead: DocumentationReadHook<PluginOptions> = async (
 			`;
 		}
 
-		let componentFileContent = source`
-			<script>
-				import { SliceZone } from "@prismicio/svelte";
+		let componentFileContent;
 
-				import { components } from "$lib/slices";
+		if (isSvelte5) {
```
angeloashmore, line 79: What do you think of switching the perspective from treating Svelte 5 as the exception to treating Svelte <5 as the exception? Doing so lets us treat the latest Svelte version as the default.

```ts
if (await checkIsSvelteVersion("<5.0.0-alpha.0", { helpers })) {
  // ...
}
```

packages/adapter-sveltekit/src/lib/checkIsSvelte5.ts
```diff
@@ -0,0 +1,49 @@
...
+const checkIsSvelteN = async (
+	version: number,
+	args: CheckIsSvelteNArgs,
+): Promise<boolean> => {
+	try {
+		const packageJsonContent = await readProjectFile({
+			filename: "package.json",
+			encoding: "utf8",
+			helpers: args.helpers,
+		});
```
angeloashmore, line 24: It might be safer and more reliable to use `await import("svelte/package.json")` or something similar (like `require.resolve`) to read the installed package's version. Using that prevents false negatives when `latest`, `next`, etc. are used.

We used a similar technique elsewhere, but I can't find it. I think it was when we needed to detect when React 17 or 18 was used.

packages/adapter-sveltekit/src/plugin.ts
```diff
@@ -200,14 +201,16 @@ export const plugin = defineSliceMachinePlugin<PluginOptions>({
...
+
 			return await readSliceTemplateLibrary({
 				...data,
 				...context,
 				dirName: path.dirname(fileURLToPath(new URL(import.meta.url))),
 				templates: [Hero, CustomerLogos, AlternateGrid, CallToAction],
 				componentFileNames: {
-					js: "javascript.svelte",
-					ts: "typescript.svelte",
+					js: isSvelte5 ? "javascript.5.svelte" : "javascript.svelte",
```
angeloashmore: Related to an above comment, we can treat Svelte <5 as the exception in the file name rather than strictly 5.

We could use a descriptor rather than a version number, like `with-let-props`.

packages/adapter-sveltekit/src/hooks/project-init.ts
```diff
@@ -173,6 +173,8 @@ const createSliceSimulatorPage = async ({
 		return;
 	}
 
+	// Due to limitations in Svelte 4 on how slots can be typed, we cannot provide a Svelte 5-flavor
+	// version of this snippet, so we have to rely on the legacy Svelte 4 syntax.
```
angeloashmore, line 177: What is the limitation? Is there a page in the Svelte docs that explain it?

packages/adapter-sveltekit/src/hooks/documentation-read.ts
angeloashmore: This change also prevents weirdness when Svelte 6 comes out, which makes the `isSvelte5` variable name seem incorrect.

angeloashmore (changes requested):

packages/adapter-sveltekit/src/hooks/project-init.ts
lihbr: Sorry, it's poorly explained in the comment.

Svelte 4 codebases have to use slots to pass data to child components, which cannot be typed in a compatible way for Svelte 5 because of the introduction of snippets. Making a major of @prismicio/svelte solves the issue on `@prismicio/svelte`'s side, but I'm not sure how we'll address this in the adapter for the Simulator without maintaining two versions of it 🤔

lihbr:
Closing in favor of #1635

PR #1628, opened 2025-05-16
fix: refresh store after syncing content relationship ids

jomifepe:
### Description

After renaming a custom type field id, that triggers an update in all content relationships, only the first update is reflected in the store, so the user is able to push just one change before the UI refreshes with all the updates that were done.

#### Demo: 3 changes (1 custom type + 2 content relationships fields) to review instead of 1 (just custom type)

https://github.com/user-attachments/assets/21eb56c3-4c66-4e4d-bff2-857e6d862905

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
+++ b/packages/slice-machine/src/features/customTypes/customTypesBuilder/CustomTypeProvider.tsx
@@ -63,11 +63,11 @@ export function CustomTypeProvider(props: CustomTypeProviderProps) {
           throw errors;
         }
 
-        syncChanges();
-
-        // Refresh the store to get the updated custom types
+        // Refresh the store with the latest server state to get the updated
+        // custom types
         stableRefreshState(await getState());
 
+        syncChanges();
         onSaveCallback?.();
       });
     },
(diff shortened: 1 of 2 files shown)


packages/slice-machine/src/modules/slices/index.ts
```diff
@@ -34,6 +34,12 @@ export const sliceUpdateSuccess = createAction("SLICE/UPDATE_SUCCESS")<{
   component: ComponentUI;
 }>();
 
+export const sliceUpdateModelsSuccess = createAction(
```
xrutayisire: ⚠️ I have another proposal for you, as this PR as quite a lot of code for what we want + creates a new action that we could avoid.

Why not just using getState + refreshState or something like that, more global:

```
const serverState = await getState();
refreshState(serverState);
```

Then in two lines of code, it's covered.

xrutayisire (changes requested):

packages/slice-machine/src/modules/slices/index.ts
jomifepe: Yeah that also works, much shorter 😄

Changed ✅

xrutayisire (approved):
LGTM

PR #1634, opened 2025-05-26
[DT-2692] feat: pick content relationship fields

jomifepe:
Relates to: [DT-2692](https://linear.app/prismic/issue/DT-2692)

### Description

When adding or editing a new field, have the ability to go through all the available fields from any custom type in the library and pick multiple.

Currently, it's only possible to pick fields from the first level, and we plan to implement two levels. Picking fields from the second level will be implemented in a different PR.

### Preview

Screenshot 2025-05-27 at 18 40 21

https://github.com/user-attachments/assets/d9c3a0da-f0ba-4aca-9e3b-bc51d78738fa

### How to QA [^1]

1. Launch `slice-machine` in this branch
2. Add a new Content Relationship field to any page/custom type
3. Use the new TreeView picker (as seen in the preview above)

[^1]:


lihbr:
Is it necessary to display/allow selecting the `uid`? It's always returned by the API and is not part of the `data` object: https://200629-sms-hoy.cdn.prismic.io/api/v2/documents/search?q=%5B%5Bat%28document.type%2C+%22kitchen_sink_4%22%29%5D%5D&ref=aDbsjxEAACwAkpaF&fetchLinks=post.title

jomifepe:
> Is it necessary to display/allow selecting the `uid`? It's always returned by the API and is not part of the `data` object: https://200629-sms-hoy.cdn.prismic.io/api/v2/documents/search?q=%5B%5Bat%28document.type%2C+%22kitchen_sink_4%22%29%5D%5D&ref=aDbsjxEAACwAkpaF&fetchLinks=post.title

That's a great point! Just double checked this with the team and it should in fact be omitted.

Nice catch! Done ✅

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -0,0 +1,194 @@
...
+            subtitle={`(${countPickedFields(form.values)})`}
+          >
+            {customTypes.map((ct) => {
+              const count = countPickedFields(form.values[ct.id]);
+
+              return (
+                <TreeViewSection
+                  key={ct.id}
+                  title={ct.label}
+                  subtitle={count > 0 ? `(${count} fields exposed)` : undefined}
```
xrutayisire: 💡 As it's cheap I would handle the plural here for the case `1 field`.

packages/slice-machine/src/legacy/lib/models/common/widgets/ContentRelationship/Form.tsx
```diff
@@ -1,105 +1,67 @@
...
-import { selectAllCustomTypes } from "@/modules/availableCustomTypes";
-
-import { hasLocal } from "../../ModelData";
 
 const FormFields = {
   label: DefaultFields.label,
   id: DefaultFields.id,
   customtypes: {
-    validate: () => yup.array().of(yup.string()),
+    validate: () =>
```
xrutayisire, line 15: ⚠️ You cannot forget the legacy custom types format that all customers have today.

It crashed when having a string format in `customtypes`

Screenshot 2025-05-30 at 12 40 31

To handle all this part of legacy management, it can be an isolated separated PR if it adds too much code.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -0,0 +1,194 @@
...
+                  badge="Custom Type"
```
xrutayisire: 💡 It's a product decision, but I would advise writing "Custom type" when the format is `custom` or undefined and `Page type` when the format is `page`. It would help devs even more

Note I'm only adding the uppercase to the first word Custom or Page, as it's what we should do, we decided that in the past, but we always forgot. You can see the side menu or breadcrumb that it's like that.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -0,0 +1,194 @@
...
+}
+
+export function ContentRelationshipFieldPicker(
+  props: ContentRelationshipFieldPickerProps,
+) {
+  const { initialValues, onChange } = props;
+  const customTypes = useCustomTypes();
+
+  const stableOnChange = useStableCallback(onChange);
+  const form = useFormik<CustomTypeFieldMap>({
```
xrutayisire, line 33: ⚠️ Actually, I come back to the usage of Formik. This is not part of a legacy widget, right?
This is a totally new code in the new `features` folder, with an isolated form.

Why continue with Formik then and not propose something else?

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -0,0 +1,194 @@
...
+  useEffect(() => {
+    stableOnChange(convertFormToCustomTypes(form.values));
+  }, [form.values, stableOnChange]);
+
+  return (
+    <FormikContext.Provider value={form}>
+      <Box overflow="hidden" flexDirection="column" border borderRadius={6}>
+        <Box
+          border={{ bottom: true }}
+          padding={{ block: 16, inline: 16, top: 12 }}
```
xrutayisire: 💡A bit harder to read to have two properties to define the same thing but with an override, this would be clearer:

```suggestion
          padding={{ inline: 16, bottom: 16, top: 12 }}
```

Same number of parameters, but no override to understand.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+export function ContentRelationshipFieldPicker(
+  props: ContentRelationshipFieldPickerProps,
+) {
+  const { initialValues, onChange } = props;
+  const customTypes = useCustomTypes();
+
+  const stableOnChange = useStableCallback(onChange);
+  const form = useFormik<CustomTypeFieldMap>({
+    initialValues: initialValues ? getInitialValues(initialValues) : {},
```
xrutayisire: 💡 Naming was a bit challenging to understand at first, I think we could improve. Because you have a `getInitialValues` function that accepts a param `initialValues`. It's really strange. Either it's a transform or a different param name.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -0,0 +1,194 @@
...
+  id: string;
+  label: string;
+  fields: { id: string; label: string }[];
+}
+
+function useCustomTypes() {
+  const customTypes = useSelector(selectAllCustomTypes);
+  const simplifiedCustomTypes = customTypes.flatMap<SimplifiedCustomType>(
+    (customType) => {
+      if (!("local" in customType)) return [];
```
xrutayisire, line 143: 💡 Super nice condition with early exit, I like it!

Could you add a small comment for someone that doesn't know at all why we have to do this?

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+      const { id, label, tabs } = customType.local;
+      return {
+        id,
+        label: label ?? id,
+        fields: tabs.flatMap((tab) => {
+          return tab.value.flatMap((field) => {
+            // filter out uid field because it's a special field returned by the API
+            // and is not part of the data object in the document
+            if (field.key === "uid") return [];
```
xrutayisire: ⚠️ We cannot do that directly, the `key` is not relevant, it's the type that is important. You CAN technically have a UID name "toto". 

For example, on a single custom type, you add a UID field and name the api ID as you want.
Actually, could you verify, maybe with Levi, that a single custom type where you add a UID will still be shown automatically? Meaning, no need to select the field here?

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+            return {
+              id: field.key,
+              label: field.value.config?.label ?? field.key,
```
xrutayisire: 💡 I'm starting to wonder if it's better the label or the key. As the label is more marketing purpose and the key dev purpose. And here as we are for devs, maybe key is nice.

Could you ask an open question in dev tools design? I would like to know what Levi thinks also.

xrutayisire (changes requested):
Here is my initial review! UI looks super nice and aligned with design! ✨

packages/slice-machine/src/legacy/lib/models/common/widgets/ContentRelationship/Form.tsx
jomifepe: Yes! You're totally correct, my bad. I deleted the support for legacy while adapting the yup schemas.

Just committed a fix for this!

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: I totally planned to do this and forgot, thanks!

jomifepe: Added ✅

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: Renamed ✅

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: Makes sense! Changed to lowercase and created a ticket to confirm this with product.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: Done ✅

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: Good point! Posted the question, let's hear from design. Will make the change afterwards if we decide on displaying the `id`.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: This was discussed privately, and the action point for this PR was to filter fields with type `UID` or id `uid`.

Changed ✅

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: Discussed privately about not depending on `formik`, also maybe refactoring to `react-hook-form` because we want to remove `formik` in the future. The action point here is that I will do a refactor to decouple it, first without using any form manager. This will be done in a separate branch/PR.

xrutayisire: Agree to have this in a separate PR 🙂

xrutayisire (approved):
LGTM! 🚀

PR #1639, opened 2025-05-30
[DT-2692] feat: improve content relationship field picker state

jomifepe:
Relates to: DT-2692

### Description

- Decouple the Content Relationship field picker from formik and improve state management.
- Use custom type and field ids instead of labels (product requirement)

### Preview (same behavior as before)

Screenshot 2025-06-02 at 16 11 58

### How to QA [^1]

[^1]:

## Summary by CodeRabbit

- **Refactor**
	- Improved the custom type picker interface for content relationships, replacing the previous form state management with a more streamlined and type-safe approach.
	- Updated the custom type selection to reflect live form values instead of just initial values.

- **New Features**
	- Introduced a utility to accurately detect plain objects, enhancing data validation within the app.

### DIFF
--- a/packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
+++ b/packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
@@ -4,23 +4,38 @@ import {
   Text,
   TreeView,
   TreeViewCheckbox,
-  TreeViewCheckboxProps,
   TreeViewSection,
 } from "@prismicio/editor-ui";
-import { FormikContext, useField, useFormik } from "formik";
-import { useEffect } from "react";
+import { SetStateAction, useEffect, useMemo, useState } from "react";
 import { useSelector } from "react-redux";
 
 import { selectAllCustomTypes } from "@/modules/availableCustomTypes";
 
-type FieldMap = Record<string, boolean>;
-export type CustomTypeFieldMap = Record<string, FieldMap>;
+type PickerCheckboxField = {
+  type: "checkbox";
+  value: boolean;
+};
 
-type CustomTypeField = string | { id: string; fields: string[] };
+type PickerCustomTypeField = {
+  [fieldId: string]: PickerCheckboxField;
+};
+
+type PickerCustomTypeFields = {
+  [customTypeId: string]: PickerCustomTypeField;
+};
+
+// copy of types-internal types
+
+type TICustomType = {
+  id: string;
+  fields?: readonly string[] | undefined;
+};
+
+type TICustomTypeFields = readonly (string | TICustomType)[];
 
 interface ContentRelationshipFieldPickerProps {
-  initialValues: CustomTypeField[] | undefined;
-  onChange: (fields: CustomTypeField[]) => void;
+  initialValues: TICustomTypeFields | undefined;
+  onChange: (fields: TICustomTypeFields) => void;
 }
 
 export function ContentRelationshipFieldPicker(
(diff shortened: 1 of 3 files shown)


xrutayisire:
@coderabbitai review

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -1,213 +1,287 @@
...
 
-type SimplifiedCustomTypeField = {
-  id: string;
-  label: string;
-};
+        function onCheckedChange(value: boolean) {
+          onCustomTypeChange((currentFields) => ({
+            ...currentFields,
+            [field]: { type: "checkbox", value },
+          }));
```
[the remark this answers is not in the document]
jomifepe: Totally valid, we might want to activate `noUncheckedIndexedAccess`, and there are already quite a few instances of this to be fixed.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -1,213 +1,294 @@
...
+  customType: TICustomType;
+  fieldCheckMap: PickerCustomType | undefined;
+  onChange: (fn: Updater<PickerCustomType>) => void;
+}
+
+function TreeViewCustomType(props: TreeViewCustomTypeProps) {
+  const { customType, fieldCheckMap, onChange } = props;
+
+  const fieldCount = countPickedFields(fieldCheckMap);
+  const fieldCountLabel = fieldCount === 1 ? "1 field" : `${fieldCount} fields`;
```
xrutayisire, line 164: 💡 Can you use the pluralize function from editor support?

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -1,213 +1,294 @@
...
-  );
-}
+    <TreeViewSection
+      key={customType.id}
+      title={customType.id}
+      subtitle={fieldCount > 0 ? `(${fieldCountLabel} exposed)` : undefined}
+      badge="Custom type"
+    >
+      {customType.fields.map((field) => {
+        const checkboxState = fieldCheckMap?.[field];
```
xrutayisire: 💡 Reducing again the usage of another name to find:

```suggestion
        const checked = fieldCheckMap?.[field].value ?? false;
```

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -1,213 +1,294 @@
...
+ */
+type TICustomTypes = readonly (string | TICustomType)[];
 
-type CustomTypeField = string | { id: string; fields: string[] };
+interface TICustomType {
+  id: string;
+  fields: readonly string[];
+}
+
+type Updater<T> = (prev: T) => T;
```
xrutayisire: 💡 This is unnecessary too small abstraction at a level that creates indirection just for `(prev: T) => T`, you can use that directly everywhere and the code would be just more readable

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -1,213 +1,294 @@
...
-    />
-  );
-}
+    <TreeViewSection
+      key={customType.id}
+      title={customType.id}
+      subtitle={fieldCount > 0 ? `(${fieldCountLabel} exposed)` : undefined}
+      badge="Custom type"
+    >
+      {customType.fields.map((field) => {
```
xrutayisire: 💡 Here it's not the full field but the `fieldId`, I would ensure to name it like that to help clarity on what we manipulate.

```suggestion
      {customType.fields.map((fieldId) => {
```

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -1,213 +1,294 @@
...
-  const stableOnChange = useStableCallback(onChange);
-  const form = useFormik<CustomTypeFieldMap>({
-    initialValues: initialValues ? convertCustomTypesToForm(initialValues) : {},
-    onSubmit: () => undefined, // values will be updated on change
-  });
-
-  useEffect(() => {
-    stableOnChange(convertFormToCustomTypes(form.values));
-  }, [form.values, stableOnChange]);
+  function onCustomTypesChange(updater: Updater<PickerCustomTypes>) {
```
xrutayisire: ⚠️ Refactor suggestion to simplify `onChange` logic by removing the `Updater` function pattern. After re-reading it seems too complicated for what it is.

The current implementation introduces a generalized `Updater<T>` function at multiple levels of the component tree to perform immutable updates on the `PickerCustomTypes` map. While this pattern allows flexibility and functional purity, it adds unnecessary abstraction and indirection to what could otherwise be straightforward updates — especially in UI-driven, form-like interactions where local mutations followed by `onChange(newState)` are simpler to understand and maintain.

Here’s how the logic could be refactored without the updater function, replacing it with direct construction of new values at each level:


### 🧩 Top-level `onChange` usage in `ContentRelationshipFieldPicker`

**Before (current):**

```tsx
function onCustomTypesChange(updater: Updater<PickerCustomTypes>) {
  onChange(convertFieldCheckMapToCustomTypes(updater(fieldCheckMap)));
}
```

**After (simplified):**

```tsx
function onCustomTypeChange(customTypeId: string, newFields: PickerCustomType) { 
  onChange(convertFieldCheckMapToCustomTypes({
    ...fieldCheckMap,
    [customTypeId]: newFields,
  }));
}
```

### 🧩 In TreeViewCustomType

***Before:***

```tsx
const onCustomTypeChange = (updater: Updater<PickerCustomType>) => {
  onCustomTypesChange((currentCustomTypes) => ({
    ...currentCustomTypes,
    [customType.id]: updater(currentCustomTypes[customType.id] ?? {}),
  }));
};
```

***After:***

```tsx
function onFieldChange(fieldId: string, isChecked: boolean) {
  const currentFields = fieldCheckMap[customType.id] ?? {}
  const newFields = {
    ...currentFields,
    [fieldId]: { type: "checkbox", value: isChecked },
  };
  onChangeCustomType(customType.id, newFields);
}
```

### 🧩 In checkbox onCheckedChange callback:

***Before:***

```tsx
<TreeViewCheckbox
  key={field}
  title={field}
  checked={checkboxState?.value ?? false}
  onCheckedChange={onCheckedChange}
/>
```

Where onCheckedChange is defined as:

```tsx
function onCheckedChange(value: boolean) {
  onChange((currentFields) => ({
    ...currentFields,
    [field]: { type: "checkbox", value },
  }));
}
```

***After:***

```tsx
<TreeViewCheckbox
  key={fieldId}
  title={fieldId}
  checked={checkboxState?.value ?? false}
  onCheckedChange={(isChecked) => onFieldChange(fieldId, isChecked)}
/>
```

### ✅ Benefits of refactoring:

- Removes cognitive overhead of Updater abstraction.
- Easier to follow and debug — mutations and structure creation are explicit.
- Reduces indirection at each component level.

_Thanks GPT for the rework of my big comment 😅_

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -1,213 +1,294 @@
...
+ *   }
+ * }
+ *
+ **/
+interface PickerCustomTypes {
+  [customTypeId: string]: PickerCustomType;
+}
+
+interface PickerCustomType {
+  [fieldId: string]: PickerCheckboxField;
```
xrutayisire, line 45: ⚠️ The leaf could always be a boolean so you simplify a lot of code everywhere.
For example, to count, you don't need anymore to have a isValidObject and check the type, you could just check if it's a boolean then you add the count, if not you continue down.

```suggestion
  [fieldId: string]: PickerCheckboxField;
```

xrutayisire (changes requested):
Review after our call and spending time checking your rework

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: I added it in the second PR for nested state because I extracted a common function: https://github.com/prismicio/slice-machine/pull/1640

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: Done

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: Got it, as discussed, I will apply these suggestions in a new PR after tackling the support for nested `customTypes` and `groups`.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: I will apply it, but that will change pretty soon with the support for nested customTypes & groups 😅

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: Got it, as discussed, I will tackle this in a different PR so that I make the change with at least content relationship support.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: I agree. Although this structure was designed to support **content relationships** and **groups**, which will be done in a separate PR. Having a type of `checkbox`, `contentRelationship`, or `group` will help me differentiate between the leaf items.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: Changed ✅

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: Done ✅

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
xrutayisire: Hum, for me, it's about the leaf right? Always having the leaf just a boolean at the very end, if it's a section, there is no checkbox anyway, right?

jomifepe: Sorry, I'm not getting your point.

To support **contentRelationships** and **groups**, instead of having:
```ts
{ [fieldId: string]: PickerCheckboxField; }
``` 

We will have something like:
```ts
{ [fieldId: string]: PickerCheckboxField | PickerContentRelationshipField | PickerGroupField; }
```

That's why I already started with an object with `type: "checkbox"`. We will have leaves that can ONLY be checkboxes, like inside nested custom types (content relationships), and for those we could simply use a `boolean`, but I don't really see the value 😕 

The usage of `isValidObject` is to check if it's safe to go down the structure of the object, regardless of having the leaves be only a `boolean` or a `PickerCheckboxField`.

xrutayisire: Hum because the way you go down can be simplified, I think that `isValidObject` is overkill no?

Like just `typeof xxx === "object" && "value" in xxx` is enough to go down, no? 
My point is every time you end up with a boolean you know you can accumulate, every time it's not a boolean, good, it means it's an object where you CAN go down.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: There was an optional chaining missing, I just added it

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: I agree that it could be simplified, my approach was to make it as safe as possible at runtime, since I did not make it type safe, for convenience.

xrutayisire: Then let see with the final PR, I think having a boolean could make think really more easy and prevent having a utils function on top that is not really use fully

xrutayisire (approved):
Not everything done, but all comments addressed and we know the plan

PR #1644, opened 2025-06-03
[DT-2692] feat: pick content relationship GROUP fields

jomifepe:
Relates to: DT-2692

### Description

Support group fields using the new Content Relationship tree view picker.

### Preview

Screenshot 2025-06-05 at 14 31 22

### How to QA [^1]

[^1]:

## Summary by CodeRabbit

- **New Features**
  - Added support for group fields in the content relationship field picker, allowing nested fields and improved selection flexibility.

- **Bug Fixes**
  - Enhanced validation for custom types, enabling more robust and recursive validation of nested fields and group structures.

- **Chores**
  - Updated internal dependencies to improve stability and maintain compatibility.

### DIFF
--- a/packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
+++ b/packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
@@ -4,23 +4,38 @@ import {
   Text,
   TreeView,
   TreeViewCheckbox,
-  TreeViewCheckboxProps,
   TreeViewSection,
 } from "@prismicio/editor-ui";
-import { FormikContext, useField, useFormik } from "formik";
-import { useEffect } from "react";
+import { SetStateAction, useEffect, useMemo, useState } from "react";
 import { useSelector } from "react-redux";
 
 import { selectAllCustomTypes } from "@/modules/availableCustomTypes";
 
-type FieldMap = Record<string, boolean>;
-export type CustomTypeFieldMap = Record<string, FieldMap>;
+type PickerCheckboxField = {
+  type: "checkbox";
+  value: boolean;
+};
 
-type CustomTypeField = string | { id: string; fields: string[] };
+type PickerCustomTypeField = {
+  [fieldId: string]: PickerCheckboxField;
+};
+
+type PickerCustomTypeFields = {
+  [customTypeId: string]: PickerCustomTypeField;
+};
+
+// copy of types-internal types
+
+type TICustomType = {
+  id: string;
+  fields?: readonly string[] | undefined;
+};
+
+type TICustomTypeFields = readonly (string | TICustomType)[];
 
 interface ContentRelationshipFieldPickerProps {
-  initialValues: CustomTypeField[] | undefined;
-  onChange: (fields: CustomTypeField[]) => void;
+  initialValues: TICustomTypeFields | undefined;
+  onChange: (fields: TICustomTypeFields) => void;
 }
 
 export function ContentRelationshipFieldPicker(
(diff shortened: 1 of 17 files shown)


xrutayisire:
@coderabbitai review

packages/slice-machine/src/legacy/lib/models/common/widgets/ContentRelationship/Form.tsx
```diff
@@ -19,21 +44,30 @@ const FormFields = {
...
+                return yup.object({
+                  id: yup.string().required(),
+                  fields: yup
+                    .array(
+                      yup.lazy((fieldValue) => {
+                        if (typeof fieldValue === "string") {
+                          return yup.string();
+                        }
+
+                        return customTypesSchema;
```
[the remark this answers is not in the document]
jomifepe: Not relevant, as this schema will be replaced soon and make use of the existing `types-internal` zod schema.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -84,40 +102,64 @@
...
-interface TICustomTypeFieldValues {
+interface TICustomTypeRegularFieldValues {
   id: string;
   fields?: readonly string[] | undefined;
 }
 
+interface TIGroupFieldValues {
+  id: string;
+  fields?: readonly (string | TIContentRelationshipFieldValue)[] | undefined;
+}
```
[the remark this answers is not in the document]
jomifepe: Totally correct, removed the duplicate type ✅

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -491,6 +649,17 @@
...
 
+      if (fieldValue.type === "group") {
+        return {
+          id: fieldId,
+          fields: Object.entries(fieldValue.value).flatMap(
+            ([fieldId, fieldValue]) => {
+              return fieldValue.value ? fieldId : [];
+            },
+          ),
+        };
```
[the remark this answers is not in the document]
jomifepe: Here I prefer to trust that it's a `boolean` because we're checking if the type is `checkbox`, so the value is being narrowed to `boolean` and it's an internal type managed by our code, so there's no risk of having _corrupt data_.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -329,10 +448,23 @@
...
           );
-        })}
-      </TreeViewSection>
-    );
-  });
+        }
+
+        const fieldState = state[field.id];
+        const fieldStateNarrowed =
+          fieldState.type === "contentRelationship" ? fieldState.value : {};
```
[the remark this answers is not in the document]
jomifepe: Valid point. This particular case is outdated, but I just checked all possible unsafe `undefined` index access.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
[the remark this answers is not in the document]
jomifepe: > ESLint warning in favor of cleaner

For clarity: There is no ESLint warning here, the `value` property is being narrowed to `boolean`.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -239,18 +295,35 @@ function TreeViewCustomType(props: TreeViewCustomTypeProps) {
...
+
+        // Group field
+        if ("fields" in field) {
+          return (
+            <TreeViewGroupField
+              key={field.id}
+              group={field}
+              onChange={onCustomTypeChange}
+              fieldCheckMap={
+                crOrGroupFieldCheckMap.type === "group"
```
xrutayisire, line 306: ❓ I'm a bit lost here, how can we have something that is not a group knowing that we did a `"fields" in field`?

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -259,27 +332,29 @@ function TreeViewCustomType(props: TreeViewCustomTypeProps) {
...
 
   const onContentRelationshipFieldChange = (
     updater: (
       prev: PickerContentRelationshipFieldValue,
     ) => PickerContentRelationshipFieldValue,
   ) => {
     onCustomTypeChange((currentCustomTypeFields) => {
-      const prevCtValue = currentCustomTypeFields[crField.id]?.value;
+      const prevCtValue = currentCustomTypeFields[crField.id] ?? {};
       const prevCtValueNarrowed =
```
xrutayisire, line 354: 💡 I know the prev things will be removed as the new way discussed in first PR don't have that and so it will be simpler to get. But can you explain what's the goal of Narrowed exactly, it's not clear for me.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -291,56 +366,205 @@ function TreeViewContentRelationshipField(
     });
   };
 
-  return crField.customtypes.map((customType) => {
-    // Invalid nested custom type, we need to have fields.
-    if (typeof customType === "string" || !customType.fields) {
-      return null;
-    }
+  if ("customtypes" in crField) {
```
xrutayisire: ❓ A bit lost, how can "customtypes" can not be there in a crField?

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -61,13 +70,32 @@ interface PickerCustomType {
...
+  type: "group";
+  value: PickerGroupFieldValue;
+}
+
+interface PickerGroupFieldWithoutContentRelationship {
+  type: "group";
+  value: PickerGroupFieldWithoutContentRelationshipValue;
+}
+
+interface PickerGroupFieldWithoutContentRelationshipValue {
```
xrutayisire: 💡 I'm lost with the naming here, why without content relationship value?

packages/slice-machine/src/legacy/lib/models/common/widgets/ContentRelationship/index.ts
```diff
@@ -61,30 +64,16 @@ const contentRelationShipConfigSchema = linkConfigSchema.shape({
...
-        ),
-      }),
-    )
+    .array()
+    .test({
+      message: "Invalid customtypes structure.",
+      test: (value) => {
+        if (!value) return true; // Still allow to be optional
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
+        return CustomTypes.decode(value)._tag === "Right";
```
xrutayisire: 💡 Related to the thread, why finally testing this? Could we not just say `.array()`?
What are we protecting by doing that here?

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
xrutayisire, line 1: ⚠️ Bug found:

The count is not right for me:
Screenshot 2025-06-09 at 11 49 38

I'm on commit: `0b66053a58e31e322d67ea9df28dd80ff808085f`

xrutayisire, line 1: ⚠️ Bug found:

I see Custom Type directly inside the Page Type but it's actually normally inside a content relationship field

Screenshot 2025-06-09 at 11 53 35
Screenshot 2025-06-09 at 11 49 06

xrutayisire, line 1: ❓ Global product question:

How do we want to handle the fact some fields are not from the same tab? Here we don't have any indication fields are in different tab.
Maybe we don't care but let's decide on that if not discussed already.

xrutayisire, line 1: ⚠️ Bug found:

When a content relationship is an empty array `[]`, here `city` has an empty array. It should still be selectable here a simple field, right?

Screenshot 2025-06-09 at 11 50 57
Screenshot 2025-06-09 at 11 50 49

xrutayisire (changes requested):
First review on this PR. 
Note that for the bugs, it may not be related to this PR and you should / can create tickets and / or tackle it on separate PRs.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: Can you check that there are no extra fields in your JSON? I noticed this one time and had some non-existent fields there. I will look into checking if the field is valid before loading it into the field check map.

jomifepe: Yes, I [mentioned this one](https://prismic-team.slack.com/archives/C07U6BBM8NP/p1749142242157699?thread_ts=1749130535.623359&cid=C07U6BBM8NP), it's just a missing section, I have the code to fix it but didn't include it here because the diff would get messed up for the review.

jomifepe: This one is confusing, but I guess we do want to show it as a simple field when empty, but also allow to pick the nested fields individually when not empty?

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: So since we're changing the level above (customType), the value that we want to mutate might not be a `contentRelationship` field, but rather a `checkbox` or a `group` so we have to check it before passing it to the updater function, which is typed with what we really want to change there (a `contentRelationship` field). I will see if this can be improved with the change you mentioned.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: It can't, you're right, this is definitely an unnecessary check because the `field` property incorrectly had a union type. Fixed ✅

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: So starting from the first level of fields, we can have the following possibilities:
```
custom type
├── regular field
├── group
│   ├── regular field
│   └── content relationship (nested custom type)
│       ├── regular field
│       └── group
│           └── regular field
└── content relationship (nested custom type)
    ├── regular field
    └── group
        └── regular field
```

The `PickerGroupFieldWithoutContentRelationshipValue` is essentially the type that represents a group that cannot have a `contentRelationship` below it (the leaf ones).

jomifepe: I just renamed them, it might be easier to understand now:
- `PickerGroupField` → `PickerFirstLevelGroupField`
- `PickerGroupFieldWithoutContentRelationship` → `PickerLeafGroupField`

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: Do you mean identifying them in the tree so that the user knows about it?

packages/slice-machine/src/legacy/lib/models/common/widgets/ContentRelationship/index.ts
jomifepe: I can leave the validation loose and see if I need it later. Done ✅

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: It's because this is not the same object. We're narrowing down the field, which initially comes from `customTypes`. When accessing a property from the `fieldCheckMap` object (in the first level, for example), TypeScript doesn't know if it's a `group`, `contentRelationship`, or a `checkbox`.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
xrutayisire: Let's validate with Product but I would expect that yes when customtypes is an empty array `[]` I can select it as a simple field for the metadata. Maybe it can be included in the PR about supporting legacy format. WDYT?

xrutayisire: Yes maybe, so as you know, a custom type can have multiple tab, like SEO etc. But when displaying the tree, the dev sees them all flatten, I think it's pretty nice to align with the content api but as a dev you may be lost by the fact they are on different tab. Did you talk about it already? If not, could you ask Benj just to be sure?

xrutayisire: So yes I have:

```
            {
              "id": "author",
              "fields": ["category", "firstname3", "lastname2"]
            }
```

And category is an existing content relationship within author btw

xrutayisire (approved):
As discussed in private, I let you manage the comments for new PRs / tickets when needed

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: I just asked Benj, and his opinion is: If we don't allow two fields with the same name in different tabs, we don't need to separate them.

jomifepe: Validating 🔄

xrutayisire: Good reasoning, nice one true, with the same id could have been a problem 😅

jomifepe: Quite a big discussion around this topic, but the bottomline is that this is a separate topic that will have it's own changes.

jomifepe: We validated that all of these cases were instances of this: https://github.com/prismicio/slice-machine/pull/1644#discussion_r2135431928

PR #1645, opened 2025-06-04
[DT-2689] E2E for CR sync after updating a custom type

jomifepe:
Resolves: DT-2689

### Description

Create some E2E tests for the core features of the new Content Relationship picker, which test:
- Adding and removing a Custom type to Content Relationship
- Syncing Content Relationships when renaming a field or group API ID
- Selecting regular and group fields
- Selecting regular and group fields inside a nested content relationship
- Selecting a level 2 nested custom type as a simple field (limit)
- Seeing an updated label of `n fields returned in the API` and `n fields selected` when selecting fields
- Seeing a `No available fields to select` when selecting a CT with empty groups and nested CTs

### How to QA [^1]

[^1]:


playwright/tests/fields/contentRelationship.spec.ts
```diff
@@ -0,0 +1,420 @@
...
+  await customTypesTablePage.goto();
+  await customTypesTablePage.openCreateDialog();
+
+  const customTypeAId = `custom_type_${generateRandomId()}`;
+  await customTypesTablePage.createTypeDialog.createType(
+    customTypeAId,
+    "reusable",
+  );
+
+  // Create custom type A
```
levimykel: Custom type B

playwright/tests/fields/contentRelationship.spec.ts
```diff
@@ -0,0 +1,420 @@
...
+    type: "Rich Text",
+    name: "My Regular Field",
+    expectedId: "my_regular_field",
+  });
+
+  await pageTypesBuilderPage.addStaticField({
+    type: "Content Relationship",
+    name: "My Content Relationship With CT A",
+    expectedId: "my_content_relationship_with_ct_a",
+  });
```
levimykel: You never actually set the relationship with CT A. I'm not sure what exactly you're testing with this. This means we can either remove the creation of CT A because it doesn't make a difference in the test, or make sure to select it for what you need.

levimykel (commented):
Just found a couple of small things. Otherwise, it looks good.

playwright/tests/fields/contentRelationship.spec.ts
levimykel: If the point is that you aren't able to select fields from this in CT D, then it might be good to add a check that `my_content_relationship_with_ct_a` isn't expandable.

playwright/tests/fields/contentRelationship.spec.ts
jomifepe: Thanks 🙏  Corrected

playwright/tests/fields/contentRelationship.spec.ts
jomifepe: You're right, it still tested the expected behavior, which was to render a simple field, but the name is misleading, so I linked to A, and it will still be a simple field because A has no fields.

Nice catch! Corrected ✅

levimykel (approved):

PR #1658, opened 2025-06-16
feat: improve render logic and use different label for page type

jomifepe:
### Description

1. Refactor the field picker rendering logic to decouple a bit from the `Link` `customtypes` structure.
2. (DT-2700) Display **Page type** or **Custom type** label depending on `format`.
3. (DT-2705) Reduce the duplication of `types-internal` types.
4. Improved naming and code readability.
  a. `CustomTypes` → `LinkCustomtypes`

### Preview

Behavior remains the same, the only visual change was `2`:

Screenshot 2025-06-16 at 12 44 17

### How to QA [^1]

[^1]:


packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -276,110 +270,123 @@ export function ContentRelationshipFieldPicker(
 }
 
 interface TreeViewCustomTypeProps {
-  customType: TICustomType;
+  customType: CustomTypeSM;
```
xrutayisire: ⚠️ As discussed, writing this here so I don't forget about not using SM types

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -276,110 +270,123 @@ export function ContentRelationshipFieldPicker(
...
       )}
-      badge="Custom type"
+      badge={customType.format === "page" ? "Page type" : "Custom type"}
     >
-      {customType.fields.map((field) => {
-        // Checkbox field
+      {customType.tabs
+        .flatMap((tab) => tab.value)
+        .map((field) => {
+          if (isUidField(field)) return null;
```
xrutayisire: ❓ One question related to that, how it will play out if a custom type hase:

1. no field at all to select?
2. only a uid to select and so filtered out here?

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -389,18 +396,26 @@ function TreeViewContentRelationshipField(
     field: crField,
     fieldCheckMap: crFieldsCheckMap,
     onChange: onCrFieldChange,
+    customTypes,
   } = props;
 
+  if (!crField.value.config?.customtypes) return null;
```
xrutayisire: 💡 Here this will be removed with the new legacy support ticket right?

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -876,12 +837,38 @@ function countPickedFields(
...
+  return field.value.type === "Group";
+}
+
+function isContentRelationshipField(
+  field: TabFields[number],
+): field is { key: string; value: Link } {
+  return (
+    field.value.type === "Link" &&
+    field.value.config?.select === "document" &&
+    field.value.config?.customtypes !== undefined
```
xrutayisire: ⚠️ For me I see a difference in the code between customtypes being undefined and customtypes being an empty array. But both should have the same code path, right?

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
```diff
@@ -876,12 +837,38 @@ function countPickedFields(
...
+
+function isUidField(
+  field: TabFields[number],
+): field is { key: string; value: UID } {
+  // Filter out uid fields because it's a special field returned by the
+  // API and is not part of the data object in the document.
+  // We also filter by key "uid", because (as of the time of writing
+  // this), creating any field with that API id will result in it being
+  // used for metadata.
+  return field.key === "uid" && field.value.type === "UID";
```
xrutayisire: ⚠️ Actually this is a `||` that we need here right?

xrutayisire (changes requested):

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: You're right, fixed ✅

jomifepe: You're right, fixed ✅

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: Just refactored to use the `types-internal` ones, as discussed.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: Correct, I will re-evaluate these conditions when tackling the support for that new UI and legacy support.

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: Same as https://github.com/prismicio/slice-machine/pull/1658#discussion_r2150507408

packages/slice-machine/src/features/customTypes/fields/ContentRelationshipFieldPicker.tsx
jomifepe: You're right, this was not preserved from the previous mapping, an empty section would be rendered until something other than a uid is added. Just committed a change to handle those cases ✅

xrutayisire (approved):
LGTM!

PR #1673, opened 2025-06-25
feat: handle invalid CR customtypes reference conversions

jomifepe:
Resolves: DT-2722, DT-2723

### Description

- When converting the `customtypes` array to the field picker component, filter out non-existing fields and wrong references (mismatching type).
- Add multiple unit tests for the `convertLinkCustomtypesToFieldCheckMap`, which takes care of these conversions.

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
+++ b/packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
@@ -936,10 +936,10 @@ function useCustomTypes(linkCustomtypes: LinkCustomtypes | undefined): {
 
 function resolveContentRelationshipCustomTypes(
   linkCustomtypes: LinkCustomtypes,
-  localCustomTypes: CustomType[],
+  allCustomTypes: CustomType[],
 ): CustomType[] {
   return linkCustomtypes.flatMap((linkCustomtype) => {
-    return localCustomTypes.find((ct) => ct.id === getId(linkCustomtype)) ?? [];
+    return allCustomTypes.find((ct) => ct.id === getId(linkCustomtype)) ?? [];
   });
 }
 
--- a/packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
+++ b/packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
@@ -1362,7 +1362,9 @@ type CustomTypeFlatFieldMap = {
 function getCustomTypeFlatFieldMap(
   customType?: CustomType,
 ): CustomTypeFlatFieldMap | { isAvailable: false; get: null; has: null } {
-  if (!customType) return { isAvailable: false, get: null, has: null };
+  if (!customType) {
+    return { isAvailable: false, get: null, has: null };
+  }
 
   const ctFieldMap = Object.values(customType.json).reduce((acc, tabFields) => {
     for (const [fieldId, field] of Object.entries(tabFields)) {
--- a/packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
+++ b/packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
@@ -956,11 +956,10 @@ export function convertLinkCustomtypesToFieldCheckMap(args: {
   const checkMap = linkCustomtypes.reduce((customTypes, customType) => {
     if (typeof customType === "string") return customTypes;
 
-    const ctFlatFieldMap = getCustomTypeFlatFieldMap(
-      allCustomTypes?.find((ct) => ct.id === customType.id),
-    );
+    const existingCt = allCustomTypes?.find((ct) => ct.id === customType.id);
+    if (allCustomTypes && !existingCt) return customTypes;
 
-    if (!ctFlatFieldMap.isAvailable) return customTypes;
+    const ctFlatFieldMap = getCustomTypeFlatFieldMap(existingCt);
 
     const customTypeFields = customType.fields.reduce((fields, field) => {
       // Check if the field exists
@@ -969,7 +968,13 @@ export function convertLinkCustomtypesToFieldCheckMap(args: {
 
       // Regular field
       if (typeof field === "string") {
-        if (allCustomTypes && existingField?.type === "Group") return fields;
+        if (
+          allCustomTypes &&
+          existingField &&
+          existingField?.type === "Group"
+        ) {
+          return fields;
+        }
 
         fields[field] = { type: "checkbox", value: true };
         return fields;
@@ -977,7 +982,13 @@ export function convertLinkCustomtypesToFieldCheckMap(args: {
 
       // Group field
       if ("fields" in field && field.fields !== undefined) {
-        if (allCustomTypes && existingField?.type !== "Group") return fields;
+        if (
+          allCustomTypes &&
+          existingField &&
+          existingField?.type !== "Group"
+        ) {
+          return fields;
+        }
 
         const groupFieldCheckMap = createGroupFieldCheckMap({
           group: field,
@@ -998,8 +1009,9 @@ export function convertLinkCustomtypesToFieldCheckMap(args: {
           allCustomTypes &&
           existingField &&
           !isContentRelationshipField(existingField)
-        )
+        ) {
           return fields;
+        }
 
         const crFieldCheckMap = createContentRelationshipFieldCheckMap({
           field,
@@ -1040,7 +1052,9 @@ function createGroupFieldCheckMap(args: {
 
     // Regular field
     if (typeof field === "string") {
-      if (allCustomTypes && existingField?.type === "Group") return fields;
+      if (allCustomTypes && existingField && existingField?.type === "Group") {
+        return fields;
+      }
 
       fields[field] = { type: "checkbox", value: true };
       return fields;
@@ -1088,26 +1102,32 @@ function createContentRelationshipFieldCheckMap(args: {
   const fieldEntries = field.customtypes.reduce((customTypes, customType) => {
     if (typeof customType === "string") return customTypes;
 
-    const ctFlatFieldMap = getCustomTypeFlatFieldMap(
-      allCustomTypes?.find((ct) => ct.id === customType.id),
-    );
+    const existingCt = allCustomTypes?.find((ct) => ct.id === customType.id);
+    if (allCustomTypes && !existingCt) return customTypes;
 
-    if (!ctFlatFieldMap.isAvailable) return customTypes;
+    const ctFlatFieldMap = getCustomTypeFlatFieldMap(existingCt);
 
     const ctFields = customType.fields.reduce((nestedFields, nestedField) => {
       // Regular field
       if (typeof nestedField === "string") {
-        if (ctFlatFieldMap.has(nestedField)) {
-          nestedFields[nestedField] = { type: "checkbox", value: true };
+        if (allCustomTypes && !ctFlatFieldMap.has(nestedField)) {
+          return nestedFields;
         }
+
+        nestedFields[nestedField] = { type: "checkbox", value: true };
         return nestedFields;
       }
 
       // Group field
       const groupFields = nestedField.fields.reduce((fields, field) => {
-        if (ctFlatFieldMap.has(`${nestedField.id}.${field}`)) {
-          fields[field] = { type: "checkbox", value: true };
+        if (
+          allCustomTypes &&
+          !ctFlatFieldMap.has(`${nestedField.id}.${field}`)
+        ) {
+          return fields;
         }
+
+        fields[field] = { type: "checkbox", value: true };
         return fields;
       }, {} as PickerLeafGroupFieldValue);
 
@@ -1116,7 +1136,6 @@ function createContentRelationshipFieldCheckMap(args: {
           type: "group",
           value: groupFields,
         };
-        return nestedFields;
       }
 
       return nestedFields;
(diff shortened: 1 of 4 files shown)


packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
```diff
@@ -943,92 +947,212 @@ function resolveContentRelationshipCustomTypes(
...
-          customTypeFields[field] = { type: "checkbox", value: true };
-        } else if ("fields" in field && field.fields !== undefined) {
-          // Group field
-          customTypeFields[field.id] = createGroupFieldCheckMap(field);
-        } else if ("customtypes" in field && field.customtypes !== undefined) {
-          // Content relationship field
-          customTypeFields[field.id] =
-            createContentRelationshipFieldCheckMap(field);
+    const existingCt = allCustomTypes?.find((ct) => ct.id === customType.id);
+    if (allCustomTypes && !existingCt) return customTypes;
```
xrutayisire: ⚠️ Hum for me it's a blocker but we will see.
The way you handle allCustomTypes is a hack for the tracking but create a potential bug.

What if there is really NO custom types anymore. Everything will be valid, but everything should be invalid in this case. 

Is it not a problem?

jomifepe: So my assumption is that when there are no custom types, `allCustomTypes` will be defined but empty, so it will still check if the fields exist and fail.

jomifepe: I can also add some sort of flag to skip the existence check

xrutayisire: Oh! Ok I see, you're using the undefined vs empty array. It's actually interesting but not trivial to see why we do the allCustomTypes check.

What about doing a variable that would auto-document the feature then.

Like:

```
const noInvalidCheck = allCustomTypes === undefined
```

Naming may not be great here but you get the idea. WDYT?

xrutayisire: Also a flag is good if you think it's clearer 🙂

packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
```diff
@@ -260,11 +261,11 @@ export function ContentRelationshipFieldPicker(
 function ContentRelationshipFieldPickerContent(
   props: ContentRelationshipFieldPickerProps,
 ) {
-  const { value, onChange } = props;
-  const { allCustomTypes, pickedCustomTypes } = useCustomTypes(value);
+  const { value: linkCustomtypes, onChange } = props;
```
xrutayisire, line 264: ✨ Nice rename, super clear like that

packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
```diff
@@ -943,92 +947,212 @@ function resolveContentRelationshipCustomTypes(
...
+
+    const customTypeFields = customType.fields.reduce((fields, field) => {
+      // Check if the field exists
+      const existingField = ctFlatFieldMap.get(getId(field));
+      if (allCustomTypes && !existingField) return fields;
+
+      // Regular field
+      if (typeof field === "string") {
+        // Check if the field matched the existing one in the custom type
+        if (allCustomTypes && existingField && existingField.type === "Group") {
```
xrutayisire: 💡 No need to check existingField here right? It's check above that for sure it's here

xrutayisire: ⚠️ Without yet checking the test I guess a potential problem.

If you have a "string" but it target a CR with a custom type that contains valid fields, it should not be a string then.

packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
```diff
@@ -943,92 +947,212 @@ function resolveContentRelationshipCustomTypes(
...
+
+        if (crFieldCheckMap) {
+          fields[field.id] = crFieldCheckMap;
+        }
+
+        return fields;
+      }
+
+      return fields;
+    }, {} as PickerCustomType);
```
xrutayisire: 💡 Can you do that instead of the `as`:

```
    const customTypeFields = customType.fields.reduce<PickerCustomType>((fields, field) => {
```

packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
```diff
@@ -943,92 +947,212 @@ function resolveContentRelationshipCustomTypes(
...
+function createGroupFieldCheckMap(args: {
+  group: LinkCustomtypesGroupFieldValue;
+  allCustomTypes?: CustomType[];
+  ctFlatFieldMap: CustomTypeFlatFieldMap;
+}): PickerFirstLevelGroupField | undefined {
+  const { group, ctFlatFieldMap, allCustomTypes } = args;
+
+  const fieldEntries = group.fields.reduce((fields, field) => {
+    // Check if the field exists
+    const existingField = ctFlatFieldMap.get(`${group.id}.${getId(field)}`);
```
xrutayisire: 💡 Is the whole idea of `ctFlatFieldMap` just for `Group`?

It doesn't seem that crazy to just access the group field from here without the whole ctFlatFieldMap to maintain.
It creates one more later to maintain, understand etc. 
Reducing the number of layers to know when opening a file is important, and here there is already the picker format to know about.

packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
```diff
@@ -943,92 +947,212 @@ function resolveContentRelationshipCustomTypes(
...
-  };
-  const crFieldCustomTypes = crField.value;
+function createContentRelationshipFieldCheckMap(args: {
+  field: LinkCustomtypesContentRelationshipFieldValue;
+  allCustomTypes?: CustomType[];
+}): PickerContentRelationshipField | undefined {
+  const { field, allCustomTypes } = args;
+
+  const fieldEntries = field.customtypes.reduce((customTypes, customType) => {
+    if (typeof customType === "string") return customTypes;
```
xrutayisire: ⚠️ Didn't check the tests yet, but this seems to be missing checks.
If the customType is just a string, it should exist right?
If the customType is just a string, it should not contain valid fields, because if yes, it should ONLY be an object with data.

packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
```diff
@@ -943,92 +947,212 @@ function resolveContentRelationshipCustomTypes(
...
       if (typeof nestedField === "string") {
-        // Regular field
-        customTypeFields[nestedField] = { type: "checkbox", value: true };
-      } else {
-        // Group field
-        const groupFieldsEntries = nestedField.fields.map(
-          (field) => [field, { type: "checkbox", value: true }] as const,
-        );
+        // Check if the field matched the existing one in the custom type
+        if (allCustomTypes && !ctFlatFieldMap.has(nestedField)) {
```
xrutayisire: ⚠️ We don't verify anything related to group here as before?

packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
```diff
@@ -1237,34 +1360,78 @@ function isContentRelationshipFieldWithSingleCustomtype(
...
+  get: (fieldId: string) => DynamicWidget | undefined;
+  has: (fieldId: string) => boolean;
+};
+
+/**
+ * Util that flattens a custom type's fields into a map of field ids to fields and
+ * returns functions to get and check if a field exists by key.
+ * For group fields, the key is separated by a dot (e.g. "group.fieldId").
+ */
+function getCustomTypeFlatFieldMap(
```
xrutayisire: 💡 On top of my other that I think it's not necessary, I think I don't like that we hide product rules within what seems to be just a flap map.

The isValid is pretty hidden but should be at the same level without indirection as the other conditions. No?

xrutayisire (changes requested):
Lot of work to make all the product rules.
I decided not to look at tests for a first review, to not be bias by them. Like, oh he has tests, so no need to ask.
I prefer to ask when I'm not sure and then look at the test, so we ensure the tests are validating what needs to be validated.

packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
jomifepe: I can, yeah, sometimes I go with these because I don't like 80 print width 🙊

jomifepe: Done ✅

packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
jomifepe: > 💡 No need to check existingField here right? It's check above that for sure it's here

The check above does not guarantee it because it has other conditions.

jomifepe: I will take a look at the case you described 👀

packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
jomifepe: Yeah, I can see that, the custom type is already an object. Just refactored, and also changed and reused the existing `getCustomTypeStaticFields` util for the same purpose ✅

Tell me WDYT.

packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
jomifepe: I just refactored and it should be clearer now!

packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
xrutayisire: I like it better, a lot of utils function at the end, but now at least I feel simpler to read 👍

packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
xrutayisire: With the new version I guess this is not really a problem anymore, a bit lost on how we could have the validation closer to the others but it's good for now

packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
jomifepe: Those are valid cases, but we don't mix up the strings (Custom Type IDs) with the "picker state", so this line just ignores them, as they won't translate into any checked checkbox or anything, right? Essentially, we only care about it when rendering the tree, when we resolve its fields (if it exists), here:

```ts
const resolvedCustomTypes = resolveContentRelationshipCustomTypes(
  field.config.customtypes,
  allCustomTypes,
);
```

These concepts are hard to grasp 😵‍💫

xrutayisire: I see but maybe it's still a use case we should look at, when we merge and convert, we don't wanna keep this customtype as string if it contains valid fields to pick, no?

Yeah it's complex here!

packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
jomifepe: You're right! That's one untested edge case. Adding it right now!

jomifepe: Done ✅ Nice catch!

xrutayisire (approved):
LGTM!

packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
jomifepe: Discussed privately, but essentially, string custom types, which are invalid in the picker check map context (which only holds the state for field checkboxes), doesn't mean anything, so they won't be considered and will be overwritten on save, because after saving, only the final state of the picker matters.

packages/slice-machine/src/features/builder/fields/contentRelationship/ContentRelationshipFieldPicker.tsx
jomifepe: Same case as https://github.com/prismicio/slice-machine/pull/1673#discussion_r2169137389

PR #1678, opened 2025-07-22
[DT-2708] fix: bind node auth listener server to localhost

jomifepe:
Resolves: DT-2708

### Description

The server that listens for authentication success in the Slice Machine init is currently binding to port 5555, but on any hostname. This PR changes it to bind to `localhost`.

### Checklist

#### Before
Screenshot 2025-07-22 at 15 43 50

#### After
Screenshot 2025-07-22 at 15 44 04

#### Full init flow demo

https://github.com/user-attachments/assets/5e9b7439-2824-4694-a665-c2efcec8c8fb

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/manager/src/auth/PrismicAuthManager.ts
+++ b/packages/manager/src/auth/PrismicAuthManager.ts
@@ -194,7 +194,7 @@ export class PrismicAuthManager {
 				server.once("listening", () => {
 					resolve();
 				});
-				server.listen(args.port);
+				server.listen(args.port, "127.0.0.1");
 			});
 
 			if (args.onListenCallback) {


xrutayisire (approved):
LGTM!

About the commit name, it's not only the init that it fixes, right?
It's also fixing every login from SM UI I guess

jomifepe:
> LGTM!
> 
> About the commit name, it's not only the init that it fixes, right? It's also fixing every login from SM UI I guess

Replying here as well for visibility:

I don’t think it needs to. The UI opens a login page with the following URL:

`https://{{DOMAIN}}/dashboard/cli/login?source=slice-machine&port=9999&path=/api/auth`

This seems to be directly using slice-machine’s port for the callback, so no need to change anything there because it's already bound to `localhost`. Although, it's true that anything that uses this process for authentication will also be fixed by this PR.

PR #1682, opened 2025-07-23
fix: properly handle sm invalid env error

xrutayisire:
Resolves: [DT-2710](https://linear.app/prismic/issue/DT-2710/clean-up-sminvalidactiveenvironmenterror-warning-in-client-logs)

### Description

- Prevent having an infinite loop when there is always an invalid env, it can happen in some weird conf of the user (case of the ticket reported in slack)
- Ensure no error is triggered to Sentry for a known error
- Ensure no network error is visible for a known error
- Add with Benj a global interception error screen in order to totally break if even after a retry the env is still invalid

### Checklist

### Preview

Screenshot 2025-07-23 at 19 28 09

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/manager/src/managers/project/ProjectManager.ts
+++ b/packages/manager/src/managers/project/ProjectManager.ts
@@ -86,14 +86,8 @@ type ProjectManagerUpdateEnvironmentArgs = {
 };
 
 type ProjectManagerFetchActiveEnvironmentReturnType =
-	| {
-			activeEnvironment: Environment;
-			error?: never;
-	  }
-	| {
-			activeEnvironment: undefined;
-			error: InvalidActiveEnvironmentError;
-	  };
+	| { type: "ok"; environment: Environment }
+	| { type: "error"; error: InvalidActiveEnvironmentError };
 
 export class ProjectManager extends BaseManager {
 	private _cachedRoot: string | undefined;
@@ -457,7 +451,7 @@ export class ProjectManager extends BaseManager {
 			this._cachedEnvironments || [],
 		);
 		if (cachedActiveEnvironment) {
-			return { activeEnvironment: cachedActiveEnvironment };
+			return { type: "ok", environment: cachedActiveEnvironment };
 		}
 
 		// If the environment is not in the cached environments list, we
@@ -479,12 +473,12 @@ export class ProjectManager extends BaseManager {
 
 		if (!activeEnvironment) {
 			return {
-				activeEnvironment: undefined,
+				type: "error",
 				error: new InvalidActiveEnvironmentError(),
 			};
 		}
 
-		return { activeEnvironment };
+		return { type: "ok", environment: activeEnvironment };
 	}
 
 	async detectVersionControlSystem(): Promise<string | "_unknown"> {
--- a/packages/manager/src/managers/project/ProjectManager.ts
+++ b/packages/manager/src/managers/project/ProjectManager.ts
@@ -86,7 +86,7 @@ type ProjectManagerUpdateEnvironmentArgs = {
 };
 
 type ProjectManagerFetchActiveEnvironmentReturnType =
-	| { type: "ok"; environment: Environment }
+	| { type: "ok"; activeEnvironment: Environment }
 	| { type: "error"; error: InvalidActiveEnvironmentError };
 
 export class ProjectManager extends BaseManager {
@@ -451,7 +451,7 @@ export class ProjectManager extends BaseManager {
 			this._cachedEnvironments || [],
 		);
 		if (cachedActiveEnvironment) {
-			return { type: "ok", environment: cachedActiveEnvironment };
+			return { type: "ok", activeEnvironment: cachedActiveEnvironment };
 		}
 
 		// If the environment is not in the cached environments list, we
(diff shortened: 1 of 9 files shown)


packages/manager/src/managers/project/ProjectManager.ts
```diff
@@ -85,9 +85,15 @@ type ProjectManagerUpdateEnvironmentArgs = {
...
-};
+type ProjectManagerFetchActiveEnvironmentReturnType =
+	| {
+			activeEnvironment: Environment;
+			error?: never;
+	  }
+	| {
+			activeEnvironment: undefined;
+			error: InvalidActiveEnvironmentError;
+	  };
```
dani-mp: for these, y prefer to have a proper key to do the intersection, like 
```
{ type: "ok", environment: Environment } | { type: "error", error :InvalidActiveEnvironmentError }
```

packages/slice-machine/src/features/environments/actions/getActiveEnvironment.ts
```diff
@@ -15,18 +15,27 @@ type GetActiveEnvironmentReturnType =
       error: unknown;
     };
 
-export async function getActiveEnvironment(): Promise<GetActiveEnvironmentReturnType> {
+export async function getActiveEnvironment(
+  isRetry = false,
```
dani-mp: i'd put this in a object, so the caller side is not just `getActiveEnvironment(true)` but something like `getActiveEnvironment({ retried: true })`

packages/slice-machine/src/legacy/components/AppLayout/index.tsx
```diff
@@ -18,7 +26,31 @@ export const AppLayout: FC<PropsWithChildren> = ({
...
+          >
+            <BlankSlate>
+              <BlankSlateIcon
+                lineColor="tomato11"
+                backgroundColor="tomato3"
+                name="alert"
+              />
+              <BlankSlateTitle>Failed to load Slice Machine</BlankSlateTitle>
+              <BlankSlateDescription>
+                {JSON.stringify(error)}
```
dani-mp, line 47: isn't it better to show `error.message` or something like that?

packages/slice-machine/src/legacy/components/Navigation/Environment.tsx
```diff
@@ -110,5 +114,9 @@ export function Environment() {
     );
   }
 
-  throw useEnvironmentsError;
+  if (useEnvironmentsError !== undefined) {
+    throw useEnvironmentsError;
+  }
+
+  throw activeEnvironmentError;
```
dani-mp, line 121: super weird that we throw here instead of just showing the error directly, no?

packages/slice-machine/src/legacy/components/AppLayout/index.tsx
```diff
@@ -18,7 +26,31 @@ export const AppLayout: FC<PropsWithChildren> = ({
   ...otherProps
 }) => {
   return (
-    <ErrorBoundary>
+    <ErrorBoundary
+      renderError={(error) => {
```
dani-mp, line 30: i guess there's no way to recover from this error?

packages/slice-machine/src/features/environments/actions/getActiveEnvironment.ts
```diff
@@ -15,18 +15,27 @@ type GetActiveEnvironmentReturnType =
...
+): Promise<GetActiveEnvironmentReturnType> {
   try {
-    const { activeEnvironment } =
+    const { activeEnvironment, error } =
       await managerClient.project.fetchActiveEnvironment();
 
+    if (error) {
+      const errorInstance = new Error(error.message);
+      errorInstance.name = error.name;
+      throw errorInstance;
```
dani-mp: why do we need to build the error again?

packages/slice-machine/src/legacy/components/AppLayout/index.tsx
xrutayisire: As it's the global main ErrorBoundary the goal is really for us (Prismic) to have everything possible to debug and understand user issues. error name is also super important and any other format type of error would be displayed thanks to that.

packages/slice-machine/src/legacy/components/Navigation/Environment.tsx
xrutayisire: No no, it's product decision, we want to have the global error boundary catching up that.
For this error the decision is not crash the whole app and not just display something on the side nav menu

packages/slice-machine/src/legacy/components/AppLayout/index.tsx
xrutayisire: Nope, these errors need support or knowledge from the user to fix it, but SM should not be used in the meantime.

Note: maybe we will improve that if we detect some recovery paths for some errors.

packages/slice-machine/src/features/environments/actions/getActiveEnvironment.ts
xrutayisire: Since we don't throw from the manager, it's not really an error but just an object and so it cannot be used with the `isInvalidActiveEnvironmentError` if I don't rebuild it first

dani-mp (approved):

PR #1687, opened 2025-08-04
[DT-2808] feat: logout button

jomifepe:
Resolves: [DT-2808](https://linear.app/prismic/issue/DT-2808)

### Description

Slice Machine never had a logout button, only a log in one. Since we deprecated [`prismic-cli`](https://github.com/prismicio/slice-machine/compare/jp/logout?expand=1), it's now even more necessary to have a replacement, so this PR creates one.

Additional features:
- Refresh UI after login to avoid displaying `Login required` while logged in.

### Checklist

### Preview

Screenshot 2025-08-04 at 16 00 21

#### Flow

https://github.com/user-attachments/assets/b97aaf2d-3d9f-437f-bf05-f17644dc2e40

### How to QA [^1]

[^1]:


packages/slice-machine/src/features/environments/useActiveEnvironment.ts
```diff
@@ -1,7 +1,15 @@
...
+import {
+  Fetcher,
+  invalidateFetcherData,
+  useRequest,
+} from "@prismicio/editor-support/Suspense";
 
 import { getActiveEnvironment } from "./actions/getActiveEnvironment";
 
+export function invalidateActiveEnvironmentData() {
+  invalidateFetcherData(getActiveEnvironment as Fetcher);
```
jomifepe: This cast should not be needed, but `invalidateFetcherData` types are causing this error:

Image


`invalidateFetcherData` should not really care about `args`, so the types might need to be adjusted on `editor` side. What's weird is that `useRequest` is typed more or less the same.

packages/slice-machine/src/features/auth/LogoutButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { invalidateActiveEnvironmentData } from "@/features/environments/useActiveEnvironment";
+import { invalidateEnvironmentsData } from "@/features/environments/useEnvironments";
+import useSliceMachineActions from "@/modules/useSliceMachineActions";
+
+export function LogoutButton() {
+  const { logout: storeLogout } = useSliceMachineActions();
+
+  async function onClick() {
+    await managerLogout();
+    storeLogout();
```
xrutayisire: 💡 I'm not sure why we need this? It adds a redux action where we wanna reduce the number of redux actions.

We have that kind of management for me already:

```
    // TODO(DT-1453): Remove the need of the global getState
    const serverState = await getState();
    // Update Redux store
    refreshState(serverState);
```

packages/slice-machine/src/features/auth/LogoutButton.tsx
```diff
@@ -0,0 +1,32 @@
+import { IconButton, Tooltip } from "@prismicio/editor-ui";
+import { toast } from "react-toastify";
+
+import { logout as managerLogout } from "@/apiClient";
+import { invalidateActiveEnvironmentData } from "@/features/environments/useActiveEnvironment";
+import { invalidateEnvironmentsData } from "@/features/environments/useEnvironments";
+import useSliceMachineActions from "@/modules/useSliceMachineActions";
+
+export function LogoutButton() {
```
xrutayisire, line 10: ❓ Did you think about `useServerState`, `Sentry.setUser` for example?

packages/slice-machine/src/features/environments/useActiveEnvironment.ts
xrutayisire: Hum here I think it could be a problem. `invalidateFetcherData` actually can care about `args`.
We need to dig that up a bit as I remember our custom useRequest was not supporting all use case.

If you check:

```
ARGS extends readonly SerializableArg[]
``

So here we pass an object and I think useRequest don't support this with an object being passed.

File: packages/slice-machine/src/legacy/components/Navigation/SideNavEnvironmentSelector/SideNavEnvironmentSelector.tsx [hunk 4]
```diff
@@ -138,6 +139,8 @@ export const SideNavEnvironmentSelector: FC<SideNavEnvironmentSelectorProps> = (
                 }
               />
             ) : undefined}
+
+            {variant === "default" && <LogoutButton />}
```
xrutayisire, line 143: ❓ Did you test the use case where we have multiple environments?

packages/slice-machine/src/features/auth/LogoutButton.tsx
jomifepe:
```
// TODO: Remove this global-state method. It is expensive and a
// potential source of bugs due to data inconsistency. SM UI relies on
// it heavily, so removal will require significant effort.
```

I see this and automatically steer away from it, but yeah, it also works 😅

packages/slice-machine/src/legacy/components/Navigation/SideNavEnvironmentSelector/SideNavEnvironmentSelector.tsx
jomifepe: I did! It works as intended, the screenshot in the description is using a repo with multiple envs.

packages/slice-machine/src/features/auth/LogoutButton.tsx
jomifepe: I can use that approach if you prefer.

packages/slice-machine/src/features/auth/LogoutButton.tsx
jomifepe: I don't really see the value of `useServerState`, I think it's better to have a programmatic call rather than a side-effect. 

Regarding `Sentry.setUser`, I'm not aware of it, should I call it to set the user `id` to `undefined`?

packages/slice-machine/src/features/environments/useActiveEnvironment.ts
jomifepe: As discussed in Slack: I tried to refactor to `react-query` but there's something wrong with the setup and I was getting an infinite re-render loop. For the sake of this task, I changed the param of `getActiveEnvironment` to be a `boolean` instead of an object with `repeat?: false`. Tell me what you think

packages/slice-machine/src/features/auth/LogoutButton.tsx
xrutayisire: I agree with you, I hate this hook
Let's do that yes, I guess if they logout we should forget who they are, even for Sentry call.

packages/slice-machine/src/modules/useSliceMachineActions.ts
```diff
@@ -148,7 +148,6 @@ const useSliceMachineActions = () => {
     );
   const updateSliceMockSuccess = (args: SaveSliceMockRequest) =>
     dispatch(sliceUpdateMockSuccess(args));
-
```
xrutayisire: Nitpick: Maybe we can revert this small line

packages/slice-machine/src/legacy/components/LoginModal/index.tsx
```diff
@@ -76,6 +78,10 @@ const LoginModal: React.FunctionComponent = () => {
         60,
       );
 
+      // refresh queries to update the UI
```
xrutayisire, line 81: ❓ Does this also handle the refresh of the getState as you did during the logout?

packages/slice-machine/src/legacy/components/LoginModal/index.tsx
```diff
@@ -76,6 +78,10 @@ const LoginModal: React.FunctionComponent = () => {
...
+      invalidateEnvironmentsData();
```
xrutayisire, line 82: ❓ Why after the startPolling? As the `await logout` is before

jomifepe: Resolved. Context: This has to happen after `startPolling` because before that the login was not complete yet.

packages/slice-machine/src/legacy/components/LoginModal/index.tsx
jomifepe: No that is already happening below.

packages/slice-machine/src/modules/useSliceMachineActions.ts
jomifepe: Done ✅

packages/slice-machine/src/legacy/components/LoginModal/index.tsx
xrutayisire: Hum where? Sorry I don't get it

packages/slice-machine/src/features/auth/LogoutButton.tsx
xrutayisire: I think both are "bad" true, just this one requires less code and much more will be replaced by something that we can use everywhere we call getstate.

Meaning, we will have a use query to get the remote custom type, we will have to just invalidate that. So a simple drop in here with the list of invalidate.

xrutayisire (approved):
LGTM

PR #1688, opened 2025-08-05
[DT-2826] feat: make field type more evident

jomifepe:
Resolves: [DT-2826](https://linear.app/prismic/issue/DT-2826)

### Description

- Display a tooltip with the field type above the icon in the field list
- Display the field type in the edit modal's header instead of the label

### Preview

Screenshot 2025-08-05 at 11 47 41

Screenshot 2025-08-05 at 11 47 52

### How to QA [^1]

[^1]:


packages/slice-machine/src/domain/fields.ts
```diff
@@ -34,12 +34,44 @@ export interface GroupField extends BaseField {
 
 export type Field = NestableField | UIDField | GroupField;
 
+export type CustomFieldType =
```
dani-mp: i'm not sure i understand what a "custom field" is in the context of our domain, could you clarify? i think we should find a different name

packages/slice-machine/src/domain/fields.ts
```diff
@@ -34,12 +34,44 @@ export interface GroupField extends BaseField {
...
+  | FieldType
+  | "LinkToMedia"
+  | "ContentRelationship"
+  | "NestedGroup";
```
dani-mp: these 3 are declared above already

packages/slice-machine/src/domain/fields.ts
```diff
@@ -34,12 +34,44 @@ export interface GroupField extends BaseField {
...
+
+export const fieldLabels: Record<CustomFieldType, string> = {
+  Text: "Text",
+  StructuredText: "Rich Text",
```
dani-mp:
```suggestion
  StructuredText: "Rich text",
```
i thought we didn't want to use title case anymore, same for the rest

packages/slice-machine/src/domain/fields.ts
```diff
@@ -34,12 +34,44 @@ export interface GroupField extends BaseField {
...
+  Choice: "Choice",
+};
+
 /**
  * Nestable fields
  */
 
 export const booleanField: NestableField = {
-  name: "Boolean",
+  name: fieldLabels.Boolean,
```
dani-mp: I would avoid introducing the `fieldLabels` object that you have to import everywhere, and just use the `name` property of the widget passed, and, if needed, i would add an optional `customName` property

then, down in the implementation, you keep evereything as it was and you just use the widget's field info to render the tooltip and the header, without the need of importing a map to compute the value on the fly

not sure if you need both `name` and `customName`, though, because we already have `type`, which i guess is the "types internal" id

packages/slice-machine/src/legacy/components/ItemHeader/index.tsx
```diff
@@ -1,13 +1,16 @@
+import { FieldType } from "@prismicio/types-internal/lib/customtypes";
 import React from "react";
 import type { Theme } from "theme-ui";
 import { Flex, Text } from "theme-ui";
 
+import { fieldLabels } from "@/domain/fields";
+
 const ItemHeader: React.FC<{
-  text: React.ReactNode;
+  type: FieldType;
```
dani-mp: can we keep text here, maybe making it a string? before, this component didn't know anything about our domain and it was purely presentational

packages/slice-machine/src/legacy/components/ListItem/Header.tsx
```diff
@@ -1,35 +1,42 @@
+import { Box, Tooltip } from "@prismicio/editor-ui";
 import * as CSS from "csstype";
 import { IconType } from "react-icons";
 import { Flex, Text, Theme } from "theme-ui";
 
+import { CustomFieldType, fieldLabels } from "@/domain/fields";
```
dani-mp: same as the comment before, we could avoid this with the solution proposed, passing from above the actual value instead of a value to compute onother one and couple it with the domain

packages/slice-machine/src/domain/fields.ts
jomifepe: I copied these from the add field dropdown and left them as they were 😕

Screenshot 2025-08-06 at 16 03 44

packages/slice-machine/src/domain/fields.ts
jomifepe: Should be resolved with the latest refactor ✅

packages/slice-machine/src/domain/fields.ts
jomifepe: Should be resolved with the latest refactor ✅

packages/slice-machine/src/domain/fields.ts
jomifepe: I like that approach! Thanks.

Should be resolved with the latest refactor ✅

packages/slice-machine/src/legacy/components/ItemHeader/index.tsx
jomifepe: Should be resolved with the latest refactor ✅

Kept the `text` prop.

dani-mp (approved):

packages/slice-machine/src/legacy/components/ListItem/Header.tsx
jomifepe: Should be resolved with the latest refactor ✅

PR #1690, opened 2025-08-07
[DT-1289] fix: prevent screenshots from being uploaded every time

jomifepe:
Resolves: [DT-1289](https://linear.app/prismic/issue/DT-1289)

### Description

Slice screenshots are being uploaded every time a slice changes, even when the screenshot remains the same. This is due to the `imageUrl` being stripped and never stored locally. Mutating the local model would mean that the user would get a git change after pushing, which is not ideal, and was most likely what motivated this decision.

To fix this, this PR proposed passing the remote slice `imageUrl` down to the manager when pushing changes, so that it's used for the comparison and we can avoid firing an image upload request for each slice variation.

### Benchmark

Let's consider a slice with 6 variations, each one with a screenshot with ~600KB.

```
Before improvements:
- 3214 ms
- 2776 ms
- 3003 ms
- 3056 ms
- 3470 ms
- 2932 ms

After improvements:
- 1751 ms
- 1922 ms
- 1746 ms
- 1859 ms
- 1589 ms
- 1817 ms
```

This change can reduce the push time of that slice up to **~40%**.

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/manager/src/managers/prismicRepository/types.ts
+++ b/packages/manager/src/managers/prismicRepository/types.ts
@@ -130,7 +130,7 @@ type SliceChange = {
 	 * push slices with the current screenshot. If a matching screenshot is not
 	 * found in this map, the current local screenshot is uploaded again.
 	 */
-	variationImageUrlMap: Record<string, string>;
+	variationImageUrlMap?: Record<string, string>;
 };
 
 export type TransactionalMergeArgs = {
--- a/packages/manager/src/managers/prismicRepository/types.ts
+++ b/packages/manager/src/managers/prismicRepository/types.ts
@@ -130,7 +130,7 @@ type SliceChange = {
 	 * push slices with the current screenshot. If a matching screenshot is not
 	 * found in this map, the current local screenshot is uploaded again.
 	 */
-	variationImageUrlMap?: Record<string, string>;
+	variationImageUrlMap: Record<string, string>;
 };
 
 export type TransactionalMergeArgs = {
--- a/packages/manager/src/managers/prismicRepository/types.ts
+++ b/packages/manager/src/managers/prismicRepository/types.ts
@@ -124,6 +124,10 @@ type SliceChange = {
 	type: "Slice";
 	status: ChangeStatus;
 	libraryID: string;
+	/**
+	 * A map of variation IDs to remote screenshot URLs. Used to detect if a
+	 * screenshot has changed when comparing with local ones.
+	 */
 	variationImageUrlMap: Record<string, string>;
 };
 
--- a/packages/manager/src/managers/prismicRepository/types.ts
+++ b/packages/manager/src/managers/prismicRepository/types.ts
@@ -125,8 +125,9 @@ type SliceChange = {
 	status: ChangeStatus;
 	libraryID: string;
 	/**
-	 * A map of variation IDs to remote screenshot URLs. Used to detect if a
-	 * screenshot has changed when comparing with local ones.
+	 * A map of variation IDs to remote screenshot URLs. These URLs are used to
+	 * detect if a screenshot has changed when comparing with local ones and to
+	 * push slices with the current screenshot.
 	 */
 	variationImageUrlMap: Record<string, string>;
 };
--- a/packages/manager/src/managers/prismicRepository/types.ts
+++ b/packages/manager/src/managers/prismicRepository/types.ts
@@ -124,6 +124,7 @@ type SliceChange = {
 	type: "Slice";
 	status: ChangeStatus;
 	libraryID: string;
+	variationImageUrlMap: Record<string, string>;
 };
 
 export type TransactionalMergeArgs = {
--- a/packages/manager/src/managers/prismicRepository/types.ts
+++ b/packages/manager/src/managers/prismicRepository/types.ts
@@ -127,7 +127,8 @@ type SliceChange = {
 	/**
 	 * A map of variation IDs to remote screenshot URLs. These URLs are used to
 	 * detect if a screenshot has changed when comparing with local ones and to
-	 * push slices with the current screenshot.
+	 * push slices with the current screenshot. If a matching screenshot is not
+	 * found in this map, the current local screenshot is uploaded again.
 	 */
 	variationImageUrlMap: Record<string, string>;
 };
--- a/packages/manager/src/managers/slices/SlicesManager.ts
+++ b/packages/manager/src/managers/slices/SlicesManager.ts
@@ -83,8 +83,9 @@ type SliceMachineManagerPushSliceArgs = {
 	sliceID: string;
 	userAgent?: string;
 	/**
-	 * A map of variation IDs to remote screenshot URLs. Used to detect if a
-	 * screenshot has changed when comparing with local ones.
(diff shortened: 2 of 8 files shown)


packages/slice-machine/src/features/sync/getUnSyncChanges.ts
```diff
@@ -76,11 +77,26 @@ export function getUnSyncedChanges(
...
-    .map((s) => ({
-      slice: s,
-      status: modelsStatuses.slices[s.model.id],
-    }))
+    .map((slice) => {
+      const status = modelsStatuses.slices[slice.model.id];
+
+      const sliceWithRemote = slices.find((s) => {
+        return hasRemote(s) && s.remote.id === slice.model.id;
+      }) as RemoteOnlySlice | undefined;
```
xrutayisire: ⚠️ Exactly like this, for me it's wrong, right? You're casting like there is no local, but you just use hadRemote, so you may still have local.

packages/manager/src/managers/slices/SlicesManager.ts
```diff
@@ -1023,13 +1038,18 @@ export class SlicesManager extends BaseManager {
...
+				const remoteImageUrl = args.variationImageUrlMap?.[variation.id];
+				const hasScreenshotChanged = !remoteImageUrl?.includes(
 					createContentDigest(screenshot.data),
 				);
 
 				// If screenshot hasn't changed, do nothing
 				if (!hasScreenshotChanged) {
-					return variation;
+					return {
+						...variation,
```
xrutayisire, line 1045: ❓ Why this change here? Why before it was not updating imageUrl with the remote why?

The comment above `// If screenshot hasn't changed, do nothing` is wrong now as we do something, we update the `imageUrl`.

packages/manager/src/managers/slices/SlicesManager.ts
```diff
@@ -760,6 +774,7 @@ export class SlicesManager extends BaseManager {
 			const modelWithScreenshots =
 				await this.updateSliceModelScreenshotsInPlace({
 					libraryID: args.libraryID,
+					variationImageUrlMap: args.variationImageUrlMap,
```
xrutayisire: ⚠️ This is always undefined no? Where is `pushSlice` used?

packages/manager/src/managers/prismicRepository/types.ts
```diff
@@ -124,6 +124,13 @@ type SliceChange = {
 	type: "Slice";
 	status: ChangeStatus;
 	libraryID: string;
+	/**
+	 * A map of variation IDs to remote screenshot URLs. These URLs are used to
+	 * detect if a screenshot has changed when comparing with local ones and to
+	 * push slices with the current screenshot. If a matching screenshot is not
+	 * found in this map, the current local screenshot is uploaded again.
+	 */
+	variationImageUrlMap?: Record<string, string>;
```
xrutayisire: 💡 Globally for all your `variationImageUrlMap`, why it's an optional param? Is it not important for us to always compare with the correct image? When do we not want to have it?

xrutayisire (changes requested):

packages/slice-machine/src/features/sync/getUnSyncChanges.ts
jomifepe: Yeah I understand that, I didn't care about it because it was only meant to be deconstructed right after. These `LocalOrRemoteX` types are a bit of pain to work with.

Improved the cast ✅

packages/manager/src/managers/slices/SlicesManager.ts
jomifepe: I agree with the comment being wrong, but this change was needed. Since we're reading the slice from the disk, and local slices don't have remote image URLs, as they are never updated to avoid producing git changes for the users. Because of that, we get the remote URL (which is already the one saved in Prismic) and send it in the payload to ensure we don't clear it.

This was not a problem before because we were updating it every time, so this imageUrl was being send after upload, now we do the same, but with the current remote URL.

Just corrected the comment ✅

packages/manager/src/managers/prismicRepository/types.ts
jomifepe: I thought about this, and decided to leave it optional to avoid changing it everywhere. If by any chance it wasn't passed down, the screenshot will still be correct, just uploaded again. Although, I changed it everywhere it mattered, so this will not happen.

I can make it required to force the developer to always think about that prop, I can see the value of it. WDYT?

packages/manager/src/managers/slices/SlicesManager.ts
jomifepe: As of now, yes, it's only being used by init, but this is only passing it down, it can be `undefined` when it reaches the `updateSliceModelScreenshotsInPlace` function.

packages/manager/src/managers/slices/SlicesManager.ts
xrutayisire: Ok I get it. And just to be 100% sure, what if the user actually update the screenshot? We replace first with the old image url and then replace it with the new one?

packages/manager/src/managers/prismicRepository/types.ts
xrutayisire: Agreed together to make it required

packages/manager/src/managers/slices/SlicesManager.ts
jomifepe: When the user updates a screenshot, the digest/hash of the local screenshot will not match the one found in the matching URL from `variationImageUrlMap`, so this condition will fail, and the upload of the new image will happen.

packages/slice-machine/src/features/sync/getUnSyncChanges.ts
xrutayisire: Hum the cast is still wrong, right? It's not a huge problem but there is potentially `local` and you are saying here it's only remote. 

Maybe to improve maintenance we could express the fact it's still a LocalOrRemote BUT we now are sure to have remote.

It's like:

```ts
const sliceWithRemote = slices.find(
        (s): s is LocalOrRemoteSlice & RemoteOnlySlice => {
          return hasRemote(s) && s.remote.id === slice.model.id;
        },
      );
```

It could be also something more verbose:
```ts
const sliceWithRemote = slices.find((s) => {
        return hasRemote(s) && s.remote.id === slice.model.id;
      });

      const imageUrlMap =
        sliceWithRemote && hasRemote(sliceWithRemote)
          ? sliceWithRemote.remote.variations.reduce<Record<string, string>>(
              (result, variation) => {
                const { imageUrl } = variation;
                if (imageUrl === undefined || imageUrl === "") return result;
                result[variation.id] = imageUrl;
                return result;
              },
              {},
            )
          : {};
```

jomifepe: Just improved it, I prefer a helper function, WDYT?

jomifepe: I can even let it infer the return type, committed.

xrutayisire (approved):
LGTM! Nice work

PR #1691, opened 2025-08-08
[DT-2837] feat: handle unauthorized auth status

jomifepe:
Resolves: [DT-2837](https://linear.app/prismic/issue/DT-2837)

### Description

Handle unauthorized errors the same way we handle invalid environment, but also display a button to logout.

### Preview

Screenshot 2025-08-11 at 12 09 35

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/legacy/components/AppLayout/index.tsx
+++ b/packages/slice-machine/src/legacy/components/AppLayout/index.tsx
@@ -46,7 +46,7 @@ export const AppLayout: FC<PropsWithChildren> = ({
               />
               <BlankSlateTitle>Failed to load Slice Machine</BlankSlateTitle>
               <BlankSlateDescription>
-                <Box alignItems="center" flexDirection="column" gap={8}>
+                <Box alignItems="center" flexDirection="column" gap={16}>
                   {JSON.stringify(error)}
                   {isUnauthorizedError(error) && (
                     <LogoutButton>Log out</LogoutButton>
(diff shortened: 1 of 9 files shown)


packages/slice-machine/src/legacy/components/AppLayout/index.tsx
```diff
@@ -74,6 +66,43 @@ export const AppLayout: FC<PropsWithChildren> = ({
   );
 };
 
+function RenderError(args: { error: unknown }) {
```
xrutayisire, line 69: 🚀 I like the component idea, you pass the error and each case you handle it how you want with a default one pretty strong in case of a report!

xrutayisire (approved):
LGTM!

packages/slice-machine/src/legacy/components/Navigation/Environment.tsx
```diff
@@ -98,17 +97,10 @@ export function Environment() {
...
-  ) {
+  if (isUnauthenticatedError(useEnvironmentsError)) {
     return (
       <SideNavEnvironmentSelector
-        variant={
-          isUnauthenticatedError(useEnvironmentsError)
-            ? "unauthenticated"
-            : "unauthorized"
-        }
+        variant="unauthenticated"
```
xrutayisire, line 103: Do we need a variant then?

jomifepe: Yes, the default value is `default`, and still also use the `offline` one above

PR #1692, opened 2025-08-08
[DT-2699] refactor(deps): remove unnecessary radix dependencies

jomifepe:
Resolves: [DT-2699](https://linear.app/prismic/issue/DT-2699)

### Description

Remove unnecessary slice-machine radix dependencies because we already get all of them from `editor-ui`.

Context: 
- https://github.com/prismicio/editor/pull/1742
- https://prismic-team.slack.com/archives/C0441S8SKQR/p1748421650495319

### How to QA [^1]

[^1]:


xrutayisire (commented):
Do you remember why they were added? I guess it was to fix something no?

jomifepe:
> Do you remember why they were added? I guess it was to fix something no?

Editor changed the individual `radix-ui/*` dependencies to a single `radix-ui` one, because it seems like the radix team improved tree shaking and there's no need to have the individual packages anymore. Because of that, we already have access to these packages by having `editor-ui` as a dependency, so we don't need to have the individual `radix-ui/*` packages installed anymore.

xrutayisire (approved):
LGTM!

PR #1693, opened 2025-08-12
[DT-2847] Improve unauthorized state handling

jomifepe:
Resolves: [DT-2847](https://linear.app/prismic/issue/DT-2847)

### Description

We recently started handling the `unauthorized` state the same way we handle invalid environment, by displaying a full-screen message and only allowing the user to log out. This helps us to clean up edge cases, as an `unauthorized` user should not keep being able to continue modeling.

While making that change, we noticed that the validations for these unwanted states are dependent on an authenticated request that is not blocking the normal UI render, so it appears for a second (depends on the connection speed) before the error screen. This is not ideal, and this PR aims to improve that. The goal is to have all _critical_ requests and checks be complete before continuing to render the UI.

#### What was done?

- Added a suspense call to the root of the project that allows us to validate if the user is `unauthorized` or has an `invalid environment` and throw an error to display an error screen handled by the top-most error boundary
- Renamed `unauthorized` auth status to `unauthenticated` and `authorized` to `authenticated`
- Removed some validations of the previous `unauthorized` state (cleanup)
- Improved the error message for both states handled in the root error screen
- Display logout button for both error states
- Prevent login modal from being closed while login is in progress. Since the auth happens in a different tab/window, this is to prevent the user from navigating throughout the app while we're transitioning states.
- Added `react-query` provider to the root of the tree and refactored two requests (environments and active environment) from `useRequest` to `useQuery`/`useSuspenseQuery`.

#### Unauthorized

Screenshot 2025-08-14 at 11 57 43

#### Invalid environment

Screenshot 2025-08-14 at 11 59 32

#### Unauthorized

https://github.com/user-attachments/assets/30374b02-6ec1-4855-8ef9-2eda8a3be028

#### Invalid environment

https://github.com/user-attachments/assets/3156edec-df1f-4b7e-96de-a111e20eed1e

#### Regular login & logout

https://github.com/user-attachments/assets/277f594a-9d58-4352-b7fc-c9ea0d915a19

### How to QA [^1]

[^1]:


packages/slice-machine/src/legacy/components/ChangesEmptyState/UnauthenticatedView.tsx
jomifepe, line 1: This file is not a new addition under legacy, it was a rename of AuthErrorPage and code removal.

packages/slice-machine/src/modules/userContext/index.ts
```diff
@@ -92,10 +93,12 @@ const getAuthStatus = (
     case undefined: {
       return AuthStatus.AUTHORIZED;
     }
-    case 403: {
-      return AuthStatus.UNAUTHORIZED;
-    }
     case 401: {
+      if (clientError.name === new UnauthenticatedError().name) {
+        return AuthStatus.UNAUTHENTICATED;
+      }
```
jomifepe, line 99: Context: `401` can also mean `unauthorized` (see [SliceMachineManager.tsx](https://github.com/prismicio/slice-machine/pull/1693/files#diff-55a2bf90479dfdc62301d24549d3776f932aa197c18ba68ab0b34fc89d1651bd)), and we don't care about handling this state in the regular UI, because we have an error page that will prevent it from being reached, so we now only handle `unauthenticated`.

packages/slice-machine/src/errorBoundaries.tsx
jomifepe, line 1: This file is also not 100% a new addition. `ErrorBoundary.tsx` was renamed, along with it's only error boundary, and new one was added.

packages/slice-machine/src/errorBoundaries.tsx
```diff
@@ -0,0 +1,150 @@
...
+  PropsWithChildren,
+  useCallback,
+  useRef,
+} from "react";
+
+import { ReloadLogoutButton } from "@/features/auth/LogoutButton";
+import { useAuthStatus } from "@/hooks/useAuthStatus";
+
+type DefaultErrorBoundaryProps = Pick<
+  // TODO(DT-1979): Export the `ErrorBoundaryProps` type from `@prismicio/editor-ui`.
```
xrutayisire: 💡 Note: I put this ticket in the low-hanging fruit project.

packages/slice-machine/src/modules/userContext/types.ts
```diff
@@ -1,6 +1,6 @@
 export enum AuthStatus {
   AUTHORIZED = "authorized",
```
xrutayisire: 💡 I would name it `AUTHENTICATED`, as it's easier to understand the diff between `authenticated` and `unauthenticated`.

packages/slice-machine/src/modules/userContext/types.ts
```diff
@@ -1,6 +1,6 @@
...
-  UNAUTHORIZED = "unauthorized",
+  UNAUTHENTICATED = "unauthenticated",
   FORBIDDEN = "forbidden",
```
xrutayisire: ⚠️ Why do we still have that? I think this should never be handled manually in the UI, right?

packages/slice-machine/src/pages/_app.tsx
```diff
@@ -68,101 +67,95 @@ const RemoveDarkMode: FC<RemoveDarkModeProps> = ({ children }) => {
...
-  Component,
-  pageProps,
-}: AppContextWithComponentLayout & AppInitialProps) {
-  const [serverState, setServerState] = useState<ServerState | null>(null);
-  const [smStore, setSMStore] = useState<{
-    store: Store;
-    persistor: Persistor;
-  } | null>(null);
+function App(props: AppContextWithComponentLayout & AppInitialProps) {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
xrutayisire: ❓ Why this `eslint-disable-next-line`?

packages/slice-machine/src/pages/_app.tsx
```diff
@@ -68,101 +67,95 @@ const RemoveDarkMode: FC<RemoveDarkModeProps> = ({ children }) => {
...
+              </TooltipProvider>
+            </ThemeProvider>
+          </RemoveDarkMode>
+        </QueryClientProvider>
       </ThemeUIThemeProvider>
     </>
   );
 }
 
+/** This is where we should check for unwanted states that should prevent the
```
xrutayisire, line 108: What I'm not at ease with here is you're writing something that is not fully true.
You're mixing environment and custom types for example.
There is nothing here related to the custom types fetching, correct? 

As you already spent enough time on it, I'm more into adding a clear comment about the fact "if env is unauthorized, custom types fetching is also unauthorized".
This sentence is correct, right?

packages/slice-machine/src/pages/_app.tsx
```diff
@@ -68,101 +67,95 @@ const RemoveDarkMode: FC<RemoveDarkModeProps> = ({ children }) => {
...
+        },
+        slices: {
+          libraries: serverState.libraries,
+          remoteSlices: serverState.remoteSlices,
+        },
+      });
+
+      return { serverState, store, persistor };
+    },
+    staleTime: Infinity,
```
xrutayisire, line 151: 💡 Let's add a comment to explain that

packages/slice-machine/src/errorBoundaries.tsx
xrutayisire, line 1: 💡 We cannot have that properly in a folder now? Like in `features` maybe with one component per folder?
It's like what we have for `auth`, `navigation` etc.

xrutayisire (changes requested):
First quick review before my meetings

packages/slice-machine/src/pages/_app.tsx
jomifepe: It's because `pageProps` is of type `any` and out eslint config doesn't like that we deconstruct that here.

packages/slice-machine/src/modules/userContext/types.ts
jomifepe: Yeah that makes sense to me, renamed ✅

packages/slice-machine/src/modules/userContext/types.ts
jomifepe: You're right, it can be removed, we're also not handling any forbidden case anymore.

Removed ✅

packages/slice-machine/src/pages/_app.tsx
jomifepe: Done ✅

packages/slice-machine/src/pages/_app.tsx
jomifepe: I'm not sure I understood. This comment is supposed to be generic. What it means is that we should use that component to validate top-level unwanted states that should prevent the user from using the app.

> "if env is unauthorized, custom types fetching is also unauthorized".
> This sentence is correct, right?

That is partially correct, as we've seen that other request may still be fired if the app was previously mounted, but yes on a first render the custom types request will never be fired if this fails first.

packages/slice-machine/src/errorBoundaries.tsx
jomifepe: Yeah I agree with that, moved ✅

packages/slice-machine/src/pages/_app.tsx
xrutayisire: Sorry, if I'm not clear on this message, we can call 😅

So for me here you're using the active env to detect all auth status.
But it's a bit of a hack, as the real authStatus is derived from getState still today.
I would make a comment about this explaining to a dev this protects us but don't derive the auth status

packages/slice-machine/src/pages/_app.tsx
xrutayisire: I would maybe do `<Component {...props.pageProps} />` 😅
Didn't want to look at it too much, but protect us so we don't have to disable ESLint for the next line and keep protection for the rest.

You think it works?

packages/slice-machine/src/pages/_app.tsx
jomifepe: Ok so I believe it's a bit of a mix of concepts, even though both are technically handling some form of auth status, they are not really connected. 

Although I added a comment to clarify why we're using the active environment requests for this effect ✅

WDYT?

packages/slice-machine/src/pages/_app.tsx
jomifepe: Yeah also works, applied ✅

packages/slice-machine/src/pages/_app.tsx
xrutayisire: Let's start with that, anyway we need to rework the whole app components one day if we keep SM like that. Out of this scope

xrutayisire (approved):

PR #1719, opened 2025-11-17
feat: figma to slice v1

jomifepe:
## Description

Code for the Figma to Prismic plugin. First version with only pasting shortcut (CTRL/CMD+V)
- Parsing the clipboard content coming from the plugin and triggering an agent for model & boilerplate code generation.
- Reuses the existing `Generate from image` modal
- The regular generation from uploaded images still uses the same service as before, only pasting from Figma uses the new agent.

## Preview

https://github.com/user-attachments/assets/e26b6684-f2b2-4e8f-981e-98cd9ec7ed28

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/manager/src/managers/customTypes/CustomTypesManager.ts
+++ b/packages/manager/src/managers/customTypes/CustomTypesManager.ts
@@ -601,21 +601,6 @@ export class CustomTypesManager extends BaseManager {
 						);
 					}
 
-					let frameworkFileExtension: string | undefined;
-					if (framework.type === "nextjs") {
-						frameworkFileExtension = "tsx";
-					} else if (framework.type === "nuxt") {
-						frameworkFileExtension = "vue";
-					} else if (framework.type === "sveltekit") {
-						frameworkFileExtension = "svelte";
-					}
-
-					if (!frameworkFileExtension) {
-						throw new Error(
-							"Could not determine framework from Slice Machine config.",
-						);
-					}
-
 					const projectRoot = await this.project.getRoot();
 					const libraryAbsPath = joinPath(projectRoot, libraryID);
 					const cwd = libraryAbsPath;
--- a/packages/manager/src/managers/customTypes/CustomTypesManager.ts
+++ b/packages/manager/src/managers/customTypes/CustomTypesManager.ts
@@ -36,6 +36,12 @@ import { UnauthorizedError } from "../../errors";
 
 import { BaseManager } from "../BaseManager";
 import { CustomTypeFormat } from "./types";
+import { existsSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { randomUUID } from "node:crypto";
+import { join as joinPath } from "node:path";
+import { mkdtemp, rename, rm, writeFile, readFile } from "node:fs/promises";
+import { query as queryClaude } from "@anthropic-ai/claude-agent-sdk";
 
 type SliceMachineManagerReadCustomTypeLibraryReturnType = {
 	ids: string[];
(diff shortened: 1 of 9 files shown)


packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -533,37 +539,339 @@ export class CustomTypesManager extends BaseManager {
...
+	
+	## Step 2: Model the Prismic slice
+	2.1. Call mcp__prismic__how_to_model_slice to learn how to structure the model for this design.
+	- Make sure the name you use for the new slice does not yet exist in the slice library at <slice_library_path>. If it does, use a different name.
+	2.2. Build the complete slice JSON model data in memory based on the guidance received and the information extracted from the image.
+	2.3. Call mcp__prismic__save_slice_data to save the model (DO NOT manually write model.json) in the slice library at <slice_library_path>.
+	
+	## Step 3: Code a boilerplate slice component based on the model
+	3.1. Call mcp__prismic__how_to_code_slice to learn how to properly structure the slice component with Prismic fields.
+	3.2. Update the slice component code at <slice_library_path>/index.${frameworkFileExtension}, replacing the placeholder code with boilerplate code with the following requirements:
```
xrutayisire: 💡Is it mandatory to have the `frameworkFileExtension` value? Like if we say index file, I would assume AI find the file, no?

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -533,37 +539,339 @@ export class CustomTypesManager extends BaseManager {
...
+	
+	<slice_library_path>
+	${libraryAbsPath}
+	</slice_library_path>
+	
+	<framework>
+	${framework.label}
+	</framework>
+	
+	# AVAILABLE TOOLS
```
xrutayisire: 🎉In love with this way to expose tools!

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -533,37 +539,339 @@ export class CustomTypesManager extends BaseManager {
...
+	
+	## Step 3: Code a boilerplate slice component based on the model
+	3.1. Call mcp__prismic__how_to_code_slice to learn how to properly structure the slice component with Prismic fields.
+	3.2. Update the slice component code at <slice_library_path>/index.${frameworkFileExtension}, replacing the placeholder code with boilerplate code with the following requirements:
+	- Must NOT be based on existing slices or components from the codebase.
+	- Must render all the Prismic components to display the fields of the slice model created at <slice_model_path>.
+	- Must be a valid ${framework.label} component.
+	- Must NOT have any styling/CSS. No inlines styles or classNames. Just the skeleton component structure.
+	- Must NOT use any other custom component or functions from the user's codebase.
+	- Avoid creating unnecessary wrapper elements, like if they only wrap a single component (e.g., <div><PrismicRichText /></div>).
```
xrutayisire: 💡Is this instruction working? I feel when we give AI an example, even if it's things NOT to do, it will use the example as the valid code haha.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -533,37 +539,339 @@ export class CustomTypesManager extends BaseManager {
...
+	FINAL REMINDERS:
+	- You MUST use mcp__prismic__save_slice_data to save the model
+	- You MUST call mcp__prismic__how_to_code_slice in Step 3.1
+	- DO NOT ATTEMPT TO BUILD THE APPLICATION
+	- START IMMEDIATELY WITH STEP 1.1 - NO PRELIMINARY ANALYSIS;`,
+						options: {
+							cwd: libraryAbsPath,
+							stderr: (data) => console.error("inferSlice error:" + data),
+							model: "claude-haiku-4-5",
+							permissionMode: "bypassPermissions",
```
xrutayisire: ⚠️ We never fully check discuss that, but do you know the implication of that param? Is it dangerous for our users? 

Could we use just "edit"?


> default: Use this for production, untrusted codebases or when you want to maintain strict oversight of what the agent does.
> acceptEdits: Good for trusted repos, rapid development workflows, or an internal tool where team trust exists and you’re comfortable letting the agent make edits without interruption.
> bypassPermissions: Only if you have a fully sandboxed environment (throw-away repo, safe context) and want maximum autonomy (at the cost of safety).

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -533,37 +539,339 @@ export class CustomTypesManager extends BaseManager {
...
+							allowedTools: [
+								"Bash",
```
xrutayisire: 💡Is it necessary to have bash?

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -533,37 +539,339 @@ export class CustomTypesManager extends BaseManager {
...
+								"Glob",
+								"Task",
+								"Edit",
+								"Write",
+								"MultiEdit",
+								"mcp__prismic__how_to_model_slice",
+								"mcp__prismic__how_to_code_slice",
+								"mcp__prismic__save_slice_data",
+							],
+							disallowedTools: [
```
xrutayisire, line 824: 🎉Nice one! But even with that, you do need to tell the AI again and again to not touch the files?

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -533,37 +539,339 @@ export class CustomTypesManager extends BaseManager {
...
+								ANTHROPIC_BASE_URL: llmProxyUrl,
+								ANTHROPIC_CUSTOM_HEADERS:
+									`x-prismic-token: ${authToken}\n` +
+									`x-prismic-repository: ${repository}\n`,
+							},
+							mcpServers: {
+								prismic: {
+									type: "stdio",
+									command: "npx",
+									args: ["-y", "@prismicio/mcp-server@0.0.20-alpha.6"],
```
xrutayisire, line 846: 💡I guess here we would put `@latest` or you think for prod we should hard code the version?

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -533,37 +539,339 @@ export class CustomTypesManager extends BaseManager {
...
+							},
+						},
+					});
+
+					let newSliceAbsPath: string | undefined;
+
+					for await (const query of queries) {
+						switch (query.type) {
+							case "result":
+								if (query.subtype === "success") {
```
xrutayisire, line 860: 👀 Funny how we get back the info, was this a problem of reliability?

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -533,37 +539,339 @@ export class CustomTypesManager extends BaseManager {
...
+	YOU ARE NOT FINISHED UNTIL YOU HAVE THESE DELIVERABLES.
+	
+	---
+	
+	FINAL REMINDERS:
+	- You MUST use mcp__prismic__save_slice_data to save the model
+	- You MUST call mcp__prismic__how_to_code_slice in Step 3.1
+	- DO NOT ATTEMPT TO BUILD THE APPLICATION
+	- START IMMEDIATELY WITH STEP 1.1 - NO PRELIMINARY ANALYSIS;`,
+						options: {
```
xrutayisire, line 802: ⚠️I think we should put a `abortController` with a timeout, WDYT? (don't have to be in this PR)

But like if after 5 min, MCP didn't finish we abort and fail. With the custom proxy we have I'm afraid the MCP could end up in a loop of 20 min without stopping 😣

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -533,37 +539,339 @@ export class CustomTypesManager extends BaseManager {
...
+					);
+
+					return InferSliceResponse.parse({ slice: JSON.parse(model) });
+				} finally {
+					if (tmpDir && existsSync(tmpDir)) {
+						await rm(tmpDir, { recursive: true });
+					}
+				}
+			} else {
+				const searchParams = new URLSearchParams({ repository });
```
xrutayisire, line 897: ❓So at the end we use infer slice for Image. What is the cost after everything is finish to also use MCP for images?

xrutayisire (commented):
Nice work!

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: Yes, I would change this to `latest` before releasing, but right now, the version with the `save_slice_data` tool is only released as an alpha.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: I have implemented a cancellation feature, which uses the `abortController`, and I plan on create a PR to add it, I can add it there. WDYT?

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: For modeling, no major cost, but coding takes longer, and that may degrade the experience for people who only want to create the model, so it's something to discuss. I left it like this to avoid changing what has been working for a while, but I plan on changing it all to use the same approach.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: This approach worked on the POC, but it doesn't stop the AI from trying. It's just extra protection to make sure it doesn't succeed at changing them directly.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: Not really, I was trying to be as clear as possible, but it works without the extension 😁

Removed

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: I feel like it is, yeah, at first I saw it creating stuff like:
```tsx
<div>
  <PrismicRichText field={slice.primary.description} />
</div>
```

I found it unnecessary, so I told it to avoid it, and I haven't seen it do it again.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: Valid point, you're right! Changed to `acceptEdits` ✅

Besides this, there are more security-oriented settings in `allowedTools` and `disallowedTools`.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: The rationale here is that I don't want to prevent it from using bash commands to list directories, find files, etc., which it does pretty often.

Although I just added some more disallowed tools for added security. WDYT?

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: I never had this fail, so I would say it's pretty reliable.

This is not entirely necessary, but since the other `inferSlice` approach deals with the model after generation, I return the model as well, and we have to change fewer things. This will probably change if we start using the agent for everything.

packages/manager/src/managers/customTypes/CustomTypesManager.ts
xrutayisire: We can merge this and you add it after no problem, faster

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -786,6 +841,11 @@ export class CustomTypesManager extends BaseManager {
 								`Write(**/model.json)`,
 								"Edit(**/mocks.json)",
 								"Write(**/mocks.json)",
+								...otherSlices.flatMap((slice) => [
+									`Write(${slice.relPath})`,
```
xrutayisire, line 830: 💡 Hum you saw problem without that? It's adding quite some code on top to get all others slices.

jomifepe: Not with editing other files, no. But I believe since we're changing the user's codebase, it's important to give it boundaries to be on the safe side. I would even give it write permissions only to the new slice directory, but it's not created yet at this stage, but could be something to improve.

The code above was also added for the disallowed names, as an effort to have it not list the slice library to come up with a unique name. The idea is to give it all the resources to avoid looking around for information, which would speed up the process.

xrutayisire (approved):
Nice work!

PR #1724, opened 2025-11-19
feat(figma-to-slice): cancellation

jomifepe:
### Description

`Cancel` button to abort all `inferSlice` processes, both for API and Claude agent.

#### Why?

If the user exits the page (e.g., reload or close the tab), the claude process would continue and eventually make the changes to the codebase anyway.

### Preview

https://github.com/user-attachments/assets/8d794780-8c0f-458f-ab6e-774ef42edb68

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/manager/src/managers/customTypes/CustomTypesManager.ts
+++ b/packages/manager/src/managers/customTypes/CustomTypesManager.ts
@@ -560,10 +560,8 @@ export class CustomTypesManager extends BaseManager {
 		const authToken = await this.user.getAuthenticationToken();
 		const repository = await this.project.getResolvedRepositoryName();
 
-		let abortController: AbortController | undefined;
-		let timeoutId: NodeJS.Timeout | undefined;
-		abortController = new AbortController();
-		timeoutId = setTimeout(
+		const abortController = new AbortController();
+		const timeoutId = setTimeout(
 			() => {
 				if (abortController && !abortController?.signal.aborted) {
 					abortController?.abort();
--- a/packages/manager/src/managers/customTypes/CustomTypesManager.ts
+++ b/packages/manager/src/managers/customTypes/CustomTypesManager.ts
@@ -965,9 +965,7 @@ FINAL REMINDERS:
 		}
 	}
 
-	async cancelInferSlice(args: {
-		requestId: string;
-	}): Promise<{ cancelled: boolean }> {
+	cancelInferSlice(args: { requestId: string }): { cancelled: boolean } {
 		const { requestId } = args;
 		const abortController = this.inferSliceAbortControllers.get(requestId);
 
--- a/packages/manager/src/managers/customTypes/CustomTypesManager.ts
+++ b/packages/manager/src/managers/customTypes/CustomTypesManager.ts
@@ -569,7 +569,7 @@ export class CustomTypesManager extends BaseManager {
 					if (abortController && !abortController?.signal.aborted) {
 						abortController?.abort();
 						console.warn(
-							`inferSlice (${source}) request timed out after 5 minutes`,
+							`inferSlice (${source}) request ${requestId} timed out after 5 minutes`,
 						);
 					}
 				},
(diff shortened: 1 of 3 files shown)


packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -548,17 +550,30 @@ export class CustomTypesManager extends BaseManager {
 	}
 
 	async inferSlice(
-		args: { imageUrl: string } & (
+		args: { imageUrl: string; requestId?: string } & (
```
xrutayisire: ❓ How can requestId be undefined?

packages/manager/src/managers/customTypes/CustomTypesManager.ts
```diff
@@ -913,14 +930,43 @@ FINAL REMINDERS:
...
-			console.info(`inferSlice (${source}) took ${elapsedTimeSeconds}s`);
+			console.info(
+				`inferSlice took ${elapsedTimeSeconds}s${
+					requestId ? ` for request ${requestId}` : ""
+				}`,
+			);
 		}
 	}
+
+	async cancelInferSlice(args: {
```
xrutayisire: 💡 Does this need to be async?

packages/slice-machine/src/features/customTypes/customTypesBuilder/CreateSliceFromImageModal/CreateSliceFromImageModal.tsx
```diff
@@ -72,11 +73,15 @@ export function CreateSliceFromImageModal(
     ["meta+v", "ctrl+v"],
     (event) => {
       event.preventDefault();
-      void handlePaste();
+      void onPaste();
     },
     { enabled: open && isFigmaEnabled },
   );
 
+  useEffect(() => {
```
xrutayisire, line 88: ⚠️ What is the intent of this code?

For me the cleanup will always use the slices value from the first render (empty array), because the effect never re-runs. On unmount, cancelActiveRequests won’t see the currently generating slices, so it won’t actually cancel active requests no?

packages/slice-machine/src/features/customTypes/customTypesBuilder/CreateSliceFromImageModal/SliceCard.tsx
```diff
@@ -15,13 +15,16 @@ export function SliceCard(props: SliceCardProps) {
   const loading = slice.status === "uploading" || slice.status === "generating";
 
   const error =
-    slice.status === "uploadError" || slice.status === "generateError";
+    slice.status === "uploadError" ||
+    slice.status === "generateError" ||
+    slice.status === "cancelled";
```
xrutayisire, line 20: ❓Not sure to get when this slice status will be set?

xrutayisire (changes requested):

packages/slice-machine/src/features/customTypes/customTypesBuilder/CreateSliceFromImageModal/SliceCard.tsx
jomifepe: My bad, I was cherry picking from another branch and missed that block of code. Added ✅

Looks like this now:

Screenshot 2025-11-19 at 17 19 12

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: Nope, nice catch. Removed ✅

packages/manager/src/managers/customTypes/CustomTypesManager.ts
jomifepe: It cannot, in the current usage. The idea was that, since it's on the manager, and it can be used anywhere, and we could just call it without thinking about cancelling it.

jomifepe: But I also see that without it, the timeout won't be defined. Making it required

jomifepe: Done ✅

packages/slice-machine/src/features/customTypes/customTypesBuilder/CreateSliceFromImageModal/CreateSliceFromImageModal.tsx
jomifepe: Yeah, this was a bit forgotten from the first implementation.

I also saw that I should handle reload/tab close, so I just updated it ✅

xrutayisire (approved):

PR #1729, opened 2025-11-24
feat: bump engines and warn about node version on init

jomifepe:
### Description

Screenshot 2025-11-28 at 15 05 32

### How to QA [^1]

[^1]:


packages/init/package.json
```diff
@@ -96,5 +96,8 @@
 	},
 	"publishConfig": {
 		"access": "public"
+	},
+	"engines": {
+		"node": ">=20.19.0"
```
xrutayisire, line 98: 💡 Why not just >=20

packages/manager/package.json
```diff
@@ -127,7 +127,7 @@
 		}
 	},
 	"engines": {
```
xrutayisire, line 129: 💡 Let's change all `engines.node` property for every packages in SM, no?
Or your goal was to minimize where we really need above 20?

Because if manager needs 20, everything needs 20 anyway. As manager is called by everyone.

packages/init/package.json
jomifepe: Only works from `20.19`, because of this:

> [2025-03-13, Version 20.19.0 'Iron' (LTS), @marco-ippolito](https://nodejs.org/en/blog/release/v20.19.0#2025-03-13-version-20190-iron-lts-marco-ippolito)
> This is a special minor release. Although v20 is in maintenance mode, meaning only patch releases are expected, an exception was made to backport require(esm) due to its importance and impact on the ecosystem.

jomifepe:
[This PR](https://github.com/prismicio/slice-machine/pull/1730) fixed the issue of not being able to run init with <20.19, so these changes are not needed for now.

packages/manager/package.json
jomifepe: Update `start-slicemachine` and `slice-machine-ui`

xrutayisire (approved):

PR #1749, opened 2025-12-15
feat(generate): better logged out feedback and generation tracking

jomifepe:
### Description

We've been seen a few errors of user trying to generate a slice and being logged out and we're trying to figure out the cause and improve the experience.

1. When `inferSlice` errors out because the user is authenticated, show the login modal again
2. Track `llmProxyUrl` to debug recurrent issue

### How to QA [^1]

[^1]:

### DIFF
--- a/packages/slice-machine/src/features/customTypes/customTypesBuilder/CreateSliceFromImageModal/CreateSliceFromImageModal.tsx
+++ b/packages/slice-machine/src/features/customTypes/customTypesBuilder/CreateSliceFromImageModal/CreateSliceFromImageModal.tsx
@@ -18,6 +18,7 @@ import {
   Text,
 } from "@prismicio/editor-ui";
 import { SharedSlice } from "@prismicio/types-internal/lib/customtypes";
+import { isUnauthenticatedError } from "@slicemachine/manager/client";
 import { useRouter } from "next/router";
 import { useEffect, useRef, useState } from "react";
 import { useHotkeys } from "react-hotkeys-hook";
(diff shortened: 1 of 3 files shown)


packages/slice-machine/src/features/customTypes/customTypesBuilder/CreateSliceFromImageModal/CreateSliceFromImageModal.tsx
```diff
@@ -297,6 +296,13 @@ export function CreateSliceFromImageModal(
...
       });
 
+      if (isUnauthenticatedError(error)) {
+        toast.error("Please log in and try again.");
+        closeModal();
+        openLoginModal();
+        return;
+      }
+
       void telemetry.track({ event: "slice-generation:ended", error: true });
```
xrutayisire, line 306: 💡 Shouldn't we put in the manager also the end so both start / end are in the same scope?

jomifepe: I thought about that, but also thought that the remaining generation code could also fail. But if we mostly care about the actual agent part I can move it there.

xrutayisire: Ok I see, it's true that we do quite more code on top of it, let's keep it like that then. Even more that we want better tracking, not less.

xrutayisire (approved):

